PR #6, opened 2011-04-18
Declared Beans enhancement : class hierarchies, positional args, field initialization 

sotty:
The fork also contains the prototypes for PMML support
(drools-pmml and drools-informer modules), but they should be moved to the drools-chance sub-project.


mdproctor:
done

PR #7, opened 2011-04-20
Pull request to fix JBRULES-2976

ljnelson:
Hello; I'm new to git and github, so I apologize if I am inadvertently ignoring etiquette or making some other mistake.

[JBRULES-2976](https://issues.jboss.org/browse/JBRULES-2976) is a bug where the `DefaultMarshaller`--the default implementation of the `Marshaller` interface--does not use the `ObjectMarshallingStrategy` that is supplied at construction.  It _stores_ it, but never does anything with it.  Instead, the strategy reachable from the supplied `StatefulKnowledgeSession` is used instead.

The patch/pull request here makes sure that the strategy is used appropriately.

All tests continue to pass in the drools-core module.

Please do let me know if this is an appropriate pull request, or what else I should do to get this fix in.

Thanks for all the work done on the Drools project.

Best,
Laird

### DIFF
--- a/drools-core/src/main/java/org/drools/marshalling/impl/DefaultMarshaller.java
+++ b/drools-core/src/main/java/org/drools/marshalling/impl/DefaultMarshaller.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010 JBoss Inc
+ * Copyright 2010, 2011 JBoss Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


etirelli:
Hi ljnelson, thanks for contributing. Your patch looks ok, so I will go ahead and merge it, but we always try to add a unit or integration test with every fix we do to the code. If you could provide such a test case would be awesome, as we can then avoid regressions in the future.

PR #18, opened 2011-05-16
JBRULES-2776 Left and Right Unlinking

leogomes:
Hi Guys,

This is the first implementation for the left and right unlinking feature discussed here: http://blog.athico.com/2010/08/left-and-right-unlinking-community.html

I've added a configuration option that allows you to activate it, since by default it's not active.

Comments are welcome.


etirelli:
Applied. Thanks!

PR #20, opened 2011-05-20
JBRULES-3039

ghost:
Remove type arguments before looking for class to avoid
ClassNotFoundException.

Also, make type names and qualified identifiers safe against
invading spaces.


etirelli:
Applied. Thanks.

PR #21, opened 2011-05-22
JBRULES-3045 Avoid ESE after simple syntax errors inside exists CE and not CE

ghost:
JBRULES-3045 Avoid ESE after simple syntax errors inside exists CE and not CE


etirelli:
Patch applied, thanks! Although, I think I need to remove the stack from the parser and use method parameters instead. The stack makes the parser code very fragile and susceptible to bugs.

PR #23, opened 2011-05-27
JBRULES-2940: Knowledge Agent puts all assets in just one package

esteban-aliverti:
- No more @Ignore'd tests


etirelli:
Applied.

PR #28, opened 2011-06-06
JBRULES-2960

esteban-aliverti:
JBRULES-2960: ClassPathResource fails to get last version if the underlying ClassLoader uses a cache.
    - Removed URLClassPathResource
    - ClassPathResource: getInputStream() uses getURL().getStream() instead of classloader.getResourceAsStream()


etirelli:
Applied.

PR #29, opened 2011-06-08
Parser nits

ghost:


etirelli:
Wolfgang, as discussed, I applied the last commit only: https://github.com/laune/drools/commit/22005a6a4129907ff736f0a61244284fd654ca97

PR #30, opened 2011-06-09
Guvnor 1476

joewhite101:
Added ability for DSL to support ~. Includes my unit test and Edson's changes to support ~ character in DSL files.


etirelli:
Applied.

PR #31, opened 2011-06-14
Jbrules 2913

esteban-aliverti:
Edson, the important commit here is 41e93d6f. I think the other commit (ecfa4935) is the duplicated commit you didn't apply from my last pull request. It is still bothering me. 
Please try to apply it again since it shouldn't do any harm. If you can't, then just apply 41e93d6f and I will create a fresh fork of drools.


etirelli:
Esteban, I did applied the commit for the JBRULES-2960 before... your repo is probably out of synch.

I just applied the commit for JBRULES-2913 individually, so I am closing this pull request.

PR #32, opened 2011-06-18
SimpleDateFormat synchronization issue

zenix:
SimpleDateFormat is not thread safe. Synchronized access to it if using static.


triceo:
Please consider pulling this into 5.2.x as well.

etirelli:
Geoffrey, since we have to release in a few hours, I will push this pull request as is. We can later tune this if necessary.

PR #34, opened 2011-06-27
Junit query

ghost:
A simple query that should return all objects fails to do so. Variations depend on the pattern Object() being absent, before or after the query call.


ghost:
Only the last commit should be taken 369433c.

etirelli:
Commit was applied: https://github.com/laune/drools/commit/369433cf94e12eb1df1c5e17d65a5fe9e5a15940

As we discussed, the pull request could not be applied.

PR #35, opened 2011-06-27
https://issues.jboss.org/browse/JBRULES-3104

sotty:
Adding support for Java @Annotations on declared beans and their fields


etirelli:
Applied.

PR #37, opened 2011-07-09
Fix potential KnowledgeAgent bugs, added a test for a TMS bug 

sotty:


etirelli:
Applied to both 5.2.x and master.

PR #39, opened 2011-07-21
JBRULES-3143 Knowledge Agent does not handle multiple packages in Builder output from single resource

sotty:
Handle resource mapping when multiple sources and packages are involved


etirelli:
Applied to both master and 5.2.x branches.

PR #40, opened 2011-07-22
JBRULES-3124, JBRULES-3063, JBRULES-2730

gwendo:
Implementation of warnings, warning filters and detection of duplicate rules and functions.


drools-compiler/src/main/java/org/drools/compiler/DroolsProblem.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.drools.builder.KnowledgeBuilderError;
+
+/**
+ * 
+ *
+ */
+public abstract class DroolsProblem implements KnowledgeBuilderError {
+
+    public abstract boolean isError();
```
mdproctor: Can we make this an enum? As currently there is error and warning. But maybe in the future we'll have other options, or possibly even different severities of errors and warnings.

drools-compiler/src/main/java/org/drools/compiler/DroolsError.java
```diff
@@ -18,21 +18,10 @@ package org.drools.compiler;
 
 import org.drools.builder.KnowledgeBuilderError;
```
mdproctor: KnowledgeBuilderError is a public interface and thus in knowlege-api. We (almost) enver expose concrete classes to users and all interfaces a user is expected to see go in knowledge-api. Because this is also used by jBPM we avoid the Drools prefix and instead use something generic - like KnowledgeBuilderError instead of DroolsBuilderError. So go with KnowledgeBuilderProblem.

drools-compiler/src/main/java/org/drools/compiler/DroolsError.java
```diff
@@ -18,21 +18,10 @@ package org.drools.compiler;
...
 
-public abstract class DroolsError implements KnowledgeBuilderError {
```
mdproctor: Please make sure we don't break backwards comptability for knowledge-api.

drools-compiler/src/main/java/org/drools/compiler/DroolsWarning.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.compiler;
+
+public abstract class DroolsWarning extends DroolsProblem {
```
mdproctor: This either needs to be an enum or an interface, if it's something  a user is to see or be aware of.  And remember it must be in knowlege-api, or it's not visible to end users.

drools-compiler/src/main/java/org/drools/compiler/DuplicateFunctionWarning.java
```diff
@@ -0,0 +1,51 @@
...
+*/
+package org.drools.compiler;
+
+import org.drools.lang.descr.FunctionDescr;
+import org.drools.rule.Function;
+
+
+
+
+public class DuplicateFunctionWarning extends DroolsWarning {
```
mdproctor: Need to decide if this DuplicateFunctionWarning is going to be exposed to end users and how. If it is, only expose an interface, via knowledge-api. This is true for all warning and error information. I won't reppeat these comments for each similar occurance.

drools-compiler/src/main/java/org/drools/compiler/PackageBuilderErrors.java
```diff
@@ -14,17 +14,17 @@ public class PackageBuilderErrors extends ArrayList<KnowledgeBuilderError>
     implements
     KnowledgeBuilderErrors,
     Externalizable {
```
mdproctor: Make sure we don't break existing knowledge-api code, we must be backwards compatible.

drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
```diff
@@ -2083,15 +2102,46 @@ public class PackageBuilder {
...
+                    }
+                }
+            }
+        }
+        return errors;
+    }
+    
+    public boolean hasWarnings() {
+        return !getWarningList().isEmpty();
+    }
```
mdproctor, line 2171: maybe hasErrors and hasWarnings should be enums? That can be combined via vararg?
hasResults( Warnings )
hasResults( Errors, Warnings )

Again maybe think about how you encapsulate severity as part of the Warning.

drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
```diff
@@ -2083,15 +2102,46 @@ public class PackageBuilder {
...
+                for (WarningFilter filter : configuration.getWarningFilters()) {
+                    if (filter.reportAsError() && filter.matches(problem)) {
+                        errors.add(problem);
+                        break;
+                    }
+                }
+            }
+        }
+        return errors;
+    }
```
mdproctor: Not sure how it would work. But as well as listing errors and warnings, with filters. Maybe people would want to have chained events, so that it does something (like through an exception) if a warning exists.

Again just thinking about how we can future proof our ideas and api.

etirelli:
I think the main point here is just to fix the interface hierarchy, by creating a KnowledgeBuilderResult super interface in drools-api, and have KnowledgeBuilderError, KnowledgeBuilderWarning and KnowledgeBuilderMessage all extend that interface. Then have the concrete classes implement them. To keep backward compatibility we will probably need to add a method to the KnowledgeBuilder interface to return results, based on a filter or not, but not change the existing hasErrors()/getErrors() method... 

Names above are just suggestions.

Also, minor thing, instead of 3-way merges on your pull request, the preferred way is to use rebase as it keeps history linear and easier to track.

gwendo, can you pop up on IRC to talk to us? or just send us an e-mail if you prefer and we can have this moving forward.

Cheers

gwendo:
I'm on holiday for a couple of days. I'll drop in on IRC once I'm back after the weekend.

/Mikael (gwendo)

---

Teamwork doesn't seem work.

25 jul 2011 kl. 17:02 skrev etirellireply@reply.github.com:

etirelli:
This pull request was applied to master branch and 5.3.x branch. It will be available on versions 5.3.0+.

While applying I refactored some class names and some method names, but the code is pretty much unchanged. Major change is to use "result" instead of "problem", so KnowledgeBuilderProblem was renamed to KnowledgeBuilderResult, etc.

Thanks for submitting the patch.

PR #41, opened 2011-08-02
Without that addition drools doesn't run with java7

anpieber:
I consider this a bug since further down that file java7 is considered


anpieber:
there is more than one remaining issue with java7. I'll try to find the remaining ones, create an issue and provide a full pull request

PR #42, opened 2011-08-02
[JBRULES-3163] Minor modifications to make the compiler run with java7

anpieber:
I'm not sure if those are all, but at least for me it works after them

### DIFF
--- a/drools-compiler/src/main/java/org/drools/commons/jci/compilers/EclipseJavaCompilerSettings.java
+++ b/drools-compiler/src/main/java/org/drools/commons/jci/compilers/EclipseJavaCompilerSettings.java
@@ -85,6 +85,7 @@ public final class EclipseJavaCompilerSettings extends JavaCompilerSettings {
         put("1.4", CompilerOptions_VERSION_1_4);
         put("1.5", CompilerOptions_VERSION_1_5);
         put("1.6", CompilerOptions_VERSION_1_6);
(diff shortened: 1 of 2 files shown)


etirelli:
Thanks. Applied.

drools-compiler/src/main/java/org/drools/commons/jci/compilers/EclipseJavaCompilerSettings.java
```diff
@@ -85,6 +85,7 @@ public final class EclipseJavaCompilerSettings extends JavaCompilerSettings {
         put("1.4", CompilerOptions_VERSION_1_4);
         put("1.5", CompilerOptions_VERSION_1_5);
         put("1.6", CompilerOptions_VERSION_1_6);
+        put("1.7", CompilerOptions_VERSION_1_6);
```
anpieber, line 88: woha. one thing I just seen... Is there a CompilerOptions_VERSION_1_7?

etirelli, line 88: yeah, fixed that. Thanks.

anpieber, line 88: Thanks for fixing it and sorry for the error in the pull-request :-( Stupid
copy-paste-modify error...

PR #45, opened 2011-08-06
New feature : "expectation" rules

sotty:
Please review and provide feedback

NOTICE: it also contains a fix for JBRULES-3153, also to be reviewed

### DIFF
--- a/drools-compiler/src/main/java/org/drools/lang/DRLParser.java
+++ b/drools-compiler/src/main/java/org/drools/lang/DRLParser.java
@@ -554,6 +554,18 @@ public class DRLParser {
                     // Going for type includes generics, which is a no-no (JIRA-3040)
                     String superType = qualifiedIdentifier();
                     declare.superType( superType );
+
+                    while ( input.LA( 1 ) == DRLLexer.COMMA ) {
+
+                        match( input,
+                                DRLLexer.COMMA,
+                                null,
+                                null,
+                                DroolsEditorType.SYMBOL );
+
+                        superType = qualifiedIdentifier();
+                        declare.superType( superType );
+                    }
                 }
             }
 
@@ -3504,8 +3516,6 @@ public class DRLParser {
      * 
      * type := ID typeArguments? ( DOT ID typeArguments? )* (LEFT_SQUARE RIGHT_SQUARE)*
      * 
-     * @param doQualify set to true if qualification is acceptable
-     * @param doGenPar  set to true if generic arguments and brackets are acceptable
      * @return
      * @throws RecognitionException
      */
--- a/drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
@@ -23,6 +23,7 @@ import java.io.Reader;
 import java.io.Serializable;
 import java.io.StringReader;
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -66,6 +67,12 @@ import org.drools.factmodel.AnnotationDefinition;
 import org.drools.factmodel.ClassBuilder;
 import org.drools.factmodel.ClassDefinition;
 import org.drools.factmodel.FieldDefinition;
+import org.drools.factmodel.traits.TraitRegistry;
+import org.drools.factmodel.traits.IThing;
+import org.drools.factmodel.traits.TraitBuilder;
+import org.drools.factmodel.traits.ITraitable;
+import org.drools.factmodel.traits.Traitable;
+import org.drools.factmodel.traits.Trait;
 import org.drools.facttemplates.FactTemplate;
 import org.drools.facttemplates.FactTemplateImpl;
 import org.drools.facttemplates.FieldTemplate;
@@ -313,7 +320,20 @@ public class PackageBuilder {
         activationType.setTypesafe( false );
         activationType.setTypeClass( Activation.class );
         builtinTypes.put( Activation.class.getCanonicalName(),
-                          activationType );        
+                          activationType );
+
+        TypeDeclaration thingType = new TypeDeclaration( IThing.class.getName() );
+        thingType.setFormat( TypeDeclaration.Format.TRAIT );
+        thingType.setTypeClass( IThing.class );
+        builtinTypes.put( IThing.class.getCanonicalName(),
+                          thingType );
+        ClassDefinition def = new ClassDefinition();
+            def.setClassName( thingType.getTypeClass().getName() );
+            def.setDefinedClass( IThing.class );
+        TraitRegistry.getInstance().addTrait( def );
+
+
+
 
     }
 
@@ -383,7 +403,7 @@ public class PackageBuilder {
 
     /**
      * Load a rule package from XML source.
-     * 
+     *
      * @param reader
      * @throws DroolsParserException
      * @throws IOException
@@ -1253,9 +1273,9 @@ public class PackageBuilder {
      *            the current package registry
      * @return the fully qualified name of the superclass
      */
-    private String resolveSuperType(String sup,
-                                     PackageDescr packageDescr,
-                                     PackageRegistry pkgRegistry) {
+    private String resolveType( String sup,
+                                PackageDescr packageDescr,
+                                PackageRegistry pkgRegistry) {
 
         //look among imports
         for ( ImportDescr id : packageDescr.getImports() ) {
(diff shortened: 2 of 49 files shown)


etirelli:
Davide, your pull request contains 3-way merges. Please use rebase to avoid the merges and resubmit the pull request.

drools-compiler/src/main/java/org/drools/compiler/PackageBuilderConfiguration.java
```diff
@@ -636,12 +654,14 @@ public class PackageBuilderConfiguration
         if ( option instanceof DefaultDialectOption ) {
             this.defaultDialect = (DefaultDialectOption) option;
         } else if ( option instanceof AccumulateFunctionOption ) {
-            this.accumulateFunctions.put( ((AccumulateFunctionOption) option).getName(),
-                                          ((AccumulateFunctionOption) option).getFunction() );
```
etirelli: Davide, are you using the code format template from the Drools project?

drools-compiler/src/main/java/org/drools/lang/DRLParser.java
```diff
@@ -3122,7 +3124,71 @@ public class DRLParser {
             }
 
             while ( input.LA( 1 ) != DRLLexer.EOF && !helper.validateIdentifierKey( DroolsSoftKeywords.END ) ) {
-                input.consume();
+
+                if ( input.LA(1) == DRLLexer.LESS && input.LA(2) == DRLLexer.NEGATION && input.LA(3) == DRLLexer.LEFT_SQUARE ) {
```
etirelli: This line should also check if the configuration parameter is set to enable your additional syntax, shouldn't it? Also, please move the whole matching part into a specific method.

drools-compiler/src/main/java/org/drools/lang/DRLParser.java
```diff
@@ -3144,6 +3214,109 @@ public class DRLParser {
...
+                    null,
+                    DroolsEditorType.SYMBOL );
+
+            if ( state.failed ) return;
+        } catch ( RecognitionException re ) {
+            reportError( re );
+        }
+    }
+
+    private void subParserParam( RHSEscapedDialectParser subParser ) {
```
etirelli: Please try to keep javadocs with the EBNF-like grammar description.

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
```diff
@@ -0,0 +1,525 @@
...
+            dump(ruleBuilder,"");
+        }
+    }
+
+    private boolean isTemporal(ExprConstraintDescr ecd) {
+        String expr = ecd.getExpression();
+        if ( expr == null ) {
+            return false;
+        }
+        boolean ans = ( expr.contains(" before ") || expr.contains(" before[")
```
etirelli: this does not work with plugable operators.  The parser as it is now only understands operators, not evaluators. The Evaluator has an isTemporal() method that is used to determine which ones are temporal. If you really need to know, at parsing time, if an operator is temporal, I imagine you will have to make the parser dependent on the evaluator registry, but I am not sure it is a good thing. Can you defer this choice until compile time?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
```diff
@@ -0,0 +1,525 @@
...
+        repairRuleBuilder.rhs(repairRHS);
+
+        if ( verbose ) {
+            dump(repairRuleBuilder,repairRHS);
+        }
+
+    }
+
+
+    protected void deMorganize(BaseDescr descr, CEDescrBuilder<?,?> argBuilder) throws RuntimeDroolsException {
```
etirelli: Not exactly sure what is your need here, but we have a LogicTransformer class that we use at rete build time to do similar things. Can we unify the code to avoid eventual inconsistencies in the future or is the bellow very specific to this situation you are implementing?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+
+
+
+
+
+
+
+    protected void visitSuccess(BaseDescr descr, CEDescrBuilder<?, ?> argBuilder) throws RuntimeDroolsException {
```
etirelli: Should we extract this logic into a visitor class that can be reused?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+
+    public void setVerbose(boolean verbose) {
+        this.verbose = verbose;
+    }
+
+
+
+    public void dump(RuleDescrBuilder rule, String rhs) {
```
etirelli: Can we have debug methods like this into a separate class to avoid bloating the actual code?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationParser.java
```diff
@@ -0,0 +1,609 @@
...
+import org.drools.lang.api.CEDescrBuilder;
+import org.drools.lang.api.PackageDescrBuilder;
+import org.drools.lang.api.PatternDescrBuilder;
+import org.drools.lang.api.RuleDescrBuilder;
+import org.drools.lang.descr.*;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class DRLExpectationParser implements RHSEscapedDialectParser {
```
etirelli: Davide, I understand that due to the necessity of accessing the existing parser methods, you probably decided to go with a hand coded parser for this. For the main parser, although not ideal, I am trying to keep up-to-date javadocs with the EBNF-like grammar spec... this is to ease maintenance later... if you have any better ideas, please suggest... otherwise, please add javadocs in here with your grammar...

drools-compiler/src/main/java/org/drools/compiler/PackageBuilderConfiguration.java
sotty: no, where can I find it? There are instructions in the Readme.md on how to setup the header and the tabbing spaces, if that is what you're referring to

drools-compiler/src/main/java/org/drools/lang/DRLParser.java
sotty: I will pass a parameter when I construct the parser, if that's ok with you. Otherwise, the parser does not have access to the configuration parameters.

Ok for moving the match into a dedicated method

drools-compiler/src/main/java/org/drools/lang/DRLParser.java
sotty: Ok, I already have the grammar

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
sotty: This could be a problem.. I can't do that at compile time, I need to extract the operators which can trigger fusion's auto-retraction mechanism. Would fusion be able to do that with a plugged operator?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
sotty: I need to negate a whole and/or formula - and that's a standard procedure. But the leaves and their negation are very specific to this situation

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
sotty: so I can extract this,  create an abstract parser with the common parts and a concrete subclass which deals with Fulfills and Viols. Would that be ok?

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
sotty: ok, no problem

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationParser.java
sotty: ok, no problem :)

drools-compiler/src/main/java/org/drools/lang/rhsdialects/DRLExpectationHelper.java
etirelli: We can discuss this over IRC or skype. The thing is, hard coding like this will not work.

PR #48, opened 2011-08-08
A new command to retrieve a FactHandle from FH external form

bauna:
A new command to retrieve a FactHandle from FH external form

Currently there is no way to obtain a FH from a previously serialized ksession. This changes adds a new Command for resolve it.

https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/10


etirelli:
Applied.

PR #50, opened 2011-08-14
performance optimization

mariofusco:
Fix for JBRULES-3177 - I did a couple of performance optimizations that, despite trivial, seem to improve the performances of the compiler. The first one in the EclipseJavaCompiler allows to open a Stream in the findType() method just once instead of twice as it did before. The second one in the JavaDialect uses a precompiled regular expression and pull out from the cycle all the invariant String concatenations.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/rule/builder/dialect/java/JavaDialect.java
+++ b/drools-compiler/src/main/java/org/drools/rule/builder/dialect/java/JavaDialect.java
@@ -10,6 +10,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.*;
 
 import org.drools.base.TypeResolver;
 import org.drools.commons.jci.compilers.CompilationResult;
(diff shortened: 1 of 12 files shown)


drools-compiler/src/main/java/org/drools/rule/builder/dialect/java/JavaDialect.java
```diff
@@ -633,27 +636,21 @@ public class JavaDialect
                                             final String prefix,
                                             final ResourceReader src) {
         // replaces all non alphanumeric or $ chars with _
-        String newName = prefix + "_" + name.replaceAll( "[ -/:-@\\[-`\\{-\\xff]",
-                                                         "_" );
+        final String newName = prefix + "_" + NON_ALPHA_REGEX.matcher(name).replaceAll("_");
+        final String fileName = packageName.replace('.', '/') + "/" + newName + "_";
```
etirelli, line 640: We should probably use a better escaping strategy in the future, something like URL escapes but without the %, to prevent name clashes. For now, lets just keep like that.

etirelli:
Applied. Thanks.

PR #51, opened 2011-08-17
Oracle fixes

mrietveld:
Unit tests that run on Oracle fail when trying to lookup the next sequence value for a generated value -- when trying to use the default hibernate_sequence. 

This pull request adds/implements the JPA @SequenceGenerator annotation to the 2 classes in drools-persistence-jpa so that sequences are explicitly created in the database. This ensures that drools-persistence-jpa will work on Oracle databases while keeping compatibility with other databases. 

See https://hudson.qa.jboss.com/hudson/view/Drools%20jBPM/job/drools-persistence-db-matrix/database=oracle11gR1/lastCompletedBuild/testReport/ for more information on how the tests fail.


etirelli:
Applied. Thanks.

PR #54, opened 2011-08-24
Fixed bug with "disconnected" factHandles (JBRULES-3187) and added a test

mrietveld:
- Fixed bug with "disconnected" factHandles (JBRULES-3187) 
- Added a test to verify that I fixed the bug!

See https://issues.jboss.org/browse/JBRULES-3187 for more information.

### DIFF
--- a/drools-core/src/main/java/org/drools/command/runtime/rule/GetFactHandlesCommand.java
+++ b/drools-core/src/main/java/org/drools/command/runtime/rule/GetFactHandlesCommand.java
@@ -45,6 +45,9 @@ public class GetFactHandlesCommand
         this.filter = filter;
         this.disconnected = disconnected;
     }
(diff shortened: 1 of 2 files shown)


etirelli:
Applied to trunk. Let me know if this should be backported to 5.2.x branch as well. Thanks.

mrietveld:
Edson, 

This fix does _not_ need to be backported: 5.2.x (and 5.2.0.M2.x) contain a different implementation of GetFactHandlesCommand which doesn't contain this logic. 

The 5.2.x version of GetFactHandlesCommand _will_ fail the test included because the FactHandles aren't cloned and remain connected in the Persistence Context (if I understand the code correctly. )

etirelli:
Ok. Thanks.

PR #56, opened 2011-09-04
JBRULES-3200 Support for dynamic typing (aka "traits")

sotty:
The feature allows to attach and remove interfaces to objects (i.e. instances)


drools-compiler/src/test/java/org/drools/factmodel/traits/StudentProxy2.java
```diff
@@ -0,0 +1,135 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.factmodel.traits;
+
+import com.sun.deploy.util.VersionID;
```
etirelli: Trying to apply your PR but getting a compilation error on this import....

sotty: Ouch, a spurious import...
Wait just a moment then, I need to check with Mark (and you) if there is 
another improvement I can make. I'll be on skype/irc in 1hr or so
Davide

etirelli:
Applied.

PR #59, opened 2011-09-09
JBRULES-3205: Add 'name' and 'description' attributes to <Resource> element

esteban-aliverti:
JBRULES-3205: Add 'name' and 'description' attributes to <Resource> element in change-sets
    - Added getters for name and description to InternalResource
    - getters implemented in BaseResourc
    - ResourceHandler takes in account these 2 attributes when creating the Resource
    - Test for ResourceHandler created


etirelli:
Applied.

PR #60, opened 2011-09-12
Jbpm 3106

mrietveld:
This majority of this pull-request is code to fix jbpm-3106 and how the (JTA) transactions are handled by the SingleSessionCommand Service. 

The rest is additional tests and typo fixes. 

It looks like, once upon a time, someone started implementing support/logic for a local transactions (e.g. non-JTA transactions, via the entity manager, etc.) However, this logic was active when JTA transactions were being used, causing the following to sometimes happen: 
1. begin transaction (layer A)
2. begin another transaction (layer B)
3. commit transaction!! (layer B)
4. commit transaction!! -- bang (layer B). 

This happened because the logic to keep track of how started or owns the transaction, was in the transactionManager itself -- but once you start the transaction, you can't tell if you started it or someone else did. The fix is to migrate the logic to the SingleSessionCommandService. 

The change on the TransactionManager interface will break jbpm-persistence-jpa (because it also depends on that interface). I will keep track of when the pull-request is applied and make sure to fix jbpm when that happens. 

Also, the primary cause of jbpm-3106 is that SingleSessionCommandService.initKSession() was being called outside of a transaction, which is a problem when drools/jbpm runs on PostgreSQL. A @Lob in postgresql is stored via the "Large Object" mechanism (http://jdbc.postgresql.org/documentation/84/binary-data.html) which stores the actual field in another table -- thus requiring a transaction because two tables will be involved. Because initKSession() modifies the SessionInfo object (which has a @Lob) field, the initKSession() method needs to be called within a transaction. 

The pull-request also contains the following: 
- fix for PersistenceUtil so that developers not using maven plugins still get the default values
- cleaned out commented-out tests that were moved to jbpm-persistence-jpa a while ago. 
- added Copyright headers. 
- one or two typo's fix (buildComman_d_Service)
- Added a test illustrating how to reload a ksession via JPA

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2011 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.drools.persistence;
 
 import java.lang.reflect.Constructor;
@@ -30,6 +45,8 @@ import org.drools.runtime.Environment;
 import org.drools.runtime.EnvironmentName;
 import org.drools.runtime.KnowledgeSessionConfiguration;
 import org.drools.runtime.StatefulKnowledgeSession;
+import org.drools.runtime.process.InternalProcessRuntime;
+import org.drools.runtime.process.ProcessRuntime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,11 +79,6 @@ public class SingleSessionCommandService
             throw new IllegalArgumentException( "Environment must have an EntityManagerFactory " +
                                                 "or a PersistenceContextManager instance" );
         }
-
-        // @TODO log a warning that all transactions will be locally scoped using the EntityTransaction
-        //        if ( env.get( EnvironmentName.TRANSACTION_MANAGER ) == null ) {
-        //            throw new IllegalArgumentException( "Environment must have an EntityManagerFactory" );
-        //        }
     }
 
     public SingleSessionCommandService(RuleBase ruleBase,
@@ -120,29 +132,26 @@ public class SingleSessionCommandService
         this.sessionInfo.setJPASessionMashallingHelper( this.marshallingHelper );
         ((InternalKnowledgeRuntime) this.ksession).setEndOperationListener( new EndOperationListenerImpl( this.sessionInfo ) );
         
-        // Use the App scoped EntityManager if the user has provided it, and it is open.
-
+        boolean transactionOwner = false;
         try {
-            this.txm.begin();
+            transactionOwner = txm.begin();
  
-            //this.appScopedEntityManager.joinTransaction();
             registerRollbackSync();
 
+            // Use the App scoped EntityManager if the user has provided it, and it is open.
             jpm.getApplicationScopedPersistenceContext().persist( this.sessionInfo );
 
-            this.txm.commit();
-
+            txm.commit(transactionOwner);
         } catch (RuntimeException re){
-            rollbackTransaction(re);
+            rollbackTransaction(re, transactionOwner);
             throw re;
         } catch ( Exception t1 ) {
-            rollbackTransaction(t1);
+            rollbackTransaction(t1, transactionOwner);
             throw new RuntimeException("Wrapped exception see cause", t1);
         } 
 
         // update the session id to be the same as the session info id
         ((InternalKnowledgeRuntime) ksession).setId( this.sessionInfo.getId() );
-
     }
 
     public SingleSessionCommandService(Integer sessionId,
@@ -153,16 +162,28 @@ public class SingleSessionCommandService
             conf = new SessionConfiguration();
         }
                 
-
         this.env = env;
         
         checkEnvironment( this.env );
         
         initTransactionManager( this.env );
 
-        initKsession( sessionId,
-                      kbase,
-                      conf );
+        boolean transactionOwner = false;
+        try {
+            transactionOwner = txm.begin();
+            
+            registerRollbackSync();
+            
+            initKsession( sessionId, kbase, conf );
+            
+            txm.commit(transactionOwner);
+        }catch (RuntimeException re){
+            rollbackTransaction(re, transactionOwner);
+            throw re;
+        } catch ( Exception t1 ) {
+            rollbackTransaction(t1, transactionOwner);
+            throw new RuntimeException("Wrapped exception see cause", t1);
+        }
     }
 
     public void initKsession(Integer sessionId,
@@ -174,12 +195,11 @@ public class SingleSessionCommandService
         }
         
         this.doRollback = false;
-
-        try {
+        try{
             this.sessionInfo = jpm.getApplicationScopedPersistenceContext().findSessionInfo( sessionId );
         } catch ( Exception e ) {
             throw new RuntimeException( "Could not find session data for id " + sessionId,
-                                        e );
+                    e );
         }
 
         if ( sessionInfo == null ) {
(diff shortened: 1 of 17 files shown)


drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
```diff
@@ -153,16 +162,28 @@ public class SingleSessionCommandService
...
         this.env = env;
         
         checkEnvironment( this.env );
         
         initTransactionManager( this.env );
 
-        initKsession( sessionId,
-                      kbase,
-                      conf );
+        boolean transactionOwner = false;
```
mrietveld, line 171: Fix for jbpm-3106

drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
```diff
@@ -360,14 +380,17 @@ public class SingleSessionCommandService
             }
 
             // always cleanup thread local whatever the result
-            SingleSessionCommandService.synchronizations.remove( this.service );
+            Object removedSynchronization = SingleSessionCommandService.synchronizations.remove( this.service );
```
mrietveld, line 383: eh, ja.. not necesary to catch what remove() returns, was helpful for debugging.

drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
```diff
@@ -360,14 +380,17 @@ public class SingleSessionCommandService
...
             
             this.service.jpm.endCommandScopedEntityManager();
 
             StatefulKnowledgeSession ksession = this.service.ksession;
             // clean up cached process and work item instances
             if ( ksession != null ) {
-                ((InternalKnowledgeRuntime) ksession).getProcessRuntime().clearProcessInstances();
+                InternalProcessRuntime internalProcessRuntime = ((InternalKnowledgeRuntime) ksession).getProcessRuntime();
```
mrietveld, line 390: nullpointer check for when commits are done and the process hasn't started yet.

drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManager.java
```diff
@@ -19,11 +34,11 @@ public interface TransactionManager {
 
     int getStatus();
 
-    void begin();
+    boolean begin();
 
-    void commit();
+    void commit(boolean transactionOwner);
```
mrietveld, line 39: This change (and the one below) break jbpm-persistence-jpa.

drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/KnowledgeStoreServiceImpl.java
```diff
@@ -51,7 +66,7 @@ public class KnowledgeStoreServiceImpl
             throw new IllegalArgumentException( "Environment cannot be null" );
         }
 
-        return new CommandBasedStatefulKnowledgeSession( (CommandService) buildCommanService( kbase,
+        return new CommandBasedStatefulKnowledgeSession( (CommandService) buildCommandService( kbase,
```
mrietveld, line 69: typo.

drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
```diff
@@ -146,23 +159,22 @@ public class JtaTransactionManager
         return null;
     }
 
-    public void begin() {
+    public boolean begin() {
```
mrietveld, line 162: Let the caller know whether or not it's the one who started the tx.

drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
```diff
@@ -146,23 +159,22 @@ public class JtaTransactionManager
...
-        } else {
-            this.localTransaction = false;
-        }
+        } 
+        return false;
     }
 
-    public void commit() {
-        if ( this.localTransaction ) {
+    public void commit(boolean transactionOwner) {
```
mrietveld, line 176: Only commit or rollback if you're actually the owner of the tx.

drools-persistence-jpa/src/test/java/org/drools/persistence/map/impl/JpaBasedPersistenceTest.java
```diff
@@ -9,28 +24,42 @@ import org.drools.KnowledgeBase;
...
+    private boolean useTransactions = false;
     
     @Before
     public void setUp() throws Exception {
         ds1 = setupPoolingDataSource();
         
         ds1.init();
-        emf = Persistence.createEntityManagerFactory( PERSISTENCE_UNIT_NAME );
+        emf = Persistence.createEntityManagerFactory( DROOLS_PERSISTENCE_UNIT_NAME );
+        if( useTransactions() ) {
```
mrietveld, line 55: When this test runs on the postgresql database, you need to put tx's around operations with SessionInfo objects.

drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
```diff
@@ -29,24 +30,38 @@ import org.drools.runtime.Environment;
...
+    
+    private PoolingDataSource ds1;
+    private EntityManagerFactory emf;
+    private Environment env;
 
     @Override
     protected void setUp() throws Exception {
         ds1 = setupPoolingDataSource();
         ds1.init();
+
```
mrietveld, line 51: Refactored to make test more readable.

drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
```diff
@@ -268,548 +254,6 @@ public class JpaPersistentStatefulSessionTest extends TestCase {
                       list.size() );
     }
 
-    public void test1() {
-        
-    }
-    
-//    public void testPersistenceWorkItems() {
```
mrietveld, line 275: These are all in jbpm-persistence-jpa now.

drools-persistence-jpa/src/test/java/org/drools/persistence/util/PersistenceUtil.java
```diff
@@ -1,36 +1,92 @@
...
+
+    public static final String DROOLS_PERSISTENCE_UNIT_NAME = "org.drools.persistence.jpa";
+    public static final String JBPM_PERSISTENCE_UNIT_NAME = "org.jbpm.persistence.jpa";
         
     protected static final String DATASOURCE_PROPERTIES = "/datasource.properties";
     private static TestH2Server h2Server = new TestH2Server();
+    
+    private static Properties defaultProperties = null;
+   
+    private static Properties getDefaultProperties() {
```
mrietveld, line 49: Default H2 props.

drools-persistence-jpa/src/test/java/org/drools/persistence/util/PersistenceUtil.java
```diff
@@ -1,36 +1,92 @@
...
+import org.slf4j.LoggerFactory;
 
 import bitronix.tm.resource.jdbc.PoolingDataSource;
 
 public class PersistenceUtil {
 
-    public static final String PERSISTENCE_UNIT_NAME = "org.drools.persistence.jpa";
+    private static Logger logger = LoggerFactory.getLogger( PersistenceUtil.class );
+
+    public static final String DROOLS_PERSISTENCE_UNIT_NAME = "org.drools.persistence.jpa";
```
mrietveld, line 41: So that we can use drools-persistence-jpa/test-jar in jbpm as well.

drools-persistence-jpa/src/test/resources/log4j.xml
```diff
@@ -12,9 +12,19 @@
 		<level value="INFO"/>
 	</logger>
 	
-	<!-- logger name="org.hibernate">
-		<level value="INFO"/>
-	</logger -->
+    <logger name="org.hibernate.cfg.annotations">
```
mrietveld, line 15: Hibernate logging can be effuse.

BooVeMan:
Hi, as we start deployment 18th of september I would very much welcome the inclusion of this patch ASAP, especially into the 5.2 branch as well as we use that in our development! (I opened the JIRA-3106)

etirelli:
Applied.

Unfortunately we can't backport into 5.2 unless it is requested by the QA team, because it affects their testing plans.

BooVeMan:
Edson, 
I'm not quite understanding why you wouldn't backport a patch that was marked blocking and affects postgres production environments! Please reconsider as I will need to backport this manually as we are going into installation with our jBPM5/PostgreSQL based solution on the 18th of this month...

PR #61, opened 2011-09-12
JBRULES-3205: Add 'name' and 'description' attributes to <Resource> element

esteban-aliverti:
JBRULES-3205: Add 'name' and 'description' attributes to <Resource> element in change-sets
    - getName() and getDescription() moved to InternalResource


etirelli:
Applied.

PR #64, opened 2011-10-13
JBRULES-3246 Traiting fails when used with EQUALITY assert behaviour

sotty:


etirelli:
Applied to master only, as per Mark's request.

PR #68, opened 2011-10-24
JBPM-3106 (2nd try): seperating EM initialization from db operations in o

mrietveld:
JBPM-3106 (2nd try): seperating EM initialization from db operations in order to have all operations within tx's.

The first effort to fix this problem caused other problems: 
- see https://github.com/droolsjbpm/drools/commit/5f452c93637d889804f3a48b268c87f67560f004#commitcomment-595711

The changes in this pull request make sure that EM initialization -- which can be done by spring, are done outside of a transaction, while all db operations are done within a tx. This is to ensure postgreSQL compatibility, since the SessionInfo object contains a @LOB, which Postgres will store (via the Large Object facility) in 2 tables -- thus requiring a tx for all operations doing anything with a SessionInfo object.


mrietveld:
Please cherry-pick this to 5.3 as well.

PR #70, opened 2011-11-02
JBPM-3420: Upgrade of Drools to JPA2 and JBoss AS 7.0.2 compatibility of Hibernate 4

oranheim:
Second try of performing a pull request to Drools 5.3.x Branch

### DIFF
--- a/drools-persistence-jpa/pom.xml
+++ b/drools-persistence-jpa/pom.xml
@@ -103,11 +103,7 @@
       <artifactId>hibernate-entitymanager</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.hibernate</groupId>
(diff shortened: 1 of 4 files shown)


mdproctor:
Unfortunately we cannot upgrade to JPA2 annotations, because of our requirements for compatability with JBoss EAP, which is still JPA1.

We cannot be pure JPA, as JPA1 does not have an equivalent of collection of elements.

One idea is that we use JPA1 annotations everywhere. For the collection of elements annotation we move that out to an xml file. We provide two XML files, one that is hibernate specific (for EAP), and one that is JPA2.

etirelli:
Based on Mark's comment, I am closing this pull request. In the future, if we can upgrade this, please resubmit the PR.

PR #71, opened 2011-11-30
Jbpm 3300 (DO NOT PULL)

mrietveld:
While this commit solves the [JBPM-3300](https://issues.jboss.org/browse/JBPM-3300) problem, it also causes the tests listed below to fail. 

_JBPM-3300 problem_: 
When we insert an object into a knowledge session and _also_ into a work item and/or a process, then users expect the modifications to the object in the process to be mirrored in the state of the object in the ksession. 

However, unmarshalling/deserialization (which uses SerializedPlaceholderResolverStrategy) breaks the "link" between the object in the ksession and the process, and recreates one object for the ksession and another for the process. Modifications to "the object" by either the ksession, process or user after that point then cause problems, obviously. 

_Failing Tests_: 
Unfortunately, I don't understand enough about  Drools internals why the tests failing below are failing. Debugging has pointed to a number of things, including (possibly?) agenda items and (concurrent) node memories. 

Help would be appreciated. :) 

```
Failed tests: 
      testMarshallEntryPointsWithSlidingTimeWindow(org.drools.integrationtests.marshalling.MarshallingTest)
      testMarshallEntryPointsWithSlidingLengthWindow(org.drools.integrationtests.marshalling.MarshallingTest)

Tests in error: 
      testCollect(org.drools.integrationtests.FirstOrderLogicTest)
      testCollectNodeSharing(org.drools.integrationtests.FirstOrderLogicTest)
      testCollectResultConstraints(org.drools.integrationtests.FirstOrderLogicTest)
      testCollectFromMVELAfterOr(org.drools.integrationtests.FirstOrderLogicTest)
      testInsertionOrderTwo(org.drools.integrationtests.BackwardChainingTest)
      testAccumulate(org.drools.integrationtests.marshalling.MarshallingTest)
```

### DIFF
--- a/drools-core/src/main/java/org/drools/impl/KnowledgeBaseImpl.java
+++ b/drools-core/src/main/java/org/drools/impl/KnowledgeBaseImpl.java
@@ -184,7 +184,9 @@ public class KnowledgeBaseImpl
         if (sss != null) {
             for (StatefulSession ss : sss) {
                 if (ss instanceof ReteooStatefulSession) {
-                    c.add(new StatefulKnowledgeSessionImpl((ReteooStatefulSession)ss, this));
+                    StatefulKnowledgeSession session = (StatefulKnowledgeSession) ((ReteooStatefulSession) ss).getKnowledgeRuntime();
+                    ((StatefulKnowledgeSessionImpl) session).kbase = this;
+                    c.add(session);
                 }
             }
         }
--- a/drools-core/pom.xml
+++ b/drools-core/pom.xml
@@ -77,7 +77,14 @@
       <artifactId>activation</artifactId>
       <scope>provided</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.slf4j</groupId>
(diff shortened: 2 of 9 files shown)


drools-core/src/main/java/org/drools/impl/KnowledgeBaseImpl.java
```diff
@@ -184,7 +184,9 @@ public class KnowledgeBaseImpl
         if (sss != null) {
             for (StatefulSession ss : sss) {
                 if (ss instanceof ReteooStatefulSession) {
-                    c.add(new StatefulKnowledgeSessionImpl((ReteooStatefulSession)ss, this));
```
mrietveld, line 187: This actually has nothing to do with the problem: it's just unnecessary and could cause problems because it means we create two stateful ksession impl's when we only needed to make one.

mrietveld:
This pull request is no longer pertinent. The issue (object cloning) is still a problem, but something that I won't get to until March because of other priorities.

PR #72, opened 2011-12-01
Jbrules 3304

mrietveld:
[JBRULES-3304](https://issues.jboss.org/browse/JBRULES-3304): 

Because the rulebase does not get inserted into the (marshaller) context that's used for unmarshalling a WorkItem(Info), when MarshallerReaderContext.resolveClass(ObjectStreamClass) get's called, it can not use the ruleBase's root class loader.

If necessary, I can supply a test case for this to show where it goes wrong – at the moment it disappeared with a stash I dropped so that I'll have to recreate it.

Based on the 5.3.x branch, please also cherry-pick to master.


etirelli:
Applied to both master and 5.3.x branch. Thanks.

PR #73, opened 2011-12-10
ExcelParser does not read double < .0001 from Excel cell 

fmeyer:
I've created a new xls based on basic workbook to reproduce this bug. 

I think we should create a nice way to test this stuff. I couldn't test the parser as properly because the strong dependence between the ExcelParser and the DefaultRuleSheetListener which lack of testing


fmeyer:
I also removed an unused piece of code.

etirelli:
@fmeyer I am unable to apply your PR cleanly. Can you please fix and re-submit? Thanks!

manstis:
Merged into droolsjbpm/master. Thanks for the pull request.

PR #75, opened 2011-12-21
JBRULES-3296: org.drools.io.impl.ClassPathResource lastRead not getting set

esteban-aliverti:
- lastRead is now modified when the IputStream of the resource is accessed. This was the behavior until the last modification of this class (where this bug was introduced).


etirelli:
Applied. Thanks.

PR #76, opened 2011-12-30
Standardizing commands

salaboy:
I've extracted an interface called IdentifiableResults to mark all the commands that uses the OutIdentifier property to return a value. 
In a different commit I've modify the SetGlobal Command to not use the out boolean and work in the same way at the other commands that uses the OutIdentifier. Notice that the out boolean was set True and no one else was using it.
This is the first step to improve and standardize the commands


ge0ffrey:
Merged

mrietveld:
Rather, I'm not seeing an implementation of the IdentifiableResult interface, and that puzzles me. :/

PR #77, opened 2012-01-03
[JBRULES-3323] Using Set in accumulate causes NPE in some scenarios.

fahman:
Added test case


etirelli:
Applied. Thanks for the test case.

PR #78, opened 2012-01-05
unit test that exercises bad expression binding type discovery

adembo:
When run, this test fails with the following assertion error:

java.lang.AssertionError: Unable to generate rule invoker. : [Accumulate: input=[AND [[Pattern: id=null; objectType=MyObj]] ]]
    org/drools/Rule_foo_3191b39fd6ce40eaac55301df07acc19AccumulateExpression0Invoker.java (19:754) : Cannot cast from Object to long


mariofusco:
This test has been included in test suite in a slightly modified version

PR #79, opened 2012-01-11
- re adding out boolean for keeping backward compatibility

salaboy:
I've just add the out parameter to the SetGlobalCommand to maintain backward compatibility. The out parameter is hidden inside the SetGlobalCommand implementation to not be used by external classes and its only used by the marshallers that exposes that attribute. For that reason I've also included the setter and getter methods.

### DIFF
--- a/drools-core/src/main/java/org/drools/command/runtime/SetGlobalCommand.java
+++ b/drools-core/src/main/java/org/drools/command/runtime/SetGlobalCommand.java
@@ -44,6 +44,9 @@ public class SetGlobalCommand
 
     @XmlAttribute(name="out-identifier")
     private String  outIdentifier;
+    // Added only for backward compatibility
+    @XmlAttribute(name="out")
+    private boolean out = false;
     
     public SetGlobalCommand() {
     }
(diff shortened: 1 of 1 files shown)


drools-core/src/main/java/org/drools/command/runtime/SetGlobalCommand.java
```diff
@@ -86,11 +89,20 @@ public void setObject( Object object ) {
...
         return this.outIdentifier;
     }
-
+    // Added only for backward compatibility
     public void setOutIdentifier(String outIdentifier) {
+        this.out = true;
         this.outIdentifier = outIdentifier;
     }
+    // Added only for backward compatibility
+    public boolean isOut() {
```
ge0ffrey, line 98: Problem is that all code ignores this now. I am trying an experiment that might fix this

ge0ffrey:
Just adding the boolean didn't work. I took an alternative approach that worked to keep backwards compatibility, see
   https://github.com/droolsjbpm/drools/commit/8c47b13de7e76370ae765a6a6ac36a88ffe6d590

PR #80, opened 2012-01-12
JBRULES-3122: ClassLoader.getResources() returns null - fixed

jespersm:
I've added a simple test case which checks the getResources() call in PackageClassLoader, and supplied the fix also. ClassLoader.getResources() cannot return a null value, according to the spec.


etirelli:
Applied to both master and 5.3.x branches. Thanks for the test and patch.

PR #82, opened 2012-01-17
Added ScannerChangeSetTest which scans CSV resource file for changes.

jsvitak:
This is a test case for
https://bugzilla.redhat.com/show_bug.cgi?id=751326

Which has most probably the same root cause as
https://bugzilla.redhat.com/show_bug.cgi?id=741219

The same problem - some resources like CSV, XLS, BPMN and PKG are not hotdeployable.


etirelli:
Applied. Thanks.

PR #83, opened 2012-01-23
Updated the documentation about negative patterns in stream mode

WolfgangWeiss:
I updated the documentation for drools-fusion to describe the behaviour of excluding objects in negative patterns. See also following discussion: http://drools.46999.n3.nabble.com/rules-users-Problem-with-Negative-Patterns-and-Temporal-Constraints-tt3663344.html#none


etirelli:
Applied.

PR #86, opened 2012-01-26
Added test case for NPE when parsing DT column headers

jsvitak:
Added test case for bug
https://bugzilla.redhat.com/show_bug.cgi?id=743262
(empty column header like action or condition throws NPE, but it should
print standard error message).


ge0ffrey:
thanks, I 'll process it monday or thuesday normally

ge0ffrey:
Merged and problem fixed

PR #88, opened 2012-01-30
Added a test case for https://bugzilla.redhat.com/show_bug.cgi?id=738992

jsvitak:
I have checked with current Drools master and it works ok, so the test
case is added because of possible regression in the future.


etirelli:
@jsvitak : I applied this PR to the master branch. Let me know if you need it cherry-picked into the 5.3.x branch or not.

jsvitak:
Yes, I would like this PR to be cherry-picked to 5.3.x. Test case still fails in BRMS 5.3 ER3, so it seems that also fix could be needed in 5.3.x branch. Thanks in advance.

etirelli:
Reopening so that it can be applied to 5.3.x branch after fixing the problem.

ge0ffrey:
Test cherry-picked to 5.3.x. It works there too without modification on my machine.
Edson already merged it to master.

PR #90, opened 2012-02-01
JBRULES-3370 Array fields are not supported in declared facts

sotty:
Well, not exactly a ground-breaking feature, but still useful :)
Also improves bean initialization with MVEL expression


etirelli:
@sotty thanks for providing this. I was reviewing the code, and with a few changes it can also support multi-dimensional arrays (e.g. Object[][], int[][][], etc). Do you think you can change it for that?

drools-compiler/src/main/java/org/drools/lang/DRLParser.java
```diff
@@ -924,6 +924,31 @@ private void field( AbstractClassTypeDeclarationBuilder declare ) {
         }
     }
```
etirelli, line 933: There is already a rule called type(). Please use that instead of creating a fieldType() rule. It supports multi-dimentional arrays and generics, although I know MVEL does not support generics yet.

sotty:
Yup, I did it a bit too quickly and overlooked a couple of details. I'll upgrade this during the weekend

PR #91, opened 2012-02-02
JBRULES-3374, BZ761707: Canceled activations by AgendaFilter stops the execution of ruleflow

lazarotti:
The Agenda routine to clean up process executions just works perfectly for
activation which has successfully fired. If it is canceled by some AgendaFilter
implementation, some of that routines are not executed and the process
(ruleflow) is interrupted. This commit fix it.


mariofusco:
Cherry-picked from fix in 5.4.x

PR #92, opened 2012-02-03
Test case for BZ#787118

tomason:
Hi,

I just finished test case for bugzilla mentioned in subject.

Tomas Schlosser


etirelli:
Thanks. Comments added to the ticket.

PR #97, opened 2012-02-14
Disconnected ExecuteCommand results

salaboy:
- check for null inside ContextImpl get()
- disconnected execute command

Note:  the execute command is not mapped using JAXB

### DIFF
--- a/drools-core/src/main/java/org/drools/command/ExecuteCommand.java
+++ b/drools-core/src/main/java/org/drools/command/ExecuteCommand.java
@@ -52,11 +52,16 @@ public class ExecuteCommand
         this.outIdentifier = identifier;
         this.disconnected = disconnected;
     }
+    
+    public ExecuteCommand(Command  command, boolean disconnected) {
+        this.command = command;
+        this.disconnected = disconnected;
+    }
 
     public ExecutionResults execute(Context context) {
         StatefulKnowledgeSession ksession = ((KnowledgeCommandContext) context).getStatefulKnowledgesession();
         
-        ExecutionResults kresults = ksession.execute( this.command );
+        ExecutionResults kresults = ((StatefulKnowledgeSessionImpl)ksession).execute(context, this.command );
         if ( this.outIdentifier != null ) {
             ((ExecutionResultImpl)((KnowledgeCommandContext) context ).getExecutionResults()).getResults().put( this.outIdentifier, kresults );
         }
--- a/drools-core/src/main/java/org/drools/command/ExecuteCommand.java
+++ b/drools-core/src/main/java/org/drools/command/ExecuteCommand.java
@@ -16,10 +16,12 @@
 
 package org.drools.command;
 
+import java.util.HashMap;
 import org.drools.command.Command;
 import org.drools.command.Context;
 import org.drools.command.impl.GenericCommand;
 import org.drools.command.impl.KnowledgeCommandContext;
+import org.drools.common.DefaultFactHandle;
 import org.drools.common.InternalFactHandle;
 import org.drools.impl.StatefulKnowledgeSessionImpl;
 import org.drools.reteoo.ReteooWorkingMemory;
@@ -34,7 +36,8 @@ public class ExecuteCommand
 
     private String   outIdentifier;
     private Command<ExecutionResults>  command;
-
+    private boolean disconnected = false;
+    
     public ExecuteCommand(Command  command) {
         this.command = command;
     }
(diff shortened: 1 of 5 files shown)


mrietveld:
Hi Mauricio, 

If you could add a unit test for this, that would be great! 

Thanks,
Marco

salaboy:
The unit test will be inside grid impl that's the only project that used disconnected results. Notice that if you don't set disconnected to true it will work as usual.

salaboy:
Test added, any impediments to merge?

drools-core/src/main/java/org/drools/command/ExecuteCommand.java
```diff
@@ -43,14 +46,40 @@ public ExecuteCommand(String identifier, Command  command) {
...
+    public ExecuteCommand(Command  command, boolean disconnected) {
+        this.command = command;
+        this.disconnected = disconnected;
+    }
 
     public ExecutionResults execute(Context context) {
         StatefulKnowledgeSession ksession = ((KnowledgeCommandContext) context).getStatefulKnowledgesession();
         
-        ExecutionResults kresults = ksession.execute( this.command );
+        ExecutionResults kresults = ((StatefulKnowledgeSessionImpl)ksession).execute(context, this.command );
```
mrietveld: If this is a CommandBasedStatefulKnowledgeSession (i.e. we're using persistence), this will break. 

I haven't been able to find a (non-hack) way to do this with the CommandBasedStatefulKnowledgeSession.

salaboy: I'm not sure if that will ever happen.. The CommandBasedStatefulKnowledgeSession creates these commands to be executed against a "real" session. For that reason, inside the context there is always a StatefulKnowledgeSessionImpl, as far as I know. Which really make sense. Executing a command against a CommandBasedStatefulKnowledgeSession will be something extremely recursive, because new commands will be created once again. I did that cast because the signature of the execute method accepting a context was not exposed in the interfaces, which basically means that nobody was using the execute command (as far as my analysis goes).

My two cents

PS: I love this kind of discussions.. here is where I learn the most :)

salaboy: By the way, I'm not getting any notification about these comments.. do you know how to enable email notifications about this?

mrietveld: Okay, I can understand that (nobody using the execute command). 

Notification: go to your github "home" page and click on "edit your profile". There's a 'Notification Center' tab on the left hand where you can choose how you want to be notified. 

Glad to hear that the discussion isn't bothering you. :) 

I think my last question is: when I imported your changes locally and removed the cast (StatefulKnowledgeSessionImpl) -- just left line 50/64 as it was, then the test did _not_ fail. 

Do you know why? And if possible, can you modify the test so that it would fail if you didn't use the StatefulKnowledgeSessionImpl method?

salaboy: I'm analyzing your first question.. 
"I think my last question is: when I imported your changes locally and removed the cast (StatefulKnowledgeSessionImpl) -- just left line 50/64 as it was, then the test did not fail."

I don't understand how do you want me to modify the tests to fail if I use the StatefulKnowledgeSessionImpl where?

mrietveld: Sorry, yeah, it wasn't that clear. 

It would be great if you could modify the test as follows: 
- (temporarily) change line 50/64 back (no cast, only command as parameter).  
  
  ```
  ExecutionResults kresults = ksession.execute( this.command );
  ```
- modify the test so that it _fails_ because the context is not passed. 
- change line 50/64 back to what it is now (cast, command _and_ context as parameter). 

This is to make sure that if someone else sees this code later and thinks "Huh, that doesn' t make sense! Let's get rid of the cast and context parameter!", that the test then immediately breaks and they realize why that needs to be done. 

Thank you for your patience, Mauricio!

salaboy: I'm trying to get rid of the ((StatefulKnowledgeSessionImpl)ksession) cast.
This was causing troubles inside drools-grid because we where using commands in a complicated way. The cast was only needed to execute the ExecuteCommand, that no one was using :). I will keep you posted on the changes. I've already improved the test to represent more accurately what we are doing inside drools-grid and right now for the test the cast is not needed. I'm looking forward to improve the test to replicate my failures in grid, but I will be offline until Tuesday, I hope I have 5 minutes to commit what I have.. I will keep you posted

mrietveld: Mauricio, No problem, this all sounds good! Good to hear that you improved the test.

salaboy:
Marco, I've just commit the test improved, plus I've added another test showing an improvement to the QueryCommand, that I've discovered when I was testing the execute command. I've double checked the casting and I remove it because inside grid we can make it work without that context. 

Please let me know if there are another issues with these fixes + tests 

PS: notice that the test for the query command fails if you remove the null checking inside the QueryCommand Class. We need to define if it's necessary to inform about that situation or just not assign the value when we are not in a batch execution (which make sense).

Cheers

PR #98, opened 2012-02-15
Test case for BZ#787118

tomason:
I have fixed the case to show another bug.


etirelli:
Hi Tomas, thanks for submitting. As I mentioned in the ticket, the behavior you observed is correct. Therefore, I am closing this pull request.

PR #99, opened 2012-02-15
JBRULES-3389 It is part of OSGi-fying of drools-camel and drools-spring artifacts.

afilipchik:
Also I created pull requests for:
droolsjbpm-build-distribution and droolsjbpm-integration


ge0ffrey:
Sorry, closing without merge. Reasons:
- We've just merged Christiano's improvements to build the OSGi stuff with the latest felix plugin, making the OSGi build a lot simpler. Any new code should follow that way.
- Versions of dependencies should only be declared in droolsjbpm-build-bootstrap/pom.xml

PR #101, opened 2012-02-22
Propagating executeCommand context

salaboy:
- Adding a test which fails if the context is not propagated


salaboy:
This should be an easy merge, this is just a continuation for the pull request nro 97.

mdproctor:
This commit is already redundant, as ti was covered by mrievt
https://github.com/droolsjbpm/drools/commit/ef4e7c9b5c84081bb803a53059b7bb40b9a4cfc5

However there is no unit test for this commit. If you could re-do your pull request, with just your tests, I'll merge the test to track regressions.

mariofusco:
This test case has been already merged

PR #102, opened 2012-02-25
Regression test case for JBRULES-2518

benbennett:
JBRULES-2518 is currently fixed in the latest versions. 

This is just regression test. 

Might be better way to grab lexar operator tokens but not a atlr expert. 
Gladly will change if someone can point me to the method to call to get tokens.


benbennett:
Was going to use this for base for the JBRULES-3399 bug I opened. Probably will try to fix JBRULES-3399  soon.

PR #106, opened 2012-03-08
[JBRULES-3416] added unit test to replicate working memory coruption

fahman:


ge0ffrey:
For jira issue
  https://issues.jboss.org/browse/JBRULES-3416

fahman:
My testcase/code is erroneous. Specifically this codefragment:

{code}
fooLeft.setBar(fooRight.getBar());
fooRight.setBar(barLeft);
session.update(fooLeftFact, fooLeft);
session.update(fooRightFact, fooRight);
{code}

following fix makes the error go away:

{code}
fooLeft.setBar(fooRight.getBar());
session.update(fooLeftFact, fooLeft);
fooRight.setBar(barLeft);
session.update(fooRightFact, fooRight);
{code}

I am retracting the pull request

fahman:
A bug after all, re-requesting the pull. (test case should be still valid since it uses only very basic stuff from drools)

mariofusco:
The jira issue https://issues.jboss.org/browse/JBRULES-3416 has been closed with "Won't fix" by Mark with the following comment: 

"After further investigation, will have to mark this as a "won't fix". Fact's work of head of index buckets and are used for equality checks. If you change a fact's values without updating the engine, then other fact updates have problems and not identify the buckets they should be in. In short either use modify, which avoids this problem, or call update after changes on one fact before making changes to another fact."

PR #108, opened 2012-03-13
JBRULES-3421 FireAllRulesCommand NPE when outIdentifier != null

mikemelton:
- Update FireAllRulesCommand to get the results reference from the session rather than the context
- Add tests for FireAllRulesCommand


mariofusco:
Merged with minor change (avoiding unnecessary cast)

PR #109, opened 2012-03-21
Solve potential NPE

mareki:
We had a NullPointerException while working with jBPM5 and Drools. The NPE was caused within JpaPersistenceContextManager#beginCommandScopedEntityManager. The local CommandScopedEntityManager which was fetched form the environment was never used. instead the global CommandScopedEntityManager was always accessed. If the global CommandScopedEntityManager is null we get a NPE.


mareki:
pull request was wrong - have to check it again

PR #115, opened 2012-04-12
Added UT for JBRULES-3320

guyramirez:


etirelli:
I added the test case to the source code.

PR #117, opened 2012-04-26
JBRULES-3483: Off-by-one error in FullFastIterator.next()

adembo:
Test case for exercising the bug.


mdproctor:
The test is not valid, as you never pass null into the iterator. As it needs the argument to traverse to the next in the list.

adembo:
That may be true, but the test still exercises the bug. There's one row in the table, so by definition it is the "last" row. That row has more than one RightTuple. The first iteration returns the first tuple, and all future iterations fail.

If you like, I can pass a mocked RightTuple whose getMemory() returns...something, but that won't change the outcome of the test.

mdproctor:
test updated and problem is now fixed.

PR #118, opened 2012-05-04
Non-deterministic # of results in query

jmvanel:
The testJMV4 in MiscTestsJMV is non deterministic : often 11 results, 5 sometimes ;
it gets 5 results (bad) with about 50% probability
if you make a grep for "My" on the test code, you get the expected 11 objects ;
with the notion of cartesian product, one can conceive more results and duplicates
(so the status of tests JMV2\* is not clear), but 5 results is clearly an error .
Tests JMV3\* succeed.
commit fe1d637d0267d564c94fff9f329b1e8e024278ca 1 parent cb5c34cab3


mariofusco:
Fixed with this commit https://github.com/droolsjbpm/drools/commit/303829ebb40ddde020511b2c4de27ef819d82976

PR #121, opened 2012-05-31
bz826952 reproducer. Event listeners are lost after reloading knowledge ...

rsynek:
...session from DB


mariofusco:
I added the test case and @Ignored it.

PR #122, opened 2012-06-07
JBRULES-3533 Added unit test for expression failure in 5.4

joewhite101:
JBRULES-3533 Added unit test for expression failure in 5.4


mariofusco:
I used a slightly modified version of the proposed test case to reproduce and fix the bug.

PR #123, opened 2012-06-11
JBRULES 3539 Error calling clone on object  from map on RHS. Adding unit test

joewhite101:
Added a unit test to MiscTest for JBRULES 3539


mariofusco:
The unit test has been slightly modified and added to MiscTest. The bug was caused by mvel and it has been fixed with this commit: https://github.com/mvel/mvel/commit/73606e898d79723fe4536e37693fd33c808b3f6e

PR #126, opened 2012-06-27
Throw exception if no transaction was found

mswiderski:
Instead of leaving ut set to null throw exception if mechanism for finding user transaction failed to find it as all subsequent operation will cause NPE (getStatus, begin, etc)


mrietveld:
Maciej, this looks fine. 
Go ahead and push the commit to master. :) 

(Closing this pull request without a merge because I Hate Merge Bubbles. :)  )

PR #127, opened 2012-07-10
Concurrency

mariofusco:
Fix (apparently) the concurrency issue caused by sliding windows on different entry points

### DIFF
--- a/drools-compiler/src/test/java/org/drools/integrationtests/MultithreadTest.java
+++ b/drools-compiler/src/test/java/org/drools/integrationtests/MultithreadTest.java
@@ -166,6 +166,85 @@ public class MultithreadTest extends CommonTestMethodBase {
         }
     }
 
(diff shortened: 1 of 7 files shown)


triceo:
Hello Mario, is there a BZ / JIRA for that? Can you tell me which releases does this affect?

mariofusco:
Hi Lukas. Don't mind about this PR. I was talking to Edson about a concurrency problem he found on the 5.4 and I sent this PR only to show him a possible quick fix (that by the way doesn't work in all cases, so we already discarded it). Yesterday evening we discussed a slightly different strategy and Edson is already working to implement it. Moreover we believe that in general we need a more robust (and possibly performant) concurrency model, so we already decided to rework it during the development of Drools 6. This bug doesn't affect Drools 5.3

triceo:
OK, thanks for the explanation.

PR #128, opened 2012-07-13
BZ-840040 BZ-840043 JBPM-3714 Implementing Shared Entity Manager support

lazarotti:
Requested by one-off patch, see BZ-840043


krisv:
(Slightly modified version of the) changes merged into branch

PR #132, opened 2012-08-01
JBRULES-3590 ClassCastException in RuleTerminalNode.retractLeftTuple()

tkobayas:
https://issues.jboss.org/browse/JBRULES-3590
Test case and fix.


mariofusco:
Fix merged in a slightly different way. Thanks for your precious help.

PR #134, opened 2012-08-05
Various KnowledgeAgent-related fixes

sotty:


drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
```diff
@@ -849,7 +849,7 @@ void compileAllRules(PackageDescr packageDescr, PackageRegistry pkgRegistry) {
         updateResults();
```
sotty, line 850: The PackageBuilder has become really messy, it might be a good idea to extract the bean generation part and rewrite/optimize it in a dedicated class

drools-core/src/main/java/org/drools/reteoo/BetaNode.java
```diff
@@ -438,15 +438,18 @@ protected void doRemove(final RuleRemovalContext context,
                 }
 
                 FastIterator it = memory.getLeftTupleMemory().fullFastIterator();
+                int x = memory.getLeftTupleMemory().size();
                 for (LeftTuple leftTuple = getFirstLeftTuple(memory.getLeftTupleMemory(), it); leftTuple != null; ) {
                     LeftTuple tmp = (LeftTuple) it.next(leftTuple);
                     if (context.getCleanupAdapter() != null) {
-                        for (LeftTuple child = leftTuple.getFirstChild(); child != null; child = child.getLeftParentNext()) {
```
sotty, line 444: The cleanUp method will disconnect the child tuple, so it won't be possible to iterate its siblings on the parent.
Might something similar have been done in other nodes?

PR #135, opened 2012-08-05
JBRULES-3155 Bug with default properties in KnowledgeStoreServiceImpl

tkobayas:
This is a part of fix for JBRULES-3155 and depends on the other part, which is https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/25 

Add and use addPropertiesLast method so that KnowledgeStoreServiceImpl can merge default properties to the end of the prop list. Now you can overwrite those default properties with KnowledgeSessionConfiguration.

### DIFF
--- a/drools-core/src/main/java/org/drools/SessionConfiguration.java
+++ b/drools-core/src/main/java/org/drools/SessionConfiguration.java
@@ -180,6 +180,12 @@ public class SessionConfiguration
         }
     }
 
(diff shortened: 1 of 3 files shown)


mariofusco:
Unit test merged. Bug fixed by this commit https://github.com/droolsjbpm/drools/commit/cb14666b3feee89f7a6b2b3eb2f7398efa52e673

PR #139, opened 2012-08-27
BZ848855 - Transaction rollback failure after concurrent access: no tran...

mswiderski:
...saction

Ensure that persistence context is cleared after transaction completion to avoid cached entities on hibernate session level which will cause org.hibernate.StaleObjectStateException.
By default when EntityManagerFactory is created using Persistence.createEntityManagerFactory it will create it with extended persistence context type which means that entities will be shared between transactions and in case there are more processes sharing same data base they will get org.hibernate.StaleObjectStateException on following operation:
- getProcessInstance - StaleObjectStateException on ProcessInstanceInfo entity
- startProcess - StaleObjectStateException - on SessionInfo entity


mswiderski:
merged into 5.3.x branch

PR #141, opened 2012-09-06
JBRULES-3619: Knowledge Agent: add the possibility to apply ByteArrayResources

esteban-aliverti:
Added 2 public methods to Kagent to apply/remove arbitrary resources no matter their types.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/agent/BaseKnowledgeAgentTest.java
+++ b/drools-compiler/src/test/java/org/drools/agent/BaseKnowledgeAgentTest.java
@@ -12,6 +12,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.drools.CommonTestMethodBase;
 import org.drools.KnowledgeBase;
+import org.drools.agent.impl.KnowledgeAgentImpl;
 import org.drools.core.util.DroolsStreamUtils;
 import org.drools.core.util.FileManager;
 import org.drools.core.util.IoUtils;
@@ -37,6 +38,7 @@ import org.junit.Before;
 import org.drools.builder.KnowledgeBuilderConfiguration;
 import org.drools.builder.KnowledgeBuilderError;
 import org.drools.builder.KnowledgeBuilderErrors;
+import org.drools.builder.ResourceType;
 
 public abstract class BaseKnowledgeAgentTest extends CommonTestMethodBase {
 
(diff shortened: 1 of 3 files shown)


drools-compiler/src/test/java/org/drools/agent/BaseKnowledgeAgentTest.java
```diff
@@ -140,6 +142,75 @@ public void afterChangeSetApplied(AfterChangeSetAppliedEvent event) {
...
+
+        KnowledgeAgentEventListener l = new KnowledgeAgentEventListener() {
+
+            public void resourceCompilationFailed(ResourceCompilationFailedEvent event) {
+                Iterator<KnowledgeBuilderError> iterator = event.getKnowledgeBuilder().getErrors().iterator();
+                while (iterator.hasNext()) {
+                    KnowledgeBuilderError knowledgeBuilderError = iterator.next();
+                    System.out.println(knowledgeBuilderError.getMessage());
+                }
+                throw new RuntimeException("Unable to compile Knowledge"+ event );
```
mdproctor, line 167: Geoffrey has added slf4j, so best to log errors there now, instead of System.out.println

ge0ffrey, line 167: Something like this:
  logger.error("Error compiling resource ({}): {}", resource.getName(), knowledgeBuilderErro.getMessage());

will give something like this:
  "Error compiling resource (myFile.drl): keyword whan not recognized"

Of course, since a RuntimeException is being thrown, it could also be a good idea not to log anything and just include all error info in the RuntimeException's message.

mdproctor:
not sure I like this, it seems very kludgy. The knowledge agent is for monitoring external resources, but this is applying them.

It's possible to make a MemoryFileSystem and actually use that, without changing KnowledgeBuilder at all.

I recently wrote a MemoryFileSystem here:
https://github.com/droolsjbpm/drools/tree/master/drools-compiler/src/test/java/org/drools/kproject/memory

And here you can see the vfs in action, programmatically, to allow me to build resources in memory without touching disk. See testCompileAndCDI, where I build up the contents of the MemoryFileSystem.
https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/test/java/org/drools/kproject/KProjectTest.java

What makes it work is the classloader, scroll down to line 455:
MemoryFileSystemClassLoader

As you can see the builder factory is using just standard knowledge builder public apis, but they are actually being resolved from in memory.
https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/main/java/org/drools/kproject/KBaseBuilder.java

You can actually mix both in memory and disk based file systems, using the in memory to "overlay" the disk.

I'm closing this for now, speak to me on IRC if you have any more Qs.

PR #142, opened 2012-09-23
JBRULES-3630: Added fix to ResourceHandler, plus a unit test with a dynamically loaded JAR.

rankinc:


mariofusco:
Fix merged with a slightly different test case

PR #143, opened 2012-09-26
Fix for https://issues.jboss.org/browse/JBRULES-2936

jespersm:
Fix ExcelParser.java as per the JIRA - Workaround for compressed unicode reading bug in JXL

https://issues.jboss.org/browse/JBRULES-2936


mariofusco:
I tested this solution with russian or arabic chars and it doesn't work

jespersm:
This change doesn't interfere with handling cyrillic or arabic chars at all, so your test is likely invalid.

ge0ffrey:
@jespersm Mario is merging (or merged) another pull request which upgraded us to the latest poi. This presumably solves these problems too.

PR #146, opened 2012-10-13
synchronized work item map

calbazasebastian:
JpaWorkItemManager is not thread safe.. I've synchronized the workItems map and done some code cleaning.

PS. I tried filling a bug for this, but it didn't let me..don't know why... also the getWorkItems method looks suspicious.. it just returns an empty set.. is that ok?

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
@@ -1,5 +1,6 @@
 package org.drools.persistence.jpa.processinstance;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


mrietveld:
The getWorkItems() method should return an empty set -- this has to do with the code [here](https://github.com/calbazasebastian/drools/blob/master/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java#L459)

The way persistence work, only workitems (or process instances, for that matter) that are being handled within a transaction, should be available during a transaction. When the transaction ends, the cache (of workitems) should be emptied. 

I think the only reason the getWorkItems() method exists is maybe because it has to because of the Interface? Normally, that class would never implement that because exposing the internal cache to other code will just lead to problems.

mrietveld:
Also, I'm not sure making the JpaWorkItemManager needs to be thread safe -- I though it was only used within the synchronized block of the SingleSessionCommandService, which means that it's never called concurrently. 

Of course, I might be mistaken in this, in which case making it thread-safe is good!

calbazasebastian:
Yeah, if it is accessed inside the session's execute method context it's ok...but I saw code in jBPM  GenericHTWorkItemHandler that uses the manager outside the context, but probably there are some more... I think it might be good to add some logic to allow calls to the manager's methods only if we are in the session's execute method context... this will force any invalid usage to be changed/fixed.

calbazasebastian:
I think I might be mistaken... the manager is accessed properly from jBPM..  no need to add thread safety here

PR #147, opened 2012-10-14
releasing the lock only after transaction completion

calbazasebastian:
Fixing SingleSessionCommandService thread safety issue when it's not ghe JTA transaction owner. This fixes the issues seen at  https://issues.jboss.org/browse/JBPM-3814

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
@@ -18,6 +18,7 @@ package org.drools.persistence;
 import java.lang.reflect.Constructor;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


mrietveld:
Could you explain a little bit more about what the pull request is supposed to fix? I've read your comments in [JBPM-3814](https://issues.jboss.org/browse/JBPM-3814) but it's unfortunately still unclear to me what exactly the cause is of the problem and why the pull request (in particular, how the ReentrantLock is used) fixes that. 

Thanks!

calbazasebastian:
Hey,
  The issue is that the session entity manager (s), which are not thread safe, can be accessed concurrently. The scenario for this is:
     - SingleSessionCommandService it's not the JTA transaction owner (the transaction was started before calling execute)
     - the same session is accessed concurrently (starting multiple processes on the same StatefulKnowledgeSessionImpl, completing human tasks, etc)
     - because of the first thing the commit doesn't happen inside the execute method, but some time later when the owner of the transaction commits, and the SynchronizationImpl.afterCompletion method is executed outside the synchronized execute method 
      -  now if another thread enters the execute method and another one is in SynchronizationImpl.afterCompletion , they both access the same entity manager... which is not thread safe.

   ReentrantLock has the same behavior as using the synchronized keyword, but in this scenario we increase the protection/lock domain from the beginning of the execute method till the transaction is actually committed.

Hope this clears things,

   Sebi

calbazasebastian:
Hey, I added a JBPM test case for this in PR https://github.com/droolsjbpm/jbpm/pull/138 .. also made some new commits after running the test. 

Marko, any news on this?

mariofusco:
You said that the SynchronizationImpl.afterCompletion(int status) method executes outside of the public synchronized <T> T execute(Command<T> command) method, and because of that, there's a race condition. I tried to reproduce the effects of this issue in Drools but I couldn't.

In the end Marco found that actually there cannot be any race condition because the afterCompletion() method implements the TransactionSynchronization.afterCompletion() method. TransactionSynchronization methods are fired when the transaction is committed or rolled back, and the transaction is commited inside of the public synchronized <T> T execute(Command<T> command) method.

mrietveld:
Sebastian, I realized this morning that the problem _does_ exist, but that it's a little more complex. I've posted a long [comment](https://issues.jboss.org/browse/JBPM-3814?focusedCommentId=12759760&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12759760) about it on the jira.

PR #148, opened 2012-10-15
JBRULES-3656: Non removed associations issue

marianbuenosayres:
Non removed associations for NotNode caused the issue at JBRULES-3656.
This update would solve the issue but I would like someone who knows a
bit more to review it before commiting it to master.


mariofusco:
This issue has been already fixed with this commit https://github.com/droolsjbpm/drools/commit/674d5e03170c10edcd1bec21d79c50963d7547d3

PR #152, opened 2012-10-19
JBRULES-3663 - using static thread pool and updating dispose logic

calbazasebastian:


mariofusco:
The request is valid, but the patch needs to be improved: initializing a thread pool as a static instance is to ask for trouble and also, the use of system property alone is not consistent with our architecture. We need the pool to be created per kbase or externally but not statically and the configuration in a similar way to other drools configurations. It would be great if you could send another pull request that follows these advices.

mrietveld:
Sebi, FYI: my sincere apologies for not getting to this in the last 5 months. :( Mario is now actively looking at it though (thank you Mario!).

PR #153, opened 2012-10-22
test case for JBRULES-3666

tkobayas:
A test case for JBRULES-3666, assuming the same EventListener should be added only once.


mariofusco:
Test case merged in a different class. 
Thanks a lot.

PR #154, opened 2012-10-25
Added test case for losing POJO fact metadata.

winklerm:
Added test case for losing fact metadata declared in POJO when adding other metadata (such as @Role) in DRL.


mariofusco:
The provided test case has been reimplemented in a more concise way

PR #155, opened 2012-10-26
JBPM-3831 - 2^x 'garbage' timers are created when rolling back a transac...

mswiderski:
...tion that involves the execution of a process that includes a timer

part 1 as jbpm pull request will follow


mrietveld:
+1 
Nice fix, Maciej.

mswiderski:
merged into master and 5.5.x branch

PR #156, opened 2012-10-27
[JBRULES-3670] 5.5.x added test case

wasperen:
This test case compiles and save a drl file with a declare. Then tries to load it and add it to a knowledge base. Refer to JBRULES-3670


mariofusco:
This issue has been fixed by this commit https://github.com/droolsjbpm/drools/commit/6edb81702253514b69def79e91aa5f6e5d9a3f14

PR #157, opened 2012-10-29
CCE storing session with activations in agenda

marianbuenosayres:
When trying to persist (serialize) a session that has an activated agenda, a ClassCastException occurs like this when calling fireAllRules(). This solution avoids this issue by not returning something different from an Activation inside ActivationIterator

### DIFF
--- a/drools-core/src/main/java/org/drools/common/ActivationIterator.java
+++ b/drools-core/src/main/java/org/drools/common/ActivationIterator.java
@@ -54,9 +54,9 @@ public class ActivationIterator
     }
 
     public Object next() {
-        Activation acc = null;
+        Object obj = null;
         if ( this.currentLeftTuple != null ) {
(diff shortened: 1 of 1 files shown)


marianbuenosayres2:
@mrietveld Here's the commit for master. Sorry for the previous error on the pull request

mrietveld:
No problem! Thanks for fixing it so quickly. :) Unfortunately, I know very little about this code, so I'll have to leave the merging to one of the drools experts. :/

salaboy:
Mariano, can you provide an isolated test case demonstrating the failing scenario that generates the exception?
Cheers

mdproctor:
Any news on a test case for this? So we can apply?

marianbuenosayres:
This issue was solved differently from proposed correction as stated on jira issue JBRULES-3674
I'll discard this pull request.

PR #159, opened 2012-10-30
CCE storing session with activations in agenda

marianbuenosayres2:
When trying to persist (serialize) a session that has an activated agenda, a ClassCastException occurs like this when calling fireAllRules(). This solution avoids this issue by not returning something different from an Activation inside ActivationIterator

### DIFF
--- a/drools-core/src/main/java/org/drools/common/ActivationIterator.java
+++ b/drools-core/src/main/java/org/drools/common/ActivationIterator.java
@@ -54,9 +54,9 @@ public class ActivationIterator
     }
 
     public Object next() {
-        Activation acc = null;
+        Object obj = null;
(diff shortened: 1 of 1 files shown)


marianbuenosayres2:
@mrietveld Here's the commit for branch 5.5.x. Sorry for the previous error on the pull request

mrietveld:
No problem! Thanks for fixing it so quickly. :) Unfortunately, I know very little about this code, so I'll have to leave the merging to one of the drools experts. :/

mariofusco:
I fixed this issue as you suggested, but in slightly different way because in that case the object inside the left tuple can be only an Activation or a Boolean.TRUE and we still want to fail fast in all other cases.

Thanks a lot for your help.

PR #160, opened 2012-10-31
BZ862325: Added tests for improper rule deactivation

winklerm:
Added 3 test cases for Bugzilla issue 862325: one for fact modification, the other two for fact retraction and insert. The test cases do not pass (including the one for fact modification which should be already fixed according to Bugzilla comment).

Please review the proposed tests, either they are not correct, or the fix to 862325 is not yet complete.

### DIFF
--- a/drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
+++ b/drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
@@ -1,0 +1,95 @@
+package org.drools.integrationtests
+
+import org.drools.integrationtests.activation.ActivationTest.ActivationTestKind
+import org.drools.integrationtests.activation.ActivationTest.ActivationTestKind.Kind
+import org.drools.integrationtests.activation.ActivationTest.ActivationTestFact
+
+// MODIFICATION
+
+// this rule must fire before the other two
+rule "rule1-modify"
+when
+    ActivationTestKind( kind == Kind.MODIFY )
+    $fact : ActivationTestFact( value == "default" )
+then
+    insert( "first_fired" );
+    modify( $fact ) { setValue( "non-default" ) }
+end
+
+rule "rule2-modify"
+when
+    ActivationTestKind( kind == Kind.MODIFY )
+    String( this == "first_fired" )
+    $fact : ActivationTestFact( value == "default" )
+then
+    // empty consequence
+end
+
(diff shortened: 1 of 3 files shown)


drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
```diff
@@ -0,0 +1,95 @@
...
+
+// MODIFICATION
+
+// this rule must fire before the other two
+rule "rule1-modify"
+when
+    ActivationTestKind( kind == Kind.MODIFY )
+    $fact : ActivationTestFact( value == "default" )
+then
+    insert( "first_fired" );
```
esteban-aliverti, line 15: This insert() creates an Activation for rule2-modify and rule3-modify

drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
```diff
@@ -0,0 +1,95 @@
...
+    modify( $fact ) { setValue( "non-default" ) }
```
esteban-aliverti, line 16: This line here cancels the activations of rule2-modify and rule3-modify. So for the modify test, only rule1-modify should be fired after fireAllRules(). Is that not the case?

drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
esteban-aliverti, line 15: These activations are not going to be executed until the current activation's execution ends

drools-compiler/src/test/resources/org/drools/integrationtests/activation/activation.drl
winklerm, line 16: Please have a look at https://bugzilla.redhat.com/show_bug.cgi?id=862325. I have understood that rules rule2-modify and rule3-modify should fire in this case. Did I get it right?

esteban-aliverti, line 16: Well, we will have to ask @etirelli. But AFAIK, the expected behavior here is that only 'rule1-modify' gets executed.

winklerm, line 16: I have asked @etirelli on irc and you are right, thanks for your comments!

PR #162, opened 2012-11-07
BZ874119: Added test case

winklerm:
Test case reproducing problem with using SlidingTimeWindow with fireUntilHalt() as described in https://bugzilla.redhat.com/show_bug.cgi?id=874119


mariofusco:
I integrated this test in CepEspTest in a more compact way.

PR #163, opened 2012-11-07
BZ874187: Added test case

winklerm:
Added test case for NPE during evaluation of accumulate when run as fireUntilHalt() as described in https://bugzilla.redhat.com/show_bug.cgi?id=874187


mariofusco:
I merged into the master a more compact version of this test

PR #165, opened 2012-11-13
JBRULES-3676: Fix debugging of Java rules

matnil:
Suggested fix for JBRULES-3676

### DIFF
--- a/drools-compiler/src/main/java/org/drools/rule/builder/dialect/DialectUtil.java
+++ b/drools-compiler/src/main/java/org/drools/rule/builder/dialect/DialectUtil.java
@@ -80,7 +80,18 @@ public final class DialectUtil {
                                             final ResourceReader src) {
         // replaces all non alphanumeric or $ chars with _
         final String newName = prefix + "_" + NON_ALPHA_REGEX.matcher(name).replaceAll("_");
-        if (ext.equals("java")) return newName + "_" + generateUUID();
+        if (ext.equals("java")) {
+            boolean isGenerateUniqueJavaRuleName = true;
+            if (System.getProperty("drools.generateUniqueJavaRuleName") != null) {
+            	isGenerateUniqueJavaRuleName = Boolean.getBoolean( "drools.generateUniqueJavaRuleName" );
+            }
+        	if (isGenerateUniqueJavaRuleName) {
+        		return newName + "_" + generateUUID();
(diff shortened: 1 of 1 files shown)


drools-compiler/src/main/java/org/drools/rule/builder/dialect/DialectUtil.java
```diff
@@ -80,7 +80,18 @@ public static String getUniqueLegalName(final String packageName,
                                             final ResourceReader src) {
         // replaces all non alphanumeric or $ chars with _
         final String newName = prefix + "_" + NON_ALPHA_REGEX.matcher(name).replaceAll("_");
-        if (ext.equals("java")) return newName + "_" + generateUUID();
+        if (ext.equals("java")) {
+            boolean isGenerateUniqueJavaRuleName = true;
+            if (System.getProperty("drools.generateUniqueJavaRuleName") != null) {
```
mdproctor, line 85: I see a new property has been added, to stop unique rule generation. This is fixing the symptom, and overriding as a system property is not ideal, and would still cause problems with rules are generated with the same name. Instead we should look why debugging has failed and attempt to reconnect the rule names, with the unique generation.

matnil, line 85: I can agree that the use of a system property is perhaps not the most beautiful solution. But I think most users could live with the limitation that "Debug As -> Drools Application" in Eclipse only works if one does not modify the knowledge base from multiple concurrent threads.

I should also mention that this fix only restores debugging for single-consequence rules. During the time the debugger has been broken "named consequences" have been added to Drools, so the Drools debugger would need to be updated to handle those also.

My initial plan was to try to reconnect the debugger with the generated class name but I found no way to do that. The problem is that Eclipse debugger needs to know in advance, while configuring breakpoints, the class name of of the rule consequence. This logic takes place in a separate JVM from the actual debug session. The code changes that were made for https://issues.jboss.org/browse/JBRULES-3267 made the class names completely random which makes it impossible to know the name of the class beforehand.

I hope you find a way to fix this problem. Let me know if I can help in any way.

mariofusco:
This is fixing the symptom, and overriding as a system property is not ideal, and would still cause problems with rules generated with the same name. We need to find a better solution.

PR #166, opened 2012-11-19
problem with temporal operator after with event with set duration

izakova:


mariofusco:
I rewrote this test in a bit more concise way, reproduced the bug and fixed it

PR #168, opened 2012-12-14
jBPM compliant with Google Application Engine

josecarloscr:
Required changes in drools to deploy and run in Google Application Engine


josecarloscr:
I want to remark this contribution has been done thanks to everis Spain S.L.

mdproctor:
We'd like to get AppEngine working, but this PR will need some more work first.
1) Avoid "if" statements in multiple places, instead create a helper class that uses a delegate to centralise the method calls in one place. That way there is only one "if" on setting the correct delegate. Also we believe there are other places the code is still broken for AppEngine. I'd also like to be sure of behaviour from switching use of getCanonicalName, to be sure it's correct in all our use cases - and not just a blind method change.

2) Separate the cmdScopedEntityManger, I suggest do a different PR for that. We'd like to discuss and understand this more before applying.

mrietveld:
This is great! I'll be looking at this to see how we can incorporate it.

mrietveld:
Huh, this has been around for 11 months.. :/ Time to retest GAE + KIE for 6.0.x.

mrietveld:
@psiroky Please close this -- this is a good start, but there are still some TODO's in the code. 

Also, when we add functionality for GAE, we should add tests, as described here: 
- https://cloud.google.com/appengine/docs/java/tools/localunittesting

psiroky:
OK, closing this one as per previous comment.

PR #169, opened 2012-12-21
DROOLS-5 Named Consequence: compilation error if used after 'not' pattern

esteban-aliverti:
- Added failing test case


mariofusco:
Test case imported in a slightly different format

PR #171, opened 2012-12-27
JBPM-3512 - Allow developer-defined correlationId in addition to process...

mswiderski:
...InstanceId
pull requests (2/3 as it goes through kie-api, drools-\* and jbpm-\* modules) that allows to assign user defined business key while starting process and then to get access to it using same business key, including bam data. Business key should be unique on runtime meaning only one instance can have given business key however bam method returns list for the same business key as there might be a need to have same business key used in different process engines, etc


mswiderski:
closing this pull request as this will be slightly redesigned to be prepared for extension

PR #173, opened 2013-01-25
test case for JBRULES-3723 IDEOGRAPHIC SPACE (U+3000) in DSL fails

tkobayas:
'IDEOGRAPHIC SPACE' (U+3000) in DSL fails
https://issues.jboss.org/browse/JBRULES-3723

This pull request includes only a test case.

As a fix, I think adding '\u3000' to DSLMap.g and generating antlr artifacts would be fine but not 100% sure.


mariofusco:
I merged this test case and fixed this issue

PR #174, opened 2013-01-25
JBPM-3785: Exception handling

mrietveld:
What this code does is introduce a mechanism to have the drools/jbpm engine(s) automatically react to (java) exceptions thrown by the underlying ksession. 

A new Interceptor class is introduced, as well as modifications to the SingleSessionCommandService in order to add the interceptor to the commandService interceptor chain within the transaction. 

(Having the interceptor act outside of the transaction is ineffectual because of the effects of the transaction commit/synchronization logic). 

This code only introduces the framework for this mechanism -- no drools-specific implementation is added here. 

A jbpm-specific implementation can be found here: https://github.com/droolsjbpm/jbpm/pull/150


mrietveld:
We ended up choosing a different way to approach this problem.

PR #176, opened 2013-02-07
[DROOLS-25] Add option to preload unmarshallstream before parsing header

magnusvojbacke:
Protobuf enforces a size limit for messages when unmarshalling from a stream.

see: https://issues.jboss.org/issues/DROOLS-25

When you marshall a large knowledge base in Drools, the messages tend to become huge. It then becomes impossible to later unmarshall them.

However, when you unmarshall Protobuf messages from a byte array, the message size limit is not enforced by Protobuf. 

As a workaround the size limitation impact on Drools, I've added the option to first preload the source stream into a byte array when unmarshalling a knowledge base.

I added the option as a static public boolean field in the ProtobufInputMarshaller class. The function is specific to the protobuf marshalling implementation. I can't see any place for it in any of the more generic classes or interfaces.

### DIFF
--- a/drools-core/src/main/java/org/drools/marshalling/impl/PersisterHelper.java
+++ b/drools-core/src/main/java/org/drools/marshalling/impl/PersisterHelper.java
@@ -16,7 +16,9 @@
 
 package org.drools.marshalling.impl;
 
+import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


etirelli:
Applied. Thanks.

PR #178, opened 2013-02-27
JBPM-3832 - Provide API method to delete session from db

mswiderski:
pull request to allow deleting sessions via api when no loner needed - 2 of 3 parts (kie-api, drools and then jbpm)


mrietveld:
I went through the code, Maciej, it looks good. Doing persistenceContext.remove( sessionInfo ) within the transaction covers all of the possible problems.

The only remotely possible problems I can foresee are ones that would have been caused by users misusing the ksession (two instances of the same ksession, for example), and those are then the users fault in that case, not the code.

mswiderski:
merged into master, thanks Marco for looking into this

PR #179, opened 2013-02-28
Replaced jexcelapi with Apache POI

frommeyerc:
I replaced jexcelapi with Apache POI to enable parsing of .xlsx files.
The current effort only includes the transition. No new features or other improvement done yet. Test are green.


mariofusco:
I fixed the pom and merged this.

PR #180, opened 2013-03-04
Adding test for BZ-917720

yurloc:
The test case demonstrates a combination of DSL configuration and DSL rule under which DefaultExpander will fall into an infinite loop. The test will not get past line 147 until the bug is fixed. The test has a timeout set so that it doesn't run forever.


mariofusco:
I merged the test case on the master and fixed this bug

PR #181, opened 2013-03-04
DROOLS-65,BZ917787: fixing conditional to allow cache update altough cli...

lazarotti:
...ent is restarted


lazarotti:
When the client is restarted the value for lastRead will be -1 always. So, the conditional:
"lastRead < lastModified && lastRead >= 0 "

... will never be true and the ResourceChangeScannerImpl will never update the local cache. I am not sure why those additional "and "conditional was added but removing it fixes the issue without side-effect.

etirelli:
Applied.

PR #183, opened 2013-03-19
Fixed the null dereferencing to work in both ways i.e. ( String.equalsIg...

marcingrzejszczak:
...noreCase(a!.b) == a!.b.equalsIgnoreCase(String) )


mariofusco:
As discussed in chat, this solution was incomplete, but it served as a very good starting point to develop a complete fix. Thanks a lot.

PR #187, opened 2013-03-29
DROOLS-64 + DROOLS-88

sotty:
DROOLS-64 (Trait modify should preserve type inheritance) + 
DROOLS-88 (Use generics in type declarations' fields)

Comes with two useful general purpose algorithms
- Hierarchy linearization (extracted from a previous, ad-hoc implementation for type declares)
- Incremental Hierarchy encoding for fast sub-sup queries/tests

Should be forward-ported to 6.x

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
@@ -116,6 +116,7 @@ import org.drools.spi.InternalReadAccessor;
 import org.drools.type.DateFormats;
 import org.drools.type.DateFormatsImpl;
 import org.drools.util.CompositeClassLoader;
+import org.drools.util.HierarchySorter;
 import org.drools.xml.XmlChangeSetReader;
 import org.xml.sax.SAXException;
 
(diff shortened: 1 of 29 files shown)


drools-compiler/src/test/java/org/drools/factmodel/traits/TraitTest.java
```diff
@@ -2024,18 +2033,22 @@ public void traitWard( TraitFactory.VirtualPropertyMode mode ) {
...
             assertTrue( core instanceof TraitableBean );
-            assertEquals( 1, ( (TraitableBean) core).getTraitMap().size() );
+            assertEquals( 1, ( (TraitableBean) core)._getTraitMap().size() );
             assertTrue( ( (TraitableBean) core ).hasTrait( Thing.class.getName() ) );
         }
 
     }
 
     @Test
+    @Ignore( "Model negative traits using TMS" )
```
mariofusco: What is a negative trait? Are we dropping any trait's features?

drools-compiler/src/test/java/org/drools/integrationtests/MiscTest2.java
```diff
@@ -1438,4 +1436,73 @@ public void testJitCastOfPrimitiveType() {
...
+                        "declare Bean \n" +
+                        "  map : Map<Foo, List<String>> \n" +
+                        "  num : List<Integer> \n" +
+                        "end \n" +
+                        "\n" +
+                        "rule R\n" +
+                        "when\n" +
+                        "" +
+                        "then\n" +
+                        "Bean b = new Bean( new HashMap<Foo, List<String>>(), Arrays.asList( 10, 20, 30 ) );\n" +
```
mariofusco: Here I tried to replace the HashMap&lt;Foo, List&lt;String&gt;&gt; with a HashMap&lt;Foo, List&lt;Integer&gt;&gt; and it compiles without any problem while it should complain that the 2 types are incompatible. Of course, I understand this is due to generics erasure and probably there's nothing we can do about this. Anyway it is something that has to be considered before to decide if we want to introduce this feature: I am afraid that generics erasure could also cause other inconsistencies that I am not able to see at the moment.

drools-core/src/main/java/org/drools/core/util/ClassUtils.java
```diff
@@ -331,7 +331,7 @@ public static String getPackage(Class<?> cls) {
         }
 
         for (Field f : clazz.getFields()) {
-            if ( !Modifier.isFinal( f.getModifiers() ) && !Modifier.isStatic(f.getModifiers()) ) {
+            if ( Modifier.isPublic( f.getModifiers() ) && !Modifier.isFinal( f.getModifiers() ) && !Modifier.isStatic(f.getModifiers()) ) {
                 props.add( new SetterInClass( f.getName(), f.getDeclaringClass() ) );
```
mariofusco: The javadoc of Class.getFields() says: "Returns an array containing Field objects reflecting all the accessible public fields of the class", so the isPublic check shouldn't be necessary.

drools-compiler/src/main/java/org/drools/compiler/PackageBuilder.java
```diff
@@ -116,6 +116,7 @@
```
etirelli, line 115: It worries me to have parsing code added to the PackageBuilder, as this should be in the parser and as it is, it duplicates code from there. It will make maintenance more challenging in the future as you know. I would rather re-use the parser there or create a type safe "TypeDesc"-like class that the parser instantiates for types instead of using raw strings.

drools-compiler/src/main/java/org/drools/lang/descr/TypeFieldDescr.java
```diff
@@ -35,6 +38,7 @@
     private String                       initExpr;
     private PatternDescr                 pattern;
     private boolean                      inherited;
+    private Set<String>                  genericTypeDeps;
```
etirelli: I am not sure this is correct. The type information is part of the PatternDescr. Generics are part of the type information, so should be handled in there...

drools-compiler/src/test/java/org/drools/integrationtests/MiscTest2.java
etirelli: I am not sure this is a generics erasure issue. Is the constructor being generated with the proper signature? It seems like it is being generated as Bean( Map m, List l ) instead of with the proper generic signature.

drools-core/src/main/java/org/drools/factmodel/BuildUtils.java
```diff
@@ -144,6 +145,30 @@ public static String getTypeDescriptor(String type) {
         return internalType;
     }
 
+    public static String getGenericTypeDescriptor( String genericName ) {
```
etirelli: It seems we are duplicating this code as well.

PR #189, opened 2013-04-10
fixed import package. added needed Private-Package to trick BND.

cvgaviao:


drools-persistence-jpa/pom.xml
```diff
@@ -34,17 +34,16 @@
...
               javax.transaction;resolution:=optional,
               javax.naming;resolution:=optional,
-              org.osgi.framework.*;resolution:=optional,
-              javax.naming;resolution:=optional,
               *
             </Import-Package>
             <Export-Package>
               org.drools.persistence.*
             </Export-Package>
+            <Private-Package></Private-Package>
```
ge0ffrey, line 46: Do we need an empty <Private-Package> on this one specifically or on all configs (so extract it to the parent pom)?

Ping @cmoulliard.

cmoulliard, line 46: Good remark, question Geoffrey. As <Private-package> is only used by BND to
package in a bundle classes which are not defined to be exported, this
option could be moved to parent pom and sed case by case in sub modules

On Fri, Apr 12, 2013 at 8:58 AM, Geoffrey De Smet
notifications@github.comwrote:

> In drools-persistence-jpa/pom.xml:
> 
> > ```
> >            *
> >          </Import-Package>
> >          <Export-Package>
> >            org.drools.persistence.*
> >          </Export-Package>
> > ```
> > -            <Private-Package></Private-Package>
> 
> Do we need an empty on this one specifically or on all configs (so extract
> it to the parent pom)?
> 
> Ping @cmoulliard https://github.com/cmoulliard.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/droolsjbpm/drools/pull/189/files#r3767567
> .

## 

Charles Moulliard
Apache Committer / Sr. Enterprise Architect (RedHat)
Twitter : @cmoulliard | Blog : http://cmoulliard.blogspot.com

cvgaviao, line 46: @ge0ffrey : Not all bundle needs that empty private line. Just the ones that need to exports packages that by default BND consider private (contains impl, private, etc).

If you add it to parent, you will never will have the chance to hide any package.

ge0ffrey, line 46: agreed

PR #190, opened 2013-04-18
added fallback lookup names for user transaction and transaction manager

mswiderski:
here is a extension to the look up mechanism for UserTransaction and TransactionManager in case it is not found in the default locations. 
This is mainly valid for jBPM as it can require access to transaction and transaction manager when application server does not allow access to default bound instances, for example when executing after timer expiration or on application server startup - these are considered non managed thread executions and thus no access to default bound names. 
Different application servers can expose different namespaces in JNDI and that's why it can be given as system property for the jvm.

Currently using jbpm prefix of the system property because it was already introduced in 5.x so just to keep backward compatibility.


mswiderski:
merged to master

PR #194, opened 2013-05-04
More bug fixes

sotty:


sotty:
See twin PR in droolsjbpm-knowledge

PR #196, opened 2013-05-13
[DROOLS-111] Corrected delegated invocation in before methods of AgendaEventListenerWrapper

AlesD:
I'm sending bugfix for issue I had previously reported.


AlesD:
I'm not familiar with Git and/or drools project release plan. Maybe this should go to other branch.

PR #198, opened 2013-05-14
DROOLS-139: Infinispan based Drools and jBPM persistence

marianbuenosayres3:
I’ve worked on an infinispan based persistence scheme for drools objects.
The whole idea is to give both a possibility for Infinispan users to have a way to persist these contents in something different than JPA, as well as to give an example of how to build your own persistence scheme for drools and jbpm objects.


mrietveld:
Marian, would you mind submitting a pull request with both modules to droolsjbpm-integration? I talked with Geoffrey briefly about this, and we'd rather put the modules there. As soon as you submit that pull request, I'll be glad to merge it.

Thanks!

PR #199, opened 2013-05-15
Adds a .signalEvent(type, event) and .signalEvent(type, event processIns...

mrietveld:
...tId) method to the (drools) WorkItemManager and impl's of that.
- This would help with exception handling in jBPM. It would allow processes to be able to trigger other parts of the process when a WorkItemHandler fails, which is even a part of the BPMN2 spec.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/command/impl/CommandBasedStatefulKnowledgeSession.java
+++ b/drools-core/src/main/java/org/drools/core/command/impl/CommandBasedStatefulKnowledgeSession.java
@@ -183,6 +183,18 @@ public class CommandBasedStatefulKnowledgeSession
                 public void internalExecuteWorkItem(WorkItem workItem) {
                     throw new UnsupportedOperationException();
                 }
+
+                @Override
+                public void signalEvent(String type, Object event) {
+                    SignalEventCommand command = new SignalEventCommand(type, event);
+                    commandService.execute(command);
+                }
+
+                @Override
+                public void signalEvent(String type, Object event, long processInstanceId) {
+                    SignalEventCommand command = new SignalEventCommand(processInstanceId, type, event);
+                    commandService.execute(command);
+                }
             };
         }
         return workItemManager;
--- a/drools-core/src/main/java/org/drools/core/process/instance/WorkItemManager.java
+++ b/drools-core/src/main/java/org/drools/core/process/instance/WorkItemManager.java
@@ -31,5 +31,9 @@ public interface WorkItemManager extends org.kie.api.runtime.process.WorkItemMan
     WorkItem getWorkItem(long id);
 
     void clear();
+    
+    public void signalEvent(String type, Object event);
+    
+    public void signalEvent(String type, Object event, long processInstanceId);
 
 }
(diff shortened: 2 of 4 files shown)


mrietveld:
Step AWAY from the merge button NOW! Move your mouse pointer away, take your hand off of the mouse and do something else with it, like maybe give yourself a light slap. What you can do is add my repo as a remote to your local remote, fetch my repo, rebase my branch onto master if necessary, compile, cherry-pick and push. But do NOT use that merge button!

mswiderski:
Marco, what would call singleEvent() on work item manager in case of exception? Trying to understand what does that and why it does not have access to ksession/kruntime directly.

P.S.
Nice comment about the merge button :)

mrietveld:
See [SignallingTaskHandlerWrapper](http://github.com/droolsjbpm/jbpm/blob/master/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/SignallingTaskHandlerWrapper.java) in jbpm-bpmn2. The way it's currently set up won't work with persistence (because persistence closes the runtime while the SignallingTaskHandlerWrapper holds a reference). See the tests for the SignallingTaskHandlerWrapper in jbpm-bpmn2 as well. ( testErrorSignallingExceptionServiceTask ).  

P.S. Thanks. :) I thought Mark or Edson made a statement about that (merge bubbles), but no one seems to have read that.

mswiderski:
Just looked at the SignallingTaskHandlerWrapper and I don't see a reason why persistence would make a difference here. On transaction completion we disconnect process instance from the kruntime but kruntime (ksession) is still active as long as you don't call dispose on it.
This is same case as we had for HumanTaskWorkItemHandlers where we kept ksession in it so it was working fine until dispose of the session. And checking the mentioned test case - it works fine with persistence and without it, here is the test case I tried: https://github.com/droolsjbpm/jbpm/blob/master/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java#L1745

mrietveld:
Okay, thanks Maciej. I think I'll probably close the pull request then. 

I have one last "concern": I'd like to be able to provide a constructor for the [SignallingTaskHandlerWrapper](http://github.com/droolsjbpm/jbpm/blob/master/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/SignallingTaskHandlerWrapper.java) that doesn't require the KieSession/KieRuntime. Given that the WorkItemManager implementations already have a reference to it, it's seems wasteful not to use it. Also, I'm a little bit concerned about the SignallingTaskHandlerWrapper using a different kruntime than what the WorkItemManager contains. 

However, I can't tell if these are unjustified concerns (like my ones about persistence). What do you think?

mswiderski:
that is completely valid point Marco, in my opinion it should always use the same ksession as the work item manager. There is single work item manager per session so best would be to make sure ksession in both places is the same instance to avoid unexpected issues.

mrietveld:
Okay, thanks. I'll merge this pull-request into master then.

mrietveld:
Merged to master.

PR #200, opened 2013-05-21
Added test case for CCE with several KieSessions - BZ 965611

winklerm:
Added reproducer test case for problem with several instances of a KieSession: https://bugzilla.redhat.com/show_bug.cgi?id=965611


mariofusco:
Merged in a slightly more compact way.

PR #201, opened 2013-05-23
Query bug test.

tomasdavidorg:
There is a bug in drl file, when queries are commented.


sotty:
I tried to reproduce it in 5.6 and it seems to work there.. if this is broken in master, could you open a JIRA ticket?
Thanks

tomasdavidorg:
https://bugzilla.redhat.com/show_bug.cgi?id=967441

mdproctor:
as this is just a test. We should probably merge it anyway, even if it works now.

mariofusco:
I couldn't reproduce this problem on the master, so I am assuming this has been fixed by a previous commit. I merged this test case in a more concise way.

PR #202, opened 2013-05-29
don't rollback transaction when session was not found

mswiderski:
avoid rollbacking transaction when session was not found and just throw runtime exception. This will give the caller (for example runtime manager) to take decision on creating new session in such situation. Exception is dedicated to indicate that session was not found and it extends RuntimeException which should rollback transaction anyway if not handled properly.

Especially valid in environments where transaction is managed outside of jbpm/drools control like CDI/EJB.


mrietveld:
+10 -- I ran into exactly this problem when I was testing deployment of the console/wb. Having a clearly labeled exception helps.

PR #203, opened 2013-05-31
NPE caused by null safe dereferencing

tomason:
Added the test to show null safe dereferencing bug


tomason:
Reported as https://bugzilla.redhat.com/show_bug.cgi?id=969420

mariofusco:
Merged and fixed

PR #206, opened 2013-06-03
Update Section-AIKnowledgeRepresentation.xml

stephen-masters:
Corrections to Drools Expert user guide, chapter 1.1 - Artificial Intelligence.

A number of text changes were made and are listed below against their line numbers.

58 - "resigned it academia" replaced with "relegated it to academia"
Guessing that this is the intended phrase.

74 - "there is renaissance" replaced with "there is a renaissance"
Renaissance is a noun.

78 - "which she refers to as a spring" removed. Redundant as the title, which is quoted contains the word "spring".

80 - "over come" replaced with "overcome"

81 - "it's" replaced with "its"
Not a contraction.

82-88 - Converted to blockquote to match style of such block quotes elsewhere, and style the attribution consistently. I haven't checked whether this makes the generated document look correct though, as I'm not sure how to generate it. 

100 - "to effectively use those rich semantics" replaced with "to use those rich semantics effectively"
Classic split infinitive.

126 - "Although each year algorithmic advances try to" replaced with "However, algorithmic advances continue to"
Although/However is a correction, but the rest of the sentence is intended to improve the clarity of the wording.

144 - No changes, but I feel that this ought to be altered. It switches into the first person, without ever making it clear who that person is. Perhaps adjust this to an attributed block quote?

162 - Needed "a"

192 - "VB" replaced with "Visual Basic"
Most developers are likely to recognise the abbreviation, but given that acronym is not explained anywhere else, it's best to use the full text.

225 - Needed a space.

231 - Removed ";" which was sitting next to a full stop. Also removed reference to Leaps as an implemented algorithm. The "Leaps" text had already been removed, but not the index link.

274 - Just needed a space.

### DIFF
--- a/drools-docs/drools-expert-docs/src/main/docbook/en-US/Chapter-Introduction/Section-AIKnowledgeRepresentation.xml
+++ b/drools-docs/drools-expert-docs/src/main/docbook/en-US/Chapter-Introduction/Section-AIKnowledgeRepresentation.xml
@@ -55,7 +55,7 @@
     venture for Prolog, and many of the failures are pinned on the problems
     of trying to run a logic based programming language concurrently on multi CPU
     hardware with effective results. Some believe that the failure of the 5GP
-    project tainted Prolog and resigned it academia, see <link
+    project tainted Prolog and relegated it to academia, see <link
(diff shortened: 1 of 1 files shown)


mdproctor:
Thank you for the detailed update, this is now applied.

PR #207, opened 2013-06-04
Corrected names of Drools book authors in recommended reading

stephen-masters:
Michal Bali and Paul Browne were both suffering from mis-typed names.


mdproctor:
Now applied. Thank you.

PR #208, opened 2013-06-05
Reproducer test for bugzillas 971001 971004 971008

tomason:
Bugs related to inline casts


mariofusco:
Merged in another PR https://github.com/droolsjbpm/drools/pull/197

PR #210, opened 2013-06-07
Added reproducer for BZ971668 - CSV not included in KieBase

winklerm:
Added reproducer test for https://bugzilla.redhat.com/show_bug.cgi?id=971668.

Test testNonEmptyKieBaseWithCSV fails, while testNonEmptyKieBaseWithXLS passes.


mariofusco:
Already fixed

PR #211, opened 2013-06-07
Drools Expert manual - Chapter 2 corrections

stephen-masters:
See individual commits for specific change explanations.


mdproctor:
thank you, we have applied all these changes now.

PR #212, opened 2013-06-11
Delete of fact - bug test.

tomasdavidorg:
There is a bug, when two queries with accumulate are defined in drl file and fact is deleted from ksession.


tomasdavidorg:
https://bugzilla.redhat.com/show_bug.cgi?id=973180

PR #213, opened 2013-06-11
Added reproducer for star import in DRL problem

winklerm:
Added reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=973264.


mariofusco:
Merged and fixed

PR #215, opened 2013-06-14
DROOLS-167: test case for Inaccurate comparison because of String to Dou...

tkobayas:
Adding a test case for DROOLS-167 : Inaccurate comparison because of String to Double coercion


mariofusco:
This is a variant of this issue https://issues.jboss.org/browse/DROOLS-68 that I already reproduced with this test case https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/test/java/org/drools/compiler/integrationtests/MiscTest2.java#L1496

I merged your test case anyway. Thanks for having reported this

PR #216, opened 2013-06-20
Added reproducer for KieBase inclusion problem

winklerm:
Added test with reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=976498. The problem did not appear in 6.0.0.Beta3.


mariofusco:
merged and fixed

PR #217, opened 2013-06-25
https://issues.jboss.org/browse/DROOLS-179

manstis:
**\* Don't merge - For information only (causes circular dependency) ***

Test demonstrating problem.

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -145,7 +145,14 @@
       <groupId>org.jboss.weld.se</groupId>
       <artifactId>weld-se-core</artifactId>
       <scope>test</scope>
-    </dependency>     
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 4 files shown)


mrietveld:
This is probably a redundant comment (it looks like there was mention of this on IRC?), but this introduces a circular dependency (drools-compiler -> jbpm-bpmn2 -> jbpm-flow-builder -> jbpm-flow -> drools-compiler). Personally, I'd add it to jbpm-bpmn2.

manstis:
I suspected as much. I'll move to jbpm-bpmn2 a little later this morning.... unless somebody suggests somewhere better between now and then. Thanks for commenting on this PR (fortunately it's not merged, so no harm done AFAIK).

manstis:
See https://github.com/droolsjbpm/jbpm/pull/169

PR #218, opened 2013-06-27
Added test for negative patterns in Fusion

winklerm:
Added reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=978979. Might be a regression, or a consequence of switching to Phreak (if the switch had already happened?).


mariofusco:
This test case has been fixed and merged.

PR #219, opened 2013-07-03
Added test reproducing problem with queries using temporal operators

winklerm:
Added reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=980775.


sotty:
This test has been added to 6.x, I think that this PR can be closed

PR #220, opened 2013-07-04
Added test for timer-fired accumulating events using fireUntilHalt()

winklerm:
Added reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=981270.


mariofusco:
Test case merged in a slightly different and more compact way.

PR #235, opened 2013-08-15
DROOLS-223 Serializing and deserializing a PackageDescr causes imports to change to import null

grahambo:
- Adding Unit Test to show issue with serialization.


sotty:
Your test and the fix was added to Drools 5.6 and is being PR-ed for 6.x. This PR can be closed, thanks for reporting this.

PR #236, opened 2013-08-20
JBPM-4101 - Retry interceptor on optimistic lock exception

mswiderski:
Adds implementation of the Interceptor that allows to retry command execution in case optimistic lock exception has been thrown. Allows configuration of the actual exception class that indicates optimistic lock exception and defaults to hibernate's org.hibernate.StaleObjectStateException.

It's not turned on by default so it needs to be added whenever is needed.

Additionally logic of SingleSessionCommandService has been moved to TransactionalInterceptor to ensure retry will be done on new transaction.

Once merged retry interceptor will be used by default by jBPM runtime manager with persistence enabled


mswiderski:
merged

mrietveld:
Apologies, Maciej, I had skimmed it but not yet gotten around to giving feedback. :/

mswiderski:
Marco, no worries, but please look at it if you find the time. You input is always valuable

PR #238, opened 2013-08-21
drools-compiler: added test for different KieSession behaviour based on its construction

winklerm:
Added reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=999388.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieSessionConstructionTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieSessionConstructionTest.java
@@ -1,0 +1,148 @@
+package org.drools.compiler.integrationtests;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.After;
+import org.junit.Before;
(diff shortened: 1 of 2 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieSessionConstructionTest.java
```diff
@@ -0,0 +1,148 @@
...
+        // working memory objects list
+        List<Command<?>> commands = new ArrayList<Command<?>>();
+
+        ListHolder listHolder = new ListHolder();
+        kieSession.setGlobal("listHolder", listHolder);
+
+        insertInstances(commands, kieBase);
+
+        // we are in the office
+        FactType hereType = kieBase.getFactType(PACKAGE, "Here");
```
mariofusco, line 98: The 2 statements above are wrong: when running testBasicQuerySessionFromContainer you're using a kieBase that is not the one from which the kieSession has been created. If you replace kieBase with kieSession.getKieBase() both tests become green.

mariofusco:
This test is wrong for the reason explained in the inline comment.

PR #241, opened 2013-08-23
DROOLS-236 NPE when ksession is serialized with facts logically inserted

tkobayas:
Tast case and fix for DROOLS-236.

This fix is to make sure that only the same class fact will create its EqualityKey. I'm not 100% clear if it is the right way to fix. Please review.


sotty:
Edson is working on a more appropriate fix. Thanks for reporting this!

sotty:
The unit test was incorporated and the bug was fixed with a more robust solution in both 5.x and 6.x. This PR can be closed. Thanks!

krisv:
Closed as requested

PR #243, opened 2013-08-27
Drools 243

calcacuervo:


sotty:
Demian, could you please check and apply this fix to all the custom Evaluators? 
I remember soundsLike, isA and str, but there may be others

calcacuervo:
Yes! There are more.

But I could only reproduce this issue with the CEP evaluators. As this
happens when applying the changeset, that tries to calculate the temporal
distance. That's why I did not applied to the others.

This night I will do some more testing in the other operators and will let
you know, and add another commit if neccesary.

Thanks Davide!

Demian

calcacuervo:
Hey Davide. I made another commits applying the same to other operators.

I was trying to think how to make a test to reproduce it.. but I think it is not easy. One could be put a precompiled pkg in test classpath, and try to unmarshall it. But this will be not easy to maintain.

Any idea?

Thanks,

Demian

sotty:
Demian, did you commit your second part of the fix? 
I know it's hard to test too, but as long as all the evaluators follow the same patter we should be fine.

sotty:
Sorry, apparently a browser glitch

sotty:
Did you forget to include the SetEvaluatorsDefinition? 
Another small issue is that now the same Operator may be added multiple times to the registry, e.g. when "soundsLike" has more than one class providing an implementation

PR #247, opened 2013-09-09
5.6.x

grrob:


sotty:
Was this PR opened by mistake? The code from 5.6 can't be applied directly to master (6.x), it has to be forward ported manually, and only if it applied.

PR #250, opened 2013-09-12
Jbpm 3814 6.0.x

mrietveld:
The solution for both JBPM-3814 and BZ 1002724.


mrietveld:
Discussed and decided to pursue a slightly different approach.

PR #251, opened 2013-09-12
Calendar test

tomasdavidorg:
Calendars don't work, when we work with two KieContainers.


tomasdavidorg:
https://bugzilla.redhat.com/show_bug.cgi?id=1007385

mariofusco:
This issue was caused by a serialization bug that has been already solved. The test case demonstrating that is working now has been merged.

PR #253, opened 2013-09-14
Drools 260

calcacuervo:
DROOLS-260. Fix and test.


sotty:
Great catch! I have cherry-picked it

triceo:
Does this make sense to apply to 6.x as well?

mdproctor:
yes, this should be applied to 6.0 too

mdproctor:
Not sure why this is open. The code seems to be applied to 5.6 and master (6.x). Please confirm and close.

etirelli:
Yes, this was applied. Closing it.

PR #254, opened 2013-09-16
Cherry-pick to master: indented keywords would fail template expansion

magnusvojbacke:
There's a limitation / bug in templates that they cannot be parsed if they contain indented keywords

Forum link where this issue was discussed: http://drools.46999.n3.nabble.com/Fwd-drools-Fixing-issue-where-indented-keywords-would-fail-template-expansion-170-td4021476.html   

This is already fixed in 5.5.x by pull request https://github.com/droolsjbpm/drools/pull/170

This new pull request proposes to cherry-pick this fix to master


manstis:
See http://github.com/droolsjbpm/drools/commit/0bd98541b.

Thanks.

PR #256, opened 2013-09-17
JBPM-3814: Fixing concurrency problems when using a singleton KieSession in a multi-threaded environment

mrietveld:
This pull-request is related to a discussion I've been having with Kris and Maciej. 

The pull-request is only to showcase the progress I've made so far, in order to get feedback from them. 

Please do **not** merge this pull-request.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/impl/EnvironmentImpl.java
+++ b/drools-core/src/main/java/org/drools/core/impl/EnvironmentImpl.java
@@ -16,14 +16,22 @@
 
 package org.drools.core.impl;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.kie.api.runtime.Environment;
+import org.kie.api.runtime.KieSession;
 
 public class EnvironmentImpl implements Environment {
 
-    private Map<String, Object> environment = new HashMap<String, Object>();
+    // The default concurrencyLevel is 16: if users have enough threads
+    // that (16^(1.5))=64 *concurrent* updates are possible/likely
+    // then the concurrencyLevel will need to be upgraded,
+    // but that situation is fairly unlikely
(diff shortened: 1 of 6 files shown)


mswiderski:
Marco, looks good to me, good to see that locking is independent of the transaction ownership as that makes the case more visible and once it's turned on it is turned on regardless of transaction control.

One thing I would like to ask is why UserTransaction has been removed from JtaTransactionManager? Isn't TransactionManager class more dedicated for application server and UserTransaction is dedicated for application use? Not sure that it will cause any issues but there might be some application servers that will not give access directly to TransactionManager but only to UserTransaction especially from non managed threads (timers, async jobs, etc).

how does that perform in real case, does it solve the issues you identified?

mrietveld:
With regards to UserTransaction: I talked to the jboss TS/JTA people back in March. In the brief conversation I had with them, the following came up: 

```
<mriet> how do different threads differentiate among the possibly multiple active tx's then? (when trying to join them, then)
<jhalliday1> huh? a thread either starts a tx from scratch, in which case it need no info, or it gets given a tx object explicitly and calls resume with it to join the existing tx.
<mriet> In other words, the only way for one thread to join another thread's tx is through an reference to an instance of the transaction -- correct?
<mriet> (this is JTA --> java, so we're talking about an instance of the UserTranscation, I assume)
<jhalliday1> yes
```

Actually, now that I reread that, I see that Jonathan (Halliday) actually mentions that the other thread needs to call resume in order to join the tx, which we don't do. 

I'll change it back to how it was. The code that I have in the pull request allowed me to (easily) verify that threads weren't sharing transactions. When the UserTransaction instance was being shared, it was hard to make sure that threads weren't sharing the same tx (I know, the spec doesn't allow it, but I just wanted to make sure.. :) ).

mrietveld:
I need to change the JpaPersistenceContextManager so that it's `cmdScopedEntityManager` field is (configurably) a ThreadLocal, since sharing that EM between threads causes problems for timer jobs.

PR #261, opened 2013-10-01
BZ-1014157-kie-ci does not resolve properties that are used for artifact...

mswiderski:
... version - test case/reproducer


mariofusco:
Merged

PR #262, opened 2013-10-02
Minor bugfix for ignored default encoding in ReaderResource

khennig:
Code changes prevent overwriting default encoding determined from InputStreamReader in ReaderResource constructor.


mdproctor:
Any chance of a unit test with the commit, so that we do not have a regression?

khennig:
… unit test, yes. Hold on please.

etirelli:
Applied! Thanks for the test case and fix!

PR #268, opened 2013-10-15
DROOLS-280 Active on lock test 

apat59:
testcase for https://issues.jboss.org/browse/DROOLS-280

lock-on-active not Active on lock, sorry


mariofusco:
Fixed

PR #269, opened 2013-10-16
Query cep entry point

tomasdavidorg:
NPE when ksession.getQueryResults() should return empty result.


tomasdavidorg:
https://bugzilla.redhat.com/show_bug.cgi?id=1019708

etirelli:
PR applied. Thanks!

http://github.com/droolsjbpm/drools/commit/315cc8f4b

PR #279, opened 2013-11-06
BZ1027221 When Guvnor is down, KnowledgeAgent keeps rebuilding kbase fro...

tkobayas:
This fix makes sure UrlResource.lastRead == cacheFile's lastModified.

Verified with manual test.


mariofusco:
Merged

PR #287, opened 2013-11-25
DROOLS-352

sotty:


mariofusco:
Merged without [DROOLS-351] Test case (already reproduced and fixed by a former commit),

PR #289, opened 2013-11-25
BZ1034030 : 0 byte local cache file in UrlResource

tkobayas:
This fix defers cache file update until completely reading InputStream. So the "0 byte" risk would be minimized.


etirelli:
Applied to 5.3.x, 5.6.x, and master. Thank you!

PR #291, opened 2013-11-26
BZ1034588 : stale local cache file in UrlResource

tkobayas:
This fix prevents UrlResource.getInputStream() from reading stream twice.


etirelli:
PR applied to 5.6.x and master. Thank you!

PR #294, opened 2013-12-05
added reproducer for BZ 1038627

winklerm:
Added reproducer for (possible) bug https://bugzilla.redhat.com/show_bug.cgi?id=1038627. Please review, if this is an intended change in behaviour or a bug.


winklerm:
Closed because the new behaviour has been explained.

PR #295, opened 2013-12-06
[DROOLS-363] Test Case

sotty:


mariofusco:
Declarative Agenda fix merged

PR #297, opened 2013-12-12
Test case for DROOLS-376 NullPointerException on JPAKnowledgeService.loa...

tkobayas:
...dStatefulKnowledgeSession() when Collection is modified.


mrietveld:
+1 Checked out and quickly looked at this. 

The NPE is being caused by this [line](https://github.com/droolsjbpm/drools/blob/6.0.x/drools-core/src/main/java/org/drools/core/reteoo/FromNode.java#L152) in `FromNode`. 

My suspicion is that the actual fix will be to modify ProtobufOutputMarshaller (to "more" correctly serialize a right tuble). 

This looks like a problem for Edson or Mario.. i don't know enough yet about this part of drools (with regads to tuples/nodes, etc.)

mariofusco:
Test merged and bug fixed.

PR #304, opened 2014-01-07
reproducer for BZ1044577

tkobayas:
This test case constantly reproduce the ArrayIndexOutOfBoundsException on my side (It shows up in STDOUT instead of surefire report log). If you can't reproduce yet, please try increasing the number of threads.


mariofusco:
Test merged and issue fixed

PR #307, opened 2014-01-24
BZ1057445 : test case

tkobayas:
A test case for BZ1057445 : isGetter for a boolean property with I18N name is not detected.

I have confirmed that MVEL-300 fixes this issue as well.


mariofusco:

PR #308, opened 2014-02-10
added testFineGrainedPropertyChange

ergh99:
added testFineGrainedPropertyChange to demonstrate a scenario that works in 5.6.0.Final but not in 6.0.1.Final

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertyReactivityTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertyReactivityTest.java
@@ -2,12 +2,14 @@ package org.drools.compiler.integrationtests;
 
 import org.drools.compiler.CommonTestMethodBase;
 import org.drools.core.factmodel.traits.Traitable;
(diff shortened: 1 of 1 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertyReactivityTest.java
```diff
@@ -1084,4 +1086,55 @@ public String getFullName(){
...
+    	Person joe = new Person();
+    	int startingWeight = joe.weight;
+    	StatelessKieSession kSession = kbase.newStatelessKieSession();
+    	//KieRuntimeLogger log = KieServices.Factory.get().getLoggers().newFileLogger(kSession, "drools-audit");
+    	kSession.execute(joe);
+    	//log.close();
+    	// higher salience rule "Set Joe's Age" should fire
+    	assertEquals(100, joe.age);
+    	// the lower salience rule should NOT fire
+    	assertEquals(startingWeight, joe.weight);
```
sotty, line 1138: Actually both rules are expected to fire

PR #309, opened 2014-02-11
Added DynamicImportPackage:* to fix OSGi class loading issues

fabianmeyer:
Due to OSGi's class loading mechanism, drools-core is not able to load classes exported by other bundles when compiling rules, which leads to a class loading exception. "DynamicImport-Package: *" fixes this issue.

### DIFF
--- a/drools-core/pom.xml
+++ b/drools-core/pom.xml
@@ -181,6 +181,9 @@
                org.jbpm.process.instance.event;resolution:=optional,
                *
             </Import-Package>
+            <DynamicImport-Package>
+              *
+            </DynamicImport-Package>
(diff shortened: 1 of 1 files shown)


cmoulliard:
This modification should be a workaround and removed in the future. Why : OSGI container will scan all the bundles to find the package (not loaded when the bundle has been started) till it discover it. That could impact performances, does not guarantee which package will be used if they are exposed by different bundles ...

fabianmeyer:
An option to configure the class loader used by the KieBuilder and the KieRuntime could solve the problem too.

mariofusco:
It is now possible to explicitly pass a custom classloader to the KieBuilder.
See https://issues.jboss.org/browse/DROOLS-763

PR #311, opened 2014-02-14
Added logging for more useful error messages

ergh99:
the only message returned by the output stream when passed a null arg is "null".  this should provide better debugging/error trapping

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
@@ -46,6 +46,9 @@ public class MemoryFileSystem
 
     private Set<String>                modifiedFilesSinceLastMark;
 
+    private final org.slf4j.Logger log =
+    		org.slf4j.LoggerFactory.getLogger(MemoryFileSystem.class);
+
     public MemoryFileSystem() {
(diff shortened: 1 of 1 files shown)


psiroky:
ok to test

psiroky:
I believe this could be merge after some minor indenting changes (replacing tabs with spaces). @mariofusco @etirelli what do you think?

psiroky:
Looking at the changes again, maybe we could lower the log level to WARN when the IOException is thrown during `out.close();`. It is not critical issue, but may indicate that the system is not in the best state. I believe we should not swallow the exception entirely.

etirelli:
+1

drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
```diff
@@ -437,6 +441,11 @@ private void writeJarEntries(Folder f,
                 out.putNextEntry( entry );
 
                 byte[] contents = getFileContents( (MemoryFile) rs );
+                if (contents == null) {
```
mariofusco, line 444: I have no idea of how this could be null. Do you have a reproducer for this?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/ffe938d1b

PR #312, opened 2014-02-17
[BZ-1034030] 0 byte local cache file in UrlResource

tkobayas:
- Stream close and Exception handling in cacheStream()


tkobayas:
Please wait for applying this PR. I think I have to raise another PR for cacheStream() for Windows and this Exception handling PR also has a defect (renameTo() before close() doesn't work for Windows). I'll fix this after the 'another PR'.

mariofusco:
I fixed this issue with this commit https://github.com/droolsjbpm/drools/commit/e9f2169e6

PR #313, opened 2014-02-17
DROOLS-430 Fails to parse DSL variable definition with single dot regex

tkobayas:
- Test case


mariofusco:

PR #316, opened 2014-02-21
BZ-1067841 Compiled resource objects leak in KnowledgeAgentImpl.register...

tkobayas:
- Test case


mariofusco:
KnowledgeAgent is no longer supported

PR #317, opened 2014-02-27
[DROOLS-444] NOT conditional element is broken when used with subnetwork...

sotty:
...s
Reproducer test cases


mariofusco:
All the commits of this PR have been merged except the first (the NOT issue has been reproduced with a simpler test and fixed) and the last (we decide to take no action for that deadlock issue and just warn the users that at the moment it is not possible to perform a marshalling inside and event).

PR #318, opened 2014-03-02
A fix for compiler.PackageBuilder.getFullTypeName.

jimwhite:
Nearly all tests (148 of 151) in drools-pmml fail for me (Mac OS X Mountain Lion - JDK 1.6) with error messages like this:

```
testFunctionsNested(org.drools.pmml.pmml_4_1.transformations.UserDefinedFunctionsTest)  Time elapsed: 0.604 sec  <<< ERROR!
java.lang.NoClassDefFoundError: org/dmg/pmml/pmml_4_1/descr/false (wrong name: org/dmg/pmml/pmml_4_1/descr/False)
    at java.lang.ClassLoader.defineClass1(Native Method)
```

By changing the 'false' value in a declare in pmml_compiler.drl to 'False' all tests but one (an apparently unrelated issue in SVMTest) pass.  I haven't any idea if this is the right way to fix this but thought I'd pass it along to save time in the event someone else runs into this.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/PackageBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/PackageBuilder.java
@@ -3208,6 +3208,8 @@ public class PackageBuilder
             return typeResolver.getFullTypeName(type);
         } catch (ClassNotFoundException e) {
             return type;
+        } catch (NoClassDefFoundError cnf) {
+            return type;
         }
     }
 


sotty:
The problem is actually before then. Simple literals such as "true" or "false" should not be looked up as potential class names. In general it's wasted work, but the contemporary presence of a non-case sensitive file system and a class called False triggered the problem. Thanks for reporting the problem

jimwhite:
The problem was that compiler.compiler.PackageBuilder.getFullTypeName only tries to catch ClassNotFoundException for this situation but this code was causing NoClassDefFoundError.  

I don't think it is a filesystem case-sensitivity issue, but perhaps JDK?  I'm using 1.6.  Oh, yeah, I get it.  Finds the class file then runs into trouble since the names don't match.

Anyhow, I've fixed my code to work as I think it was intended to.

sotty:
Yes, "it finds the class but names don't match". With literals, it should not even get there:
https://github.com/sotty/drools/commit/4ddfb88e757c65da1cb1d8517499d1eabce112e1

jimwhite:
Yeah, I get that.  The expression needs to be parsed by the appropriate expression parser, but there doesn't seem to be one around here.  I'd be inclined to keep the NoClassDefFoundError check since I don't see what it could hurt and chasing some other problem that might pop up would be a waste.

PR #320, opened 2014-03-10
improved KieCDIExtension to allow to inject into Instance<> injection po...

mswiderski:
...ints


mswiderski:
closing this as will prepare PR for master with dedicated BZ for it

PR #323, opened 2014-03-12
[Drools-450] Cannot use decimal formatters for integers in an excel decision table

mfalaize:
Force locale of DataFormatter to use english formatting to avoid number format errors.


manstis:
PR's commits squashed and applied: http://github.com/droolsjbpm/drools/commit/288824d1d

Thanks for reporting, fixing and providing a test.

PR #328, opened 2014-04-07
BZ1081848 KnowledgeAgent keeps rebuilding kbase from cached files when G...

tkobayas:
...uvnor is down and a client restarts

Test case and fix.


mariofusco:
Fix merged manually

PR #330, opened 2014-04-29
Pull request for kie module deployment builder

mrietveld:
I developed this code for the `kie-remote` modules in the 6.0.x branch, because it was very helpful when unit testing: I could programmatically call the `KieModuleDeploymentHelper` instance to create a kjar and then use the REST/JMS services to deploy it. 

However, for obvious reasons, the code doesn't belong in the (`droolsjbpm-integration/kie-remote/kie-services-client`) but I'd like to keep it, since I think it would be useful. 

If the PR is approved, I'll add documentation as well (to `kie-docs`). 

The 6.0.x code (in `kie-services-client`) can be found [here](https://github.com/droolsjbpm/droolsjbpm-integration/tree/6.0.x/kie-remote/kie-services-client/src/main/java/org/kie/services/client/deployment).

### DIFF
--- a/kie-ci/pom.xml
+++ b/kie-ci/pom.xml
@@ -71,6 +71,19 @@
       <artifactId>httpcore</artifactId>
     </dependency>
 
+    <!-- test --> 
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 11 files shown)


mrietveld:
Updated PR to remove reference to `org.sonatype.aether.artifact.Artifact` class. 

After the eclipse.aether/maven 3 migration, you may have to change 2 lines in the test (l. 87, l. 172), but that should be pretty easy.

PR #331, opened 2014-04-29
DROOLS-473: Drools spawning a lot of JIT threads

fpavageau:
This pull request does 3 things:
- It replaces the unbounded thread pool by a bounded one (with a configurable bound)
- It adds a name to the threads in the pool
- It replaces the `SynchronizedQueue` used in the unbounded thread pool by a `LinkedBlockingQueue` (with a configurable bound as well) to retain an asynchronous behavior even when all the threads are already active

I'd add the documentation for the system properties configuring the bounds, but I'm not sure where it should go (beyond "somewhere in drools-expert-docs"), if you have any suggestions.


mariofusco:
This issue has been fixed on master with this commit https://github.com/droolsjbpm/drools/commit/f386ceb09 following your suggestions.
Thanks a lot.

PR #333, opened 2014-05-10
PseudoClockDoubleNotTest DROOLS-479

richardambridge:
DROOLS-479

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PseudoClockDoubleNotTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PseudoClockDoubleNotTest.java
@@ -1,0 +1,146 @@
+package org.drools.compiler.integrationtests;
+
+import java.util.concurrent.TimeUnit;
+import org.drools.compiler.Cheese;
+import org.drools.compiler.CommonTestMethodBase;
+import org.drools.core.time.SessionPseudoClock;
+import org.junit.Test;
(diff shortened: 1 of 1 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/PseudoClockDoubleNotTest.java
```diff
@@ -0,0 +1,146 @@
...
+        Thread t = new Thread(new FireThread(ksession));
+        t.start();
+
+        try {
+            //Insert a smelly cheese.  Template 1 should fire after 3 seconds passed
+            Cheese c = new Cheese("smelly");
+            ksession.insert(c);
+            if (doPseudo) {
+                SessionPseudoClock clock = ksession.getSessionClock();
+                clock.advanceTime(10, TimeUnit.SECONDS);
```
sotty, line 112: The problem is actually a race between the FireThread and the main thread. In Drools 6, the rule is evaluated and enabled when "fire" kicks in for the first time. The rule is then scheduled +3s later as soon as the time advances. In your case, when you advance the pseudo clock. 
However, it may be the case that the main thread will advance the clock BEFORE the Fire Thread has activated the rule. If so, the rule will be considered the NEXT time you advance (again) the time. To make this test work, you'll have to make sure that the FireThread is awakened at least once before you move the pseudo clock.

mariofusco:
Reproduced in a slightly different way and fixed

PR #335, opened 2014-05-14
[DROOLS-483] fix the repeated look-up of absent services

fpavageau:
Some services accessed through the `ServiceRegistry` are not provided by
Drools modules but by jBPM, and have a higher probability of not
actually being in the classpath. `ProcessBuilderFactory` and
`ProcessRuntimeFactory` (try to) load such services when respectively
called during the construction of `KnowledgeBuilderImpl` and
`StatefulKnowledgeSessionImpl`. Trying to load a class is an expensive
operation, especially in the context of a web application where the
classloader will scan all the jars. With Tomcat for example, this
operation is moreover synchronized and can lead to lock contention.

There's also the issue of using exceptions for the regular control flow,
because filling an exception's stacktrace is also quite expensive.

`ProcessBuilderFactory` is modified to only try to load the service once,
and then either stores the service or the exception. The exception can
then be rethrown on subsequent accesses, to keep reporting on the failure
to load the service in `KnowledgeBuilderImpl`.

`ProcessRuntimeFactory` is simply modified to try to load the service once
and return null is the service cannot be loaded, simplifying the control
flow in `StatefulKnowledgeSessionImpl`.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/ProcessBuilderFactory.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/ProcessBuilderFactory.java
@@ -8,6 +8,7 @@ public class ProcessBuilderFactory {
 
     private static final String PROVIDER_CLASS = "org.jbpm.process.builder.ProcessBuilderFactoryServiceImpl";
 
+    private static IllegalArgumentException initializationException;
     private static ProcessBuilderFactoryService provider;
 
     public static ProcessBuilder newProcessBuilder(KnowledgeBuilder kBuilder) {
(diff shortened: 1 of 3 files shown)


mariofusco:
I merged what you did and fixed it when running in OSGi. Thanks a lot.

fpavageau:
Great, thanks a lot to you. :+1: 

I have a related question though, if I may: where are the `resetInitialization()` methods called from? Is it configured somewhere in the OSGi manifest (or an OSGi convention so it happens automagically)?

I couldn't find it in drools or droolsjbpm-knowledge.

mariofusco:
The resetInitialization() methods are called in the jbpm osgi activators for those services: https://github.com/droolsjbpm/jbpm/commit/dc8e836e6490a5a9b60097e56968976d481a96bb

Sorry again for the huge delay in merging this :)

PR #337, opened 2014-05-16
Automating version number in protobuf marshalling

mrietveld:
(I had this laying around and figured I should submit it before I forgot about it.. ;) )

The downside of the code is that the replacer-plugin will automatically modify the code after a release. However, that's also kind of the upside. 

Because the `-SNAPSHOT` of the version isn't used (parse-version), the code replacement won't happen until _after_ a release.

### DIFF
--- a/drools-core/pom.xml
+++ b/drools-core/pom.xml
@@ -193,13 +193,51 @@
           <excludePackageNames>org.kie.asm*,org.kie.objenesis.*,org.kie.commons.jci.*</excludePackageNames>
         </configuration>
       </plugin>
-
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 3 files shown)


mariofusco:
I developed a different mechanism for automatic version recognition in this class https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/util/Drools.java 

Also see how I am using it here https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieModuleCacheHelper.java#L41

Is it ok to replace what you did in the PersisterHelper with this?

mrietveld:
Sure, that looks fine. I'm just glad the issue is being addressed. Thanks!

PR #338, opened 2014-05-27
BZ1101471 A rule is not activated with ruleflow-group and 'not' pattern

tkobayas:
-Test case

I'm afraid that Misc2Test may not be the right place. I had to edit pom.xml to make ruleflow work. But the issue seems to exist in Drools side rather than jBPM.

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -26,6 +26,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-bpmn2</artifactId>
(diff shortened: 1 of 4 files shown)


tkobayas:
As additional information, I found strange behaviours in the test. Please take a look at ruleflow_with_not.drl. I left 3 comments in the DRL.

// (A) If you uncommnet this dummy rule, rules will work expectedly

// (B) If you comment-out one or two of these 'not' pattern, rules will work expectedly

// (C) If you set Trailer.TypeStatus.SHIPPING here, rules will work expectedly

So if you make minor modifications to the DRL, the issue goes away. I hope it helps the investigation.

mariofusco:
Reproduced using agenda-groups instead of ruleflow-groups

PR #341, opened 2014-06-02
Added null check for the pkg before trying to get the rule.

mikedev9000:
Prevent NullPointerException within KieContainerImpl.updateToVersion(ReleaseId).


mikedev9000:
Fixes DROOLS-511.

mariofusco:
I haven't understood how to reproduce this problem. Could you please provide a reproducer?

PR #342, opened 2014-06-02
Re-adding a call to setKnowledgeRuntime within StatefulKnowledgeSessionImpl constructor.

mikedev9000:
It seems as though this call was inadvertently removed
as part of commit 1f8dc327. The method makes a call that
results in the ksession's mbean being created and registered.

Without this call, no ksession mbeans are created.


mikedev9000:
Fixes DROOLS-512.

mikedev9000:
I'm closing this PR in order to submit a new one for the 6.2.x branch after resolving conflicts.

PR #344, opened 2014-06-06
BZ-1096128: add reproducer for included and including KieModule change

winklerm:
Adding reproducer for problem with changes in a KieModule which includes KieBase from another KieModule not being picked up. 

I have experienced this behaviour also with workbench and KieScanner.


mariofusco:
Test merged and bug fixed

PR #349, opened 2014-06-25
First round of fixes for FindBugs warnings

psiroky:
PR contains 2 things:
- first is the exclusion of ANTLR-generated classes from the FindBugs analysis, as is does not make sense to fix issues there (generated code) and they would only clutter the final report
- second one is the fix for several dozens of warnings related to default encoding usage. Instead of platform dependent encoding UTF-8 is used.


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
```diff
@@ -250,14 +253,14 @@ private static String getPomPropertiesFromZipFile(String rootPath) {
...
-        FileReader reader = null;
+        Reader reader = null;
         try {
             File file = KieBuilderImpl.findPomProperties( new File( rootPath ) );
             if ( file == null ) {
                 log.warn( "Unable to find pom.properties in " + rootPath );
                 return null;
             }
-            reader = new FileReader( file );
+            reader = new InputStreamReader( new FileInputStream( file ), Charset.forName( "UTF-8" ) );
```
psiroky: Need to use the InputStreamReader, because FileReader does not offer a way how to specify the encoding.

psiroky:
Created constant IoUtils.UTF8_CHARSET and using that instead of Charset.forName("UTF-8"). All tests passed locally.

PR #352, opened 2014-07-03
Drools-543

esteban-aliverti:
Added a test case and a simple solution (Credits go to @sotty)


esteban-aliverti:
The solution is to fast-fail in the case something already went wrong before the rule compilation.

mariofusco:
I reproduced the problem using your test case (that I merged) but I believe the solution you're suggesting is not the right thing to do. We don't want to skip rule compilation if there is any problem with a type declaration because we want to give to user a complete report of the issues present in the drl he is trying to compile. Actually skipping stuff is exactly what is causing this NPE: since type A is not valid it skips compilation of type C causing the NPE during a later type declaration lookup. My fix will then go in the opposite direction, i.e. performing compilation of a type declaration even if a former one had a problem.

esteban-aliverti:
Much better and elegant solution indeed.

PR #353, opened 2014-07-05
Serialization

fbertola:
Serialization for large rule bases.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/mvel/MVELCompilationUnit.java
+++ b/drools-core/src/main/java/org/drools/core/base/mvel/MVELCompilationUnit.java
@@ -162,7 +162,7 @@ public class MVELCompilationUnit
     public void writeExternal( ObjectOutput out ) throws IOException {
         out.writeUTF( name );
 
-        out.writeUTF( expression );
+        out.writeObject( expression );
 
         out.writeObject( globalIdentifiers );
         out.writeObject( operators );
@@ -183,7 +183,7 @@ public class MVELCompilationUnit
     public void readExternal( ObjectInput in ) throws IOException,
                                               ClassNotFoundException {
         name = in.readUTF();
-        expression = in.readUTF();
+        expression = (String) in.readObject();
 
         globalIdentifiers = (String[]) in.readObject();
         operators = (EvaluatorWrapper[]) in.readObject();


mariofusco:
If I am not wrong that limit is 65535 bytes, meaning that in the very worst case (3 bytes for each char) you can have a 21K+ characters long expression. I'd be very surprised if somebody may need to use a so long expression. Do you have an example?

fbertola:
We have a rule that remove stopwords from an object wrapping a string:

```
rule "remove stop words"
when
        $ne1:NE(identifier in (
    <large list of stopwors>
then 
    <...>
```

that rule is generated from a template and could easily exceed the limits of `writeUTF`.

mariofusco:
Honestly I believe you're misusing the 'in' operator or at least there are better ways to achieve the same result. For instance did you evaluate the possibility of putting all those stopWords in a global list and using it with the 'memberOf' operator as in the following test case?

```
@Test
public void testGlobalInList() throws Exception {
    String drl =
            "global java.util.List list; \n" +
            "\n" +
            "rule R2 when\n" +
            "    $s : String( this memberOf list )\n" +
            "then\n" +
            "   System.out.println( 'aa' );" +
            "end\n" ;

    KieHelper helper = new KieHelper();
    helper.addContent( drl, ResourceType.DRL );
    KieSession ksession = helper.build().newKieSession();
    ksession.setGlobal( "list", Arrays.asList( "a", "b" ) );

    ksession.insert("a");
    ksession.fireAllRules();
}
```

fbertola:
Yes, I'm aware that there are better ways of achieving the goal but I cannot actually modify the rules.
Besides, your test would imply that the code that use the rules wold know what kind of list the rules need and this is not the case.

Thanks for the help!

mariofusco:
First of all, I am no longer convinced that the change you're suggesting fixes your problem. For what I can see here http://docs.oracle.com/javase/7/docs/api/serialized-form.html#java.lang.String it seems that writeObject(String) delegates to writeUTF(). Did you try the fix you're suggesting? Are you sure it solves your problem?

Second we really don't want to promote a bad practice like having a constraint longer than 64K. There are lots of alternative ways to fix this. For instance if you don't like the solution based on global I suggested above you could also add the stop words to a StopWords fact and use that in the rule. Moreover you can even use a "setup" rule to add all the stop words to his StopWord fact if you want everything generated.

fbertola:
The documentation is _IMHO_ a bit misleading. If you see the code of `ObjectOutputStream.writeObject` you'll notice that, if the class of the argument is `String`, it calls `ObjectOutputStream.writeString` and **not** directly `ObjectOutputStream.writeUTF` which is a bit different:

```
    private void writeString(String str, boolean unshared) throws IOException {
        handles.assign(unshared ? null : str);
        long utflen = bout.getUTFLength(str);
        if (utflen <= 0xFFFF) {
            bout.writeByte(TC_STRING);
            bout.writeUTF(str, utflen);
        } else {
            bout.writeByte(TC_LONGSTRING);
            bout.writeLongUTF(str, utflen);
        }
    }
```

Clearly, the solution is `writeLongUTF`.
As for your second point, I understand and I'll respect that. I'll try to sort things out differently.

Thanks a lot for your patience though :)

PR #354, opened 2014-07-06
Testcase + fix for DROOLS-546 FactType.get/set should throw specific exception (not NPE) for unknown fields

bvoisin:
I also created a base exception for fact model problems since I could not find one : FactModelException.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/compiler/TypeDeclarationTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/compiler/TypeDeclarationTest.java
@@ -3,6 +3,7 @@ package org.drools.compiler.compiler;
 import org.junit.Assert;
 import org.drools.core.common.EventFactHandle;
 import org.drools.core.definitions.impl.KnowledgePackageImpl;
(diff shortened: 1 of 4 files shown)


mariofusco:
Test case merged. NPE fixed by https://github.com/droolsjbpm/drools/commit/315566acb9360eaee2370287f3be8762e61446c6

bvoisin:
First of all, thanks for the great work on Drools.

Nevertheless, I disagree with you on this updated fix (315566a, ignoring the setter if the field is unknown). I believe that the NPE was a symptom of an user's bug (typo or other error) and should NOT be ignored. Ignoring this field naming error will result in hours lost by users trying to find out why a rule fires or not as expected.

My proposal (throwing a UnknownFieldException) was to improve the symptom to provide a clear information to the user to help him/her find the typo as fast as possible.

The reason I proposed this fix, was because I recently had to debug the application to understand a NPE I was having, and discover a typographic error (badly named field). With my proposed fix, the error message would have been clear enough for me to correct my error without the need to debug. Without the NPE, I will in a similar case loose much more time trying to find the typo.

Still, I might have missed something and would be very happy to hear your thoughts on the subject.

PR #356, opened 2014-07-10
Fixing FindBugs warnings: MS_SHOULD_BE_FINAL

psiroky:
Basically adding "final" where it should be in first place (static non-private fields). Please do _not_ merge this PR yet, we need to discuss if all the changes are safe.

Most of the changes are trivial (e.g. adding final to logger object), but few of them could potentially break something, please see the comments inline


drools-compiler/src/main/java/org/drools/compiler/rule/builder/dialect/java/JavaDialect.java
```diff
@@ -93,44 +93,36 @@
...
     private static Map<Class<?>, EngineElementBuilder> builders;
 
-    public static void setPatternBuilder( PatternBuilder PATTERN_BUILDER ) {
-        JavaDialect.PATTERN_BUILDER = PATTERN_BUILDER;
-    }
-
-    public static void setGEBuilder( GroupElementBuilder GE_BUILDER ) {
-        JavaDialect.GE_BUILDER = GE_BUILDER;
-    }
-
```
psiroky, line 133: @sotty, could you please look at these setters? You added them some time ago (https://github.com/droolsjbpm/drools/commit/3858161e7c5f5ab6188667241beebb5c89ab8275), but they do not seem to be used. Is it safe to remove them?

drools-compiler/src/main/java/org/drools/compiler/rule/builder/dialect/mvel/MVELDialect.java
```diff
@@ -89,39 +89,31 @@
...
     private static Map<Class< ? >, EngineElementBuilder> builders;
 
-    public static void setPatternBuilder(PatternBuilder PATTERN_BUILDER) {
-        MVELDialect.PATTERN_BUILDER = PATTERN_BUILDER;
-    }
-
-    public static void setGEBuilder(GroupElementBuilder GE_BUILDER) {
-        MVELDialect.GE_BUILDER = GE_BUILDER;
-    }
-
```
psiroky, line 124: Same as above (in the JavaDialect class).

drools-core/src/main/java/org/drools/core/io/impl/UrlResource.java
```diff
@@ -58,7 +58,7 @@
 
     private static final int    DEFAULT_BUFFER_SIZE      = 1024 * 4;
 
-    public static File          CACHE_DIR                = getCacheDir();
+    public static final File    CACHE_DIR                = getCacheDir();
```
psiroky, line 61: @etirelli, is it safe to make the cache dir final? It seems, at least from the upper case usage, that this is a constant.

drools-core/src/main/java/org/drools/core/type/DateFormatsImpl.java
```diff
@@ -47,7 +47,7 @@
      * This is here for any static classes, such as MVEL DataConverters, to access.
      * It is expected this will be set, before access.
      */
-    public static ThreadLocal<DateFormats> dateFormats = new ThreadLocal<DateFormats>() {
+    public static final ThreadLocal<DateFormats> DATE_FORMATS = new ThreadLocal<DateFormats>() {
```
psiroky, line 50: This also seems to be a constant, please review and let me know if its not and should be reverted (removing the final).

drools-compiler/src/main/java/org/drools/compiler/rule/builder/dialect/java/JavaDialect.java
psiroky, line 133: Per @sotty's comment, the above was added as part of the integration with drools-chance . It will be reworked in 6.x, so it should be safe to remove those methods now.

drools-core/src/main/java/org/drools/core/io/impl/UrlResource.java
etirelli, line 61: I think so. It is only initialized once.

PR #358, opened 2014-07-15
[DROOLS-498] Test case for duration-based test memory leak

andersonkent:
As requested by Mark Proctor and Davide Sottara, relating to DROOLS-498.  This test case sets up and causes a memory leak caused by a duration-based rule and a non-duration based rule that match the same events.

The use case is as follows.  We wish to suppress events that are reported and clear themselves in a short window of time.  If the time window expires, the event is forwarded into our system.

Three events are inserted in order, separated by 500ms:
StockTick(0, "DROO", 1.00)  
StockTick(1, "OORD", 1.00) // Will cause previous event to be deleted.
StockTick(2, "DROO", 1.00) // Gets deleted after 2 seconds, but StockTick object remains in PhreakPropagationContext.


sotty:
Test case finished and PRed here
https://github.com/sotty/drools/commit/49da68315983c0b040d076402d2bc99f61fb7b49
See also comments on 
https://issues.jboss.org/browse/DROOLS-498

PR #360, opened 2014-07-22
[DROOLS-560] Added unit test to reproduce the problem.

mikedev9000:


mariofusco:
Test case merged and bug fixed.

PR #362, opened 2014-07-25
Partial fixes for Windows 

sotty:
Tested on a WIN machine up to module drools-pmml
drools-pmml fails (parsing error)
the rest has not been tested


mariofusco:
Squashed to a single commit and merged

PR #365, opened 2014-07-25
[DROOLS-495] maven-plugin now executes also process-test-resources phase

psiroky:
- this phase is needed in order to copy the test resources into target/,
  so they can be actually used for testing.


psiroky:
I think it would be useful to cherry-pick this also into 6.1.x (assuming there is still time before release).

PR #366, opened 2014-07-28
Fixed errors caused by improper handling of line feeds in Windows environment

lanceleverich:
Fixed errors caused by improper handling of line feeds in Windows environment


manstis:
I've been watching these fixes; why not use System.getProperty("line.separator")?

Only a passing observation.

sotty:
some of the methods are used when loading files, processing streams, managing bytes created internally and, ultimately, even by test cases. Only some of them use the OS-specific line separator.
Moreover, there may be files that are copied between different systems without replacing the separator. 
I'm not sure, there may be a missing commit here, but I'm not sure that this solution is definitive.
@manstis btw, we only checked drools, all the other projects are waiting.........

manstis:
Ahh, yes, depending on where and how files were created (and whether they've been copied from Linux to Windows or vice-versa) explains things... We have enough trouble compiling kie-wb on Windows anyway (command-line length limitations) not to worry about something as "trivial" as line separators ;)

PR #375, opened 2014-08-28
6.0.x

pragyakirti:


manstis:
I wonder if the 666 commits is trying to tell us something?

Looks like your PR hasn't been sync'ed with the 6.0.x branch. I suspect this will be rejected.

PR #377, opened 2014-09-01
Add test case for reproduce Exception when remove package

chenem2000:
- testRemoveHasSameCondition method is test  remove package which rule
  has same condition.
- testExeRulAfterRemoveHasSameCon method is test fire rule after remove
  package, and the package has rule with same condition.


mariofusco:
Test cases merged and bugs fixed

PR #378, opened 2014-09-02
[DROOLS-242] dump generated DRL from dtables when drools.dump.dir specified

psiroky:


mariofusco:
The implementation is ok. As for the tests, why do you delete the dump directory at the beginning of the test and not at the end? I guess it will be even better to put that dumpDir creation/removal part in 2 methods annotated with @Before and @After

psiroky:
I think advantage of deleting if before the test is that the directory will actually stay there after the tests, so when debugging, one can look there and actually see what was or was not generated. But it definitely makes sense to put this kind of code into method annotated by @Before.

psiroky:
@mariofusco thanks for the comments! I updated the tests to use the @Before and @After methods.

PR #385, opened 2014-09-26
BZ-1101295 enhance test to reproduce the problem

winklerm:
The original test did not reproduce the problem on older Drools versions, proposing an enhancement to fix it.


mariofusco:
Constraint jitting can be tested in an easier way by setting the TEST_JITTING flag in MvelConstraint to true.

PR #386, opened 2014-09-26
Fixes for kie-ci tests

psiroky:
- fix failing test + un-ignore other KieScanner tests
- configure the tests to use custom maven repo to avoid using ~/.m2. This way the tests are more isolated and the result of the tests does not depend on the contents of local ~/.m2 repo on the particular machine.

### DIFF
--- a/kie-ci/pom.xml
+++ b/kie-ci/pom.xml
@@ -169,4 +169,52 @@
     </dependency>
   </dependencies>
 
+  <build>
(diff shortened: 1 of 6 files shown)


psiroky:
@mariofusco, I am not 100 % sure about the fix, please review the changes

psiroky:
Rebased onto the latest changes in master so the commits can be easily merged.

PR #387, opened 2014-10-01
DROOLS-620

rogeral:
- This address issue of interval rule actions not being processed
- Three test cases added
- testIntervalRuleInsertion demonstrates ignored fact insertion
- testIntervalRuleModify demonstrates ignored modify
- testIntervalRuleSetFocus demonstrates ignored setFocus call


mariofusco:
I reproduced this bug in a simpler way and fixed it.
Thanks a lot for your help.

PR #389, opened 2014-10-12
Add test case for reproduce exception when remove and fire rule

chenem2000:
- testRemoveHasSameConElement method will throw exception when fire rule after remove rule has same condition element.
- testFireAfterRemoveWithSameCondition method will throw exception , but change the fire rule sentence order or add rule order,  no exception throw.
- testRemoveWithSameRuleNameInDiffPackage test remove rule has same name in different package.
- testSameEval test fire rule has same eval sentence.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/AddRemoveRulesTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/AddRemoveRulesTest.java
@@ -1,10 +1,12 @@
 package org.drools.compiler.integrationtests;
 
 import org.junit.Test;
+import org.kie.api.KieBase;
 import org.kie.api.definition.KiePackage;
 import org.kie.api.definition.rule.Rule;
 import org.kie.api.io.ResourceType;
 import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.StatelessKieSession;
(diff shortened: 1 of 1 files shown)


psiroky:
Thanks for the PR! 

There are two small things that would be great if you could fix:
1) Could you please rebase the commits into single one? It seems most of the commits are either merge commits or small fixes, that can be incorporated to the previous commit.

2) You seem to be using different chars for indenting (I see both spaces and tabs). Could you please update the tests to use only 4 spaces to be consistent with the rest of the codebase?

chenem2000:
sorry , my mistake.
I am not familiar with git.
Is it now ok ?

psiroky:
I just ran the tests and they are still failing.

@mariofusco could you please briefly look at this? It seems the tests are finding real bugs regarding removing rules from KieBase.

The tests are using the old method `removeKnowledgePackage()` but since the new method `removeKiePackage()` calls exactly that one, it should not matter.

mariofusco:
The provided test cases evidenced 3 different bugs that I reported in these jira:
- https://issues.jboss.org/browse/DROOLS-891
- https://issues.jboss.org/browse/DROOLS-893
- https://issues.jboss.org/browse/DROOLS-894

and fixed with the commits linked in the respective tickets.

All test cases have been also merged.

Thanks a lot for having reported these issues,
Mario

PR #394, opened 2014-10-23
[DROOLS-512] Re-adding call to setKnowledgeRuntime in StatefulKnowledgeS...

mikedev9000:
...essionImpl constructor.

It seems as though this call was inadvertently removed
as part of commit 1f8dc32. The method makes a call that
results in the ksession's mbean being created and registered.

Without this call, no ksession mbeans are created.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
+++ b/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
@@ -398,7 +398,7 @@ public class StatefulKnowledgeSessionImpl extends AbstractRuntime
 
         //runtimeServices = new HashMap<ResourceType, KieRuntimeManager>();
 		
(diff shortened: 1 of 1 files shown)


drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -397,6 +397,8 @@ public StatefulKnowledgeSessionImpl(final int id,
         }
 
         //runtimeServices = new HashMap<ResourceType, KieRuntimeManager>();
+		
+        setKnowledgeRuntime(this);
```
mikedev9000: I just realized that setKnowledgeRuntime(this) results in a second call to create the processRuntime, which is already created in teh constructor. I'll update this PR to ensure the constructor only create the processRuntime once.

mariofusco:
I embedded this fix in another small refactor. Thanks.

PR #395, opened 2014-10-24
Build drools-core before drools-beliefs.

mikedev9000:
Without this change, a compilation error can occur, because
drools-beliefs depends on drools-core.


mikedev9000:
Closing this PR. It turns out that the compilation error I encountered had nothing to do with the module ordering.

PR #398, opened 2014-11-25
[DROOLS-654] An integration test to reproduce Drools bug 654

tmescic:
An integration test that reproduces DROOLS-654 bug. The test is called testKJarUpgradeSameSessionInStreamMode is essentially a copy of test testKJarUpgradeSameSession(), the difference being that the former creates a kSession using a KieBaseConfiguration (with STREAM event processing mode). If a session is created that way, facts will be removed from ksesson, and rules wil not be triggered.


mariofusco:
This is not a bug.
See comment in https://issues.jboss.org/browse/DROOLS-654

PR #399, opened 2014-12-17
Fixed wrong quotes in documentation

pgericson:
It was the wrong quote character that will be removed when converted to html, with will fix it.


psiroky:
The documentation has been moved to https://github.com/droolsjbpm/droolsjbpm-knowledge. I fixed the typos there. Thanks!

PR #400, opened 2015-01-12
Documentation typo fix.

baldimir:
KieSession names must be unique. Also session with name KSession2_2 is mentioned later in documentation, so this was a copy-paste mistake I think.


psiroky:
Thanks for the PR.!The location of the drools-docs changed recently and this one is no longer used and needs to  be removed. I committed the fix to the correct location: https://github.com/droolsjbpm/droolsjbpm-knowledge/commit/30335d565a5f8bd78b5c83d17d456e22d38af740

PR #403, opened 2015-01-20
DROOLS-693: UrlResource does not handle Basic Authentication correctly

manstis:


manstis:
_drools_
(master) http://github.com/droolsjbpm/drools/commit/7c1900629

(6.2.x) http://github.com/droolsjbpm/drools/commit/3b725b5e5

PR #405, opened 2015-02-10
BZ-1190903 NPE when trying to access modified Rule associated with a non...

lazarotti:
...-default session


mariofusco:
This issue has been already reproduced and fixed

PR #406, opened 2015-02-11
DROOLS-707 - failing test - NullPointer when changing order of the rules

fpeloi-bmj:
I added a test with 2 asserts: 
- the first one with a Drools rule that fails as a NPE is thrown
- the second runs fine and the workaround is to move one line of the rule at the top


mariofusco:
I reproduced this issue in a more compact way and fixed it.

Thanks

PR #407, opened 2015-02-13
Update AbductionTest.java

lboullo0:
Added getFactClass() to get the Class of FactType otherwise o.getClass().equals( .... ) is always false


lboullo0:
Beware that now the test fails at assertEquals( 2, bs.size() );
java.lang.AssertionError: expected: <2> but was: <1>

sotty:
Thanks for catching this.
I have incorporated it in a commit that adds a new feature to the abductive queries. 
I am also investigating whether the correct behavior is preserved or not

sotty:
Fixed here:
https://github.com/sotty/drools/commit/7d4968808371a1bb9b61b50a6651707af432b56b

PR #409, opened 2015-02-24
reproducer for missing resource provider for MemoryKieModule

mswiderski:


mariofusco:
I took your test case, extended it a bit and pushed it on master together with the fix

PR #411, opened 2015-02-25
DROOLS-707 - more test failures when changing order of rules

fpeloi-bmj:
I found other 2 cases where I get not consistent behavior with variables and order of the rule lines.
I created this pull request with a few cases, if I remember correctly, the line that throws a NPE is exactly the same as the problem has been fixed already in DROOLS-707.


fpeloi-bmj:
This was just a failing test to replicate the bug, the patch of this bug has gone live with 6.2.0.Final, and there were other failing tests done by Mario. I think this pull request can be closed.

psiroky:
@fpeloi-bmj thanks for the info, closing the PR.

PR #417, opened 2015-03-11
[BZ-1200784] add reproducer

winklerm:
Adding reproducer for https://bugzilla.redhat.com/show_bug.cgi?id=1200784.


mariofusco:

PR #418, opened 2015-03-16
DROOLS-740 NoSuchMethodError at runtime while evaluating String and Long

tkobayas:
- test case


mariofusco:

PR #419, opened 2015-03-18
Update MVELSafeHelper.java

andbi:
No comments.


mariofusco:
I fixed it in a slightly different way https://github.com/droolsjbpm/drools/commit/1beb00549

Thanks for having reported that

PR #420, opened 2015-03-23
6.2.x

amitkhanchi:
fix for https://issues.jboss.org/browse/GUVNOR-2175 and https://issues.jboss.org/browse/GUVNOR-2175

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -1177,7 +1177,7 @@ public class RuleModelDRLPersistenceImpl
                                                buf );
                     break;
                 case BaseSingleFieldConstraint.TYPE_TEMPLATE:
-                    buildTemplateFieldValue( type,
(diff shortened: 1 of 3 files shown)


krisv:
Note that this should be merged into master rather than 6.2.x

amitkhanchi:
I will open another pull request in that case. Closing this one.

manstis:
Thanks. Can you please add some tests too - see https://github.com/droolsjbpm/drools/blob/master/drools-workbench-models/drools-workbench-models-guided-template/src/test/java/org/drools/workbench/models/guided/template/backend/RuleTemplateModelDRLPersistenceTest.java for examples. Have you tested "makeValueList" for data-types other than String? (I see you'll surround values with quotes.. I wonder if you've tried this for fields that are numerical, Dates, BigDecimal, BigInteger etc? - I know DRL will produced, but IDK if the DRL will compile)

amitkhanchi:
I have opened the pull request with a test added for master at https://github.com/droolsjbpm/drools/pull/421
I have tested the generated DRL with all types except boolean as that type won't make any sense with in operator and it works.  It will work with date type as long as the  values represent the same point in time as the field does so, it would rarely be used with date type. Even in a regular DRL, try using date field for equality and you would get what I am trying to say. Mostly, date types would be considered for a range constraint rather than equality even though equality constraint can be enforced as long it is same point in time for both a regular DRL or in this case with 'in' operator.

PR #423, opened 2015-03-26
DROOLS-751 - Added missing nullchecks.

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-751


mdproctor:
We don't do null checks, as we don't do defensive programming. This is the symptom, the cause will be else where and we need to fix the cause.

We'll need a unit test to progress this.
http://docs.jboss.org/drools/release/5.6.0.Final/droolsjbpm-introduction-docs/html/gettingstarted.html

baldimir:
But it's obvious from the code that previous can be null, because it's set to null in the same methods that I changed. See for example https://github.com/droolsjbpm/drools/blob/49d67ad97b0bb8a65964932f739da5b64189cf37/drools-core/src/main/java/org/drools/core/common/LeftTupleSetsImpl.java#L127

PR #424, opened 2015-03-29
 Replace deprecated KnowledgeBase with KieBase in AccumulateTest

ligangty:
```
  I'm reading the drools code and found that KnowledgeBase interface
  has been annotated with deprecated, so try to replace it with new
  KieBase. AccumulateTest is first one. Note that there is still
  left because of the self-owned method addKnowledgePackages of
  KnowledgeBase.
```


psiroky:
+1 to merge this. @mariofusco do you agree?

PR #425, opened 2015-03-30
Perform evaluation before launching a jit compilation request. This will...

mvtorres:
... alleviate race conditions when jit compiling rules with Lazily initialized objects.

### DIFF
--- a/drools-core/src/main/java/org/drools/rule/constraint/MvelConstraint.java
+++ b/drools-core/src/main/java/org/drools/rule/constraint/MvelConstraint.java
@@ -206,11 +206,14 @@ public class MvelConstraint extends MutableTypeConstraint implements IndexableCo
                 }
             }
 
+			boolean result = conditionEvaluator.evaluate(object, workingMemory, leftTuple);
             if (!isDynamic && invocationCounter.getAndIncrement() == JIT_THRESOLD) {
                 jitEvaluator(object, workingMemory, leftTuple);
             }
+            return result;
+        }else{
+			return conditionEvaluator.evaluate(object, workingMemory, leftTuple);
         }
-        return conditionEvaluator.evaluate(object, workingMemory, leftTuple);
     }
 
     protected void createMvelConditionEvaluator(InternalWorkingMemory workingMemory) {


mariofusco:
It is not clear to me which problem you're trying to solve with this pull request. 
Moreover we're no longer maintaining version 5.x.x.
Please try to better clarify what you're try to achieve with this and I'll evaluate to forward port this to Drools 6.x.

Thanks,
Mario

mvtorres:
Hi Mario,

The problem I was trying to solve was exceptions when the jit thread and main thread tries to initialize hibernate objects during rule evaluation.

Here's the thread I opened in google groups.

https://groups.google.com/forum/#!searchin/drools-usage/jit/drools-usage/WEg7cHssFXY/CZXWxpLtj08J

Thanks

mariofusco:
As I wrote we're no longer maintaining version 5.x.x so I cannot accept this PR anyway.

That said I am still not understanding in which cases the problem you're trying to fix with this PR could arise and how this PR can "alleviate" it. Finally if we have a problem (and note that I am NOT saying that you aren't right in that and there couldn't be any problem) I'd like to have a better understanding of where the problem is and possibly find a definitive fix instead of just "alleviating" it.

It would be great if you could provide a reproducer for this issue you're reporting so I could give a look at it. Ideally the reproducer should use 6.x API, but if this is not possible also one using drools 5.6 could be ok and I'll try to migrate it to the new APIs.

Thanks for your help,
Mario

mvtorres:
Hi Mario,

Understood. We're on 5.5 so it was easiest for me to push the fix on 5.6.

The problem happens when 2 threads(jit thread, regular calling thread) tries to initialize "lazy" properties of hibernate entities inserted into drools. Hibernate sessions are not thread safe, and can produce exceptions when accessed concurrently.

With the existing codebase on 5.5/6.x the jit thread is "launched" prior to rule evaluation. This can cause the calling thread, and jit thread to call lazy property initialization at the same time. This fix just reorders the call such that the rule is first evaluated, thereby initializing these lazy properties, before launching the jit thread.

I currently don't have a test that's isolated and consistent, but we observed this issue happening intermittently on our integration tests.

Should I open a jira issue?

Thanks

mariofusco:
Ok, at least now the problem is more clear to me. So yes, please open a jira. If you could provide a reproducer it would be great otherwise I'll try to develop one on my own. 

One last question: does the fix you suggested only "alleviated" that problem (meaning that from time to time you still experienced it, even if less frequently) or definitively solved it?

Thanks again,
Mario

mvtorres:
I opened DROOLS-762.

So far after the fix, I have not seen it happen on the integration test where it happens frequently. It may take a while for me to isolate a reproducer, since it happens on our integration test. 

Thanks.

PR #426, opened 2015-04-08
Guided decision tree one line rule unmarshall test

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-758


manstis:
Wrote some additional tests and merged here: http://github.com/droolsjbpm/drools/commit/7c48450a2

PR #427, opened 2015-04-09
DROOLS-759 GuidedDecisionTreeModelUnmarshallingVisitor doesn't support unmarshalling rules that contain line with only spaces in LHS

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-759

Contains reproducer of the issue and possible fix.


manstis:
Reformatted, squashed and merged: http://github.com/droolsjbpm/drools/commit/7c4ba5b03

Thanks for the PR.

PR #428, opened 2015-04-09
DROOLS-761 Guided decision trees - new test for testing of unmarshalling drls containing different line endings

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-761

Contains new test and some code generalization.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtree/src/test/java/org/drools/workbench/models/guided/dtree/backend/GuidedDecisionTreeDRLPersistenceUnmarshallingTest.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtree/src/test/java/org/drools/workbench/models/guided/dtree/backend/GuidedDecisionTreeDRLPersistenceUnmarshallingTest.java
@@ -3322,12 +3322,12 @@ public class GuidedDecisionTreeDRLPersistenceUnmarshallingTest {
     private GuidedDecisionTree getAndTestUnmarshalledModel( final String drl, final String baseFileName,
             final int expectedParseErrorsSize ) {
         final GuidedDecisionTree model = GuidedDecisionTreeDRLPersistence.getInstance().unmarshal( drl,
-                baseFileName,
-                dmo );
+                                                                                                   baseFileName,
+                                                                                                   dmo );
 
-        assertNotNull(model);
+        assertNotNull( model );
         assertEquals( expectedParseErrorsSize,
-                model.getParserErrors().size() );
+                      model.getParserErrors().size() );
         return model;
     }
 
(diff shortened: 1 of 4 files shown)


drools-workbench-models/drools-workbench-models-guided-dtree/src/test/java/org/drools/workbench/models/guided/dtree/backend/GuidedDecisionTreeDRLPersistenceUnmarshallingTest.java
```diff
@@ -3535,13 +3287,48 @@ public void testOneLineRules() throws Exception {
                        Integer.class.getName(),
                        DataType.TYPE_NUMERIC_INTEGER );
 
+        getAndTestUnmarshalledModel( drl, "test", 1 );
+    }
+
+    @Test
```
triceo: I was wondering whether this test wouldn't be better done using @Parameterized. With "\n" or "\r\n" being the parameter value.

It would definitely be cleaner that way, as one invocation of the test method would only test one case.

drools-workbench-models/drools-workbench-models-guided-dtree/src/test/java/org/drools/workbench/models/guided/dtree/backend/GuidedDecisionTreeDRLPersistenceUnmarshallingTest.java
```diff
@@ -3535,13 +3287,48 @@ public void testOneLineRules() throws Exception {
...
+                       DataType.TYPE_THIS );
+        addModelField( "Person",
+                       "integerField",
+                       Integer.class.getName(),
+                       DataType.TYPE_NUMERIC_INTEGER );
+
+        final String valueValid = "10";
+        final String valueInvalid = "\"something\"";
+
+        final String drl = "rule \"test\" {separator}" +
```
triceo: Also, I was thinking that {separator} and {value} could be constants, so as to not repeat ourselves.

drools-workbench-models/drools-workbench-models-guided-dtree/src/test/java/org/drools/workbench/models/guided/dtree/backend/GuidedDecisionTreeDRLPersistenceUnmarshallingTest.java
baldimir: Ok, I'll refactor it a bit.

baldimir:
Closing because of conflicts. I will create a new one that can be merged correctly.

PR #429, opened 2015-04-13
Guided decision trees - new test for testing of unmarshalling drls containing different line separators

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-761


manstis:
Re-formatted and merged: http://github.com/droolsjbpm/drools/commit/532cd00ba

PR #431, opened 2015-04-21
DROOLS-383, switch over String in function

jcgarciam:
Eclipse compiler has a least known documented configuration property "org.eclipse.jdt.core.compiler.compliance"
which need to be set along with the 'org.eclipse.jdt.core.compiler.source' and 'org.eclipse.jdt.core.compiler.codegen.targetPlatform'
properties.


psiroky:
ok to test

psiroky:
@mariofusco please take a look at this. All the tests passed so it seems to be good.

I am thinking if we could easily add test for this. I guess it should work on all JDKs (6, 7, 8) assuming we would set correct lang level for Drools?

psiroky:
Merged into master as https://github.com/droolsjbpm/drools/commit/42a3811fac0111978f65514801c080e48bee41fc. Thanks for the PR!

PR #432, opened 2015-04-26
Rename root POM from drools-multiproject to just drools

psiroky:
This should be quite safe change as only artifacts from this repo should depend on the root POM. 

@mbiarnes @ge0ffrey @etirelli could you please confirm that we want this change in? (I was following the comment in the pom, but it is a bit old, so things might have changed).

### DIFF
--- a/drools-beliefs/pom.xml
+++ b/drools-beliefs/pom.xml
@@ -5,7 +5,7 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.drools</groupId>
-    <artifactId>drools-multiproject</artifactId>
+    <artifactId>drools</artifactId>
     <version>6.3.0-SNAPSHOT</version>
   </parent>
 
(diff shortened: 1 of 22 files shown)


ge0ffrey:
+1 This is safe now

psiroky:
retest this please

mbiarnes:
+1

mbiarnes:
I am testing it right now

mbiarnes:
drools doesn't build - because of: https://gist.github.com/mbiarnes/f1a9e601af233d45ade2
this is another reason. Send a mail on last Friday.
we should wait until this is fixed

psiroky:
@mbiarnes  the build error you linked is for `droolsjbpm-tools` repo. This is PR is for Drools which is built earlier in the chain, so I think this is quite safe change and would not influence the build of `droolsjbm-tools`. WDYT?

etirelli:
+1

PR #433, opened 2015-04-29
ref #775 - implemented a reflection cache for property listener check

fbertola:
With this simple cache I've noticed a noticeable performance increase while matching rules with objects with many fields and methods.


psiroky:
ok to test

mariofusco:
As agreed with Federico the class introspection is a red herring and the performance problem is not caused by that

PR #434, opened 2015-05-06
removed dependency to drools-docs to prevent circular dependency

mbiarnes:
Should fix JIRA https://issues.jboss.org/browse/DROOLS-779


psiroky:
Merged into master. The test failure is not related to this change.

PR #436, opened 2015-05-20
[DROOLS-751] - add reproducer for regression introduced by the fix

winklerm:
Added reproducer for a regression introduced by adding the propagation queue.

The test succeeds when run against the code just before the commit 8b0ee2e. When the commit is applied, none of the two rules fires (both are expected to fire two times).

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
@@ -5460,6 +5460,66 @@ public class CepEspTest extends CommonTestMethodBase {
         ksession.fireAllRules();
     }
 
+    @Test
+    public void testAfterAndBeforeOperators() {
+        // DROOLS-751 regression
+        String drl =
+                "import " + Event1.class.getCanonicalName() + "\n" +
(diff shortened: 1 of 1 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -5460,6 +5460,66 @@ public void testSerializationWithEventInPast() {
...
+        final List<String> ruleBeforeList = new ArrayList<String>();
+
+        ksession.setGlobal("ruleBeforeList", ruleBeforeList);
+        ksession.setGlobal("ruleAfterList", ruleAfterList);
+
+        try {
+            for (int i = 0; i < 3; i++) {
+                Event1 tc = new Event1("code"+i, i);
+                stream.insert(tc);
+                clock.advanceTime(8, TimeUnit.SECONDS);
```
winklerm, line 5511: If you add
`
ksession.fireAllRules();
`
on line 5511 (it must be before the clock is advanced), the test passes.

Not sure if this is intended, though.

mariofusco:
We discussed this and think that this behavior is correct.

PR #437, opened 2015-05-21
DROOLS-751 Different threads accessing LeftTuple fields that reference LeftTupleSet

baldimir:
Reproducer for NullPointerException that raises when many threads insert and delete the same fact(s).

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
@@ -98,7 +98,7 @@ public class MultithreadTest extends CommonTestMethodBase {
                         FactHandle[] facts = new FactHandle[objectCount];
                         for (int i = 0; i < objectCount; i++) facts[i] = ksession.insert(new Bean(i));
                         ksession.fireAllRules();
-                        for (FactHandle fact : facts) ksession.retract(fact);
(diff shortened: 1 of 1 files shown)


mrietveld:
Hi @baldimir, I think it's a very good idea to test these type of situations. 

I'll leave the decision about merging this to @mariofusco or another core drools engineer, but personally, I don't think this is the proper approach. 

This type of test will provide a lot of false positives: because of how threads work, there is no guarantee that this test will expose problems. Furthermore, this is a very expensive test for the build. 

What I would do is to look into the (excellent!) byteman library as well as looking more at the java.util.concurrent classes. I would be glad to help you design a byteman-based test that tests this issue. 

Thanks!

baldimir:
Hi @mrietveld, thanks for byteman tip! I'll definitely look at it. 

Only for clarification... this test is only a modification of existing test, with that in mind, that the modification raises the chances that the test fails on NPE (by raising the number of threads used.. 1000 is only a bulk number so it fails every time it's run). Original test method was called testConcurrentInsertions(). 

Talked about this issue with @mariofusco, he'll look at it (it's bound to the LeftTupleSet JIRA [1]). It shouldn't fail on NPE. I agree that threads can be pretty nondeterministic in providing deterministic results :) So meanwhile I'll look at the byteman stuff an try to do something with it. 

Again, thanks for the tip!

[1] https://issues.jboss.org/browse/DROOLS-751

mariofusco:
Test case merged

PR #438, opened 2015-05-21
BZ1218308: [GSS](6.1.z) Non existing imports in rdrl files prevents the rule to be opened in Business Central

manstis:
Additional test (and fix) for an edge-case reported by @yurloc 

Furthermore (all be it a hack for possible other bugs) if there's any Exception thrown during unmarshalling we fall-back to using a FreeFormLine to hold the DRL (so at least the user can open and edit the rule).

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -120,6 +120,8 @@ import org.drools.workbench.models.datamodel.workitems.PortableObjectParameterDe
 import org.drools.workbench.models.datamodel.workitems.PortableParameterDefinition;
 import org.drools.workbench.models.datamodel.workitems.PortableStringParameterDefinition;
 import org.drools.workbench.models.datamodel.workitems.PortableWorkDefinition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -3651,9 +3657,15 @@ private ExpressionFormLine parseExpression( final RuleModel m,
...
-                                                                   currentFact.getType() ) );
+                    //If the bound type is not in the DMO it probably hasn't been imported.
+                    //So we have little option than to fall back to keeping the value as Text.
+                    if ( currentFact == null ) {
+                        expression.appendPart( new ExpressionText( expressionPart ) );
+                    } else {
+                        expression.appendPart( new ExpressionVariable( expressionPart,
+                                                                       currentFact.getClassName(),
+                                                                       currentFact.getType() ) );
+                    }
```
Rikkola, line 3668: Something I like to do. When all the options for the if do the same thing, in this case they append a part, I usually like to extract a method. This way when reading the code afterwards the methods stay shorter and are faster to read.

``` java
expression.appendPart( getPart(currentFact,expressionPart) );

private ExpressionPart getPart(Something currentFact,ExpressionPart expressionPart){
  // now the if goes here
  if(currentFact == null){
    return new ExpressionText( expressionPart );
  } else {
    return new ExpressionVariable( expressionPart,
      + currentFact.getClassName(),
      + currentFact.getType() );
}
```

Rikkola:
https://github.com/droolsjbpm/drools/commit/9c371d95d

PR #439, opened 2015-05-22
BZ1218308: [GSS](6.1.z) Non existing imports in rdrl files prevents t…

manstis:
…he rule to be opened in Business Central. Remove unneeded log4j.xml configuration.


psiroky:
Thanks Mike. Cherry-picked into 6.2.x as well.

PR #440, opened 2015-05-22
[DROOLS-751] - add reproducer for possible concurrency issue

winklerm:
Another possible issue - this test started to fail (but not everytime) after adding the propagation queue.

The test executes the same code three times, and sometimes (roughly in 50 % of cases) the rules do not fire. This time we use fireUntilHalt() so this is not an issue with a missing fireAllRules() call.

Could you please take a look if I am missing something? Thanks!


mariofusco:
Test case merged with this commit https://github.com/droolsjbpm/drools/commit/00c540640

PR #441, opened 2015-05-25
[BZ-1221862] fix marshalling of StartProcessCommand

psiroky:
- the out-identifier is optional and may not be set.
  The Xstream converter should first check if it exists,
  and only then set the non-null value.


psiroky:
@etirelli  please take look as this is related to KIE Server BZ. This seems to be the best solution I could find, but I might be missing some details about the XML marshalling.

psiroky:
https://github.com/droolsjbpm/drools/commit/adb6ae291bc2cb981babbb906546f066cf6ada05

PR #442, opened 2015-05-27
DROOLS-794 Advance time in SessionPseudoClock number of times in a row

baldimir:
Added method for time advancement in SessionPseudoClock number of times in a row.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/time/SessionPseudoClock.java
+++ b/drools-core/src/main/java/org/drools/core/time/SessionPseudoClock.java
@@ -38,6 +38,16 @@ public interface SessionPseudoClock extends SessionClock {
      * @param unit the used time unit
      * @return the current absolute timestamp
      */
-    public long advanceTime( long amount, TimeUnit unit );
+    long advanceTime( long amount, TimeUnit unit );
+
+    /**
+     * Advances the clock time number of times in the specified unit amount specified.
+     *
+     * @param amount the amount of units to advance in the clock
+     * @param unit the used time unit
+     * @param numberOfTimes the number of times the clock is advanced by specified amount
+     * @return the current absolute timestamp
+     */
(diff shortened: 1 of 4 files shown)


etirelli:
Thanks for the PR but I am not sure what is the use case for this? A call to advance time by 10sec+10sec for instance, will have the same effect AFAIKT than a call to advance time by 20sec, as all the callbacks are executed in order during clock advance.

Also, the patch seems to imply moving time backwards would be ok, but it would cause problems all over the place as there is an assumption that time only moves forward. 

Please let me know if I am missing something.

baldimir:
This raised during discussion with @mariofusco about this PR [1]. 10 + 10 is not the same as 20 only. Because i.e. timer events (rule timers, etc.) can catch up each time time is advanced. As an example - the test from the mentioned PR uses advanceTime(40) and it fails. When it uses a for cycle for advancing time by 1 with iteration count 40, it passes. Please ask @mariofusco for more info. 

To the backwards time movement. I did this on purpose because it can be done also now in present advanceTime method (i.e. advanceTime(-40, SECONDS)). But I can change this for this PR if it's needed. 

[1] https://github.com/droolsjbpm/drools/pull/440/files

psiroky:
ok to test

baldimir:
Changed the backwards time possibility. Now it throws IllegalArgumentException.

etirelli:
@baldimir I will discuss this with @mariofusco , but from your description, it seems to me we need to fix the advance time call, not create a new one... let me talk to him first.

mariofusco:
The advance of the pseudo-clock for a specific number of times in a row is necessary only in very specific use-cases. Due to its limited usage and considering that the same effect can be obtained with a trivial for-loop, after a chat with @etirelli we decided to not add this new method to our API.

PR #445, opened 2015-06-02
GUVNOR-2187 : Real time verification and validation for decision tables

Rikkola:
Documentation is here. https://github.com/droolsjbpm/kie-wb-distributions/pull/58
Going to do a PR to kie-wb-common and drools-wb. Those two will be depending on this.


manstis:
http://github.com/droolsjbpm/drools/commit/f79217ad6

PR #451, opened 2015-06-23
added missing ASL licenses

mbiarnes:
added ASL license in files (*.java. *.drl) where it was missing
did a mvn clean install -Dfull executing the tests
the build was successful

IMPORTANT: after adding the license to files which didn't have them two test fail.
https://gist.github.com/mbiarnes/4b5dd6acfbeb656f3984


psiroky:
Merged into master as https://github.com/droolsjbpm/drools/commit/bea40eed28154911647cf26387d3f4a786e0f70e. Some headers in drools-templates were removed as they cause test to fail - new PR with these will be created.

PR #452, opened 2015-06-24
DROOLS-828 - Agenda group clear command doesn't work

sutaakar:
Test case for bug DROOLS-828.


mariofusco:
Test case incorporated by https://github.com/droolsjbpm/drools/commit/66b87a112

PR #455, opened 2015-07-06
drools-templates: add license headers to test .drl files

psiroky:
The license headers cause some of the drools-templates tests to fail. After brief discussion with @manstis it seems like a bug as the headers should be ignored by the parser.

### DIFF
--- a/drools-templates/src/test/resources/templates/rule_template_1.drl
+++ b/drools-templates/src/test/resources/templates/rule_template_1.drl
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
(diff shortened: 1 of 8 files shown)


manstis:
@psiroky Hi.. I looked at the source, and the `Header` actually contains everything from the start of the "template definition" to the `template "xxx"` section, including comments etc. Therefore we can either (a) change the tests to check for the license or (b) exclude the license from the _test_ resources. WDYT is best?

psiroky:
If this is the expected behavior (not ignoring the leading comment) then I think the best is to remove the license header from the resources (basically ignoring changes in this PR).

manstis:
It is expected behaviour. Closing PR.

PR #456, opened 2015-07-06
DROOLS-836: Decision sheet substitution incorrect for more than 9 parameters

manstis:
See https://issues.jboss.org/browse/DROOLS-836


drools-templates/src/main/java/org/drools/template/model/SnippetBuilder.java
```diff
@@ -57,83 +55,90 @@
...
+        final String[] cellVals = split( cellValue );
         String result = this._template;
 
-        for (int paramNumber = 0; paramNumber < cellVals.length; paramNumber++) {
-            final String replace = PARAM_PREFIX + (paramNumber + 1);
-            result = replace(result,
-                             replace,
-                             cellVals[paramNumber].trim(),
-                             256);
+        //Replace in reverse order so $10 is replaced before $1 etc
```
manstis, line 123: @mariofusco This is the guts of the change; substitute parameters for values in descending order as $10 was being replaced when $1 was processed. Reversing the order of parameter substitutions ensures $10 is replaced before $1 etc.

manstis, line 123: @mariofusco The rest of the changes in this file are "formatting" changes.

mariofusco:
Merged https://github.com/droolsjbpm/drools/commit/f56fe87550be9a3700823dab9b2b8fa7af9fe0e1

PR #457, opened 2015-07-08
DROOLS 842 (fix) and 843 (test)

sotty:


mariofusco:
[DROOLS-842] merged
[DROOLS-843] rejected (invalid test case)

PR #458, opened 2015-07-09
BZ1241358 SingleSessionCommandService Interceptor stack is lost when …

tkobayas:
…rollback
- Keep the Interceptor stack and just recreate TransactionInterceptor

Proposal fix. Please review.


mswiderski:
fixed it slightly different and added test case. jBPM test case will  be included as well, thanks @tkobayas

PR #459, opened 2015-07-09
BPMSPL-151 - Change Impact prediction (preparation work on fully resolving imports in rules files)

mrietveld:
This PR updates the `EclipseJavaCompiler` class to use the newer ECJ `Compiler`
constructor instead of the deprecated one. 

The change is very simple and easy enough to figure out if you look at the deprecated
methods that were being used: 
1. How to use `CompilerOptions`: 
   https://github.com/eclipse/eclipse.jdt.core/blob/R4_3_maintenance/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/Compiler.java#L161
2. How/why to set the `CompilerOptions.parseLiteralExpressionsAsConstants` field: 
   https://github.com/eclipse/eclipse.jdt.core/blob/R4_3_maintenance/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java#L495-498

I also included a small change to use the `JavaCompilerFactory` constructor (in the `KieBuilder` class), 
since the `JavaCompilerFactory.getInstance()` method has also been deprecated.


mrietveld:
For others reading this later: I reverted the KieBuilderImpl change (from `JavaCompilerFactory.getInstance()` to `new JavaCompilerFactory()`) after talking to Mario because we were unsure about possible side-effects. 

The change is also incidental and not important to me.

PR #460, opened 2015-07-15
GUVNOR-2194 : Guided Decision Table V&V - Improved Reporting

Rikkola:
Removed the analysis information from the dtable model and moved the Analysis and Issue classes into drools-wb.

This PR alone will break drools-wb. This PR is needed to fix drools-wb https://github.com/droolsjbpm/drools-wb/pull/46


manstis:
http://github.com/droolsjbpm/drools/commit/016266832

PR #461, opened 2015-07-20
GUVNOR-2176 : Guided decision table verification and validation

Rikkola:
Some bug fixes for dtables V&V.
Tests are in drools-wb.


Rikkola:
I guess I'll merge this one liner in myself

PR #462, opened 2015-07-20
Propagation queue halt problem reproducer. 

baldimir:
Reproducer for problem mentioned in BZ [1]. Tested in latest 6.2.x branch. 

It looks like that there's some bad race condition when Timers are involved. The test sometimes fails on second test iteration. From what I examined, the iteration calls SynchronizedPropagationList.halt() method and there's no call to notifyAll later, so the queue processing is halted till the end of the test. However, it fails not in deterministic way, so sometimes the test passes.   

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1218291

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepFireUntilHaltTimerTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepFireUntilHaltTimerTest.java
@@ -131,7 +131,7 @@ public class CepFireUntilHaltTimerTest {
             // give time to fireUntilHalt to process the insertions
             Thread.sleep(1000);
 
-            Thread.sleep(60000);
+            Thread.sleep(30000);
 //            clock.advanceTime(30, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


mariofusco:
I fixed the race condition on both master https://github.com/droolsjbpm/drools/commit/4ed10fac9 and 6.2.x https://github.com/droolsjbpm/drools/commit/9686624cf

I agreed with @baldimir to not merge this test case because it takes too long to run.

mrietveld:
@baldimir Maybe change it to a Byteman test that isolates the race condition? :)

baldimir:
Yes the Byteman that you mentioned in one of my previous PRs :) I still haven't time to sit to it.. (...putting a sticker note on my LCD...)
byteman

mrietveld:
@baldimir HAhahaha.. Awesome!:D 

I've used it a little bit, but haven't done any real tests with it. If I get a good example for you, I'll share it.

baldimir:
Thanks, I'll appreciate that.

PR #463, opened 2015-07-21
Fix to not loose the pomModel when the MemoryKieModule is cloned for incremental builds

wmedvede:


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
```diff
@@ -290,7 +290,7 @@ void cloneKieModuleForIncrementalCompilation() {
         }
         trgMfs = trgMfs.clone();
         init();
-        kModule = kModule.cloneForIncrementalCompilation( releaseId, kModuleModel, trgMfs );
+        kModule = kModule.cloneForIncrementalCompilation( releaseId, kModuleModel, trgMfs, kModule );
```
manstis, line 293: Why do you need pass `kModule` to itself? (i.e. you're calling `kModule.cloneXYZ...(kModule)`

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
```diff
@@ -103,6 +106,11 @@ MemoryKieModule cloneForIncrementalCompilation(ReleaseId releaseId, KieModuleMod
         for (KieBaseModel kBaseModel : getKieModuleModel().getKieBaseModels().values()) {
             clone.cacheKnowledgeBuilderForKieBase(kBaseModel.getName(), getKnowledgeBuilderForKieBase( kBaseModel.getName() ));
         }
+        clone.setPomModel( sourceKModule.getPomModel() );
```
manstis, line 109: No need to use `sourceKModule` surely just use `this`?

wmedvede, line 109: whatever option for me is good, as I  explained in the email the kModule has aditional info that is populated by the KieBuilder like, dependencies. This PR is really more a suggestion for some thinigs I've detected that shoud eventually be copied when the clone is created. The obvious is the pomModel, but likelly some other info needs to be cloned too.

So this PR is a "suggestion", of course Mario will find the best option.

mariofusco:
I implemented the changes you suggested also taking count of @manstis comments

PR #464, opened 2015-07-22
Adding possibility to create Clear commands using CommandFactory.

sutaakar:


psiroky:
@mariofusco, @etirelli looks like a safe change. WDYT?

etirelli:
+1

mariofusco:
The problem is that for now the execution of the ClearAgendaGroup command won't cause any effect. That's because it is supposed to cancel all the activations for that group, but given the lazy nature of phreak in reality there won't be any activation to be cancelled, since they're created only immediately before being fired. 

I agreed with Mark that when a clear agenda-group is requested it should flush the propagation queue, create all activation without firing them and then cancel them. I opened a jira for that https://issues.jboss.org/browse/DROOLS-865 and when this enhancement will be developed I'll also merge this PR.

psiroky:
@mariofusco thanks for the additional info!

PR #465, opened 2015-07-23
[Don't merge yet] Add dependency on jaxb-core

psiroky:
- the jaxb-impl was split into two: jaxb-impl and jaxb-core
  so we need to depend on both to make the dependant modules
  work
- this needs to be merged after we upgrade to ip-bom 6.0.0.CR27, because the `jaxb-core` dep is not declared in CR26


mbiarnes:
ok - will merge this when I upgrade jboss-ip-bom to 6.0.0.CR27

PR #466, opened 2015-07-24
DROOLS-858 NoSuchMethodError on JIT Compilation of non-primitive comparison

sdirbach:
DROOLS-858 NoSuchMethodError on JIT Compilation of non-primitive 
- Testcase & Fix


sdirbach:
There was still a bug in the fix implementation, new pull request will come shortly.

PR #468, opened 2015-07-24
UPDATED: DROOLS-858 NoSuchMethodError on JIT Compilation of non-primitive comparison

sdirbach:
UPDATED: DROOLS-858 NoSuchMethodError on JIT Compilation of non-primitive
- Testcase & Fix

### DIFF
--- a/drools-core/src/main/java/org/drools/core/rule/constraint/ASMConditionEvaluatorJitter.java
+++ b/drools-core/src/main/java/org/drools/core/rule/constraint/ASMConditionEvaluatorJitter.java
@@ -549,11 +549,19 @@ public class ASMConditionEvaluatorJitter {
             }
 
             if (!toType.isAssignableFrom(fromType)) {
-                mv.visitTypeInsn(NEW, internalName(toType));
-                mv.visitInsn(DUP);
-                load(regNr);
-                coerceByConstructor(fromType, toType);
-                store(regNr, toType);
+                try {
+                    if (toType == Character.class || toType.getConstructor(String.class) != null) {
+                        mv.visitTypeInsn(NEW, internalName(toType));
+                        mv.visitInsn(DUP);
+                        load(regNr);
+                        coerceByConstructor(fromType, toType);
+                        store(regNr, toType);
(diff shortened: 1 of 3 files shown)


drools-core/src/main/java/org/drools/core/rule/constraint/ASMConditionEvaluatorJitter.java
```diff
@@ -549,11 +549,23 @@ private void castOrCoerceTo(int regNr, Class<?> fromType, Class<?> toType, Label
...
+                        mv.visitInsn(DUP);
+                        load(regNr);
+                        coerceByConstructor(fromType, toType);
+                        store(regNr, toType);
+                    } else {
+                        mv.visitInsn(ACONST_NULL);
+                        mv.visitVarInsn(ASTORE, regNr);
+                        mv.visitJumpInsn(GOTO, nullLabel);
+                    }
+                } catch (NoSuchMethodException nme) {
```
mariofusco, line 564: I don't see how a NoSuchMethodException could be thrown at this point. I'm afraid you're confusing this code that generates the bytecode with what happens when the bytecode itself is executed.

sdirbach, line 564: Hi, Class.getConstructor(...) throws a NoSuchMethodException if no constructor with the given signature exists [1]. Remember, the toType can be a class for any Object. I had to include the try-catch, because from what I know, there is no other possibility to find out, if a given constructor exists. 

[1] http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getConstructor%28java.lang.Class...%29

mariofusco, line 564: Ok, I see. Sorry, I misunderstood the reason why you added that try/catch.

mariofusco:
I fixed it with this commit https://github.com/droolsjbpm/drools/commit/80eb0e326 following what you suggested.

Thanks a lot for your help with this.

PR #471, opened 2015-07-26
Run tests in parallel for specified modules

psiroky:
- running tests in multiple forks results
  in significant speed-up
- only some modules are able to run tests in
  parallel, others like -jpa would need to be
  fixed first to not depend on each other.
- it also makes most sense for modules which build
  time is high (most of the time is spent in tests).


drools-reteoo/src/test/java/org/drools/reteoo/integrationtests/MultithreadTest.java
```diff
@@ -142,6 +142,7 @@ public String toString() {
     }
 
     @Test(timeout = 60000)
+    @Ignore("Flakey - fails randomly, probably due to race condition somewhere")
```
mrietveld: @psiroky IMHO, we should not be `@Ignore`'ing tests simply because they fail randomly. The fix here is to fix the test, not ignore it. 

The rest of the PR looks great! But -1 for merging it with this ignore. 

If the test really is no longer vaild, feel free to delete it.

psiroky: I am not intending to merge the PR with that ignore. I just wanted to share the idea and also use CI build to verify everything is OK. 

I will talk to @mariofusco about the failing test as I don't have the knowledge to fix this (the cause looks to be deep down in the engine). However, I am not sure if this even priority since it is in the old `drools-reteoo` module. Anyway, I agree that the test should be fixed or deleted.

pom.xml
```diff
@@ -66,6 +70,21 @@
...
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <configuration>
+            <forkCount>${surefire.forkCount}</forkCount>
+            <reuseForks>true</reuseForks>
```
triceo, line 81: I like to set `reuseForks` to false. This way, you get fresh new JVM for each test - a lot slower,  but eliminates whatever environment conflicts between tests, making them more stable.

psiroky, line 81: The default is I believe `true`, so we have been using one fork (JVM) for a long now, without any troubles. I would like to speed-up the tests, that's the reason for reusing the forks. I will test this locally to see if there is a significant slowdown.

I can make this configurable through property, so that it could be set from command line if needed (e.g. for debugging).

psiroky, line 81: Based on a quick local build, with `<reuseForks>false</reuseForks>` the build takes around 40% more time, which is significant I think.

I believe the tests should be written in a way that they should work independently of other tests, so reusing the forks should not be an issue. If we find such tests we need to fix them.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/d3cced001

PR #472, opened 2015-07-29
JaxbUnknownAdapter: return empty ArrayList when unmarshalling JaxbListWrapper without elements

sutaakar:
In current implementation if there is unmarshalled JaxbListWrapper with null elements (happens when marshalling empty List) then Arrays.asList throws NPE, which upper layers process as null element. This fix assures that JaxbListWrapper is unmarshalled to List instance even with empty elements.


psiroky:
Would it be possible to add a new test as well?

sutaakar:
Added tests covering changed functionality.

psiroky:
ok to test

psiroky:
@mariofusco could you please review this and possible merge?

mrietveld:
@psiroky I'll look at this ASAP.

mrietveld:
Sorry for the delay!!

Merged @ https://github.com/droolsjbpm/drools/commit/9a277f1c4a7386fdc6d1e76f00562900fb7f9384

PR #475, opened 2015-08-01
DROOLS-868 Custom metadata declared at field-level is not available at runtime

iambic69:
In addition to adding the call to fieldDef.addMetaData(), I removed the not-null check on typeResolver since there are no test cases that define what the behaviour should be if typeResolver is null (i.e. it's never null for any of the tests).


mariofusco:
Merged with https://github.com/droolsjbpm/drools/commit/0515b427f

PR #476, opened 2015-08-12
Shorten ByteArrayResource.toString() to avoid flooding logs

psiroky:
- only up-to first 10 bytes are included in the toString()

### DIFF
--- a/drools-core/pom.xml
+++ b/drools-core/pom.xml
@@ -120,6 +120,12 @@
       <artifactId>commons-codec</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <scope>test</scope>
+    </dependency>
+
   </dependencies>
 
   <profiles>
--- a/drools-core/src/main/java/org/drools/core/io/impl/ByteArrayResource.java
+++ b/drools-core/src/main/java/org/drools/core/io/impl/ByteArrayResource.java
@@ -44,8 +44,10 @@ public class ByteArrayResource extends BaseResource
     public ByteArrayResource() { }
 
     public ByteArrayResource(byte[] bytes) {
-        if ( bytes == null || bytes.length == 0 ) {
-            throw new IllegalArgumentException( "bytes cannot be null" );
+        if ( bytes == null ) {
+            throw new IllegalArgumentException( "Provided byte array can not be null" );
+        } else if ( bytes.length == 0 ) {
+            throw new IllegalArgumentException( "Provided byte array can not be empty" );
         }
         this.bytes = bytes;
     }
@@ -125,8 +127,14 @@ public class ByteArrayResource extends BaseResource
     }
     
     public String toString() {
-        return "ByteArrayResource[resource=" + Arrays.toString(this.bytes) + "]";
+        return "ByteArrayResource[bytes=" + firstNBytesToString(10) + ", encoding=" + this.encoding + "]";
     }
 
+    private String firstNBytesToString(int nrOfBytes) {
+        // this.bytes cannot be empty or null (enforced by constructors)
+        String str = Arrays.toString(Arrays.copyOf(this.bytes, Math.min(this.bytes.length, nrOfBytes)));
+        // append the dots ("...") only if the resource has more bytes than requested
+        return this.bytes.length > nrOfBytes ? str.substring(0, str.length() - 1) + ", ...]" : str;
+    }
 
 }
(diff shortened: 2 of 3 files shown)


drools-core/src/main/java/org/drools/core/io/impl/ByteArrayResource.java
```diff
@@ -125,8 +127,28 @@ public int hashCode() {
     }
     
     public String toString() {
-        return "ByteArrayResource[resource=" + Arrays.toString(this.bytes) + "]";
+        return "ByteArrayResource[bytes=" + firstNBytesToString(10) + ", encoding=" + this.encoding + "]";
+    }
+
+    private String firstNBytesToString(int nrOfBytes) {
```
mswiderski, line 133: @psiroky maybe using directly StringUtils.abbreviate would be cleaner: https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringUtils.html#abbreviate(java.lang.String,%20int)

psiroky, line 133: Thanks for the tip. I wasn't aware of this method.

It should be possible to turn the whole byte array into String and then abbreviate that string. It would be probably slightly slower than the current approach, but I don't think that is a big issue.

We would also need to bring the `commons-lang3` into `compile` scope (now it is only in `test` socpe. But again I don't think that is an issue. So all in all, it looks like a very good idea.

@mariofusco what do you think?

psiroky, line 133: Just tried this and it is not as easy as it seemed :)

We actually want first 10 `byte`s of the array and we don't know how long is the whole string after calling `Arrays.toString()`. The `StringUtils.abbreviate()` needs number of chars, but we don't have that number handy. We would need to somehow compute it.

I believe the current approach is actually easier to grasp.

mswiderski, line 133: what you mean we don't have that number handy - isn't that first 10 that you want to show? Anyway trimming it will make it pretty much unreadable and it's just information that there are more data.

Fine to use what you have now, just wanted to point out that StringUtils method.

psiroky, line 133: We want to show the first 10 bytes, but the string form of the byte can have one, two, three or even four chars, depending on the value (e.g. "1", "64", "127", "-127").

mariofusco, line 133: Probably it will be a bit less clear, but I believe that this will achieve the same result in a shorter and faster way:

```
String s = Arrays.toString(Arrays.copyOf(this.bytes, Math.min(this.bytes.length, 10)));
return this.bytes.length > 10 ? s.substring( 0, s.length()-1 ) + ", ...]" : s;
```

drools-core/pom.xml
```diff
@@ -120,6 +120,12 @@
       <artifactId>commons-codec</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <scope>test</scope>
+    </dependency>
```
mariofusco, line 127: For now I'd avoid adding this dependency at all if possible.

drools-core/src/test/java/org/drools/core/io/impl/ByteArrayResourceToStringTest.java
```diff
@@ -0,0 +1,80 @@
...
+                                        "ByteArrayResource[bytes=[10, 20, 30, 40, 50, 60, 70, 80, 90, 100, ...], encoding=UTF-8]"
+                                },
+                        }
+        ));
+    }
+
+    // using List<Byte> instead of directly byte[] to make sure the bytes are printed as part of the test name
+    // see above ({index}: bytes[{0}], encoding[{1}]) -- Array.toString only return object id
+    @Parameterized.Parameter(0)
+    public List<Byte> bytes;
```
mariofusco, line 65: Isn't it possible to have a byte[] instead of a List<Byte> here? This will also allow to easily avoid the commons-lang3 dependency.

psiroky:
Merged both into master and 6.3.x.

drools-core/src/test/java/org/drools/core/io/impl/ByteArrayResourceToStringTest.java
psiroky, line 65: See few lines above.

PR #477, opened 2015-08-12
Bug 1247559 - [GSS] (6.1.z) Guided Rule editor does not seem to reloa…

Rikkola:
…d "eval()" functions correctly while used inside the constraint of a fact

One more Guided Rule Editor bug. Here is the test for it.


mariofusco:
Merged and fixed by https://github.com/droolsjbpm/drools/commit/02580f9a5ecf6163e9f100f4cf16e6114af9715a

PR #479, opened 2015-08-14
BZ-1252327 - TransactionLockInterceptor cannot unlock on transaction …

mswiderski:
…timeout

@tkobayas could you please have a look and maybe give it a go on your environment? I did some tests and it seems to be solving the issue. Though next request using same thread will fail with hibernate attempting to clean up from previous premature completion and you can see following exception:

```
Caused by: org.hibernate.HibernateException: Transaction was rolled back in a different thread!
```

it happens only once per failure of transaction timeout as hibernate somehow delays the call to afterCompletion. Sort of outside of our control

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/jta/TransactionLockInterceptor.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/jta/TransactionLockInterceptor.java
@@ -16,7 +16,11 @@
 
 package org.drools.persistence.jta;
 
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.locks.ReentrantLock;
+
+import javax.transaction.Status;
 
 import org.drools.core.command.impl.AbstractInterceptor;
 import org.drools.persistence.OrderedTransactionSynchronization;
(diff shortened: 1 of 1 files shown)


drools-persistence-jpa/src/main/java/org/drools/persistence/jta/TransactionLockInterceptor.java
```diff
@@ -81,28 +87,62 @@ public TransactionLockInterceptor(Environment environment, String releaseTxKey)
             if (locked) {
                 logger.debug("About to register lock release handler by {}", Thread.currentThread().getName());
                 release((TransactionManager) environment.get(EnvironmentName.TRANSACTION_MANAGER));
+            } else if (forceUnlock.contains(Thread.currentThread().getId())) {
+                forceUnlock.remove(Thread.currentThread().getId());
```
mrietveld: personal preference: use the boolean result of the `CopyOnWriteArraySet.remove()` method instead of `contains`, even though using both the `contains` and `remove` methods is technically _not_ a race condition here, because the thread id will always be unique. 

However, when I see both being used in concurrent code, I automatically suspect a race condition, whereas if you just have `remove`, that's not possible..

tkobayas:
Thanks, Maciej. I posted a comment on the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1252327#c11 . Please let me know your thought.

mswiderski:
yes, have seen it but didn't have time to reproduce and look for a fix yet. will do it today or tomorrow. thanks!

mswiderski:
@mrietveld thanks Marco, I applied the suggested change and rely only on remove method instead of both contains and remove.

@tkobayas see my comment in bz and if you could give it ago with the suggested check to see if that unlocks it would be great.

mswiderski:
merged into master and 6.3.x

PR #480, opened 2015-08-18
DROOLS-885 ExecutorProviderImpl's ThreadPoolExecutor always run with …

tkobayas:
…one thread

The fix proposal is just setting corePoolSize so it may not the behavior which you like (It doesn't shrink smaller than corePoolSize). Feel free to change.


triceo:
Toshiya, can you make this not depend on `Thread.sleep()` so much? The behavior of this test will be largely random because of that.

If you need to ensure that all the threads are executing at any given moments, you can use `java.util.concurrent` classes to achieve that - I can see this implemented by a countdown latch or a barrier.

tkobayas:
Updated the test case. Does this look okay?

mariofusco:
I merged the fix without the test case, because I also didn't like the sleeps and I didn't see your updated version. Anyway I don't think it is likely that we could have a regression here, so the test case can be omitted. 

By the way another great catch Toshyia, thanks again.

PR #481, opened 2015-08-18
JBPM-4739 - Create a Process Instance and Task webservice

mrietveld:
Also meant for cherry-pick to 6.3.x.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/command/runtime/process/SignalEventCommand.java
+++ b/drools-core/src/main/java/org/drools/core/command/runtime/process/SignalEventCommand.java
@@ -16,6 +16,8 @@
 
 package org.drools.core.command.runtime.process;
 
+import java.util.Map;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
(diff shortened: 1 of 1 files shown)


drools-core/src/main/java/org/drools/core/command/runtime/process/SignalEventCommand.java
```diff
@@ -107,8 +121,10 @@ public Void execute(Context context) {
     }
 
     public String toString() {
-        if (processInstanceId == -1) {
-            return "ksession.signalEvent(" + eventType + ", " + event + ");";
+        if (processInstanceId == -1 && correlationKey == null) {
+            return "ksession.signalEvent(" + eventType + ", " + event + ");"; 
+        } else if (correlationKey != null) {
+            return "ksession.signalEvent(" + correlationKey + ", " + eventType + ", " + event + ");";
```
mswiderski, line 135: minor thing but would be better to align toString with actual method signature as it sort of simulates it - meaning processInstanceId/correlationKey to have as last parameter (signal by correlationKey seems not to be in place yet if ever planned to be there)

besides that looks good to me

mrietveld, line 135: I agree, but that would also mean modifying the toString() method below -- and I try not to change anything that's not related to the change I want to make.

triceo:
As far as I know, BPMSPL-166 is no longer approved for the release - please do not cherry-pick to 6.3.x.

mrietveld:
Yup, just saw that today. Thanks @triceo!

(BTW, that is a _big_ mug of beer! ;D)

triceo:
Nah, _no mug_ is too big. :-)

PR #482, opened 2015-08-20
fix javadoc about lnglevel

tkobayas:
drools.dialect.java.compiler.lnglevel is the correct system property name.

https://github.com/tkobayas/drools/blob/master/drools-core/src/main/java/org/drools/core/rule/builder/dialect/asm/ClassLevel.java#L26


psiroky:
Merged into master and 6.3.x. Thanks for the PR!

PR #483, opened 2015-08-20
drools-pmml: don't print unnecessary info to sysout during tests

psiroky:


sotty:
Agree it should not go to sysout.
The test should make some assertions though:
- validate against the xsd schema
- check that the NN parameters in the XML matches the expected content

psiroky:
@sotty I agree about the assertions, but I am afraid I don't have the knowledge to add those (at least not without some non-trivial time spent studying the pmml module). I merely tracked down the source of those messages and disabled them. Would it be possible for you to take a look at adding those assertions?

sotty:
I have created a PR with an improved version of this test

psiroky:
Thanks @sotty, I am closing this PR in favor of #484

PR #484, opened 2015-08-20
[PMML][Test] Avoid sysout, add assertions

sotty:


psiroky:
Just a note about the check failure: One test failed `org.drools.reteoo.integrationtests.MultithreadTest.testSlidingTimeWindows`, but it is failing intermittently, so definitely not related to the changes in this PR.

PR #486, opened 2015-08-29
Improve rule datamodel

yurloc:
- Added RuleMetadata equals and hashCode
- Changed getParameters() so that it doesn't modify object state

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/BaseSingleFieldConstraint.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/BaseSingleFieldConstraint.java
@@ -165,7 +165,10 @@ public class BaseSingleFieldConstraint
         if (constraintValueType != that.constraintValueType) return false;
         if (expression != null ? !expression.equals(that.expression) : that.expression != null) return false;
         if (operator != null ? !operator.equals(that.operator) : that.operator != null) return false;
-        if (parameters != null ? !parameters.equals(that.parameters) : that.parameters != null) return false;
+        if (parameters != null ?
+                !parameters.equals( that.parameters ) && !parameters.isEmpty() && that.parameters != null
+                : that.parameters != null && !that.parameters.isEmpty() )
+            return false;
         if (value != null ? !value.equals(that.value) : that.value != null) return false;
 
         return true;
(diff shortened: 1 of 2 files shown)


psiroky:
retest this please

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/BaseSingleFieldConstraint.java
```diff
@@ -138,7 +138,7 @@ public void deleteParameter( String key ) {
 
     public Map<String, String> getParameters() {
         if ( this.parameters == null ) {
-            this.parameters = new HashMap<String, String>();
+            return new HashMap<String, String>();
```
triceo: Is this change necessary? This changes the behavior in that the map will not be referenced in the class; and therefore changes to the map outside of the class will have no effect inside the class. This seems harmful - why give people the option of updating the map without it actually achieving anything?

yurloc: @triceo Thank you for pointing this out. While the code before the change may have been "ugly" because it allowed to construct a clone that was _not equal_ to original, my change makes the behavior more "dangerous". I will provide a better solution.

The change addresses a situation where `RuleModel` cloned by [RuleModelCloneVisitor](https://github.com/droolsjbpm/drools-wb/blob/a2af613/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/RuleModelCloneVisitor.java) was not equal to the original. It is because `equals()` of `BaseSingleFieldConstraint` checks `parameters` field like this:

``` java
public boolean equals(Object o) {
    ...
    if (parameters != null ? !parameters.equals(that.parameters) : that.parameters != null)
            return false;
    ...
}
```

and parameters are cloned like this:

``` java
    for ( Map.Entry<String, String> entry : original.getParameters().entrySet() ) {
        clone.setParameter( entry.getKey(), entry.getValue() );
    }
```

When original had no parameters `getParameters()` call created new `HashMap` instance in the original but clone's `parameters` remained `null` (because `setParameter()` was never called) and so clone was not equal to the original.

manstis:
@yurloc Please advise when these PRs are good to go...

yurloc:
In the end I changed equals() to consider two objects equal even if one has `parameters == null` and the other one's `parameters` is an empty map. I believe this one is the only possible solution without affecting behavior of [get|set]Parameters() which already have a lot of usages.

@manstis ready

manstis:
(master) http://github.com/droolsjbpm/drools/commit/730386c3c

(6.3.x) http://github.com/droolsjbpm/drools/commit/93677291d

PR #487, opened 2015-09-01
DROOLS-899: KIE-CI sends MavenRequest without considering the repositories defined in user settings file

manstis:
I moved resolution of additional repositories from MavenRepository to a new class MavenRepositoryUtils that is then used by MavenEmebedder. I could not reference MavenRepository from MavenEmbedder as the former instantiates a MavenEmbedder when constructing Aether and a circular loop develops. MavenRepository makes use of the MavenRepositoryUtils class but also appends repositories defined in the Aether class.

I really need to test this with the clr.mortgages example and execution server before this is merged... but that'll require either temporaribly backporting to 6.2.x or running Execution Server in 6.3.x.... that'll teach me to bail out of Edson's presentation last week :(

### DIFF
--- a/kie-ci/src/test/resources/org/kie/scanner/embedder/settings_with_repositories.xml
+++ b/kie-ci/src/test/resources/org/kie/scanner/embedder/settings_with_repositories.xml
@@ -1,0 +1,38 @@
+<settings>
+
+  <servers>
+    <server>
+      <id>kie-wb-m2-repo</id>
+      <username>testadmin</username>
+      <password>admin1234;</password>
+      <configuration>
+        <wagonProvider>httpclient</wagonProvider>
+        <httpConfiguration>
+          <all>
+            <usePreemptive>true</usePreemptive>
+          </all>
+        </httpConfiguration>
+      </configuration>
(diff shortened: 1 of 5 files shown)


manstis:
I also need to write a test for this before it's merged...

manstis:
Test written and merged into this PR.

kie-ci/src/test/resources/org/kie/scanner/embedder/settings_with_repositories.xml
```diff
@@ -0,0 +1,38 @@
+<settings>
+  <localRepository>/home/mwinkler/QA/jenkins/WORKSPACE/business-central-maven-repo</localRepository>
```
psiroky: Is this correct? Looks like the path is not used, otherwise the tests using it would fail.

manstis:
@psiroky The line was not used in the context of the test using the file; but I've removed it none the less to avoid future confusion.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/8e3e0c9ab

PR #489, opened 2015-09-02
Fix two rule model issues related to FormLineExpression

yurloc:
- Avoid StackOverflowError caused by a loop in ExpressionPart#equals().
- Extend ExpressionVisitor to handle ExpressionFieldVariable classes.

Not ready to merge yet. Help with implementing `ToStringExpressionVisitor#visit(ExpressionFieldVariable)` needed.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionPart.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionPart.java
@@ -104,10 +104,8 @@ public abstract class ExpressionPart
         if (classType != null ? !classType.equals(that.classType) : that.classType != null) return false;
         if (genericType != null ? !genericType.equals(that.genericType) : that.genericType != null) return false;
         if (name != null ? !name.equals(that.name) : that.name != null) return false;
-        if (next != null ? !next.equals(that.next) : that.next != null) return false;
         if (parametricType != null ? !parametricType.equals(that.parametricType) : that.parametricType != null)
             return false;
-        if (prev != null ? !prev.equals(that.prev) : that.prev != null) return false;
 
         return true;
     }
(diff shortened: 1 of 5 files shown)


drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionPart.java
```diff
@@ -104,10 +104,8 @@ public boolean equals(Object o) {
         if (classType != null ? !classType.equals(that.classType) : that.classType != null) return false;
         if (genericType != null ? !genericType.equals(that.genericType) : that.genericType != null) return false;
         if (name != null ? !name.equals(that.name) : that.name != null) return false;
-        if (next != null ? !next.equals(that.next) : that.next != null) return false;
         if (parametricType != null ? !parametricType.equals(that.parametricType) : that.parametricType != null)
             return false;
-        if (prev != null ? !prev.equals(that.prev) : that.prev != null) return false;
```
yurloc, line 110: When FormLineExpression has multiple ExpressionParts this caused StackOverflowError because `next.equals()` called `next.previous.equals()` (which is `this.equals()`) and that created an infinite loop.

Moreover it is redundant to consider two ExpressionParts equal only if all parts of the chain are equal. Equality of the collection of ExpressionParts is handled in [ExpressionFormLine](https://github.com/droolsjbpm/drools/blob/70405fb2d/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionFormLine.java#L162).

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/visitors/ToStringExpressionVisitor.java
```diff
@@ -112,6 +113,11 @@ public void visit( ExpressionCollectionIndex part ) {
         moveNext( part );
     }
 
+    public void visit( ExpressionFieldVariable part ) {
+        // FIXME implement this
+        visit( (ExpressionText) part);
+    }
```
yurloc, line 122: @manstis please, how should ExpressionFieldVariable be converted to string?

manstis, line 122: @yurloc Hi Jiri.. looking at the code `ExpressionFieldVariable` is used when a "binding" relates to a field and not a FactType. I could only find it's use [here](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/ExpressionBuilder.java#L216). Consequentially it will be OK to use the same String construction as for `ExpressionVariable`.

yurloc:
@manstis Ready. Thanks for reviewing!

manstis:
_drools_
(master) https://github.com/droolsjbpm/drools/commit/90d36eb72

(6.3.x) https://github.com/droolsjbpm/drools/commit/2e480f674

manstis:
Thanks @yurloc

PR #492, opened 2015-09-10
[DROOLS-904] Trait proxy classes are not available on session deseria…

mswiderski:
…lization -regenerated protobuf"


mariofusco:
Merged with https://github.com/droolsjbpm/drools/commit/c1e310289

PR #494, opened 2015-09-11
Serialization fixes (for Remote Serivces) and a command for work item ids

mrietveld:
In particular, I need to push the serialization commit (or a modified version) before Monday.


mrietveld:
The `JaxbListWrapper` class was kept because it seems to be more commonly used by clients. The `JaxbCollectionWrapper` was basically an exact copy of the list wrapper class.

PR #498, opened 2015-09-14
JBPM-4764 - NPE on fact delete after conditional event using RuleAwar…

mswiderski:
…eProcessEventLister - reproducer


mariofusco:
Test case merged and bug fixed with this commit https://github.com/droolsjbpm/drools/commit/6ae5f4d57f806a0ee22e98bd70f50bce4e28b7b0

PR #499, opened 2015-09-16
BZ1263358: KieBuilder does not read pom.xml from KieFileSystem if setPomModel() is used before buildAll()

manstis:
This includes a test demonstrating the issue we had in kie-wb; where we set the PomModel on multiple successive KieBuilder instances which led to the KJAR resulting from KieBuilder.buildAll() not containing a pom.xml or pom.properties file.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
@@ -15,6 +15,18 @@
 
 package org.drools.compiler.kie.builder.impl;
 
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Properties;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
 import org.drools.compiler.builder.impl.KnowledgeBuilderConfigurationImpl;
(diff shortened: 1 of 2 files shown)


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
```diff
@@ -432,6 +441,9 @@ public PomModel getPomModel() {
      */
     public void setPomModel( PomModel pomModel ) {
         this.pomModel = pomModel;
+        if ( srcMfs.isAvailable( "pom.xml" ) ) {
```
manstis, line 444: This is the actual fix!

drools-compiler/src/test/java/org/drools/compiler/builder/KieBuilderTest.java
```diff
@@ -293,22 +295,73 @@ public void testInvalidPomXmlContent() throws ClassNotFoundException, Interrupte
...
+            readPom1 = readPom1 + (char) charCode;
+        }
+        reader1.close();
+
+        assertEquals( pom,
+                      readPom1 );
+
+        //Read pom.xml from second KieBuilder's KieModule
+        InternalKieModule kieModule2 = (InternalKieModule) ( (KieBuilderImpl) kieBuilder2 ).getKieModuleIgnoringErrors();
+        final Reader reader2 = kieModule2.getResource( "META-INF/maven/org.kie.test/pomModelReuse/pom.xml" ).getReader();
```
manstis, line 354: w/o the fix this results in a NPE as the KieModule does not have a pom.xml

PR #500, opened 2015-09-17
fix drools parserhelper.reportError's outofindexexception

ThomasLau:
On my mac, this code may result in a exception

```
@BeforeClass
    public static void setup() throws IOException, RuleException {
        executor = new PostExecutor();
        String ruleFileDircetory = "src/test/resources/rule-post/";
        File ruleDirectory = new File(ruleFileDircetory);
        String apath = ruleDirectory.getAbsolutePath();
        File[] rules = ruleDirectory.listFiles();
        for (File ruleFile : rules) {
            StringBuilder sb = new StringBuilder();
            TestUtils.readFile(new FileInputStream(ruleFile), sb);
            String ruleStr = sb.toString();
            Statement st = new Statement();
            st.setRuleId(ruleFile.getName());
            st.setVersion(1);
            st.setContent(ruleStr);
                executor.updateKnowledge(st);
        }
    }
```

the out put will be:

```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
    at java.util.ArrayList.rangeCheck(ArrayList.java:653)
    at java.util.ArrayList.get(ArrayList.java:429)
    at org.antlr.runtime.CommonTokenStream.LT(CommonTokenStream.java:106)
    at org.drools.compiler.lang.ParserHelper.reportError(ParserHelper.java:388)
    at org.drools.compiler.lang.DRL6Parser.compilationUnit(DRL6Parser.java:134)
    at org.drools.compiler.lang.AbstractDRLParser.compilationUnit(AbstractDRLParser.java:86)
    at org.drools.compiler.compiler.DrlParser.compile(DrlParser.java:238)
    at org.drools.compiler.compiler.DrlParser.parse(DrlParser.java:155)
    at org.drools.compiler.compiler.DrlParser.parse(DrlParser.java:144)
    at org.drools.compiler.builder.impl.KnowledgeBuilderImpl.drlToPackageDescr(KnowledgeBuilderImpl.java:445)
    at org.drools.compiler.builder.impl.KnowledgeBuilderImpl.addPackageFromDrl(KnowledgeBuilderImpl.java:433)
    at org.drools.compiler.builder.impl.KnowledgeBuilderImpl.addKnowledgeResource(KnowledgeBuilderImpl.java:653)
    at org.drools.compiler.builder.impl.KnowledgeBuilderImpl.add(KnowledgeBuilderImpl.java:2164)
    at org.drools.compiler.builder.impl.KnowledgeBuilderImpl.add(KnowledgeBuilderImpl.java:2153)
    at org.thomas.drools.BasicExecutor.updateKnowledge(BasicExecutor.java:84)
    at org.thomas.drools.PostExecutorTest.setup(PostExecutorTest.java:35)
    at org.thomas.drools.PostExecutorTest.main(PostExecutorTest.java:44)
java.lang.RuntimeException: org.drools.compiler.compiler.DroolsParserException: Unexpected exception raised while parsing. This is a bug. Please contact the Development team :
Index: 1, Size: 1
```

i think  this is because AntLR use code below to parse the .DS_Store file, but not so correctlly:

```
@Override
    public Token LT(int k) {
        //System.out.println("enter LT("+k+")");
        if ( p == -1 ) setup();
        if ( k == 0 ) return null;
        if ( k < 0 ) return LB(-k);
        int i = p;
        int n = 1; // we know tokens[p] is a good one
        // find k good tokens
        while ( n<k ) {
            // skip off-channel tokens
            i = skipOffTokenChannels(i+1);
            n++;
        }
        if ( i>range ) range = i;
        return tokens.get(i);
    }
```

the code assumed that EOF should be last token, but in this case it won't. so i think ParserHelper.reportError need to check it .

tks.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/lang/ParserHelper.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/lang/ParserHelper.java
@@ -336,8 +336,13 @@ public class ParserHelper {
     }
 
     public void reportError( Exception e ) {
-        errors.add( errorMessageFactory.createDroolsException( e,
-                                                               input.LT( 1 ) ) );
+        if (input.index()-input.range()<2) {
+            errors.add( errorMessageFactory.createDroolsException( e,
+                                                                   input.LT( -1 ) ) );
+        }else{
+            errors.add( errorMessageFactory.createDroolsException( e,
+                                                                   input.LT( 1 ) ) );
+        }
     }
 
     /** return the raw DroolsParserException errors */
--- a/drools-compiler/src/test/java/org/drools/compiler/lang/DRLParserHelperTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/lang/DRLParserHelperTest.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.drools.compiler.lang;
+
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+import org.kie.api.io.ResourceType;
+import org.kie.internal.builder.KnowledgeBuilder;
+import org.kie.internal.builder.KnowledgeBuilderFactory;
+import org.kie.internal.io.ResourceFactory;
+
+public class DRLParserHelperTest {
+
+    /**
+     * this test will throw a {@link RuntimeException} caused 
+     * by an {@link IndexOutOfBoundsException}<br>
+     * 
+     */
+    @Test
+    public void test() throws Exception {
+        byte[] content=new byte[]{0x04,0x44,0x00,0x00,0x60,0x00,0x00,0x00};
+        //"content" may come from any .DS_Store file of Mac,
+        //i just use this 8-Byte instead to reproduce the case
+        
+        KnowledgeBuilder knowledgeBuilder = KnowledgeBuilderFactory
+                .newKnowledgeBuilder();
+        try {
+            knowledgeBuilder.add(ResourceFactory.newByteArrayResource(content), 
+                    ResourceType.DRL);
+        } catch (Exception e) {
+            assertTrue(e instanceof RuntimeException);
+            e.printStackTrace();
+        }
+    }
+}


mariofusco:
Sorry but I haven't understood how to reproduce this and then I'm also not understanding which use case this pull request is going to fix. Could you please also add a unit test reproducing this issue in your pull request?

ThomasLau:
@mariofusco  sorry for late , i just now add a test,pls check.

mrietveld:
@ThomasLau why would you want to add the `.DS_Store` file to a knowledge base?

ThomasLau:
@mrietveld i think i was not willing to add the .DS_Store file,i code like this
**String ruleFileDircetory = "src/test/resources/rule-post/"** and it runs well on Linux or Windows,but fails on Mac OS.

mrietveld:
@ThomasLau Ah, I can understand that. However, it doesn't make sense for us to pass some of the most internal code we have (the _parser_) in order to accommodate this feature. This precedent would then mean that we would modify the parser for every possible file that users wanted to exclude. 

To be clear: the purpose of the parser classes is to parse files. Basic OO philosophy dictates that those classes should only be used for parsing, not for other purposes. That's why I don't think that we want to add code that does other things to those classes. 

One possible solution for you is to simply filter on the `.DS_Store` class. For example:  

```
@BeforeClass
    public static void setup() throws IOException, RuleException {
        executor = new PostExecutor();
        String ruleFileDircetory = "src/test/resources/rule-post/";
        File ruleDirectory = new File(ruleFileDircetory);
        String apath = ruleDirectory.getAbsolutePath();
        File[] rules = ruleDirectory.listFiles();
        for (File ruleFile : rules) {
            // exclude .DS_Store files
            if( ruleFile.getName().equals(".DS_Store") ) {
                continue;
            }
            StringBuilder sb = new StringBuilder();
            TestUtils.readFile(new FileInputStream(ruleFile), sb);
            String ruleStr = sb.toString();
            Statement st = new Statement();
            st.setRuleId(ruleFile.getName());
            st.setVersion(1);
            st.setContent(ruleStr);
                executor.updateKnowledge(st);
        }
    }
```

psiroky:
I am not sure I understand this issue correctly, but I believe that the Drools parser should never return `java.lang.IndexOutOfBoundsException` to the end user.

mrietveld:
@psiroky Hmm.. that's a good point.

mariofusco:
I honestly don't understand why you're trying to parse the .DS_Store file with the drools parser. Also the test case you created doesn't make much sense in my opinion. It is not said that you should be able to send any arbitrary combination of bytes to the parser otherwise we should fill it with all sort of verbose and useless defensive programming.

ThomasLau:
@mariofusco hi, I honestly don't understand why you're always asking why am i trying to parse the .DS_Store file, as the error message write  **"java.lang.RuntimeException: org.drools.compiler.compiler.DroolsParserException: Unexpected exception raised while parsing. This is a bug. Please contact the Development team "**, the outofindexexception was made by Drools,not AntLR.

```
public void reportError( Exception e ) {
      errors.add( errorMessageFactory.createDroolsException( e, input.LT( 1 ) ) );
}
```

sotty:
I see two separate issues here:
- @ThomasLau it is not safe to try to parse all the files in a folder as DRLs unless you can guarantee that there will only be DRL files. The DS_Store is but one example of hidden/system files that should not be excluded. Rather than an exclusion filter, I would use an inclusion one on ".drl" files. 
  The parser will not be implemented defensively.
  However, this said
- @mariofusco that reportError method is MOSTLY invoked to contextualize a caught RecognitionException, but in the compilationUnit method it is used to handle generic exceptions as well. In that case, the assumption that there is at least one more token in the stream may not be appropriate.
  So, maybe the right fix is to handle generic exceptions differently

ThomasLau:
@sotty yes,it is unwise to list files without a file filter. i agree that this exception _may_ should be handled properly.

mariofusco:
Fixed by https://github.com/droolsjbpm/drools/commit/6498da712

PR #501, opened 2015-09-18
BZ1264321: Guided Rule Editor: replace or replaceAll condition parameters on a String missing on re-opening rule

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1264321

The issue was caused by a `ExpressionMethod`s parameters being stored as a Map keyed off the data-type; so `replace` has two parameters both of which are `String` so the first parameter was being overwritten by the second leading to the `replace` method invocation only having one parameter.

I changed this to be keyed off `ExpressionMethodParameterDefinition` which holds index and data-type information.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionMethodParameterDefinition.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionMethodParameterDefinition.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2014 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.workbench.models.datamodel.rule;
+
+import org.drools.workbench.models.datamodel.util.PortablePreconditions;
+
+/**
+ * Meta Data for a ExpressionMethod's parameter definitions.
+ */
+public class ExpressionMethodParameterDefinition {
+
+    private int index;
+    private String dataType;
+
+    public ExpressionMethodParameterDefinition() {
+        //Empty constructor for Errai marshalling. Cannot use @MapsTo since we don't want any Errai JAR dependencies here.
+    }
+
+    public ExpressionMethodParameterDefinition( final int index,
+                                                final String dataType ) {
+        this.index = index;
+        this.dataType = PortablePreconditions.checkNotNull( "dataType",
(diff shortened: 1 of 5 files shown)


manstis:
@mariofusco I've assigned this PR to you as the problem also manifested itself when unmarshalling `DRL`->`RuleModel`.. something you know a little about (and there's nobody else around that knows anything about any of this).

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/ExpressionMethodParameterDefinition.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return dataType.equals( that.dataType );
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = index;
+        result = 31 * result + dataType.hashCode();
+        return result;
```
yurloc, line 71: @manstis Missing `result = ~~result;`? (Just guessing.)

manstis, line 71: Dohhhhhhh!

triceo, line 71: I know I've already seen an explanation for this, but can't find it for the life of me. @manstis, maybe you could point me someplace where the `~~` is explained?

yurloc, line 71: @triceo https://groups.google.com/forum/#!topic/google-web-toolkit/GDDMpq5X6-8

triceo, line 71: Interesting - one would expect this to be well documented somewhere in GWT itself. In any case - is it reasonable to expect people to just know about this? Since forgetting about this will automatically create a bug, is there something we can do to make sure this doesn't happen?

Some options come to mind - maybe a checkstyle rule, or creating (and enforcing the use of) some sort of an `GwtHashCodeBuilder`.

manstis, line 71: IDK if it's "well known".. it took a lot of hunting to ascertain why we had
some bugs in kie-wb that ended up being duff hashCode implementations.

We could possibly add a checkstyle rule.. but I don't think we fail the
build for them atm.. and they're of less use unless people are forced to
take action.

Enforcing use of a GwtHashCodeBuilder takes the same discipline as looking
for use of ~~

triceo, line 71: @manstis Thanks - I didn't mean to say that it was well known; merely that I would expect the authors of GWT to tell people about this strange requirement. :-)

Wrt. the discipline - you are right. Only with the hashcode builder, we could use Javadoc to document why the `~~` is even necessary. I'll discuss with Petr whether or not there is some smart way of making sure we don't miss this in the future.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/2166a1e92

PR #502, opened 2015-09-18
BZ1264339: Guided Rule Editor: List contains condition missing bound value on re-opening rule - unit test.

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1264339

Unit Test for use-case. No issue was found in code-base; tested in UI too.


manstis:
_drools_
(master) http://github.com/droolsjbpm/drools/commit/d8eefede4

Merged myself - as it's only(!) a Unit Test.

PR #503, opened 2015-09-19
[DROOLS-923] make KieRepository thread safe

mrietveld:
@mariofusco You have a better idea of the concurrency problems that this code might deal with, but I saw a couple of use cases that I think should be fixed regardless (in particular, there are some places where `putIfAbsent` or other `ConcurrentMap` variants could be used instead of the normal `Map` methods).  

Let me know if you have questions, there are a lot of different race conditions that this code deals with: I tried to add good explanations in the comments, but that's not always enough (or, of course, I might have made a mistake.. :) ).

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -151,8 +151,92 @@
       <groupId>org.jboss.weld.se</groupId>
       <artifactId>weld-se-core</artifactId>
       <scope>test</scope>
-    </dependency>     
+    </dependency>
+         
+   <!-- test: byteman -->
+    <dependency>
+      <groupId>org.jboss.byteman</groupId>
+      <artifactId>byteman</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.byteman</groupId>
+      <artifactId>byteman-submit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.byteman</groupId>
+      <artifactId>byteman-install</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.byteman</groupId>
+      <artifactId>byteman-bmunit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
+  
+  <build>
+    <plugins>
+      <plugin>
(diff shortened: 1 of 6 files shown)


mrietveld:
@mariofusco Added tests. The commits are structured as follows: 
1. [28e796bc](https://github.com/mrietveld/drools/commit/28e796bc?w=1) Contains tests that expose some of the more prominent race conditions and currently fail. 
2. [dd4e2078](https://github.com/mrietveld/drools/commit/dd4e2078?w=1) Changes that fix the problems
3. [65450ac4](https://github.com/mrietveld/drools/commit/65450ac4?w=1) Changes to the tests to accompany the new code (since the tests would deadlock on some synchronize blocks otherwise)

mariofusco:
When I try to run your test inside maven I get the following failures:

Tests in error: 
  KieModuleRepoTest.org.drools.compiler.kie.builder.impl.KieModuleRepoTest »  BM...
  KieModuleRepoTest.removeStoreArtifactMapTest:252->waitFor:58 »  test timed out...

Moreover if I try to run the same tests inside intellij I get the following Exception. As you anticipated this could be a problem with BMUnit that at the moment can work only with Eclipse ClassLoader.

java.io.FileNotFoundException: Rule file not found for Byteman test case org.drools.compiler.kie.builder.impl.KieModuleRepoTest#byteman/newerVersionDeployOverwritesTest.btm
    at org.jboss.byteman.contrib.bmunit.BMUnit.loadScriptFile(BMUnit.java:195)
    at org.jboss.byteman.contrib.bmunit.BMUnitRunner$7.evaluate(BMUnitRunner.java:270)
    at org.jboss.byteman.contrib.bmunit.BMUnitRunner$6.evaluate(BMUnitRunner.java:241)
    at org.junit.internal.runners.statements.FailOnTimeout$StatementThread.run(FailOnTimeout.java:74)

Finally please send this PR against master branch instead of 6.3.x

mrietveld:
@mariofusco Which version of Intellij (and which JDK) are you using? Using JDK 7 or 8 on Intellij 14, it works for me.

mariofusco:
We agreed that this code is too complex and reliability and maintainability are much more important than performances in this specific part of code.

This PR has been rectified accordingly with this https://github.com/droolsjbpm/drools/pull/518

PR #505, opened 2015-09-24
DROOLS-921 fix event propagation from LIA node

sotty:


mariofusco:
Fixed by https://github.com/droolsjbpm/drools/commit/9b479f9b7

PR #507, opened 2015-10-13
Fix for NPE

mrietveld:
The `typeDeclaration` variable is null at this point in the code. I'm pretty sure that it should be the `newDecl` variable.


mariofusco:
I'm pretty sure you're right, but do you have a test case reproducing the NPE? How did you discover this problem?

mrietveld:
My IDE (eclipse) flags NPE's. :)

See: https://github.com/droolsjbpm/drools/pull/507/files#diff-0c1d41c38edec652a3783f36c0829dfcL916

PR #508, opened 2015-10-14
DROOLS-948 testEventWithShortExpiration() in CepEspTest and CepJavaTypeTest fails

baldimir:
- Added sleep times.


mrietveld:
-1 We should not be adding tests with sleep times in them: `Thread.sleep(..)` _ensures_ that this test will cause a false negative. 

I believe it should be possible to add `RuleEventListener`s (or similar) that use `java.util.concurrent.*` objects to mimic the chronology that you want tested.

baldimir:
Agree. I'll change the tests accordingly.

PR #509, opened 2015-10-14
BZ1269366: RuleDescr has null rule name when rule's name includes [0-9]_[0-9]

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1269366

This PR contains a `@Ignore` failing unit test.


mariofusco:
See my comment here https://bugzilla.redhat.com/show_bug.cgi?id=1269366#c2

PR #510, opened 2015-10-14
BZ1271534: drools-compiler fails to report error for invalid rule with field binding

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1271534


mariofusco:
Reproduced and fixed by https://github.com/droolsjbpm/drools/commit/ce7e91825e64b49b74b5d3db3b9df960bd4e4439

PR #511, opened 2015-10-15
ClasspathKieProject should not deal ".jar" always like a file

Dawuid:
Some containers allow deployments of artifacts in "expanded" mode (instead of a unique jar/ear/war/... file, al artifacts are deployed full decompressed, so they are directories). This is default with last JBoss Tools and Wildfly (unless you use compressed option for deployments). In that case drools fails to load jar contents.

With this patch jar files are checked that they are actually files. In other case they are treated like directories.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
@@ -211,7 +211,7 @@ public class ClasspathKieProject extends AbstractKieProject {
             rootPath = IoUtils.asSystemSpecificPath( rootPath, rootPath.lastIndexOf( ':' ) );
         }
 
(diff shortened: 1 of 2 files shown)


mrietveld:
Would you be able to add a unit test for this code?

etirelli:
The patch looks good, we just need a unit test to go with it.

Dawuid:
Working on that unit test. I will upload in the coming days.

mariofusco:
I agree the patch looks ok. I'll merge it as soon as we will have that unit test.

Dawuid:
I'm trying to make a unit test based in any similar but I'm not finding anyone in drools-compiler or kie-ci (strangely no one is stopping in my breakpoints). Could you point me out any test that I can use like "inspiration"?

mrietveld:
`getPomProperties` is a public static method, I would write a unit test that calls this method and tests the results, possibly using Mockito methods. Does that make sense?

Dawuid:
At last unit test is provided! :-)

psiroky:
ok to test

psiroky:
I've slightly tweaked the test and merged into master.

Thanks @Dawuid!

PR #513, opened 2015-10-19
[DROOLS-954] fix removal of a rule with a sub-subnetwork

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/IncrementalCompilationTest.java
```diff
@@ -2043,4 +2047,38 @@ public void testSegmentMergeOnRuleRemovalWithNotExistingSegment() throws Excepti
...
+
+        KieServices ks = KieServices.Factory.get();
+
+        ReleaseId releaseId1 = ks.newReleaseId("org.kie", "test-upgrade", "1.1.1");
+        KieModule km = createAndDeployJar(ks, releaseId1, drl1);
+
+        KieContainer kc = ks.newKieContainer(km.getReleaseId());
+        KieSession ksession = kc.newKieSession();
+
+        kc.updateToVersion(releaseId1);
```
etirelli, line 2071: Out of curiosity, why do you call updateToVersion() using the same releaseId the current container was created with?

etirelli:
+1 to merge

drools-compiler/src/test/java/org/drools/compiler/integrationtests/IncrementalCompilationTest.java
mariofusco, line 2071: That's just a copy/paste mistake. I'll remove it, thanks.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/1a6710b7f

PR #514, opened 2015-10-19
[DROOLS-951] preserve PathMemory linking mask if the path is not chan…

mariofusco:
…ged after a rule addition or removal


etirelli:
Mario, I think a quick change in the test itself is necessary, or maybe create two tests. The current test only works because the Message class does not implement equals()/hashcode(). If someone for any reason implements theses methods, it will break the test as all insertLogicals() should be equality based, not identity based.

Suggestion: add the equals()/hashcode() to the class, then use different strings for the message in each of the rules. For bonus points, create a second test with equal strings as you have now, but then since you have multiple rules justifying the same message, the number of message objects in memory should always be 1, until the last rule is removed.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/bc9dfb8b2

PR #515, opened 2015-10-19
[DROOLS-944] fix FieldReaders serialization

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/rule/builder/QueryElementBuilder.java
```diff
@@ -499,11 +496,7 @@ public static boolean isVariable( String str ) {
             }
         }
 
-        if ( str.endsWith( ".class" ) ) {
-            return false;
-        }
-
-        return true;
+        return !str.endsWith( ".class" );
```
psiroky, line 499: This is shorter, but imo bit harder to read :) (but I am fine with leaving it as is)

drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -8205,4 +8205,55 @@ public void testCompilationFailureWithNonExistingField() {
...
+        private Map<String, Object> objects = new HashMap<String, Object>();
+
+        public Map<String, Object> getObjects() {
+            return objects;
+        }
+
+        public void setObjects(Map<String, Object> objects) {
+            this.objects = objects;
+        }
+    }
```
psiroky, line 8211: I am missing an assert in this test. I know the JIRA is about NPE during de-serialization, but still, could we add an assert that would verify the kbase is the same before and after the serialize+de-serialize calls?

drools-core/src/main/java/org/drools/core/base/BaseClassFieldReader.java
```diff
@@ -184,4 +190,22 @@ public int getHashCode(Object object) {
...
+                                                    ClassNotFoundException {
+        index = in.readInt();
+        valueType = (ValueType) in.readObject();
+        String clsName = in.readUTF();
+
+        try {
+            fieldType = in instanceof DroolsObjectInput ?
+                        Class.forName( clsName, false, ( (DroolsObjectInput) in ).getClassLoader() ) :
+                        Class.forName( clsName );
+        } catch (ClassNotFoundException e) { }
```
psiroky: Could you please describe why we don't care about the CNFE? I am sure there is a good reason and I would like to know as my knowledge of Drools is still very limited...

drools-core/src/main/java/org/drools/core/base/BaseClassFieldReader.java
```diff
@@ -184,4 +190,24 @@ public int getHashCode(Object object) {
...
+        index = in.readInt();
+        valueType = (ValueType) in.readObject();
+        String clsName = in.readUTF();
+
+        try {
+            fieldType = in instanceof DroolsObjectInput ?
+                        ClassUtils.getClassFromName( clsName, false, ( (DroolsObjectInput) in ).getClassLoader() ) :
+                        ClassUtils.getClassFromName( clsName );
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException( e );
```
psiroky, line 210: Is there more info we could provide in the exception message? I mean what could be or is the reason the CNFE was thrown, e.g. de-serializing fact which is not on current classpath (just an example, I am not exactly sure what is the cause in this case).

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/f0b787326

PR #516, opened 2015-10-20
ENTESB-4090 - Fixing fuse integration (Fuse hack breaks kie-remote-client usage in Fuse code)

mrietveld:
Hi Mario, 

This PR is caused by [ENTESB-4090](https://issues.jboss.org/browse/ENTESB-4090) -- in short David Ward (Fuse/Switchyard) hacked the `kie-remote-client` code about 18 months ago, and "added" functionality that wasn't there: basically, he disabled the internal checks and forced the client API to support the the BatchExecutionCommand. 

..and then he didn't file a bugzillla, or a jira or tell anyone. So 1 month ago, Fuse QE found the bug (because the internal (`kie-remote-client`) code changed, of course) and filed this https://issues.jboss.org/browse/ENTESB-4090

So, this PR does the following: 
1. Fixes serialization for the `BatchExecutionCommandImpl` and all of the commands it references. 
2. Of course, then that showed that the "filling" algorithm for `FlatQueryResults` and related classes was broken, so I fixed that as well (Now I know more about the internals of the drools Query API than I wanted to.. :_D )

All of this code thus makes it possible for Fuse to use the `BatchExecutionCommand`, which they use in order to seamlessly integrate with a BPMS instance. 

(Well, then there are changes that I need to finish in the (1) `kie-remote-`\* modules, and then changes in the (2) `fuse-bxms-integ` modules, but (1) depends on this, and (2) depends on (1).. ) 

If you want me to split this PR into more commits, just let me know. 

Thanks.


mrietveld:
The following tests are also failing on master: 

```
org.drools.pmml.pmml_4_2.predictive.models.NaiveBayesTest.testNaiveBayes
org.drools.pmml.pmml_4_2.predictive.models.NaiveBayesTest.testNaiveBayesWithGaussianDistr
```

mrietveld:
Added minor fixes to the `ModifyCommand` and the `UpdateCommand` (serialization/JAXB issues, that are tested in the `kie-remote-`\* modules.

mswiderski:
any need to verify this change against KIE Server?

drools-core/src/main/java/org/drools/core/QueryResultsImpl.java
```diff
@@ -93,30 +97,18 @@ public int size() {
         }
         Declaration[] parameters = getParameters();
 
-        Set<String> set  = new HashSet<String>();
```
mrietveld, line 96: @mariofusco This is probably what's causing the drools-pmml test failure: I'll go ahead and remove this change since it's unrelated to what I needed.

mrietveld:
@mswiderski Yes, this should be verified against exec-server, good point!

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/afd21c8c036d72ea66bc0114b5f1b869a47e5536

PR #518, opened 2015-10-21
KieModuleRepo work

mrietveld:
Hi Mario, 

I've left the old commits in order to document some of the work that happened. 

However, almost all of the work done in the 3rd commit has been thrown away: I basically just made every method on the `KieModuleRepo` _synchronized_, and that fixes any race conditions that there may have been. 

The easiest thing to do is probably to compare the files from the last commit with master. 

The main reason to do this was to keep the code as readable as possible, given that the concurrency code I wrote was performant and good, but complex and maybe harder to maintain. 

I couldn't find a good place to document the 2 new properties (for the LRU cache sizes): any suggestions? 

Thanks!

PS. once you approve this, I'll submit a PR for 6.3.x as well.

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -215,7 +215,6 @@
         </configuration>
       </plugin>
      
-      <!--  Re-enable once we upgrade byteman to 3.0.2
       <plugin>
         <groupId>org.jboss.byteman</groupId>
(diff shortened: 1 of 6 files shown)


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b44ab2f85

PR #521, opened 2015-10-23
[BZ-1267510] treat .gdst and .scgd as standard KIE resources

psiroky:
- .gdst (guided dtables) and .scgd (guided score cards) were
  treated in a special way using FormatConverters which caused
  issues during compilation as the resulting DRLs were copied
  in the compiled kjar. The behavior was changed to be in line
  with other resources like .xls dtables where the DRL representation
  is create at runtime, on the fly.
- RuleTemplateConverter was refactored to RuleTemplateProvider to
  have the name in sync with the others
- the FormatConverter abstraction was removed completely. All supported
  resources should be treated equaly to avoid different behavior
  for different resources (very much like the related BZ reported)


psiroky:
Related PRs:
https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/103
https://github.com/droolsjbpm/droolsjbpm-integration/pull/213

psiroky:
@manstis would you be able to take a quick look? Besides @mariofusco (on PTO) you are probably the only one that has the needed info as we discussed this issue over IRC.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDecisionTableProviderImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.InputStream;
+
+public class GuidedDecisionTableProviderImpl implements GuidedDecisionTableProvider {
+
+    @Override
+    public String loadFromInputStream(InputStream is) throws IOException {
+        String xml = new String(IoUtils.readBytesFromInputStream(is), IoUtils.UTF8_CHARSET);
+        GuidedDecisionTable52 model = GuidedDTXMLPersistence.getInstance().unmarshal(xml);
+
+        return GuidedDTDRLPersistence.getInstance().marshal(model);
```
manstis: I'm a little concerned whether this change will support Guided Decision Tables that use DSLs. The old `FormatConverter` could potentially alter the asset's file name (e.g. from .drl to .dslr if the decision table used DSLs). I don't see a similar mechanism with `ResourceTypes` - or whether it matters!?!

manstis: Perhaps a unit test using a .gdst that contains DSLs would put my mind at rest?

psiroky: Good point. Don't we already have test for this? I will take a look and in case there is none, I will add it to see what happens. 

You are right that the current code will always treat the generated String as DRL. Which will likely not work in case of DSL usage.

psiroky:
I've fixed the GDST parsing. If it contains DSL sentences it is parsed as DSLR, otherwise as DRL. I also added test which parses and compiles GDST with DSL.

psiroky:
jenkins, retest this

manstis:
I don't see your test for when the .gdst doesn't have DSLs.. Has it been removed?

manstis:
You might also like to consider moving `ConversionResult` since it'll be used by Guided Templates too (of course you can do that when you get to apply a similar change for those too)..

psiroky:
There seems to be no test for .gdst without DSL. I haven't created one and it seems there actually wasn't one before (or I am not looking at correct place). It would definitely be useful though.

psiroky:
About the ConversionResult. I don't like the name much, but wasn't able to come up with anything better. I will move it when fixing the BZ for rule templates.

manstis:
Heh, OK. I assumed your original PR contained one for .gdst's w/o DSLs.. since we then had the conversation about .gdst's with DSLs. I'm sure I saw some XML in your original PR.. and I'm sure you created some new XML this afternoon.... I must be going mad.

psiroky:
I've added test for .gdst without DSL. The tests are now `IntegraiontTest`s, so executed by failsafe instead of surefire (just a detail). I also added more asserts to make sure the parser actually created expected rules from the .gdst (checking just rule names, the marshalling logic is tested elsewhere).

psiroky:
jenkins, retest this

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/eda1a8b54bf1f4c39d13a1c543227595036663a0

PR #522, opened 2015-10-23
BZ-1274394: add reproducer

winklerm:
Added reproducer for BZ-1274394.

Later I found that the failing test (LogicalInsertTest#testChangeElement) uses the class Person which does not define its equals and hashcode methods correctly. When I added the age attribute to the equals method, the tests passes.

This seems to be a change between Drools 6.2.x and 6.3.x, is this intentional?


mariofusco:
As agreed this test case is invalid due to that wrong equals method. When the equals method is fixed the test succeeds as expected.

PR #523, opened 2015-10-23
BZ-1274696: add reproducer

winklerm:
Added a reproducer for a possible bug, or a change in behavior since 6.2.x. It is very likely that failing fast when attempting to delete an expired event is expected, in that case, please feel free to close this PR.

On the other hand, if this behavior is expected, is there a way how to delete the event "safely"? For instance, a method callable from the rule RHS, such as:

```
boolean deleteSafely(FactHandle handle) {
  if (kieSession.getFactHandles().contains(handle)) {
    delete(handle);
    return true;
  } else {
    return false;
  }
}
```

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
@@ -570,6 +570,38 @@ public class CepEspTest extends CommonTestMethodBase {
     }
 
     @Test(timeout=10000)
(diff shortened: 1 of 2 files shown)


winklerm:
A possible workaround (but might not be completely safe) might be to replace the `delete($event)` in the rule with:

```
if (kcontext.getKieRuntime().getFactHandles().contains($event)) {
  delete($event);
}
```

mariofusco:
Reproduced and fixed by https://github.com/droolsjbpm/drools/commit/3e455f3cf

PR #525, opened 2015-10-23
Bug 1266076 - DSL sentences not working in guided rule templates - 6.3.x

Rikkola:
Same, but for master: https://github.com/droolsjbpm/drools/pull/524
(cherry picked from commit a89b3185b047c533c68ce56e4b984ba440d53258)


manstis:
@Rikkola out of interest does this still work OK for Templates that do _not_ have DSLs?

I seem to think if we run the `Expander` over regular DRL we get errors (as it tries to expand "Person( age == 10 )" as opposed to "> Person( age == 10 )". This is why we have RuleModel.hasDSLSentences()```. I made a similar point on https://github.com/droolsjbpm/drools/pull/521

Rikkola:
I'll add a test for that, just to be sure.

Rikkola:
@manstis you are right, it does try that.

For guided rules we have different save formats if the rule uses DSL. We now need to add it for the guided template editor or do what Guided Decision Tables does (need to check what that is).

Closing this anyway.

PR #529, opened 2015-10-26
BZ1264339: Guided Rule Editor: List contains condition missing bound value on re-opening rule

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1264339

There is already a test; but the test was wrong so this PR fixes the test and fixes the issue.


Rikkola:
+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b25038feb

PR #530, opened 2015-10-26
[DROOLS-965] associate a RuleTerminalNode to its own Rule

mariofusco:


etirelli:
Looks good Mario. +1.

PR #533, opened 2015-10-27
[BZ-1266076] add support for guided rule templates with DSL

psiroky:
- guided templates use the same mechanism as guided dtables. If it
  contains DSL sentences, the generated string is treated as DSLR,
  otherwise as standard DRL


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/6bc7cdb4a

PR #535, opened 2015-10-27
[6.3.x] [BZ-1266076] add support for guided rule templates with DSL

psiroky:
- guided templates use the same mechanism as guided dtables. If it
  contains DSL sentences, the generated string is treated as DSLR,
  otherwise as standard DRL


psiroky:
Same changes already merged into master by https://github.com/droolsjbpm/drools/pull/533

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e88a97ee8

PR #536, opened 2015-10-28
[BZ-1271909] log all declarations even if not used in consequence

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/WorkingMemoryLoggerTest.java
```diff
@@ -36,4 +44,33 @@ public void testOutOfMemory() throws Exception {
...
+
+        Message message = new Message();
+        message.setMessage("Hello World");
+        ksession.insert(message);
+        ksession.fireAllRules();
+
+        for (LogEvent logEvent : logger.getLogEvents()) {
+            if (logEvent instanceof ActivationLogEvent) {
+                assertTrue( ((ActivationLogEvent) logEvent ).getDeclarations().contains( "$myMessage" ));
+            }
```
etirelli, line 73: Mario, just to prevent future regressions, wouldn't it be good to check for both $m and $myMessage variables here?

etirelli:
Looks good. +1 to merge!

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/2292d17dc

PR #537, opened 2015-10-29
[BZ-1276311] improve PatternBuilder performances

mariofusco:


etirelli:
Mario, +1 to merge. It is a good step forward.
I wonder if we can improve it even more by sharing a StringBuilder among the methods to prevent local concatenation (that behind the scenes is instantiating its own string builder every time)?

mariofusco:
I'm not doing this because I need to store the different subexpressions in the respective descriptors. This allows me to avoid resplitting the original complex constraint using regular expressions like it did before (probably in a bugged way as you made me notice).

PR #542, opened 2015-11-03
[BZ-1276379] fix kie module files discovery when dependencies are installed in JBoss EAP Module

mariofusco:


psiroky:
+0. I am not fully familiar with the issue and the underlying logic. The code itself is good. It would be nice to have some tests.

PR #543, opened 2015-11-04
[DROOLS-970] fix Date coercion when the constraint uses a declaration or an instance field

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
```diff
@@ -953,10 +949,26 @@ private String normalizeExpression( RuleBuildContext context, Pattern pattern, R
...
         return normalizeEmptyKeyword( subExpr, operator );
     }
 
+    private boolean isDateType( RuleBuildContext context, Pattern pattern, String leftValue ) {
+        Declaration declaration = pattern.getDeclarations().get( leftValue );
+        if (declaration != null) {
+            return declaration.getValueType() == ValueType.DATE_TYPE;
+        }
+
+        if (pattern.getObjectType() instanceof FactTemplateObjectType) {
```
etirelli, line 963: Do we still support FactTemplates? I think we should clean up the code and remove all references to FactTemplates as we haven't supported this since the Drools 4 days I believe.

etirelli:
+1 to merge, looks good.

mariofusco:
@etirelli I think that we should also consider to cherry-pick this commit to 6.3.x not only because it fixes the bug reported here https://issues.jboss.org/browse/DROOLS-970 but, probably more important, because while investigating it I've found another way to alleviate the build time performance degradation reported here https://bugzilla.redhat.com/show_bug.cgi?id=1276311

After this fix I rerun the benchmark mentioned in this comment https://bugzilla.redhat.com/show_bug.cgi?id=1276311#c1 obtaining the following result:

747.449 ± 3.265  ms/op

etirelli:
@mariofusco ok, submit another PR targeting 6.3.x and linking to the performance issue.

drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
mariofusco, line 963: I think so, or at least without that check this test case fails https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java#L6968

etirelli, line 963: Ok, leave the code as is for now, but we need to talk to Mark and clean up the code. I am sure there are problems with fact templates and I am not aware of anyone using them, so it is likely this is just code garbage now.

mdproctor, line 963: One reason we have not removed them, is I do believe there is a new for an abstracted type system in Drools, it's just type declarations need more work. I've left them in, as they expose a lot of the plumbing points, if we ever decide to re-investigate this work.

PR #546, opened 2015-11-05
DROOLS-951: add missing @Test annotation

winklerm:
Added missing @Test annotation. When run locally, the test passed.


etirelli:
+1

PR #547, opened 2015-11-05
[DROOLS-971] avoid IndexOutOfBoundsException in IoUtils when reading an empty InputStream

mariofusco:


mariofusco:
jenkins, retest this

psiroky:
+1 (just to be safe, I think we should wait on Jenkins build results)

PR #552, opened 2015-11-09
6.3.x activation group issue report

tatecn:


drools-compiler/pom.xml
```diff
@@ -38,6 +42,10 @@
       <groupId>org.kie</groupId>
       <artifactId>kie-internal</artifactId>
     </dependency>
+	<dependency>
+		<groupId>org.jbpm</groupId>
+		<artifactId>jbpm-bpmn2</artifactId>
+	</dependency>
```
mariofusco, line 48: You cannot add kie-ci and jbpm as a dependency to drools-compiler because this will cause a circular dependency.

mariofusco:
I'm sorry, but this pull request is not acceptable. You cannot make drools to depend on jbpm otherwise we will have a circular dependency. Moreover the problem you're trying to reproduce is not clear to me.

My suggestion is to open a issue on our jira explaining the problem you found and send a PR with a reproducer to the jbpm project.

PR #553, opened 2015-11-10
Fix for DROOLS-975 - added TransactionManagerFactory

lukenjmcd:
Added TransactionManagerFactory to remove coupling between Drools/jBPM classes and the JtaTransactionManager.

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManagerFactory.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManagerFactory.java
@@ -23,7 +23,7 @@ import org.slf4j.LoggerFactory;
 /**
  * Abstract factory for creating {@link org.drools.persistence.TransactionManager} instances..
  * 
- * The implementation class can be specified using the {@code opt.kie.tmfactory.class} system property.
+ * The implementation class can be specified using the {@code org.kie.txm.factory.class} system property.
  * 
  */
 public abstract class TransactionManagerFactory {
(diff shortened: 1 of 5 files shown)


drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManagerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ * 
+ * The implementation class can be specified using the {@code opt.kie.tmfactory.class} system property.
+ * 
+ */
+public abstract class TransactionManagerFactory {
+    private static final Logger logger = LoggerFactory.getLogger(TransactionManagerFactory.class);
+    private static volatile TransactionManagerFactory INSTANCE;
+
+    static {
+        String factoryClassName = System.getProperty("opt.kie.tmfactory.class", JtaTransactionManagerFactory.class.getName());
```
mswiderski: could we change the name of system property to be consistent with other system properties to something like this

```
org.kie.txm.factory.class
```

mswiderski:
ok to test

mswiderski:
in general looks good to me, one question - have you looked at droolsjbpm-integration repository to check if there is no need for any other alignments after introducing this factory? Important parts there are:
- kie-spring
- drools-osgi
  so might be worth double checking if it does work there the way it was designed - with factory

drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManagerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+            logger.error("Unable to instantiate "+factoryClassName, e);
+        }
+    }
+    
+    /**
+     * Sets the factory used for creating {@link TransactionManager}s.
+     * 
+     * @param factory
+     */
+    public static final void set(TransactionManagerFactory factory) {
```
mrietveld: Please remove the `public static final set` method: IMHO, this is far too dangerous and unneccessary, AFAICT.

lukenjmcd:
A couple questions (this is my first time using github or submitting to an open-source project) 

1 - is it my responsibility to make the changes being proposed here (change property name, remove set() method)?
2 - if yes to question 1, do I need to create a new pull request with those changes or is there some documentation about the process to follow for this?

Thanks.

psiroky:
@lukenjmcd it would be great if you could do those suggested changes. There is no need to create new PR, you can amend (update) this one. See for example http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/ for more details on how to do that.

mrietveld:
@lukenjmcd It's enough to modify your local branch and then `git push -f origin <pr-branch-name>`, where origin is your github repository name. (I'm assuming that "upstream" is the alias for droolsjbpm github repository).

lukenjmcd:
I've made the suggested changes.  I received a failure notification but I get "server not found" error when trying to see the details.  The tests pass on my local system.  

I also made a PR for a change in droolsjbpm-integration/kie-infinispan/drools-infinispan-persistence

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/d67f775502821d40be9404e7d3ced44f9976e97c

PR #554, opened 2015-11-11
DROOLS-976

calcacuervo:
This is a simple test to reproduce DROOLS-976


manstis:
The PR description states it's a failing test.. I doubt we'd want to merge
w/o @Ignoring test (or fixing issue)!

calcacuervo:
@manstis  yes. It is the failing test for the bug, not fixed

mariofusco:
I rewrote the test case using the new kie api and fixed this issue.

PR #557, opened 2015-11-16
BZ-1208056 - Concurrent process execution will fail if it references …

mswiderski:
…a new user/group - remove systout


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b9f1bb926

PR #558, opened 2015-11-16
[DROOLS-978] avoid to evaluate an agenda item in the eager list when an incremental compilation removed the corresponding rule

mariofusco:


etirelli:
Looks good @mariofusco . +1 to merge.

PR #561, opened 2015-11-20
[DROOLS-983] fix events expiration when using inheritance

mariofusco:


mariofusco:
ok to test

drools-core/src/main/java/org/drools/core/common/EventFactHandle.java
```diff
@@ -180,7 +181,23 @@ public void decreaseActivationsCount() {
...
+
+    public int getOtnCount() {
+        return otnCount;
+    }
+
+    public void setOtnCount( int otnCount ) {
+        this.otnCount = otnCount;
+    }
+
     public EventFactHandle clone() {
```
etirelli, line 201: Mario, I am not sure this is enough. On deserialization, don't we need to know exactly on which nodes the event is expired and which ones it is not to rebuild state? If that is the case, you might need to record the actual OTN references instead of just a counter, although I know it can be heavy, so we might need to look for alternatives. Can you create a test case for it please??

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -1763,18 +1764,19 @@ public void execute(InternalWorkingMemory workingMemory) {
...
                         this.node.retractObject(factHandle, context, workingMemory);
                     } else {
                         workingMemory.getEntryPoint(factHandle.getEntryPoint().getEntryPointId()).delete(factHandle);
                     }
+                    factHandle.decreaseOtnCount();
 
                     context.evaluateActionQueue(workingMemory);
                     // if no activations for this expired event
-                    if (factHandle.getActivationsCount() == 0) {
+                    if (factHandle.getOtnCount() == 0) {
```
etirelli, line 1777: In order to set the expired flag, checking for the OTN count is enough, but before deleting the event it also needs to check for activation count == 0.

etirelli:
Mario, please consider my comments before merging.

etirelli:
@mariofusco +1 to merge

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/62ac284f1

PR #563, opened 2015-11-25
GUVNOR-2286: Guided Rule Editor doesn't load rules correctly with checks in list elements values

manstis:
See https://issues.jboss.org/browse/GUVNOR-2286


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b89c11863

PR #565, opened 2015-11-25
DROOLS-956 - Add a central kie-test-util module

mrietveld:
This has been approved by both @krisv and @etirelli. 

@psiroky Would you mind looking at this?

### DIFF
--- a/kie-test-util/pom.xml
+++ b/kie-test-util/pom.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>drools</artifactId>
+    <groupId>org.drools</groupId>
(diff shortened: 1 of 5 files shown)


kie-test-util/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.kie</groupId>
+  <artifactId>kie-test-util</artifactId>
+
+  <name>KIE :: Test Utility Classes</name>
+  <!-- This module should only be added as a "test" scope dependency -->
+
+  <dependencies>
+    <!-- Please do not add any drools/jbpm/kie dependencies here. -->
```
psiroky, line 18: I would be even more strict here, we can not allow any drools/jbpm/kie  deps. WDYT about changing it to something like `Important! Do not add any drools/jbpm/kie dependencies here`

kie-test-util/src/main/java/org/kie/test/util/network/AvailablePortFinder.java
```diff
@@ -0,0 +1,161 @@
...
+                    } catch (IOException e) {
+                        /* should not be thrown */
+                    }
+                }
+            }
+        }
+
+        return result;
+    }
+}
```
psiroky, line 161: This is great! We can finally remove all those copy&pasted duplicates and use the one true port finder!

psiroky:
+1, merged into master. Thanks @mrietveld!

PR #567, opened 2015-11-29
DROOLS-992 Improved state machine for thread interactions

mdproctor:


drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1308,11 +1342,52 @@ public int fireAllRules(AgendaFilter agendaFilter,
...
+        public static RestHandler FIRE_ALL_RULES = new FireAllRulesRestHandler();
+        public static RestHandler FIRE_UNTIL_HALT = new FireUntilHaltRestHandler();
+
+
+        PropagationEntry handleRest(InternalWorkingMemory wm, ExecutionState currentState);
+
+        public static class FireAllRulesRestHandler  implements RestHandler {
+            @Override
+            public PropagationEntry handleRest(InternalWorkingMemory wm, ExecutionState currentState) {
+                synchronized (this) {
```
mariofusco, line 1362: I believe this synchronization is on the wrong target or more likely totally unnecessary. 'this' is the FireAllRulesRestHandler instance and I don't think it could be used by multiple threads at the same time.

drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1308,11 +1342,52 @@ public int fireAllRules(AgendaFilter agendaFilter,
...
+                    PropagationEntry head = wm.takeAllPropagations();
+
+                    // if halt() has called, the thread should not be put into a wait state
+                    // instead this is just a safe way to make sure the queue is flushed before exiting the loop
+                    if (head == null && currentState == ExecutionState.FIRING_UNTIL_HALT) {
+                        wm.waitOnRest();
+                        head = wm.takeAllPropagations();
+                    }
+                    return head;
+                }
```
mariofusco, line 1386: This is a minor thing, but I don't like that the WorkingMemory has to expose its own propagation list. I'm afraid this could cause potential problem in future. Wouldn't be better to move this sync'ed block inside a new method of the WorkingMemory itself and just call that method from here?

drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
mdproctor, line 1362: You are right, it should be on the wm.agenda. It needs to be on what ever is synced, when you change the engine state.

drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
mdproctor, line 1386: We can possible make it more explicit like wm.getStackMachineLock(), which returns Object. Even if it just returns the PropagationList underneath.

drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1308,11 +1342,52 @@ public int fireAllRules(AgendaFilter agendaFilter,
...
+
+        PropagationEntry handleRest(InternalWorkingMemory wm, ExecutionState currentState);
+
+        public static class FireAllRulesRestHandler  implements RestHandler {
+            @Override
+            public PropagationEntry handleRest(InternalWorkingMemory wm, ExecutionState currentState) {
+                synchronized (this) {
+                    PropagationEntry head = wm.takeAllPropagations();
+                    if (head == null) {
+                        wm.setEngineInactive();
```
mariofusco, line 1365: It is wrong to call wm.setEngineInactive() here because it will call agenda.setEngineInactive() that will do a notify(). Immediately after the end of the fireLoop will call immediateHalt() that will perform a second notify() thus allowing 2 threads to enter the fireLoop. I replaced this wm.setEngineInactive() with agenda.halt() and this (apparently) fixed the concurrency problem that I found yesterday. I believe this problem was present also in my original implementation but your refactor simplified the state machine and then made this issue more evident.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/13bb9a748

PR #569, opened 2015-12-02
[DROOLS-992] Improved state machine for thread interactions

mariofusco:


drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1321,21 +1333,24 @@ private int fireLoop(AgendaFilter agendaFilter,
...
                     limitReached = (fireLimit > 0 && fireCount >= fireLimit);
                 } else {
                     returnedFireCount = 0; // no rules fired this iteration, so we know this is 0
+                    group = null; // set the group to null in case the fire limit has been reached
                 }
 
                 head = workingMemory.takeAllPropagations();
 
-                if ( returnedFireCount == 0 && head == null ) {
+                if ( returnedFireCount == 0 && head == null && ( group == null || !group.isAutoDeactivate() ) ) {
```
mariofusco, line 1353: @mdproctor I only have a doubt about this further !group.isAutoDeactivate() check that I had to add. This is only to make the org.drools.workbench.models.testscenarios.backend.ScenarioRunnerTest.testRuleFlowGroupActivation to pass without falling in an endless loop. That one is the only test that does 

((InternalAgendaGroup)ksession.getAgenda().getRuleFlowGroup( "asdf" )).setAutoDeactivate( false );

What is that autoDeactivate flag for? Is it something used by jbpm?

mariofusco:
ok to test

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/9a49d5e5a

PR #571, opened 2015-12-03
deal with spurious wakeups of Object.wait()

Zillode:
> As in the one argument version, interrupts and spurious wakeups are possible, and this method should always be used in a loop. [1]

[1] http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#wait(long)


psiroky:
ok to test

PR #572, opened 2015-12-03
[DROOLS-993] fix use of globals with temporal operators

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -5898,4 +5898,90 @@ public void testSubclassWithLongerExpirationThanSuperWithSerialization() throws
...
+
+        PseudoClockScheduler clock = ksession.getSessionClock();
+        ksession.setGlobal("clock", clock);
+
+        List<String> list = new ArrayList<String>();
+        ksession.setGlobal("list", list);
+
+        SimpleEvent event1 = new SimpleEvent("1", 0L);
+        ksession.insert(event1);
+        clock.advanceTime(15, TimeUnit.SECONDS);
```
etirelli, line 5939: Mario, these two examples go against the design of the engine. Remember that globals used in constraints must be immutable, or the results of executing such rules becomes undeterministic. Even if the goal of the feature here is to enable something useful for some edge cases, I don't think we should implement exemples in tests where mutable globals are used as constraints.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -5898,4 +5898,90 @@ public void testSubclassWithLongerExpirationThanSuperWithSerialization() throws
...
+
+        PseudoClockScheduler clock = ksession.getSessionClock();
+        ksession.insert(clock);
+
+        List<String> list = new ArrayList<String>();
+        ksession.setGlobal("list", list);
+
+        SimpleEvent event1 = new SimpleEvent("1", 0L);
+        ksession.insert(event1);
+        clock.advanceTime(15, TimeUnit.SECONDS);
```
etirelli, line 5982: Same as previous comment.

etirelli:
Mario, before merging these changes, I think it would be good to fix the tests to use constant globals.

mariofusco:
@etirelli Good point Edson, I changed the test.

Merged by https://github.com/droolsjbpm/drools/commit/75cb39d15

PR #573, opened 2015-12-03
[DROOLS-984] @expires( -1 ) should disable Event expiration

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationConfigurator.java
```diff
@@ -230,7 +230,9 @@ private static void configureExpirationOffset( KnowledgeBuilderImpl kbuilder, An
                 }
                 return;
             }
-            type.setExpirationOffset(TimeIntervalParser.parse( expiration )[0]);
+            long offset = TimeIntervalParser.parseSingle( expiration );
+            // @Expires( -1 ) means never expire
+            type.setExpirationOffset(offset == -1L ? Long.MAX_VALUE : offset);
```
etirelli, line 235: Mario, my spider senses are all tingling with this line, because Long.MAX_VALUE would be a valid (although very large) expiration offset. Are you sure this will not break the expiration scheduling down the line?

drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -5898,4 +5898,49 @@ public void testSubclassWithLongerExpirationThanSuperWithSerialization() throws
     @Role(Role.Type.EVENT)
     @Expires( "20s" )
     public static class SubClass extends SuperClass { }
+
+    @Test
+    public void testNoExpiration() {
```
etirelli, line 5903: Mario, there are 3 more scenarios I think need testing in this case: 
1. an event that is used in a sliding window and that is annotated with @expires(-1). 
2. an event that is not used in any temporal constraint and that is annotated with @expires(-1).
3. an event that has no @expires annotation, but is used as part of a temporal constraint. (we might have this test already)

drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationConfigurator.java
mariofusco, line 235: @etirelli Long.MAX_VALUE milliseconds > 30 millions years, so I think we shouldn't have any problem. Moreover we are already not creating any expiration when the offset is equal to Long.MAX_VALUE https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PropagationEntry.java#L105

mariofusco:
@etirelli I added the tests you suggested.

Merged by https://github.com/droolsjbpm/drools/commit/1f88647d2

drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationConfigurator.java
etirelli, line 235: Yeah, I was concerned about not creating the expiration job, so if it is not, great. :)

PR #574, opened 2015-12-03
tests: use better names for non existing fields/classes + small typo …

psiroky:
…fixes
- using names that are quite different from the expected ones
  makes it easier to spot them in the log. 'mesage' vs 'message'
  is very easy to overlook


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/cf744e9d5

PR #578, opened 2015-12-05
Fix failing tests

psiroky:
@mariofusco this should fix the test failure we have been seeing in CI (at least it does for me when running locally). We will have to run multiple builds in CI to be sure though.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
@@ -15,6 +15,7 @@
 
 package org.drools.compiler.integrationtests;
 
+import org.assertj.core.api.Assertions;
 import org.junit.Test;
 import org.kie.api.KieServices;
 import org.kie.api.builder.KieModule;
@@ -190,9 +191,10 @@ public class KieContainerTest {
             kieSession.setGlobal("list", list);
             kieSession.fireAllRules();
             kieSession.dispose();
-            // There can be 2 items in the list if an updateToVersion is triggered during a fireAllRules
-            // in that case it may fire with both the old and the new rule
-            assertTrue( list.size() >= 1 && list.size() <= 2);
+            // There can be multiple items in the list if an updateToVersion is triggered during a fireAllRules
+            // (updateToVersion can be called multiple times during fireAllRules, especially on slower machines)
+            // in that case it may fire with the old rule and multiple new ones
+            Assertions.assertThat(list).isNotEmpty();
             if (list.get(0).equals("rule9")) {
                 break;
             }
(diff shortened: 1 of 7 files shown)


psiroky:
jenkins, retest this

psiroky:
Looks like the current fix is not enough, some tests still failing.

psiroky:
I've executed the build 20 times in a row on one of the CI machines and all the runs were successful. This does not mean there aren't any other tests influencing each other, but it is definitely a step forward.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
```diff
@@ -190,9 +191,10 @@ public void run() {
             kieSession.setGlobal("list", list);
             kieSession.fireAllRules();
             kieSession.dispose();
-            // There can be 2 items in the list if an updateToVersion is triggered during a fireAllRules
-            // in that case it may fire with both the old and the new rule
-            assertTrue( list.size() >= 1 && list.size() <= 2);
+            // There can be multiple items in the list if an updateToVersion is triggered during a fireAllRules
+            // (updateToVersion can be called multiple times during fireAllRules, especially on slower machines)
+            // in that case it may fire with the old rule and multiple new ones
+            Assertions.assertThat(list.size()).isGreaterThanOrEqualTo(1);
```
triceo: Unless I'm mistaken, you can `assertThat(list)` and then use something like `.isNotEmpty()`.

psiroky: I maybe be missing something (I am not much familiar with AssertJ), but wouldn't that fail for lists with size greater than 1?

triceo: The original code is asserting that the list's size `isGreaterThanOrEqualTo(1)`. In other words - if there is at least one element in the list. Changing this to what I propose is just a more descriptive way of saying the same.

psiroky: `isNotEmpty()` works, I was originally looking at `hasSize()` which I though was in the first comment.... Anyway, thanks I will fix it.

triceo: Ah, my apologies - I fixed my original comment 5 seconds after I posted it, I had no idea someone was able to read it in the meantime. :-)

psiroky: Looks like some caching issue as I read the comment like 15 mins after it was created (at least that what github was showing). Not important, though. Should be now fixed.

etirelli:
Looks good @psiroky . +1

psiroky:
Merged into master.

PR #580, opened 2015-12-07
[6.3.x] Fix few intermittently failing tests

psiroky:
- KieModuleRepoTest was influencing other tests,
  executed after
- specified runOrder for surefire. Hourly means
  that surefire will execute the tests
  in alphabetical order on even hours and in
  reverse alphabetical order on odd hours. This
  brings certain level of predictability as to
  in which order the tests are executed, but at the same
  time allows to detect at least some issues related
  to order of exeuction of the tests (which is
  bad as unit tests should not be dependent on
  each other). see http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#runOrder for more info
- increase test timeouts to accomodate slower machines

Same changes already applied to master. @etirelli since these are only test fixes, I believe it is save to apply them to 6.3.x as well.

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -129,6 +129,11 @@
       <artifactId>glazedlists_java15</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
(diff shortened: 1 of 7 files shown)


mariofusco:
+1 merged

PR #584, opened 2015-12-15
[DROOLS-1007] fix removal of a rule containing a OR

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/IncrementalCompilationTest.java
```diff
@@ -2424,4 +2424,31 @@ private String getExistenzialRule(String rulename, String condition) {
...
+        KieContainer kc = ks.newKieContainer(km.getReleaseId());
+        KieSession ksession = kc.newKieSession();
+
+        ksession.fireAllRules();
+
+        ReleaseId releaseId2 = ks.newReleaseId("org.kie", "test-upgrade", "1.1.2");
+        km = createAndDeployJar( ks, releaseId2 );
+        kc.updateToVersion(releaseId2);
+
+        ksession.fireAllRules();
```
etirelli, line 2452: @mariofusco ideally we should have asserts on every test. If the error was an exception being raised, using a try/catch with Assert.fail()/Assert.success() instead of throwing the Exception in the method helps with "documenting" the behaviour. Alternatively, whenever possible, you can also use mock listeners and make assertions on them.

drools-core/src/main/java/org/drools/core/common/BaseNode.java
```diff
@@ -161,21 +160,16 @@ public void setPartitionId(final RuleBasePartitionId partitionId) {
...
     }
-    
-    /**
-     * Returns the map of associations for this node
-     */
-    public Map<Rule, RuleComponent> getAssociations() {
-        return this.associations;
+
+    public void addAssociation( Rule rule, RuleComponent ruleComponent ) {
+        addAssociation( rule );
```
etirelli, line 170: @mariofusco can we remove this method and keep only the one with the single parameter above? Simply ignoring the second parameter seems to be error prone.... unless this is necessary for polymorphism on subclasses.

etirelli:
Mario, see my comments. Other than that +1 to merge. Good job!

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/3739b5c4d

PR #587, opened 2015-12-18
DO NOT MERGE YET - [DROOLS-653] avoid wrapping the LeftTuple created by a subnetwork into a RightTuple

mariofusco:
This pull request is made of 2 commits. 

The first commit is a further refactor of the Left/RightTuple intended to allow to have a Tuple implementation that is both a Left and a RightTuple. I'm keeping it separated because it is a nice and useful refactor and I'd like to merge it regardless if we will be able to avoid wrapping the LeftTuple created by a subnetwork into a RightTuple.

The second commit contains a new Tuple implementation called SubnetworkTuple that now is the Tuple generated by RightInputAdapterNode. Since the SubnetworkTuple implements both LeftTuple and RightTuple is no longer necessary to wrap the JoinNodeLeftTuple originally created by the RightInputAdapterNode into a RightTuple in order to propagate it on the right.

This isn't yet ready to be merged since there is one last test failing. In particular BackwardChainingTest.testInsertionOrderTwo throws the following NPE

java.lang.NullPointerException
    at org.drools.core.common.TupleSetsImpl.removeInsert(TupleSetsImpl.java:143)
    at org.drools.core.reteoo.QueryElementNode$UnificationNodeViewChangedEventListener.rowUpdated(QueryElementNode.java:531)
    at org.drools.core.phreak.PhreakQueryTerminalNode.doLeftUpdates(PhreakQueryTerminalNode.java:117)
    at org.drools.core.phreak.PhreakQueryTerminalNode.doNode(PhreakQueryTerminalNode.java:50)
    at org.drools.core.phreak.RuleNetworkEvaluator.innerEval(RuleNetworkEvaluator.java:310)
    at org.drools.core.phreak.RuleNetworkEvaluator.evalStackEntry(RuleNetworkEvaluator.java:225)
    at org.drools.core.phreak.RuleNetworkEvaluator.outerEval(RuleNetworkEvaluator.java:168)
    at org.drools.core.phreak.RuleNetworkEvaluator.evaluateNetwork(RuleNetworkEvaluator.java:120)
    at org.drools.core.phreak.RuleExecutor.reEvaluateNetwork(RuleExecutor.java:194)
    at org.drools.core.phreak.RuleExecutor.evaluateNetworkAndFire(RuleExecutor.java:73)
    at org.drools.core.common.DefaultAgenda.fireNextItem(DefaultAgenda.java:1003)
    at org.drools.core.common.DefaultAgenda.fireLoop(DefaultAgenda.java:1346)
    at org.drools.core.common.DefaultAgenda.fireAllRules(DefaultAgenda.java:1284)
    at org.drools.core.impl.StatefulKnowledgeSessionImpl.internalFireAllRules(StatefulKnowledgeSessionImpl.java:1303)
    at org.drools.core.impl.StatefulKnowledgeSessionImpl.fireAllRules(StatefulKnowledgeSessionImpl.java:1293)
    at org.drools.core.impl.StatefulKnowledgeSessionImpl.fireAllRules(StatefulKnowledgeSessionImpl.java:1274)
    at org.drools.compiler.integrationtests.BackwardChainingTest.testInsertionOrderTwo(BackwardChainingTest.java:2279)

@mdproctor It would be great if you could review what I have done so far and, if you have time, also to give a look at that last failing test, since I haven't been able to understand what's going wrong there.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
@@ -817,7 +817,7 @@ public class AccumulateTest extends CommonTestMethodBase {
         execTestAccumulateMax( "test_AccumulateMaxMVEL.drl" );
     }
 
-    @Test(timeout = 10000)
+    @Test//(timeout = 10000)
     public void testAccumulateMultiPatternJava() throws Exception {
         execTestAccumulateReverseModifyMultiPattern( "test_AccumulateMultiPattern.drl" );
     }
@@ -1421,13 +1421,13 @@ public class AccumulateTest extends CommonTestMethodBase {
         wm.setGlobal( "results",
                       results );
 
-        final Cheese[] cheese = new Cheese[]{new Cheese( "stilton",
-                                                         10 ), new Cheese( "stilton",
-                                                                           2 ), new Cheese( "stilton",
-                                                                                            5 ), new Cheese( "brie",
-                                                                                                             15 ), new Cheese( "brie",
-                                                                                                                               16 ), new Cheese( "provolone",
-                                                                                                                                                 8 )};
+        final Cheese[] cheese = new Cheese[]{ new Cheese( "stilton", 10 ),
+                                              new Cheese( "stilton", 2 ),
+                                              new Cheese( "stilton", 5 ),
+                                              new Cheese( "brie", 15 ),
+                                              new Cheese( "brie", 16 ),
+                                              new Cheese( "provolone", 8 ) };
+
         final Person bob = new Person( "Bob",
                                        "stilton" );
         final Person mark = new Person( "Mark",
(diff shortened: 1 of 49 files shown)


mariofusco:
@mdproctor The problem in that last failing test is caused by the fact that the same SubnetworkTuple is both a Left and a Right Tuple and then it attempts to add it first to a left staged list ( https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/reteoo/QueryElementNode.java#L421 ) and then to a right one ( https://github.com/mariofusco/drools/blob/d653/drools-core/src/main/java/org/drools/core/phreak/RuleNetworkEvaluator.java#L650 ).

At the moment a tuple can only be in a single staged list at time, so the second add remove the tuple from the first, causing a NPE when it tries to remove the tuple from the first list here https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/reteoo/QueryElementNode.java#L531

So the question is: is it a structural thing that a SubnetworkTuple, being a Left and a Right Tuple, could be added to the left and right staged lists at the same time or it is something necessary only for the QueryNode? In the first case I should allow the SubnetworkTuple to be added to both lists, but this means that everytime a tuple is added to a staged list we should specify if it's added on left or on the right and this could be a quite invasive change. In the second case I could more pragmatically do an hack only for the QueryNode. What do you think?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/c87ce8f19

PR #588, opened 2015-12-18
[BZ-1199965] enable use of Java 8 syntax in RHS

mariofusco:
Since we have upgraded to ECJ 4.4.2 ( see https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/9de5cd80f30f060629e33587654be550fca3f0c7 ) it is now possible to use Java 8 syntax.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
@@ -8160,7 +8160,7 @@ public class Misc2Test extends CommonTestMethodBase {
                                              .newKieSession();
 
         ksession.insert( "\"#" );
-        assertEquals(1, ksession.fireAllRules());
+        assertEquals( 1, ksession.fireAllRules() );
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -8177,6 +8177,43 @@ public void testPatternMatchingWithFakeNullSafe() {
...
+        checkJava8InRhs("i -> list.add(i)");
+    }
+
+    @Test
+    public void testMethodReferenceInRHS() {
+        checkJava8InRhs("list::add");
+    }
+
+    private void checkJava8InRhs(String expr) {
+        if (!System.getProperty("java.version").startsWith( "1.8" )) {
```
triceo, line 8194: http://junit.sourceforge.net/javadoc/org/junit/Assume.html
Perfectly designed for this use case.

etirelli:
@mariofusco doesn't java 8 support requires us to change our java parser?

drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
mariofusco, line 8194: Thanks @triceo I'll update the test using that assert method.

mariofusco:
@etirelli Unless I'm not missing some specific use case I think this change will be enough to allow Java 8 syntax in out RHS, that anyway is what the all the users who reported this issue so far are looking for. Of course enabling lambdas in the LHS is a totally different story and I'm afraid we won't be able to achieve this till we will keep using mvel. I think the long term plan (Drools 7?) is to totally get rid of mvel, is that correct?

etirelli:
@mariofusco I was talking about RHS. I thought we would need to update our java parser to be java 8 compatible. But if it is working, then great. +1 to merge.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/5e88d07b7

PR #589, opened 2015-12-21
[DROOLS-1012] register the ClassLoaderResolver on the ServiceRegistry in kie-ci-osgi Activator

mariofusco:


etirelli:
@mariofusco is there a test for these changes?

mariofusco:
@etirelli No, I tried the fix in Fuse with the reproducer attached to the jira https://issues.jboss.org/browse/DROOLS-1012

I don't know how to develop a test case to reproduce this issue. Probably I could try to add a test using Arquillian in the fuse-bxms integration pack, but I'm not sure if this is doable and for sure it will take a while.

etirelli:
Maybe just create a unit test that checks the Activator class is being properly initialized with the classloader? would that make sense to prevent regressions?

Other than that +1 to merge.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e48ee1730

reynoldsm88:
Hey I am the original filer of this ticket. I wanted to test out this change with my reproducer which is actually on GH here: https://github.com/reynoldsm88/brms-fuse-integration.

When I pull down master there are compilation problems at HEAD. Checking out this specific commit there are also compilation problems in drools-core that seem unrelated to this issue. Not sure what to do, but just letting you know.

reynoldsm88:
Just to let you know, I was working off an out of date droolsjbpm-knowledge. After recompiling that project the above mentioned issue was resolved.

PR #592, opened 2015-12-24
DROOLS-1015 Wrong MvelConstraint compilation with Unicode class name …

tkobayas:
…and the same name property
- test case

This commit includes a Unicode name class file. I'm not sure if it causes any trouble in our project CI (I hope not!).


tkobayas:
CI failed with the 2 test cases expectedly.

mariofusco:
This issue is caused by mvel class literals and can be reproduced in an easier way as documented in https://issues.jboss.org/browse/DROOLS-1015

PR #594, opened 2016-01-04
[ENTESB-4493] use DEBUG logging to inform about type of Maven parser

psiroky:
- this message gets printed everytime a POM is read/parsed. That can be
  easily dozens of times for anything bigger than just simple examples.
  It may easily flood the logs.


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/972282495

PR #596, opened 2016-01-11
DROOLS-1023 Change kmodule.xsd targetNamespace to become versionless …

ge0ffrey:
DROOLS-1023 Change kmodule.xsd targetNamespace to become versionless (to avoid hardcoding "6.0.0" in it)

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieBuilderTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieBuilderTest.java
@@ -101,6 +101,89 @@ public class KieBuilderTest extends CommonTestMethodBase {
     }
 
     @Test
+    public void testValidXsdTargetNamespace() {
+        String drl1 = "package org.drools.compiler\n" +
+                "rule R1 when\n" +
+                "   $m : Message()\n" +
+                "then\n" +
+                "end\n";
+
+        String kmodule = "<kmodule xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" \n" +
+                "         xmlns=\"http://www.drools.org/xsd/kmodule\">\n" +
+                "  <kbase name=\"kbase1\">\n" +
+                "    <ksession name=\"ksession1\" default=\"true\"/>\n" +
+                "  </kbase>\n" +
+                "</kmodule>";
+
+        KieServices ks = KieServices.Factory.get();
+
+        ReleaseId releaseId1 = ks.newReleaseId( "org.kie", "test-kie-builder", "1.0.0" );
+        Resource r1 = ResourceFactory.newByteArrayResource( drl1.getBytes() ).setResourceType( ResourceType.DRL ).setSourcePath( "kbase1/drl1.drl" );
+        KieModule km = createAndDeployJar( ks,
+                                           kmodule,
+                                           releaseId1,
+                                           r1);
+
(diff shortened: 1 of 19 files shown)


ge0ffrey:
The PR build fails because the droolsjbpm-knowledge PR's changed weren't applied.
I didn't run all drools tests locally though. But I verified it worked with the new and the old namespace (although there's no test currently for the old namespace...)

psiroky:
The changes from droolsjbpm-knowledge were applied, looking at the build log these upstream repositories were cloned and build before this one:

```
git://github.com/droolsjbpm/droolsjbpm-build-bootstrap:master
git://github.com/ge0ffrey/droolsjbpm-knowledge:kmodule-xsd-patch
```

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieBuilderTest.java
```diff
@@ -64,7 +64,7 @@ public void testResourceInclusion() {
                 "end\n";
 
         String kmodule = "<kmodule xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" \n" +
-                "         xmlns=\"http://jboss.org/kie/6.0.0/kmodule\">\n" +
+                "         xmlns=\"http://www.drools.org/xsd/kmodule\">\n" +
```
mariofusco, line 67: @ge0ffrey I think what you did broke backward compatibility anyway. The old namespace should still work but if I try to use the old one here this test fails.

ge0ffrey, line 67: That test should indeed work with the old namespace. Let's duplicate that
test so the old namespace is tested too (as well as the new one).

I am on PTO today, I'll try to take a look tomorrow

2016-01-12 20:06 GMT+01:00 Mario Fusco notifications@github.com:

> In
> drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieBuilderTest.java
> https://github.com/droolsjbpm/drools/pull/596#discussion_r49498241:
> 
> > @@ -64,7 +64,7 @@ public void testResourceInclusion() {
> >                  "end\n";
> > 
> > ```
> >      String kmodule = "<kmodule xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" \n" +
> > ```
> > -                "         xmlns=\"http://jboss.org/kie/6.0.0/kmodule\">\n" +
> > -                "         xmlns=\"http://www.drools.org/xsd/kmodule\">\n" +
> 
> @ge0ffrey https://github.com/ge0ffrey I think what you did broke
> backward compatibility anyway. The old namespace should still work but if I
> try to use the old one here this test fails.
> 
> —
> https://github.com/droolsjbpm/drools/pull/596/files#r49498241.

## 

Geoffrey De Smet

ge0ffrey:
@mariofusco there were some serious issues with my PR, but they should be all solved now.
I've added 3 unit tests to test the new schema, the old schema and an invalid schema.
The original valid code in KieModuleModelImpl didn't actually work (despite that I swear it worked for me when using it in OptaPlanner with the old and new schema). In any case, I fixed it to work, but unfortunately this required duplicating the stream initialization (which isn't a real performance cost anyway). See changes on KieModuleModelImpl in my last commit for more info.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/2149530174e121a437bbef93196a4eef9121f9fd

PR #598, opened 2016-01-12
6.3.x

moh2sen2006:
I've written a rule for alerting ssh event with "failed password". This rule is here:
<code>
rule "Hello World"
  when
    accumulate(m:Message(eventType=="Failed password") over window:time( 59s );s:count(m);s>3)
  then
     System.out.println( "success" );
     Alert alert=new Alert("ssh","test");
     insert(alert);
end
</code>
This is working for the first scenario. But I want to extend this rule. I want to accumulate ssh event with "failed password" for with the same Src_ip address. For example, if I have 4 failed password ssh event in 59s from three different src_ip, the rule does not match, but when I have 4 failed password ssh event in 59s from one src_ip, rule matches. How I should rewrite this rule for this scenario.


manstis:
I suspect (1) user should email forums, (2) this PR should be rejected and
closed.

psiroky:
Closing. @moh2sen2006 please post your question on https://groups.google.com/forum/#!forum/drools-usage

PR #600, opened 2016-01-13
kie-test-utils: added (logback) filtered pattern layout

mrietveld:
This is a class that's been useful to me during development of both GUVNOR-2173 and JBPM-4349. 

Note: this is a class in a _test utility_ jar, so that this module should always be used in a maven test scope, (which thus also applies to any transitive depenencies the module itself has as well).

### DIFF
--- a/kie-test-util/src/main/java/org/kie/test/util/logging/FilteredPatternLayout.java
+++ b/kie-test-util/src/main/java/org/kie/test/util/logging/FilteredPatternLayout.java
@@ -1,0 +1,195 @@
+package org.kie.test.util.logging;
+
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+
+import ch.qos.logback.classic.pattern.EnsureExceptionHandling;
+import ch.qos.logback.classic.pattern.ExtendedThrowableProxyConverter;
+import ch.qos.logback.classic.pattern.ThrowableProxyConverter;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.IThrowableProxy;
+import ch.qos.logback.classic.spi.StackTraceElementProxy;
+import ch.qos.logback.classic.spi.ThrowableProxyUtil;
+import ch.qos.logback.core.CoreConstants;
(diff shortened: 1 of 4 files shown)


triceo:
@mrietveld , have you considered using http://logback.qos.ch/manual/filters.html? It seems to have a lot of features.

triceo:
Oh, nevermind, this deals with stack traces. The filters in Logback itself probably can't do that.

mrietveld:
@triceo Yup.. :) 

The functionality has actually since been added: it might be in 1.1.3? That was released in march 2015, just 3 weeks after the commit below was added.. We're currently on 1.0.9. 

For more info, see: 
- [LOGBACK-540](http://jira.qos.ch/browse/LOGBACK-540?focusedCommentId=13939&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13939)
- https://github.com/qos-ch/logback/commit/a71b2902cff5911bbb16d4e2950476b0dfeebeee (Note that the commit is from 2012, but that it was only pushed in early 2015). 
- https://github.com/qos-ch/logback/pull/200

triceo:
@psiroky Any chance we could upgrade Logback? The current version is over 3 years old...

psiroky:
We should be able to upgrade once we branch for 7.x.

kie-test-util/src/main/java/org/kie/test/util/logging/FilteredPatternLayout.java
```diff
@@ -0,0 +1,195 @@
...
+     */
+    private final Set<String> filteredPackages = new HashSet<String>();
+
+    /**
+     * Adds a new filtered frame. Any stack frame starting with <code>"at "</code> + <code>filter</code> will not be
+     * written to the log.
+     *
+     * @param filter a class name or package name to be filtered
+     */
+    public void setFilter(String filter) {
```
psiroky, line 64: I think `addFilter` would better describe the intent. It is adding the filter to the set.

kie-test-util/src/main/java/org/kie/test/util/logging/FilteredPatternLayout.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+
+        /**
+         * Check if the given string starts with any of the filtered patterns.
+         *
+         * @param step checked String
+         * @return <code>true</code> if the begininning of the string matches a filtered pattern, <code>false</code> otherwise
+         */
+        private boolean startsWithAFilteredPAttern(StackTraceElementProxy step) {
```
psiroky: This looks like typo - `PAttern` -> `Pattern`

psiroky:
I think we could merge this in and once we upgrade to logback 1.1.3 we would remove it?

kie-test-util/src/main/java/org/kie/test/util/logging/FilteredPatternLayout.java
mrietveld, line 64: Unfortunately, changing the method name breaks the functionality.

mrietveld:
Removing this after 1.1.3 is fine.

mrietveld:
Jenkins, please test.

mrietveld:
@psiroky I've updated the diff with the changes you suggested. If you could review it again, please do. Thanks!

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/c7ff991e2

PR #604, opened 2016-01-18
[DROOLS-1013] fix jitting of a compareTo constraint

mariofusco:


etirelli:
@mariofusco do we need a test to cover this or is there a test already?

other than that, +1 to merge

mariofusco:
@etirelli there is already test that indeed was failing by turning the TEST_JITTING flag on

PR #606, opened 2016-01-21
[ENTESB-4802] add workaround for Fuse 6 + Java 8 + ECJ issue

psiroky:
- Fuse 6.2.1 is distributed with PAX Web JSP bundle. That bundle
  exports ECJ packages with version 3.2.5, which is quite old
  and does not work properly with Java 8. We need to be able to
  override that version, and use the ECJ Drools is being tested
  with. Without the explicit dependency on the version [0,1), the
  latest available one (3.2.5, which is fact much older) gets wired
  and thus results in failures. Proper fix is to pursuade Eclipse
  JDT team to export the packages with specific versions, and then
  fix the drools-compiler metadata to use those versions

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -195,7 +195,10 @@
               javax.enterprise.*;resolution:=optional,
               org.antlr.*;resolution:=optional,
               org.codehaus.janino.*;resolution:=optional,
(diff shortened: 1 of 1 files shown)


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/85a9a7991

PR #607, opened 2016-01-21
[DROOLS-1037] fix or expression rewriting

mariofusco:


etirelli:
Mario, I don't think we should allow expressions in the form of:

(x in (...)) == false

The "in"/"not in" operators are special operators and there is no need to complicate the language by "making then comparable to a boolean". 

If you disagree, lets talk on IRC.

mariofusco:
Edson, in theory I agree with you and I also think that this expression should not be allowed. I said this to Martin on IRC when we discussed this issue but as he wrote here https://bugzilla.redhat.com/show_bug.cgi?id=1300699#c4 he told me that "this syntax is working in BRMS 6.1, and that the customer has 2000+ rules with an unknown number of places where this syntax is used. They are not willing to take the effort of refactoring all their rules, and see this as a regression which blocks their migration project".

Unfortunately he is right saying that it worked in BRMS 6.1 and indeed what he considers a "regression" has been caused by this fix https://github.com/droolsjbpm/drools/commit/dc81e1f1ca7f56a29b4fb6c0ad0e14bb3c49551c

The other thing to be considered is that an expression like 

(name in (\"Alice\", \"Charlie\", \"David\"))==false

still works when not used in conjunction with an OR. That said I see 2 options:
1. we merge this "fix" and then solve the "regression".
2. we decide that an expression like the former one is invalid and then we prevent it to compile in any case.

As I anticipated in theory I'd go with option 2. The problem is that if we do that we risk to have even more customers to complain of what is anyway considered a regression. My suggestion is to merge this fix and have this working in Drools 6.x. If I'll have time (as I hope) to get rid of mvel in Drools 7 I'll have to review all these expression anyway.

Let me know how to proceed on this.

etirelli:
LOL, this is funny... sometimes things break without us knowing but it is the first time I see something working without us knowing. 
Yeah, if it works in 6.1, I think we will have to merge your fix, document and start supporting... 
+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/f7c3f2561

PR #608, opened 2016-01-22
[DROOLS-787] remove unnecessary synchronization in Process*Factory classes

psiroky:
- the providers are initialized only once by default, so most of
  the synchronization could be removed. This use case
  covers the standard (non-OSGi) use cases as the provider class
  will be on classpath when doing the initializaiton. In OSGi,
  we need a way to force re-initialization as the provider class
  might not have been available when calling the initialize method
  the first time (e.g. the jBPM bundles were not yet started). But
  once activating the jBPM bundles, we want to make sure the Drools
  engine is aware of the jBPM-related factories.
- simple getters/setters do not need synchronization as assignment
  of object reference is atomic
- synchronization is left only in the reInitialize*() methods.
  Since they are called only once in the Activators it should
  not bring any noticible performance penalty

@mariofusco, @mswiderski  would be great if you could take a look to make sure I haven't overlooked something.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/ProcessBuilderFactory.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/ProcessBuilderFactory.java
@@ -24,36 +24,49 @@ public class ProcessBuilderFactory {
     private static final String PROVIDER_CLASS = "org.jbpm.process.builder.ProcessBuilderFactoryServiceImpl";
 
     private static IllegalArgumentException initializationException;
-    private static ProcessBuilderFactoryService provider;
+    private static ProcessBuilderFactoryService provider = initializeProvider();
 
     public static ProcessBuilder newProcessBuilder(KnowledgeBuilder kBuilder) {
-        return getProcessBuilderFactoryService().newProcessBuilder(kBuilder);
-    }
-
-    public static synchronized void setProcessBuilderFactoryService(ProcessBuilderFactoryService provider) {
-        ProcessBuilderFactory.provider = provider;
-    }
-
-    public static synchronized ProcessBuilderFactoryService getProcessBuilderFactoryService() {
-        if (provider == null && !initialized()) {
-            try {
-                loadProvider();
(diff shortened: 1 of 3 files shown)


psiroky:
This PR needs to be merged together with https://github.com/droolsjbpm/jbpm/pull/379

drools-compiler/src/main/java/org/drools/compiler/compiler/ProcessBuilderFactory.java
```diff
@@ -24,36 +24,47 @@
...
 
-    private static void loadProvider() {
-        ServiceRegistryImpl.getInstance().addDefault( ProcessBuilderFactoryService.class, PROVIDER_CLASS );
-        setProcessBuilderFactoryService(ServiceRegistryImpl.getInstance().get( ProcessBuilderFactoryService.class ) );
+    /**
+     * This method is used in jBPM OSGi Activators as we need a way to force re-initialization when starting
+     * the bundles.
+     */
+    public static synchronized  void reInitializeProvider() {
+        provider = initializeProvider();
```
mswiderski, line 64: should it set to null initializationException when reinit..ing? Otherwise https://github.com/droolsjbpm/drools/pull/608/files#diff-76a54b4320bd06996ea2c9f2e849190dR30 will return an error if it failed in previous attempts, meaning it was set to not null value

psiroky, line 64: Yes, you are right. Very good point. I will see if I could add test that checks for this. Thanks!

psiroky, line 64: I fixed it, but I could not find an easy to way to test this (as the needed class is in jBPM repo, so can't rally instantiate it).

mswiderski:
looks good to me so +1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b28fd0a2b

PR #610, opened 2016-01-22
[DROOLS-1030] initialize jpa ksession before unmarshalling it

mariofusco:


mswiderski:
excellent, so it does work :)

we could just run jbpm tests on top of it to make sure it does work fine there too. 
Besides that +1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/4f5589824

PR #611, opened 2016-01-23
kie-test-util: project version test utility class (for "future" deprecated methods tests)

mrietveld:


mrietveld:
@psiroky If you get a chance, could you look at this?

psiroky:
I've tested with versions `7.0.0.Beta1` and `7.0.0.20160128-098765` and the test fails with

```
java.lang.NumberFormatException: multiple points
    at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1914)
    at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
    at java.lang.Double.parseDouble(Double.java:538)
    at org.kie.test.util.StaticMethodTestHelper.projectVersionIsLessThan(StaticMethodTestHelper.java:75)
    at org.kie.test.util.StaticMethodTestHelperTest.versionTestTest(StaticMethodTestHelperTest.java:11)
```

We need to support those versions as well. If we could separate the comparator logic from the pom parsing it would also be easier to test different version strings.

There is also missing licence header, but that's just minor thing.

mrietveld:
@psiroky Thanks, I've updated the util to cover those versions and also added license headers .

PR #612, opened 2016-01-25
[DROOLS-1000] avoid race conditions when adding/removing listeners from a KieBase

mariofusco:


etirelli:
Mario, unfortunately I don't think synchronizing that code is enough to guarantee thread safety as the following method leaks the thread-unsafe collection, even if it is read-only:

public Collection<KieBaseEventListener> getKieBaseEventListeners() {
        return Collections.unmodifiableCollection( kieBaseListeners );
    }

I think you need to replace the HashSet by a thread-safe collection in this case. 

You can leave the synchronized blocks if the operations in there need to be atomic.

mrietveld:
@mariofusco I usually end up using `Collections.newSetFromMap(new ConcurrentHashMap<E, Boolean>());` in cases like this..

mariofusco:
I agree the solution I'm proposing is not correct. Retrying following @mrietveld suggestion

PR #614, opened 2016-01-26
[DROOLS-1010] fix equality of MvelConstraints when constraints are used in rules with different packages

mariofusco:


drools-core/src/main/java/org/drools/core/rule/constraint/MvelConstraint.java
```diff
@@ -628,6 +628,34 @@ public boolean equals(Object object, InternalKnowledgeBase kbase) {
...
+        for (int i = 0; i < expression.length(); i++) {
+            if ( lastStart == -1 ) {
+                if ( !isQuoted && Character.isJavaIdentifierStart( expression.charAt(i) ) ) {
+                    lastStart = i;
+                }
+            } else if ( !Character.isJavaIdentifierPart( expression.charAt(i) ) ) {
+                tokens.add(expression.subSequence(lastStart, i).toString());
+                lastStart = -1;
+            }
+            if (expression.charAt(i) == '"' || expression.charAt(i) == '\'') {
```
etirelli, line 644: As we discussed, just add a comment making clear your intention to ignore quoted strings in this method. Otherwise it is easy for someone simply reading the code to think there is a problem in here.

etirelli:
+1 to merge

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e67d1ca3c

PR #615, opened 2016-01-26
[DROOLS-1000] avoid race conditions when adding/removing listeners from a KieBase

mariofusco:


etirelli:
+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e7b642390

PR #616, opened 2016-01-27
[DROOLS-1041] workaround for Javadoc generation error on JDK 8

psiroky:


drools-core/pom.xml
```diff
@@ -258,7 +258,11 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
-          <excludePackageNames>org.kie.asm*,org.kie.objenesis.*,org.kie.commons.jci.*</excludePackageNames>
```
psiroky, line 261: The packages are not present in drools-core so it looks like a leftover which can be removed.

omajid:
Hey guys. Sorry to just barge in, but if you ever run into more OpenJDK issues, please feel free to cc me. I can work with the OpenJDK folks to get things resolved. These fixes should even make it into the proprietary JDK builds after a while.

PR #620, opened 2016-01-27
JBPM-4900 - Enhance TaskList Search capability

mswiderski:
since this introduces changes for jbpm tests are included in jbpm repo

### DIFF
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/ProcessMarshallerWriteContext.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/ProcessMarshallerWriteContext.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2010 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.core.marshalling.impl;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Map;
+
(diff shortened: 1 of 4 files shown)


drools-core/src/main/java/org/drools/core/marshalling/impl/ProcessMarshallerWriteContext.java
```diff
@@ -0,0 +1,86 @@
...
+import org.drools.core.impl.InternalKnowledgeBase;
+import org.drools.core.marshalling.impl.MarshallerWriteContext;
+import org.kie.api.marshalling.ObjectMarshallingStrategyStore;
+import org.kie.api.runtime.Environment;
+
+/**
+ * Extension to default <code>MarshallerWriteContext</code> that allows to pass additional
+ * information to marshaller strategies, such as process instance id, task it, state
+ */
+public class ProcessMarshallerWriteContext extends MarshallerWriteContext {
```
mariofusco, line 34: Is it necessary to have this class in drools-core or can it be moved to drools-persistence-jpa?

mswiderski, line 34: it needs to be in drools as this class relies on it https://github.com/droolsjbpm/drools/blob/master/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java - of course one modified in this PR

mrietveld, line 34: @mswiderski But the class itself can then still be moved from `drools-core` to `drools-persistence-jpa`, right?

mswiderski, line 34: if we would do it we force users who would like to build their custom marshaller strategies to always depend on drools-persistence-jpa even tough they might not use spa for their marshaller. drools-core needs to be there always.

mrietveld, line 34: Ah, that makes sense. 

+1 for leaving it in drools-core. (And eventually to move it to kie-core?)

mswiderski, line 34: agreed, it should be common so kie core sounds like a good idea

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/5ea84fb45

PR #621, opened 2016-01-28
[DROOLS-1044] allow inclusion of resources with same name from different KieModules

mariofusco:


etirelli:
Looks good, +1.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/cc3a9af55

PR #626, opened 2016-02-02
[DROOLS-1049] fix config after upgrading jaxb2-maven-plugin from 1.5 to 2.2

psiroky:
Related PRs:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/141
https://github.com/droolsjbpm/droolsjbpm-integration/pull/283

### DIFF
--- a/drools-pmml/pom.xml
+++ b/drools-pmml/pom.xml
@@ -88,7 +88,7 @@
             </goals>
             <configuration>
               <sources>
-                <source>${basedir}/target/generated-sources</source>
+                <source>${project.basedir}/target/generated-sources</source>
               </sources>
             </configuration>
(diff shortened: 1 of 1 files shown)


mrietveld:
+1

Note: what's interesting is that `javax.xml.bind.annotation.XmlElementRef` has a new attribute as of Java 8 -- and the difference between 6 and 8 showed up when I compared the original to the PR result. 

In particular, this attribute is new: [@XmlElementRef.required](http://docs.oracle.com/javase/8/docs/api/javax/xml/bind/annotation/XmlElementRef.html#required--)

However, it looks like the jaxb maven plugin is correctly setting this value, so no worries.

PR #627, opened 2016-02-02
[DROOLS-1050] make POI to load spreadsheets from Files when possible

mariofusco:


manstis:
@mfusco, looks ok but some (any, one?) would be nice.

drools-compiler/src/main/java/org/drools/compiler/compiler/DecisionTableFactory.java
```diff
@@ -28,8 +29,12 @@
 
     private static DecisionTableProvider provider;
     
-    public static String loadFromInputStream(InputStream is, DecisionTableConfiguration configuration) {
-        return getDecisionTableProvider().loadFromInputStream( is, configuration );
+    public static String loadFromInputStream(InputStream is, DecisionTableConfiguration configuration ) {
+        return loadFromResource(ResourceFactory.newInputStreamResource( is ), configuration);
```
mrietveld: @mariofusco There's some exception handling done in the drools-decisiontables for when reading the input stream fails. 

For example, [here](https://github.com/droolsjbpm/drools/blob/master/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/xls/ExcelParser.java#L99-L106) and [here](https://github.com/droolsjbpm/drools/blob/master/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/xls/ExcelParser.java#L99-L106). 

Replacing that with `ResourceFactory.newInputStreamResource( is )` gets rid of that error handling -- or does the `ResourceFactory` method cover that?

mrietveld:
@mariofusco Ignore my comment on the code (deleted it). 

+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b410f29a0

PR #628, opened 2016-02-03
Added unit tests for node sharing and improved getMatchingNode() to avoid recursive calls to equals.

chdasif:


drools-examples/pom.xml
```diff
@@ -107,6 +107,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>12.0</version>
```
triceo: IIRC our parent POMs bring their own version of Guava. Can we standardize on that?

psiroky: Yeah, we definitely have to use the Guava version already coming from parent.

mariofusco: Please remove the Guava version so it will use the one in our parent POM.

chdasif: Ok, i will remove the version.

chdasif:
Conan, As discussed, can you please review this?

mdproctor:
This branch is not currently for merging. it's just so we can easy track and provide review for the community person, as they progress.

drools-compiler/src/test/java/org/drools/compiler/CommonTestMethodBase.java
```diff
@@ -94,27 +96,32 @@ protected StatelessKnowledgeSession createStatelessKnowledgeSession(KnowledgeBas
...
 
 	protected KnowledgeBase loadKnowledgeBaseFromString(
 			KieBaseConfiguration kBaseConfig, String... drlContentStrings) {
-		return loadKnowledgeBaseFromString(null, kBaseConfig, phreak,
+		return loadKnowledgeBaseFromString(null, kBaseConfig, phreak, null,
 				drlContentStrings);
 	}
 
-	protected KnowledgeBase loadKnowledgeBaseFromString( KnowledgeBuilderConfiguration config, KieBaseConfiguration kBaseConfig, RuleEngineOption phreak, String... drlContentStrings) {
+	protected KnowledgeBase loadKnowledgeBaseFromString( KnowledgeBuilderConfiguration config, KieBaseConfiguration kBaseConfig, RuleEngineOption phreak, NodeFactory nodeFactory, String... drlContentStrings) {
```
mariofusco, line 124: Probably you could just add an overloaded version of this method instead of changing all its invocation with an additional null value

drools-core/src/main/java/org/drools/core/common/BaseNode.java
```diff
@@ -189,4 +189,15 @@ public int getAssociationsSize(Rule rule) {
...
+    //Default implementation
+    public boolean thisNodeEquals(final Object object) {
+        return false;
+    }
+
+    //Default implementation
+    public int nodeHashCode()
+    {
+        return this.hashCode();
+    }
```
mariofusco, line 202: I'm not getting the meaning of this default implementations (especially the thisNodeEquals one). Since BaseNode is an abstract class I believe it would be safer to leave them out, thus obliging the concrete nodes to implement them properly.

drools-core/src/main/java/org/drools/core/reteoo/AlphaNode.java
```diff
@@ -193,26 +197,44 @@ public int calculateHashCode() {
...
-               constraint instanceof MvelConstraint ?
-               ((MvelConstraint)constraint).equals( other.constraint, getKnowledgeBase() ) :
-               constraint.equals( other.constraint );
+        return this.nodeHashCode() == other.nodeHashCode() &&
+                constraint instanceof MvelConstraint ?
+                ((MvelConstraint) constraint).equals(other.constraint, getKnowledgeBase()) :
+                constraint.equals(other.constraint);
+    }
+
+    public int nodeHashCode() {
```
mariofusco, line 222: I believe this is totally unnecessary for AlphaNodes: there the hashcode is already calculated eagerly only once on node creation and then cached.

drools-core/src/main/java/org/drools/core/reteoo/BetaNode.java
```diff
@@ -570,8 +574,8 @@ public boolean equals(final Object object) {
         final BetaNode other = (BetaNode) object;
 
         return this.getClass() == other.getClass() &&
```
mariofusco, line 576: Since the hashCode is now cached also on BetaNodes I think it would be faster if as first thing you compare the 2 hashcode and return false in case they're different.

drools-core/src/main/java/org/drools/core/reteoo/LeftInputAdapterNode.java
```diff
@@ -474,6 +476,13 @@ public int hashCode() {
     }
 
     public boolean equals(final Object object) {
+        final LeftInputAdapterNode other = (LeftInputAdapterNode) object;
```
mariofusco, line 479: This implementation assumes that the equals method is always passed with another LeftInputAdapterNode, but if this is not the case (as it indeed my happen) throws a ClassCastException

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/1b3c16ba5969b70890214320b006367966418cc2

PR #630, opened 2016-02-03
[DROOLS-1051] KieScanner does not initialize if missing to locate ...

tarilabs:
...a kjar dependency of test scope.

See Jboss JIRA https://issues.jboss.org/browse/DROOLS-1051

Suppose a kjar has a test scope dependency which may not be resolvable
at runtime; the KieContainer will be created correctly (warning raised
in log) but the initialization of the KieScanner fail with NPE.
Expected behavior: also the KieScanner shall be able to initialize if
the KieContainer is working correctly.

Description of proposed patch.
Problem is that if the dependency "dep" is not resolvable at runtime,
the method artifactResolver.resolveArtifact(...) will return a null, and
in turn a NPE, and in turn will fail the initialization of the
KieScanner.
Proposal is to null guard the rest of the above mentioned code - if the
variable "artifact" is not resolved, not considered as a KJar.

The ideal would be to have also the scope in the dependency/artifact
information, so to check for this specifically, but frankly couldn't
locate this data to propose a better patch for now - in case just reject
it and let me know a few pointers I'm keen in getting it corrected
because we have this issue on our projects.

Many thanks.

### DIFF
--- a/kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java
+++ b/kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java
@@ -15,6 +15,7 @@
 
 package org.kie.scanner;
 
+import org.apache.maven.model.Dependency;
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
 import org.drools.compiler.kie.builder.impl.InternalKieScanner;
 import org.drools.core.util.FileManager;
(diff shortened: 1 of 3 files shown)


mariofusco:
@tarilabs the pull request looks ok, but I believe you should also check the scope of the dependency ( DependencyDescriptor.getScope() ) and make the KieScanner to fail fast if the dependency is not optional.

tarilabs:
Missed that method as I developed this change against the `6.2.x` branch, and modified the commit just before submitting the PR to have it against master... Thanks for pointing it out ! =)

I will submit an update to this PR, to take that into account accordingly.

tarilabs:
@mariofusco updated, and tidied by squash, to reflect your comments. Kindly let me know if sufficient or needs further improvements somewhere, please? Thank you;

tarilabs:
Ciao @psiroky , posting this comment just to notify you as well of the latest update on this PR, as I noticed your email on the Drools dev forums ;-) I hope this helps all related issues.

psiroky:
@tarilabs thanks. I am currently looking into this one and the other issue mentioned in the forums. It seems that these changes do not directly fix the issue with test/provided tests in all the cases.

In any case, thanks for the PR. These changes are definitely looking good, we will just need to add a test case.

tarilabs:
Unless I've overlooked something, providing a test case requires to build the test pom a bit differently, in order to be able to define scope and systemPath of the dependencies. Before I get "too creative" I've just proposed the test case as a separate commit on this PR.

Kindly let me know if this is OK I can squash everything before the merge, otherwise I would need some pointers to amend the test, please. Hope this helps and thanks for the feedbacks so far

psiroky:
ok to test

psiroky:
The test looks good. If you could squash the commits, that would be great.

kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java
```diff
@@ -142,6 +143,27 @@ public void testKScannerWithDependencies() throws Exception {
...
+        MavenRepository repository = getMavenRepository();
+        KieServices ks = KieServices.Factory.get();
+        ReleaseId releaseIdWithDep = ks.newReleaseId( "org.kie", "test-with-dep", "1.0-SNAPSHOT" );
+
+        Dependency dep1 = dependencyWithScope( "org.kie", "test-dep1-in-test-scope"    , "1.0-SNAPSHOT", "test"     );
+        Dependency dep2 = dependencyWithScope( "org.kie", "test-dep2-in-provided-scope", "1.0-SNAPSHOT", "provided" );
+        Dependency dep3 = dependencyWithScope( "org.kie", "test-dep3-in-system-scope"  , "1.0-SNAPSHOT", "system"   );
+        dep3.setSystemPath(fileManager.getRootDirectory().getAbsolutePath());
+        // to ensure KieBuilder is able to build correctly:
+        repository.deployArtifact(ks.newReleaseId("org.kie", "test-dep3-in-system-scope", "1.0-SNAPSHOT"), new byte[]{}, new byte[]{});
```
psiroky: Due to a recent changes on master, this needs to be changed to `installArtifact()` to make the test work. You can see the failure when you rebase your changes on top of the current master.

tarilabs:
:+1: rebased on master, squashed, and thanks for the heads up about `installArtifact()`

psiroky:
@tarilabs thanks. Looks good to me.

@mariofusco could you please take a final look?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/80a06ea4b

PR #631, opened 2016-02-04
[DROOLS-465] avoid adding local repository among remote ones

mariofusco:


manstis:
+1 to merge (no tests again?!? but I understand it may be difficult to write such a test).

psiroky:
It seems that those failed tests are directly related to this change.

mariofusco:
Yes this commit fixes the issue reported here https://issues.jboss.org/browse/DROOLS-465 but also breaks SNAPSHOTs resolution. 

I'm still working on this. Please leave this PR open for now.

PR #632, opened 2016-02-04
[DROOLS-465] avoid adding local repository among remote ones

mariofusco:


manstis:
@mfusco, "Do [you] feel lucky? Well, do ya, punk?"... (Dirty Harry)

manstis:
Assuming all green, +1 to merge.

manstis:
@mariofusco 

BTW, in Guvnor we also _deploy_ to any `RemoteRepository` defined in the Project's `pom.xml` `<distributionManagement>` section: https://github.com/droolsjbpm/guvnor/blob/master/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java#L372

IDK if it's something you should consider (if you do, we could probably change Guvnor to use KIE's `MavenRepository` instead - although we'd need the ability to programmatically add a `RemoteRepository` representing "Business Central" that's not defined in any `pom.xml` or `settings.xml`)

psiroky:
It seems there are no tests for the deploy() method. I think it would be useful to add some.

I am also still not sure why it is needed to create the KPOM from scratch in some of the places (instead of reusing one instance), but that does not seem important.

mariofusco:
@psiroky The deploy() methods are no longer in use inside kie-ci. All kie-ci tests do an install on the local repository now. 

I left those method there for now, because I discussed this with @manstis and we would like to concentrate all the install/deploy facilities inside kie-ci while at the moment business-central does this on its own using directly the aether API.

I don't see a way to add a test case for those deploy() methods anyway because the only repository where I could do a deploy is the local one and for what I've found yesterday doing a deploy into the local repository is considered an antipattern.

Let me know if I can merge this PR or if you have any further suggestion on how to test those deploy() methods.

manstis:
I do some "test" deploys here: https://github.com/manstis/guvnor/blob/BPMSPL-252/guvnor-project/guvnor-project-backend/src/test/java/org/guvnor/common/services/project/backend/server/RepositoryResolverTest.java

Basically I setup a temporary "remote" Repository (in /tmp/...) and provide details in an overriding settings.xml file.. but if KIE has no need of "deploy" at the moment it's probably unnecessary.

psiroky:
Yes, what @manstis mentioned should work. Basically creating tmp remote filesystem-based repository (preferable in target/) and deploy there. 

If think that we need to test the deploy if the method is meant to stay in the API.

mariofusco:
@psiroky @manstis I cannot test a deploy in that way because at the moment kie-ci can only make a deploy on the local repository (using it as it was a remote one). I need to change the kie-ci API in order to make it possible to define the repository where the deploy will have to be performed. In this way also business-central could use directly this feature provided by kie-ci.

However I believe that this change is out of the scope of this fix so I created another jira ( https://issues.jboss.org/browse/DROOLS-1057 ) to keep track of this. If it's ok for you I'll merge this fix as it is for now, even considering that it's true that the deploy method is not tested now, but anyway it is also not used at all and its signature and implementation has to change.

manstis:
@mariofusco @psiroky I'm happy to +1, given Mario's comments.

psiroky:
Agreed. +1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e40c2ccd84e3a2d27160b0054b40d2f6c63db4f5

PR #633, opened 2016-02-05
BPMSPL-252: [GSS-RFE] Ability to check duplicate of ReleaseId (GroupId, ArtifactId and Version) with existing projects

manstis:
See https://issues.jboss.org/browse/BPMSPL-252

This change just exposes the `Aether( MavenProject project )` constructor.

### DIFF
--- a/kie-ci/src/main/java/org/kie/scanner/Aether.java
+++ b/kie-ci/src/main/java/org/kie/scanner/Aether.java
@@ -15,6 +15,10 @@
 
 package org.kie.scanner;
 
+import java.io.File;
+import java.util.Collection;
+import java.util.HashSet;
+
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
 import org.apache.maven.settings.Settings;
@@ -36,15 +40,11 @@ import org.kie.scanner.embedder.MavenSettings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.File;
-import java.util.Collection;
-import java.util.HashSet;
-
-import static org.kie.scanner.embedder.MavenProjectLoader.loadMavenProject;
+import static org.kie.scanner.embedder.MavenProjectLoader.*;
 
 public class Aether {
 
-    private static final Logger log = LoggerFactory.getLogger(Aether.class);
+    private static final Logger log = LoggerFactory.getLogger( Aether.class );
 
     private String localRepoDir;
     private final boolean offline;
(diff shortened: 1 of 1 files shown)


manstis:
Part of the following ensemble:

https://github.com/droolsjbpm/drools/pull/633
https://github.com/droolsjbpm/guvnor/pull/240
https://github.com/droolsjbpm/kie-wb-common/pull/236
https://github.com/droolsjbpm/drools-wb/pull/117
https://github.com/droolsjbpm/kie-wb-distributions/pull/187

manstis:

psiroky:
Could we use the upstream JIRA project(s) instead of BPMSPL in the commit message?

manstis:
@psiroky Yes, if I had another JIRA reference.. but I don't. I seem to remember you've requested this before.. I guess I need to create a new JIRA somewhere and update all the branch names and PR titles?

psiroky:
Yes, we should create upstream JIRAs for the BPMSPL ones, unless they already exist. 

I see this spans multiple PRs/repos, so it might be painful to update all of those. I will leave this decision for the gatekeepers, if they see this as something which needs to be done :)

manstis:
@psiroky <sigh> I must try better to remember this... as it causes a load of unnecessary pain (for me) :(

manstis:
_Documentation_
https://github.com/droolsjbpm/kie-docs/pull/40

csadilek:
@manstis @psiroky Yes, in general we have to make sure the JIRAs used in commit messages are accessible to the community. Not sure it's worth updating the 5 pending PRs though in this case as the commit message is descriptive enough. So, I'd prefer going ahead. Once this is merged, I can merge the 4 others.

psiroky:
@csadilek I think all these PRs should be merged together (to prevent build failures). So would you prefer merging all of these yourself (including this one) or coordinating with other gatekeepers to push here and immediately after that you would push to the other repos?

csadilek:
@psiroky it doesn't look like merging this PR here will cause a build failure anywhere or am I missing something? This just makes a constructor public. Once this is merged by you or Mario, I will merge the others which are depending on it.

psiroky:
TBH I haven't really looked at the changes, sorry, You are right that this one is fine. But what about others, similar to this (but actually breaking something)? Should multiple gatekeepers coordinate, or one would merge everything? Just asking as this is probably not defined well, so would be great to clear it a bit (I don't mind either way).

psiroky:
+1. Merged into master.

csadilek:
@psiroky thanks. Yes, good point in general. Don't see another way than the gatekeepers coordinating. I mean it could be a single gatekeeper taking care of the merge if everything was plus one'd but I would not feel comfortable merging anything in drools. So, I'd prefer some coordination.

PR #634, opened 2016-02-05
Make KieFileSystem.write() smart enough to write java files under src…

ge0ffrey:
…/main/java/... instead of src/main/resources/...


manstis:
A test would be nice :)

ge0ffrey:
Test added :)

manstis:
It's not my call - I'm just being nosey on a Friday afternoon - but IMO an assertion that the KieModule's content would be nice (that the path /src/main/java contains the Java file and /src/main/resources contains the DRL...) but I'm happy to be `@Ignore("manstis is irrelevant")` :)

ge0ffrey:
@manstis see my extra commit, that does that assertion for the java class (other tests do it for drl resources already)

PR #639, opened 2016-02-11
Memory file uniqueness (ignored test, which is reproducer asked by Mario)

ge0ffrey:
This also includes this PR:
  https://github.com/droolsjbpm/drools/pull/638
So I recommend merging that PR first!


mariofusco:
Merged and fixed by https://github.com/droolsjbpm/drools/pull/642

PR #641, opened 2016-02-11
use getDefaultPath() instead of hardcoded comparison

ge0ffrey:


etirelli:
Geoffrey, yesterday I tried to simply do this change, but it broke a number of tests. Since I don't have the time to investigate the test failures at the moment, I postponed the change. Not sure if @mariofusco has time to look into it, but we can't merge this until we fix the other failures.

ge0ffrey:
Jenkins CI agrees. Looks like the resourceType is often null in tests, causing an NPE. I 'll close this PR for now as it's invalid.

PR #642, opened 2016-02-11
fix MemoryFileSystem

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
```diff
@@ -356,6 +361,11 @@ public void write(String pResourceName,
     public void write(String pResourceName,
                       byte[] pResourceData,
                       boolean createFolder) {
+        if (pResourceData.length == 0 && pResourceName.endsWith( java.io.File.separator )) {
```
ge0ffrey, line 364: the resource names always use "/", never "\" IIRC, because they simulate the contract of ClassLoader.getResource() which also never uses backslashes.
File.separator resolves to "\" on windows.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/874ebb638

PR #643, opened 2016-02-12
DROOLS-1062 Reproducer for regression of ProjectClassLoader.getResources().

ge0ffrey:
Does not include fix, only reproducer.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
@@ -26,17 +26,20 @@ import org.kie.api.builder.KieModule;
 import org.kie.api.builder.ReleaseId;
 import org.kie.api.definition.type.FactType;
 import org.kie.api.io.Resource;
+import org.kie.api.io.ResourceType;
 import org.kie.api.runtime.KieContainer;
 import org.kie.api.runtime.KieSession;
 
+import java.io.IOException;
 import java.io.StringReader;
 import java.lang.reflect.Constructor;
+import java.net.URL;
 import java.util.ArrayList;
+import java.util.Enumeration;
 import java.util.List;
 
 import static org.drools.compiler.integrationtests.IncrementalCompilationTest.createAndDeployJar;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.*;
 
 public class KieContainerTest {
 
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Note that this PR is failing because it contains a reproducer for the regression but not a fix.
The regression blocks PLANNER-522 which in turn blocks some of the optaplanner-wb work.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
```diff
@@ -251,6 +254,58 @@ public void testMemoryFileSystemFolderUniqueness() {
...
+                         "</kmodule>";
+
+        // Create an in-memory jar for version 1.0.0
+        ReleaseId releaseId = kieServices.newReleaseId("org.kie", "test-delete", "1.0.0");
+        createAndDeployJar(kieServices, kmodule, releaseId, resource1, resource3);
+
+        KieContainer kieContainer = kieServices.newKieContainer(releaseId);
+
+        ClassLoader classLoader = kieContainer.getClassLoader();
+        assertEnumerationSize(2, classLoader.getResources("org/drools/testdrl"));
```
ge0ffrey, line 296: This size should expect 1 instead of 2 (see screenshot of https://issues.jboss.org/browse/DROOLS-1062 for reason).
Currently it fails because the size is actually 0.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
```diff
@@ -251,6 +254,58 @@ public void testMemoryFileSystemFolderUniqueness() {
...
+        String kmodule = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
+                         "<kmodule xmlns=\"http://www.drools.org/xsd/kmodule\">\n" +
+                         "  <kbase name=\"testKbase\" packages=\"org.drools.testdrl\">\n" +
+                         "    <ksession name=\"testKsession\"/>\n" +
+                         "  </kbase>\n" +
+                         "</kmodule>";
+
+        // Create an in-memory jar for version 1.0.0
+        ReleaseId releaseId = kieServices.newReleaseId("org.kie", "test-delete", "1.0.0");
+        createAndDeployJar(kieServices, kmodule, releaseId, resource1, resource3);
```
psiroky, line 291: I presume the `resource2` should be added as well here?

psiroky:
@mariofusco, @ge0ffrey I found the issue and a way to fix it. I will create a new PR with this test (fixed) and the proposed fix.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieContainerTest.java
ge0ffrey, line 291: Yes, good catch. thanks.

I 'll fix this and the other problem in the PR

psiroky:
I've created https://github.com/droolsjbpm/drools/pull/644 which contains this reproducer + proposed fix. I believe this PR could be closed now.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/55d6445e1

PR #644, opened 2016-02-13
[DOOLS-1062] test case + proposed fix

psiroky:
@mariofusco, @ge0ffrey this PR contains the test case from #643 and suggested fix in MemoryKieModule.

With these changes, those two failing optaplanner tests are now passing.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
@@ -167,7 +167,10 @@ public class MemoryFileSystem
     }
 
     public boolean existsFolder(String path) {
-        return folders.get( path ) != null;
+        if (path == null) {
+            throw new NullPointerException("Path can not be null!");
+        }
+        return folders.get( path.length() > 0 ? MemoryFolder.trimLeadingAndTrailing( path ) : path ) != null;
     }
 
     public boolean existsFile(String path) {
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
@@ -16,13 +16,18 @@
 package org.drools.compiler.kie.builder.impl;
 
 import org.drools.compiler.commons.jci.readers.ResourceReader;
+import org.drools.compiler.compiler.io.Folder;
+import org.drools.compiler.compiler.io.Resource;
 import org.drools.compiler.compiler.io.memory.MemoryFileSystem;
 import org.drools.compiler.kproject.models.KieModuleModelImpl;
 import org.drools.core.common.ResourceProvider;
 import org.kie.api.builder.ReleaseId;
 import org.kie.api.builder.model.KieBaseModel;
 import org.kie.api.builder.model.KieModuleModel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -30,11 +35,16 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 
 public class MemoryKieModule extends AbstractKieModule
         implements
         ResourceReader {
+
+    private static final Logger logger = LoggerFactory.getLogger(MemoryKieModule.class);
+
+    private static final String MEMORY_URL_PROTOCOL = "mfs";
 
     private final MemoryFileSystem mfs;
     private final long creationTimestamp = System.currentTimeMillis();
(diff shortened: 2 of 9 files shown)


drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
```diff
@@ -167,7 +167,7 @@ public boolean existsFolder(MemoryFolder folder) {
     }
 
     public boolean existsFolder(String path) {
-        return folders.get( path ) != null;
+        return folders.get( path.length() > 0 ? MemoryFolder.trimLeadingAndTrailing( path ) : path ) != null;
```
triceo, line 173: NPE is lurking here. Are we 100 % sure `path != null`?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
```diff
@@ -184,4 +219,51 @@ public void connect() throws IOException {
...
+        }
+
+        private InputStream folderMembersToInputStream(Folder folder) {
+            StringBuilder sb = new StringBuilder();
+            Collection<? extends Resource> members = folder.getMembers();
+            if (members != null) {
+                for (Resource resource : members) {
+                    // take just the name of the member, no the whole path
+                    sb.append(resource.getPath().toRelativePortableString(folder.getPath()));
+                    // append "\n" to be in sync with the JDK's ClassLoader (returns "\n" even on Windows)
```
triceo, line 262: There is a whole bunch of different JDKs. Are we sure this statement holds across all of them?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
```diff
@@ -30,12 +35,17 @@
...
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 
 public class MemoryKieModule extends AbstractKieModule
         implements
         ResourceReader {
 
+    private static final Logger logger = LoggerFactory.getLogger(MemoryKieModule.class);
```
triceo, line 45: As a constant, shouldn't this be ALL_CAPS? The very next line has an all-caps constant.

drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
psiroky, line 173: Since the method is public we can never be sure. I don't know what caller could expect passing null (NPE?), but there should be no harm adding one more null check there.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
psiroky, line 262: If the JDKs have different behavior we can't really tailor to all of them anyway (having different code for different JDKs would be just horrible). Will try to see if this somehow defined in the JVM specification.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
psiroky, line 45: Loggers are getting different treatment from me. I am not exactly sure where this originated (I guess I must have seen it somewhere). It makes it slightly easier to use since you don't have start with upper case letter.

There is about 130 other classes just in Drools using the same. If we would ever decided to move to ALL_CAPS, it should be done in all the places.

ge0ffrey:
+1 (only reviewed the tests)

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
ge0ffrey, line 45: That's probably because most were like this at some point:

```
 protected final transient Logger logger = Loggerfactory.getLogger(getClass());
```

But for exotic cases that's a memory/performance overhead. Although it's much more extending classes friendly.

psiroky, line 45: Interesting read about the different static/non-static logger issues http://wiki.apache.org/commons/Logging/StaticLog (older, but still applicable I think).

Apart from the slight performance benefits of static loggers (would be nice to see some kind of benchmark for that), it also points of one interesting point: _avoid static references to log objects from code that might be deployed in a shared classpath_. The reasons for that are described in the article. However, I am not exactly sure if that could be the case for Drools?

Anyway, it would certainly be nice to standardize on a single way of creating Loggers, although from the information I have now I don't know what that should be.

ge0ffrey, line 45: We don't use commons-logging because of it's classpath issues (and it's not
just the static loggers that it has a problem with IIRC). I don't think
that that commons-logging advice also applies to slf4j/logback, but it's
worth finding out.

2016-02-14 15:24 GMT+01:00 Petr Široký notifications@github.com:

> In
> drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/MemoryKieModule.java
> https://github.com/droolsjbpm/drools/pull/644#discussion_r52843417:
> 
> >  import java.util.Collection;
> > 
> >  public class MemoryKieModule extends AbstractKieModule
> >          implements
> >          ResourceReader {
> > -    private static final Logger logger = LoggerFactory.getLogger(MemoryKieModule.class);
> 
> Interesting read about the different static/non-static logger issues
> http://wiki.apache.org/commons/Logging/StaticLog (older, but still
> applicable I think).
> 
> Apart from the slight performance benefits of static loggers (would be
> nice to see some kind of benchmark for that), it also points of one
> interesting point: _avoid static references to log objects from code that
> might be deployed in a shared classpath_. The reasons for that are
> described in the article. However, I am not exactly sure if that could be
> the case for Drools?
> 
> Anyway, it would certainly be nice to standardize on a single way of
> creating Loggers, although from the information I have now I don't know
> what that should be.
> 
> —
> https://github.com/droolsjbpm/drools/pull/644/files#r52843417.

## 

Geoffrey De Smet

psiroky, line 45: I believe it does. The article is on common-logging wiki, but it describes the general issues with static loggers which apply to slf4j as well (afaik).

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/55d6445e1

PR #645, opened 2016-02-15
Fix OutOfBoundException on MemoryFileSystem.existsFile("")

ge0ffrey:


drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
```diff
@@ -170,11 +170,11 @@ public boolean existsFolder(String path) {
         if (path == null) {
             throw new NullPointerException("Path can not be null!");
         }
-        return folders.get( path.length() > 0 ? MemoryFolder.trimLeadingAndTrailing( path ) : path ) != null;
+        return folders.get(MemoryFolder.trimLeadingAndTrailing(path)) != null;
     }
 
     public boolean existsFile(String path) {
-        return fileContents.containsKey( MemoryFolder.trimLeadingAndTrailing( path ) );
+        return fileContents.containsKey(MemoryFolder.trimLeadingAndTrailing(path));
```
psiroky, line 177: Not directly related to this PR, but should we check if the path is null? I've added the check to existsDir() and forgot to add it here. I think that we should use the same for both (whatever we choose). WDYT?

ge0ffrey, line 177: I'd say no (but this is opinion), not until we see a valid use case that pass a null value to this method. We can add a javadoc with "@param path never null" to clearly indicate the contract of this method.

A null check like that is "defensive code". Defensive code is great for public api's, but a performance consideration and maintenance worry for internal code. I strongly prefer just javadoccing params and return values as either "never null" xor "sometimes null".

Furthermore, a null check can be handle in 2 ways: fail fast by throwing a clear IllegalState|ArgumentException (this should be your default behaviour!) or make the method null-safe. In this particular case, making the method null-safe is doable and reasonable (just return null), but in most cases it's not reasonable because it causes bug later down the line (without a correct error message or a outright misbehavior without an error message at all).

psiroky:
+1. Merged by https://github.com/droolsjbpm/drools/commit/75b81dcbd26a11d5b370ea5bdf102c2d62ebd0c3.

drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
ge0ffrey, line 177: For more info on my opinion, see optaplanner docs section 17.2 Development guidelines:
  http://docs.jboss.org/optaplanner/release/latest/optaplanner-docs/html_single/index.html#developmentGuidelines

psiroky, line 177: I've added the null check for consistency.

psiroky, line 177: I would say passing null into existsFile method is not legal (but opinion as well). We could of course just return `false` in this case to be null-safe.

Even though the method should not be used outside Drools/jBPM, it still is public, so anyone can pass anything into it even it they should not. Documenting it sounds like a good idea. We could also consider something like @NotNull annotations (I could not find any really standard library for this. There are multiple ones providing these)?

PR #648, opened 2016-02-15
Backport MSF changes to fix IndexOutOfBoundsException on "" for trim path

ge0ffrey:
I needed to backport all 3 commits.

@mariofusco This includes the first commit which you originally didn't desire to have on 6.4.x too...

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -125,6 +125,11 @@
     </dependency>
 
     <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>net.java.dev.glazedlists</groupId>
       <artifactId>glazedlists_java15</artifactId>
       <scope>test</scope>
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/io/Folder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/io/Folder.java
@@ -17,7 +17,6 @@ package org.drools.compiler.compiler.io;
 
 import java.util.Collection;
 
-
 public interface Folder extends Resource {
     String getName();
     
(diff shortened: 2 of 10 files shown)


psiroky:
Build failed, because I've used Java 7/8 API in the commit: `java.nio.charset.StandardCharsets.UTF_8` as I thought we won't need to backport this to 6.4.x. We will need to use IoUtils.UTF_8 (or what is it called).

ge0ffrey:
jenkins please retest this

It failed with:
[ERROR] Failed to execute goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.13:check (check-java-1.6-compatibility) on project drools-compiler: Signature errors found. Verify them and ignore them with the proper annotation if needed. -> [Help 1]

ge0ffrey:
Ok, I 'll fix that and update the PR. I had build it locally with JDK 8, instead of 7.

psiroky:
The error comes from animal-sniffer, so it should fail even on JDK 8. Could you please double check that? (make sure to run with -Dfull). If it does not fail, we need to fix it.

ge0ffrey:
@psiroky Just tried it again: it definitely does fail without -Dfull. And I don't do -Dfull when committing as it takes too long...

psiroky:
You mean it does _not_ fail without -Dfull (seems to be the case when looking at -kie-parent-metadata)? In any any case, that's why we have these PR jobs...

ge0ffrey:
yes indeed, success without -Dfull. This PR jobs are great indeed!

2016-02-15 19:01 GMT+01:00 Petr Široký notifications@github.com:

> You mean it does _not_ fail without -Dfull (seems to be the case when
> looking at -kie-parent-metadata)? In any any case, that's why we have these
> PR jobs...
> 
> —
> https://github.com/droolsjbpm/drools/pull/648#issuecomment-184327271.

## 

Geoffrey De Smet

PR #652, opened 2016-02-19
GUVNOR-2030: Guided Decision Table: Empty field results in all Conditions being cancelled

manstis:
See https://issues.jboss.org/browse/GUVNOR-2030

This PR adds extraction of "Template Keys" to the `DRL`->`RuleModel` parsing to support use of `IPattern` and `IAction` in BRL Fragments in XLS to Guided Decision Table conversion.


Rikkola:
+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/07748111b

PR #653, opened 2016-02-22
[BZ-1310261] fix incremental compilation when a rule makes reference to a global defined in a .gdrl file

mariofusco:


manstis:
@mariofusco The change seems quite reasonable; however what guarantee is there that neither DSL or GDRL have changed? Is it not possible that either (or both) the DSL and GDRL file content _has_ changed in a KJAR update?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -254,16 +256,27 @@ private void updateKBase( InternalKnowledgeBase kBase, InternalKieModule current
...
         for (String file : newKM.getFileNames()) {
-            if (ResourceType.DSL.matchesExtension(file) && !cs.contains(file)) {
+            if ( includeIfUnchanged( file ) && !cs.contains( file ) ) {
                 dslFiles.add(file);
             }
         }
         return dslFiles;
     }
 
+    private static final ResourceType[] TYPES_TO_BE_INCLUDED = new ResourceType[] { ResourceType.DSL, ResourceType.GDRL };
```
mariofusco, line 269: @manstis The problem was caused by the fact that the resources of certain types (the ones that aren't directly translated into rules but act as a support for the conversion to drl) have to be included in the incremental compilation process even if they didn't change. I already did this but only for .dsl files. Now I also added .gdrl and checked that this fixes the reported problem, but I suspect that other types of resources may need the same treatment. Can you please help me to check if there are other ResourceTypes that should be added to this list?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -254,16 +256,27 @@ private void updateKBase( InternalKnowledgeBase kBase, InternalKieModule current
         }
     }
 
-    private List<String> getUnchangedDslFiles(InternalKieModule newKM, KieJarChangeSet cs) {
+    private List<String> getUnchangedResources( InternalKieModule newKM, KieJarChangeSet cs ) {
         List<String> dslFiles = new ArrayList<String>();
         for (String file : newKM.getFileNames()) {
-            if (ResourceType.DSL.matchesExtension(file) && !cs.contains(file)) {
+            if ( includeIfUnchanged( file ) && !cs.contains( file ) ) {
```
mariofusco, line 262: @manstis if the .dsl/.gdrl files are changed they are in the change set, so they aren't added to the list here but managed as any other changed resource.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
manstis, line 269: @mariofusco Hi... OK. Understood. The Workbench doesn't have any more supporting `ResourceType` - so as far as this change is concerned it looks OK. However consider this scenario: A user has 2 DRL files; one containing a `function` and the other a `rule`. The `rule` makes use of the `function`. The `rule` is updated as part of an incremental change..... do you need the `function` too? Does this also apply for `queries`, declared types etc?

manstis:
+1 to merge (there may be a wider issue that would probably be best tracked as a separate JIRA)

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
mariofusco, line 269: @manstis Your scenario is not a problem. Functions, queries and declared types are in the KieBase that is going to be updated. Since the resource from which they have been created is not changed they are not removed and then they are still available for other rules that makes use of them. Conversely the content of .dsl and .gdrl is not directly translated in an asset stored in the KieBase, but just used as support for compilation of other files and then discarded. That is why it has to be readded during a subsequent incremental compilation even if it didn't change and also why I was asking if you're aware of other types of resources with this same characteristic.

manstis:
+1 to merge. Thanks for the explanations @mariofusco

PR #654, opened 2016-02-22
Faster check if pojo has property change listener support.

vkrot:
Checking if class has `addPropertyChangeListener` via `Class#getMethod` is very slow, as it throws exception, populating its stacktrace is quite an expensie operation. In many usecases all objects injected to session may not have this method, so this operation is not rear. Looking for method by name is faster and increases overall performance. In our application this gave ~ 1.7 better throughput.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/ClassObjectTypeConf.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/ClassObjectTypeConf.java
@@ -35,6 +35,7 @@ import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -149,14 +150,13 @@ public class ClassObjectTypeConf
     }
 
     private boolean checkPropertyListenerSupport( Class<?> clazz ) {
-        Method method = null;
-        try {
-            method = clazz.getMethod( "addPropertyChangeListener",
-                                      ADD_REMOVE_PROPERTY_CHANGE_LISTENER_ARG_TYPES );
-        } catch (Exception e) {
-            // intentionally left empty
+        for (Method method : clazz.getMethods()) {
+            if ("addPropertyChangeListener".equals(method.getName()) &&
+                    Arrays.deepEquals(ADD_REMOVE_PROPERTY_CHANGE_LISTENER_ARG_TYPES, method.getParameterTypes())){
+                return true;
+            }
         }
-        return method != null;
+        return false;
     }
 
     /**


mariofusco:
I'm not sure this will be a performance improvement in all cases. It really depends on how many methods the bean under inspection has. I suspect that if there are many of them the for-loop may be actually slower, but it's hard to guess. Can you please clarify how did you measure that performance improvement and how many methods the class you used for your benchmark has? Also it would be great to check what happens when you try to increase the number of those methods.

vkrot:
I created a simple JMH test - https://github.com/vkrot/drools-pr-jmh
It fires rules for pojos with 10, 30 and 100 public methods, with and without property change listener.
I run it twice, with and without my patch
Results without patch:

```
DroolsBenchmark.testPojo100WithPropertyChangeListener     thrpt   60  110843,172 ± 4620,967  ops/s
DroolsBenchmark.testPojo100WithoutPropertyChangeListener  thrpt   60   90321,611 ± 3367,654  ops/s
DroolsBenchmark.testPojo10WitPropertyChangeListener       thrpt   60  111485,053 ± 4977,910  ops/s
DroolsBenchmark.testPojo10WithoutPropertyChangeListener   thrpt   60   91307,725 ± 3965,244  ops/s
DroolsBenchmark.testPojo30WithPropertyChangeListener      thrpt   60  112825,160 ± 5108,829  ops/s
DroolsBenchmark.testPojo30WithoutPropertyChangeListener   thrpt   60   92673,035 ± 3984,959  ops/s
```

results with patch:

```
DroolsBenchmark.testPojo100WithPropertyChangeListener     thrpt   60  127594,990 ± 3582,671  ops/s
DroolsBenchmark.testPojo100WithoutPropertyChangeListener  thrpt   60  125212,307 ± 3639,989  ops/s
DroolsBenchmark.testPojo10WitPropertyChangeListener       thrpt   60  189475,308 ± 8918,434  ops/s
DroolsBenchmark.testPojo10WithoutPropertyChangeListener   thrpt   60  183095,107 ± 9224,133  ops/s
DroolsBenchmark.testPojo30WithPropertyChangeListener      thrpt   60  171798,659 ± 6746,250  ops/s
DroolsBenchmark.testPojo30WithoutPropertyChangeListener   thrpt   60  172786,004 ± 7061,823  ops/s
```

to reproduce it check out benchmark and run 

```
mvn clean package
java -jar target/benchmarks.jar
```

As you can see, you are right, performance degrades as you add more methods. But even for really bad usecase (I suppose 100 methods per class is not very common) it still outperforms original version.
And it is still better for case when pojo does have property change listener, because `InitalFactImpl` is always inserted and it does not have property change listener.

By default this runs with patch. To run version without patch - delete ClassObjectTypeConf class.

mrietveld:
`Class.getMethods()` is a relatively expensive operation since all superclasses have to be scanned going up the tree and put together in an array. I think that this case has not been taken into account in your performance tests? 

In cases in which the POJO has a deep hierarchy of parent tests, `Class.getMethods` will always be slower than `Class.getMethod(methodName, argTypes)`.

mrietveld:
It may be worth looking at the following code: 

OpenJDK 6_14 [Class.getMethod(String name, Class<?>... parameterTypes)](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/Class.java#Class.getMethod%28java.lang.String%2Cjava.lang.Class[]%29)

OpenJDK 6_14 [Class.getMethods()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/Class.java#Class.getMethods%28%29)

mariofusco:
@mrietveld Sorry I just merged this PR before reading your comments :(
Do you think I should revert this?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/d24cb6019

mrietveld:
@mariofusco Personally, yes, I do not think this is a good idea and yes I think this should be reverted.

In short, this is code that tries to be smarter than the JDK itself. There's no way to tell if there are optimizations in the JDK (of which there are of course multiple implementations and versions) and which ones will be faster. Furthermore, (AFAIK?), this is not a pressing problem with regards to optimization. As Knuth has said: "Premature optimization is the root of all evil"...

vkrot:
This code does not try to be smarter, I tried to eliminate throwing/catching exceptions at each call, because it is expensive to create exception, fill the stacktrace. I will find detailed benchmark later. Another solution - just cache already checked classes,  this will also eliminate exceptions and will not suffer from deep hierarchies problem. What do you think?

PR #656, opened 2016-02-24
BZ1301051: Column values are deleted after changing the operator in a Guided Decision Table

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1301051

This PR simply removes the values set for `valueString` and `dataType`. This is necessary as when a `DTCellValue52` is serialised from the server to the client the fields are set with reflection not the constructor and hence a NUMERIC `DTCellValue52` would still have a `valueString==""`.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/shared/model/DTCellValue52Test.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/shared/model/DTCellValue52Test.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/shared/model/DTCellValue52Test.java
```diff
@@ -0,0 +1,131 @@
...
+import org.drools.workbench.models.datamodel.oracle.DataType;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+// These tests check legacy (pre-6.4.CR1) objects that have been de-serialized can be correctly read.
+// DTCellValue52.valueString was set to an empty String statically and hence when the object was de-serialized
+// it was possible for instances to have both a 'valueString (=="")' and '<another>Value' for types other than
+// DataTypes.DataType.TYPE_STRING. For example. valueBoolean=true, valueString="", dataType=BOOLEAN.
```
Rikkola, line 31: Not sure if it matters, but I usually use the multiline comment for javadocs. This way the editor pick it up. You can read the comment when hovering the mouse over the class for example.

Rikkola:
+1

PR #660, opened 2016-02-25
Move btm1.tlog and btm2.tlog from module dir into target/btm directory

ge0ffrey:


ge0ffrey:
Don't merge yet, looks like this fails drools-persistence-jpa

ge0ffrey:
jenkins please retest this

ge0ffrey:
Ready to merge now

manstis:
I suspect Mario or Petr will want the commits squashed?

ge0ffrey:
At BCN we clearly decided that squashing of commits for 1 feature is optional, as long as they are all in the same PR. I 've done a backport PR.

I don't like squashing: it takes extra time, it deletes valuable intermediate state from the git history. If multiple contributors worked together on the same feature it misinterprets who actually wrote which subsystem (and that is bad - not for accountability or proper attribution like OpenHUB - but for being able to discuss change proposals with the original author).

manstis:
Hehe, yes such valuable commits that your first broke drools-persistence-jpa and the second fixed it. Very valuable...but I don't really care. Your call.

ge0ffrey:
It's valuable to know that the static getConfiguration() trick didn't work for all use cases ;)

mswiderski:
alternatively we could use null journal (pretty much no op journal) that is suitable for tests, similar as we do in jBPM for all the tests to avoid unrecoverable issues when InterruptedException is thrown which kill file based journal

https://github.com/droolsjbpm/jbpm/blob/master/jbpm-persistence-jpa/src/test/resources/bitronix-default-config.properties

ge0ffrey:
@mswiderski that's a great idea!

ge0ffrey:
Looks like you guys still clean the btm files though:
  https://github.com/droolsjbpm/jbpm/blob/master/jbpm-persistence-jpa/pom.xml#L235

mswiderski:
yes, as you can still reconfigure it to use disc journal locally and to avoid any accidental commits with this files it was left there. Of course it does not guarantee it will be cleaned before it's committed but still might help...

ge0ffrey:
@mswiderski I've tested adding both the journal null and defining the tlog files in the target directory. That works. So no need for cleaning any more (even with local journal override).
Shall you guys also reuse this kie-test-util like drools-persistence-jpa and optaplanner-persistence-jpa? That way we can be consistent, use the same PersistenceUtil implementation and add DB specific exceptions once.

ge0ffrey:
Improved with maciej's suggestions.

Squashed, so the bureaucracy can expand to meet the needs of the expanding bureaucracy. :)

ge0ffrey:
Build failure by jenkins is false negative: Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Failure to find org.jboss.errai.bom:errai-internal-bom:pom:4.0.0-SNAPSHOT in https://repository.jboss.org/nexus/content/groups/public/ was cached in the local repository, resolution will not be reattempted until the update interval of jboss-public-repository-group has elapsed or updates are forced

psiroky:
Yeah, looks like Nexus (repository.jboss.org) is down.

psiroky:
+1 to merge. Thanks @ge0ffrey!

PR #661, opened 2016-02-25
[DROOLS-1025] Incremental Rule Refactor + new tests

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -8309,4 +8310,44 @@ public void testIn() {
         assertEquals( "Bob", list.get( 0 ) );
     }
 
+    @Test
+    public void testXXX() {
```
triceo, line 8314: Could we maybe have a more descriptive name? :-)

manstis, line 8314: I can always help with descriptive test names; I have quite a reputation ;-)

mariofusco, line 8314: This PR is incomplete, Mark and me are still working on it. I used that test only to compare the behaviour of drools with the same set of rules but with and without incremental compilation. I'll remove that test when I'll be done with this.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/8d52a6a34

PR #671, opened 2016-03-01
[6.4.x] [DROOLS-1065] fix ClasspathKieProject to recognize "*.jar" dirs

psiroky:
- test creates a kmodule jar file and a kmodule jar directory
  (ending name in ".jar") for testing that both are processed correctly

Backport of https://github.com/droolsjbpm/drools/pull/511. We were asked by community member to backport this to 6.4 (see the JIRA). I believe this is a safe change, so I would agree with backporting.

@mariofusco please take a look.


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/e30ca18c3

PR #672, opened 2016-03-02
DROOLS-1025 - AddRemove rules tests - generated rules tests using not(not()) constraint.

baldimir:
- Added class with generated AddRemove rule tests that use not(not(...)) constraint in rules. 
- Extracted reproducers of incremental compilation failures into AddRemoveRulesTest(last 3 tests in the class).
- Added exception catch in tests so we know which test permutation failed in case of an exception.


mariofusco:
Incorporated by https://github.com/droolsjbpm/drools/pull/679

PR #673, opened 2016-03-03
-Flushing had an incorrect smemIndex++, which would skip the sm that …

mdproctor:
…needed flushing.

-Refactored the flushing methods to have more shared code
-Removed the redundant lian argument to various network evaluation methods
-Removed the break from findPmemToBeFlushed, as this was causing pmems not to be flushed, that needed it
-removed redundant endNode argument


mariofusco:
Incorporated by https://github.com/droolsjbpm/drools/pull/679

PR #678, opened 2016-03-08
[JBPM-5019] DisposeCommand is not part of the public API

agiertli:


agiertli:
The failed test does not seem to be related to this PR. Can somebody please confirm and let me know whether some fix on my side is needed or not?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/2546412fd

PR #680, opened 2016-03-09
[6.4.x] [DROOLS-1074] fix usage of org.kie.spring.KModuleSpringMarshaller

psiroky:
- fixed the reflective call to the method from kie-spring
  (removed not needed argument)
- fixed the URL/path parsing method to strip the suffix
  KieModuleModelImpl.KMODULE_JAR_PATH only if the path
  actually ends with it (otherwise we are incorrectly stripping
  content from paths which do not end with that suffix)

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
@@ -158,11 +158,11 @@ public class ClasspathKieProject extends AbstractKieProject {
         }
     }
 
-    private static void fetchKModuleFromSpring(URL kModuleUrl, String fixedURL){
+    private static void fetchKModuleFromSpring(URL kModuleUrl, String fixedURL) {
         try{
             Class clazz = Class.forName("org.kie.spring.KModuleSpringMarshaller");
-            Method method = clazz.getDeclaredMethod("fromXML", java.net.URL.class, String.class);
-            method.invoke(null, kModuleUrl, fixedURL);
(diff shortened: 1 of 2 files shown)


mariofusco:
Is there any reason why you sent this PR against 6.4.x branch and not on master?

psiroky:
I'll send one for master as well once this gets merged.

I had to develop and test on 6.4.x, because fuse-bxms-integ does not yet use KIE 7.0.0-SNAPSHOTs.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/51509dc5d

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
```diff
@@ -359,7 +359,7 @@ public static String fixURLFromKProjectPath(URL url) {
         } else {
             if (url.toString().contains("-spring.xml")){
                 urlPath = urlPath.substring( 0, urlPath.length() - ("/" + KieModuleModelImpl.KMODULE_SPRING_JAR_PATH).length() );
-            } else {
+            } else if (url.toString().endsWith(KieModuleModelImpl.KMODULE_JAR_PATH)) {
```
psiroky, line 362: @mariofusco looking at the code again, should I use `urlPath.endsWith()` instead of `url.toString().endsWith()`? Since the suffix is removed from the `urlPath` it seems I should use that. WDYT?

psiroky:
I forgot to mention it, but https://github.com/droolsjbpm/droolsjbpm-integration/pull/350 is directly related. Build should not fail, since the class is called via reflection, but it will fail during runtime when trying to call non-existing method.

PR #677, opened 2016-03-08
DROOLS-1082: Correct how ClassAwareObjectStore manages cached objectstores when insert / queries come in unexpected order

jelford:
This commit simplifies how ClassAwareObjectStore its internal datastructures, and ensure the linked structure is properly updated when read and writes come out of order.

The problem was that the way the linking structure was put together went down a different branch depending on whether the nodes had already been initialized (e.g. by an earlier read). This change reduces the branching, so it's clearer what transformations need to be applied for each access.
(cherry picked from commit d92aad7)

This commit should close https://issues.jboss.org/browse/DROOLS-1082, which I recently opened. It adds some basic unit tests to ClassAwareObjectStore that are mainly focused on the behaviour being changed.

PR #682, opened 2016-03-10
[DROOLS-1082] fix ClassAwareObjectStore when a class is queried befor…

mariofusco:
…e any instance of that class is inserted


jelford:
@mariofusco #677 also included some changes that were meant to simplify ClassAwareObjectStore (making the intent of those for-loops a bit clearer), to hopefully avoid similar mistakes in future.

Can I re-suggest them - as I while they constitute a larger overall change, I think they were worthwhile from a maintenance perspective?

mariofusco:
See comment in https://github.com/droolsjbpm/drools/pull/677

PR #683, opened 2016-03-10
[DROOLS-1074] fix usage of org.kie.spring.KModuleSpringMarshaller

psiroky:
- fixed the reflective call to the method from kie-spring
  (removed not needed argument)
- fixed the URL/path parsing method to strip the suffix
  KieModuleModelImpl.KMODULE_JAR_PATH only if the path
  actually ends with it (otherwise we are incorrectly stripping
  content from paths which do not end with that suffix)

Same change already on 6.4.x (https://github.com/droolsjbpm/drools/pull/680)


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/7d46e95ce

PR #684, opened 2016-03-10
[DROOLS-1089] flush reentrant working memory actions when eager evalu…

mariofusco:
…ation is enforced


mswiderski:
jbpm test case included in https://github.com/droolsjbpm/jbpm/pull/420

+1 as it fixes the issue and does not break anything else

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/f79c2115f

PR #685, opened 2016-03-11
flush reentrant working memory actions when eager evaluation is enforced

mswiderski:
this is a copy of #684 to allow test of jbpm test case for drools-1089 - don't merge it - it will be closed when jBPM is verified and #684 will be merged as expected.


mariofusco:
@mswiderski Can we close this now?

mswiderski:
yes, closing it

PR #686, opened 2016-03-11
GUVNOR-2455: Guided Decision Tree: Unable to re-open trees using data-type suffixes (d, f, L) in Actions

manstis:
See https://issues.jboss.org/browse/GUVNOR-2455

The issue was that `double`, `float` and `long` data-types have `d`, `f` and `L` appended to their value in the generated `DRL` that lead to the fields being interpreted as `Formula` (that are unsupported by the Guided Decision Tree editor) when unmarshalling DRL->Model.


Rikkola:
+1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/94235a293

PR #691, opened 2016-03-16
DROOLS-1025 Added TestContext to AddRemoveRules tests + new NPE fail discovered.

baldimir:
- Created new TestContext, so we can batch TestOperations and observe state of the engine during tests. 
- Found new NPE - reproducer is test method AddRemoveRulesTest.testAddRemoveFacts(). 

Stacktrace: http://pastebin.com/Cngs78d9


baldimir:
New PR because of conflicts here https://github.com/droolsjbpm/drools/pull/692

PR #692, opened 2016-03-17
DROOLS-1025 - Added tests that add and remove rules while adding and removing facts.

baldimir:
- Created new TestContext, so we can batch TestOperations and observe state of the engine during tests.
- Found new NPE - reproducer is test method AddRemoveRulesTest.testAddRemoveFacts().
- New NPE in new tests in class AddRemoveRulesAddDeleteFactsTest.


mariofusco:
Incorporated by https://github.com/droolsjbpm/drools/pull/704

PR #696, opened 2016-03-21
BZ 1319742: add reproducer

winklerm:
Add reproducer for issue when creating KieBase with a XSD resource using KieClasspathContainer.


mariofusco:
Merged and fixed by https://github.com/droolsjbpm/drools/commit/5624ddf6a

PR #698, opened 2016-03-21
BZ 1319777: add reproducer

winklerm:
Add reproducer for NPE when marshalling a persistent KieSession with partially matched forall.


mariofusco:
Merged and fixed by https://github.com/droolsjbpm/drools/commit/b5062bb3997f1c2a535a97b9ce411e228b2e270d

PR #701, opened 2016-03-22
[BZ-1319742] skip xsd files not having a jaxb configuration

mariofusco:


psiroky:
@mariofusco just a small note about the squashing. I think in cases like this (reproducer provided via PR) it would be better to have 2 commits. The first one would be from Marek with the reproducer and the second one from you with the fix. This would clearly separate who did what. Currently there is only a commit from Marek so he is also be the git author of the fix, which is a bit misleading. This is more about future us looking at the history and wondering who to go to ask about particular parts of the sources.

mariofusco:
@psiroky I agree, I'll avoid the squashing in future.

PR #702, opened 2016-03-23
[DROOLS-1026] improve equals/hashCode performances for all rete nodes

mariofusco:


chdasif:
Mario,
My initial tests are looking good. The performance has increased significantly. I have tried loading KB with sharing on/off. The difference is reasonable.
20k rules with sharing on took 52 sec vs sharing off that took 46 sec to load. i also debugged and it seems that hashcode comparison is helping an early exit. so that's good.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/08d6538b3

PR #704, opened 2016-03-24
[DROOLS-1025] fix left tuples deletion when there are more than 2 removed sinks

mariofusco:


mariofusco:
Merged by https://github.com/droolsjbpm/drools/compare/a8fc81a...0eb011a

PR #705, opened 2016-03-29
[DROOLS-1108] make kproject ClassLoader accessible from inside a rule

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -8366,4 +8367,50 @@ public void testNonSerializableInEvaluatorWrapper() throws Exception {
...
+
+    @Test
+    public void testClassLoaderGetResourcesFromWithin() throws IOException {
+        // DROOLS-1108
+        KieServices kieServices = KieServices.Factory.get();
+        String drl1 = "package org.drools.testdrl;\n" +
+                      "global java.util.List list;\n" +
+                      "rule R when\n" +
+                      "then\n" +
+                      "   java.net.URL url = drools.getProjectClassLoader().getResource(\"META-INF/foo.xml\");\n" +
```
mswiderski: where's the drools "variable" coming from? is it similar to kcontext

mariofusco:
@mswiderski drools and kcontext are synonym (exactly the same object) and then they can be used interchangeably

mswiderski:
great, that's what I wanted to hear (read) :)

etirelli:
Mario, this seems to be a step forward, but what happens if a pojo class inside the kjar tries to do get getClass().getResource(...)? Will it work?

mariofusco:
@etirelli I think that if the pojo belongs to the same kjar it should work because it will be loaded by the same ProjectClassLoader, but I'll add another test to double check this.

etirelli:
Looks good Mario! +1 to merge! 

Thanks!

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/dac087491e86e95482965b1436ffcf8452fe5cd0

PR #708, opened 2016-03-30
avoid synchronization on segment creation

mariofusco:


mariofusco:
ok to test

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/1e50f13e6

PR #710, opened 2016-03-31
[BZ-1319808] reproducer for resolving LATEST artifact from remote repo

psiroky:


psiroky:
Closing, as there is no easy solution right now to fix this.

PR #714, opened 2016-04-04
[DROOLS-1059] race condition when retrieving a query node trying to create a segment from a prototype

mariofusco:


mariofusco:
DO NOT MERGE (for now)

@baldimir I tried to reproduce the issue reported here https://issues.jboss.org/browse/DROOLS-1059 by creating a more complex version of the reproducer attached to that jira, but with no luck. I also tried to find any possible concurrency issue in that area by statically reading the code, but I couldn't find anything. The small change I made in SegmentUtilities is just intended to simplify a piece of code that was unnecessary complex, but the old version and mine are functionally equivalent. Please take a look at my reproducer and check if you're able to modify it in order to reproduce this issue.

@mdproctor I really don't see how that exception could be possibly thrown: there's no dynamic  KieBase change involved here, so there shouldn't be any problem retrieving that QueryLiaNode regardless of the concurrency. Am I overlooking something?

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/b815ca37e6749bf4c028060552dea10504670951

@baldimir I investigated this issue a bit more and I'm pretty sure that I found what was causing this issue (even if I haven't been able to reproduce it). I discussed my findings with @mdproctor and he also agreed that the problem had to be there.

PR #716, opened 2016-04-04
[DROOLS-1112] prevent wrong node sharing for 2 betas with same left input but different right one

mariofusco:


mariofusco:
ok to test

PR #718, opened 2016-04-06
GUVNOR-2173 - Change Impact: Modifications for use in for drools-wb

mrietveld:
This is a small method that's useful in a number of the drools-wb indexers. 

@manstis Would you mind reviewing?


manstis:
@mrietveld other than code formatting, first parameter being non-final (the others are) and lack of tests it looks fine.. I assume they'll be a test or two coming? (You can mock the hell out of it... so it'll be simply).

mrietveld:
@manstis Thanks for harping about tests -- I'll add one and clean up the code and let you know when I've done that.

PR #719, opened 2016-04-06
[BZ-1323310] allow custom user maven settings in kie-ci to also accept an URL

mariofusco:


mariofusco:
@etirelli according to what written by Matthew in his last comment https://bugzilla.redhat.com/show_bug.cgi?id=1323310#c4 one of the necessary step to allow seamless integration between kie-ci and Fabric is allowing to also pass a URL (other than a regular java File) as custom user settings.

I'm not sure if this change will be enough to close that issue (still waiting for Matthew's reply about that), but in all case I believe that this is a nice and useful improvement for kie-ci, also regardless of Fabric integration.

mariofusco:
ok to test

etirelli:
Looks good, +1 to merge.

kie-ci/src/main/java/org/kie/scanner/embedder/MavenEmbedder.java
```diff
@@ -111,8 +113,8 @@ protected MavenExecutionRequest buildMavenExecutionRequest( MavenRequest mavenRe
             mavenExecutionRequest.setGlobalSettingsFile( new File( mavenRequest.getGlobalSettingsFile() ) );
         }
 
-        if ( mavenRequest.getUserSettingsFile() != null ) {
-            mavenExecutionRequest.setUserSettingsFile( new File( mavenRequest.getUserSettingsFile() ) );
+        if ( mavenRequest.getUserSettingsSource() instanceof FileSettingsSource ) {
```
paoloantinori, line 116: @mariofusco hi Mario, shouldn't this code be changed to avoid the `instanceof` call and instead always build a new temporary/cached file from the generic `InputStream` that class `SettingSource` exposes?

mariofusco:
@paoloantinori you're right, I will implement this fix. Anyway this is totally unrelated with the issue reported in the bugzilla, correct?

paoloantinori:
Well, tbh I'm not entirely sure. Since it's using only `FileSettingsSource` what happens if the settings were specified with the newer support for `return new UrlSettingsSource( new URL( customSettings ) );` ?

Is it losing information in that case?

mariofusco:
@paoloantinori That MavenEmbedder is used only when parsing a kjar's pom to find the dependencies eventually declared in parent poms, so it shouldn't have any impact on retrieval of a single kjar as reported in that ticket. Anyway I agree this has to be fixed, I'll push a fix asap.

paoloantinori:
I see. Ok!

PR #720, opened 2016-04-07
[BZ-1323167] lazily instance KieScanner used by KieRepository

mariofusco:


mariofusco:
@etirelli I'm now creating the KieScanner used by the KieRepository lazily and this should solve the problem with bundles installation sequence reported here https://bugzilla.redhat.com/show_bug.cgi?id=1323167

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieRepositoryImpl.java
```diff
@@ -61,20 +61,32 @@
...
+                                                                                                DEFAULT_VERSION));
+
     public static final KieRepository INSTANCE = new KieRepositoryImpl();
 
     private final KieModuleRepo kieModuleRepo;
 
-    private InternalKieScanner internalKieScanner;
+    private static class KieScannerHolder {
+        // Use holder class idiom to lazily initialize the kieScanner
+        private static final InternalKieScanner kieScanner = getInternalKieScanner();
```
psiroky, line 74: This looks like constant, could we call it `KIE_SCANNER`?

etirelli:
+1

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieRepositoryImpl.java
mariofusco, line 74: It is "incidentally" a constant inside that inner class in order to implement the class holder lazy initialization idiom. Anyway, yes I think that it is more correct to write it in upper case. I will fix it, thanks

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/aebe63fab

PR #726, opened 2016-04-11
[DROOLS-1078] fix class introspection in case of an overridden getter with a covariant return

mariofusco:


mariofusco:
ok to test

mariofusco:
@etirelli Jenkins reported an intermittent problem with Misc2Test.testFunctionInvokingFunction. At the moment I'm not able to reproduce that failure (even running with 100 threads instead of 10) and I'm still trying to figure out if we could have a race condition somewhere in that area. Anyway this failure is totally unrelated with the fix in this PR.

etirelli:
Fix looks good Mario. +1

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/bac3ffe30

PR #727, opened 2016-04-11
[DROOLS-1064] fix NPE when a wrong field name is the same of a declaration

mariofusco:


etirelli:
If tests succeed, +1 mario.

mariofusco:
Fixed by https://github.com/droolsjbpm/drools/commit/cb94ef640

PR #728, opened 2016-04-11
[DROOLS-1057] Change kie-ci API to make it able to deploy on any remote repository

mariofusco:


mariofusco:
@manstis Note that the reported test failure is in drools-compiler (I know we have some random failures there and trying to figure out what to do about it) and then totally unrelated with this. I also tried to honour <distributionManagement> in kproject's pom as you suggested.

kie-ci/src/main/java/org/kie/scanner/MavenRepository.java
```diff
@@ -188,32 +191,38 @@ public Version resolveVersion( String artifactName ) {
...
-        File jarFile = bytesToFile( releaseId, jarContent, ".jar" );
-        File pomFile = bytesToFile( releaseId, pomContent, ".pom" );
-        deployArtifact( releaseId, jarFile, pomFile );
+    private RemoteRepository getRemoteRepositoryFromDistributionManagement( File pomfile ) {
+        MavenProject mavenProject = parseMavenPom( pomfile );
+        DistributionManagement distMan = mavenProject.getDistributionManagement();
+        if (distMan == null) {
+            return null;
+        }
+        DeploymentRepository deployRepo = mavenProject.getVersion().endsWith( "SNAPSHOT" ) ?
```
manstis: See https://github.com/droolsjbpm/guvnor/blob/master/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java#L381 (if the Artifact is a SNAPSHOT but no SnapshotRepository is defined; the non-SNAPSHOT Repository should be used)

kie-ci/src/main/java/org/kie/scanner/MavenRepository.java
```diff
@@ -188,32 +191,38 @@ public Version resolveVersion( String artifactName ) {
...
+                                          distMan.getSnapshotRepository() :
+                                          distMan.getRepository();
+        if (deployRepo == null) {
+            return null;
+        }
+        return new RemoteRepository.Builder( deployRepo.getId(), "default", deployRepo.getUrl() ).build();
```
manstis: Does this take into account authentication for a server defined in settings.xml. See https://github.com/droolsjbpm/guvnor/blob/master/guvnor-m2repo-editor/guvnor-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/GuvnorM2Repository.java#L518 where we setup authentication for the RemoteRepository defined in DistributionManagement.

kie-ci/src/main/java/org/kie/scanner/MavenRepository.java
```diff
@@ -188,32 +191,38 @@ public Version resolveVersion( String artifactName ) {
     public void deployArtifact( ReleaseId releaseId,
                                 InternalKieModule kieModule,
                                 File pomfile ) {
-        File jarFile = bytesToFile( releaseId, kieModule.getBytes(), ".jar" );
-        deployArtifact( releaseId, jarFile, pomfile );
+        RemoteRepository repository = getRemoteRepositoryFromDistributionManagement( pomfile );
+        if (repository == null) {
```
manstis, line 209: It's not an exception if no Repository is defined in DistributionManagement; it simply means don't deploy to any additional repositories.

kie-ci/src/main/java/org/kie/scanner/embedder/MavenSettings.java
```diff
@@ -38,9 +40,23 @@
...
+            settings = initSettings(userSettingsSource);
+            mavenConf = new MavenRepositoryConfiguration(settings);
+            Aether.instance = null;
+            MavenProjectLoader.mavenProject = null;
+            MavenRepository.defaultMavenRepository = null;
+        }
+    }
+
+    // USE ONLY FOR TESTING PURPOSES
+    public static void reinitSettings() {
```
manstis: Would package protected be a better option (IDK if tests would need to move to another package)?

manstis:
@mariofusco OK, comments made.. I seem to remember there being a typo last night too (when I had a quick look but I can't find it now). You'll also probably want to squash your commits.

kie-ci/src/main/java/org/kie/scanner/MavenRepository.java
mariofusco, line 209: In this overloaded version of the deployArtifact method no RemoteRepository is passed, so I assume that I should take that repo from the distribution management. If also there's no distribution management I don't know where the artifact should be deployed so I just throw that Exception. Should I just silently ignore this issue and don't perform any deploy? Should I log a warning message?

mariofusco:
@manstis I pushed another commit implementing what you suggested in your comments.

kie-ci/src/main/java/org/kie/scanner/MavenRepository.java
manstis, line 209: OK, fair enough.. assuming the JavaDocs make it clear how each overloaded method works.. then it should be OK... oh.. you do have JavaDocs....?

kie-ci/src/main/java/org/kie/scanner/embedder/MavenSettings.java
manstis: No thoughts on this being package protected?

manstis:
@mariofusco Thanks.. I've made a couple of other minor comments.. otherwise it seems fine.

Once you've considered my last comments I can +1 this PR...

kie-ci/src/main/java/org/kie/scanner/embedder/MavenSettings.java
mariofusco: Done.

mariofusco:
@manstis I added a few javadocs, in MavenRepository to clarify how the deployArtifact and installArtifact methods work. I know that many other javadocs are missing. I'll try to add them asap.

manstis:
@mariofusco Looks good - thanks.. want to squash before I +1?

mariofusco:
@manstis I will squash and merge it. Thanks

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/2f9b2c740

PR #730, opened 2016-04-12
[DROOLS-1110] synchronize access to statefulsessions in kbase

mariofusco:


mariofusco:
@etirelli I discussed this fix with @mdproctor and he also thinks it is ok. Please let me know if I can merge into master and backport to 6.4.x

PR #731, opened 2016-04-12
Fixes after Weld 1.x -> 2.x upgrade

psiroky:
@mariofusco, @mdproctor could you please review? This is kind of urgent as the kie-wb is currently not working without the Weld 2.x upgrade.

Depends on:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/173

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/cdi/KieCDIExtension.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/cdi/KieCDIExtension.java
@@ -47,9 +47,9 @@ import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.Bean;
-import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.InjectionPoint;
+import javax.enterprise.inject.spi.PassivationCapable;
 import javax.enterprise.inject.spi.ProcessInjectionTarget;
 import javax.enterprise.util.AnnotationLiteral;
 import javax.inject.Named;
@@ -63,6 +63,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 
 public class KieCDIExtension
     implements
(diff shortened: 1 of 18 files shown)


drools-compiler/src/main/java/org/drools/compiler/cdi/KieCDIExtension.java
```diff
@@ -165,7 +165,7 @@ public void init() {
 
     public void addKBaseInjectionPoint(InjectionPoint ip, KBase kBase, Class< ? extends Annotation> scope, ReleaseId releaseId, KReleaseId kReleaseId) {
         if ( kBaseNames == null ) {
-            kBaseNames = new HashMap<KieCDIEntry, KieCDIEntry>();
+            kBaseNames = new HashMap<>();
```
mariofusco, line 168: Is this fix intended to be pushed only on master? If so that's ok otherwise we should avoid using the diamond operator (and other non Java 6 compatible features)

psiroky, line 168: Yes, this will go only into master. I thought I removed these fixes (I did few more of those, but then decided that we should probably do these tweaks for the repo at the same time). I can remove this one to make it more consistent in the meantime...

mariofusco:
The failing test is not caused by this PR, but since it is CDI related it would be great why it is intermittently failing. 

+1 for me

drools-examples-cdi/cdi-example-scopes/src/test/java/org/drools/example/cdi/scopes/RequestScopedRulesJUnitTest.java
```diff
@@ -21,34 +32,33 @@
...
+
+    private ContextControl ctxCtrl = BeanProvider.getContextualReference(ContextControl.class);
+
     @Inject
     private MyRequestScopedBean myRequestBean;
 
+    @Before
+    public void setup() {
+        // make sure the RequestScoped context is stopped, otherwise the test fails
+        ctxCtrl.stopContext(RequestScoped.class);
```
psiroky: @mariofusco this should fix the randomly failing test. From what I can tell, the test was somehow influenced by the tests running before it. However, I was not able to figure out what exactly was causing it. This setup code makes sure we start with a clean plate before the test.

mariofusco: @psiroky This fix is ok for me, even if I just checked all other tests and they seem to correctly close all contexts they use at the end. Unless there isn't a former test that doesn't terminate correctly? (but in that case also that test should have been failed). Probably a cleaner solution would be to move all the stopContext invocations in @After methods?

psiroky: Yeah, I was counting those start() -> stop() methods as well and could find the cause.

Let me try to move the stopContext() invocations into @After methods to see if works.

PR #734, opened 2016-04-14
BZ 1327202: add reproducer

winklerm:
Added reproducer for BZ 1327202.


mariofusco:
Included in https://github.com/droolsjbpm/drools/pull/738

PR #737, opened 2016-04-15
[DROOLS-1123] make TraitObjectTypeNode serializable

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/factmodel/traits/TraitTest.java
```diff
@@ -5941,4 +5943,44 @@ public ExtEntity( String id, int num ) {
...
+                     "then list.add( 1 ); end " +
+
+                     "rule Test2 " +
+                     "when " +
+                     "  IStudent( this isA StudentImpl.class ) " +
+                     "then list.add( 2 ); end " +
+
+                     "";
+
+        KnowledgeBase kbase = getKieBaseFromString( drl );
```
psiroky, line 5967: Just a thought (not really something to hold off the PR): I think we should starting using the KIE API even in the tests. KnowledgeBase is deprecated and with every usage, we get a compiler warning.

PR #738, opened 2016-04-15
[BZ-1327202] submit queries on the propagation queue

mariofusco:


mariofusco:
Merged by https://github.com/droolsjbpm/drools/compare/1cf6f3d...626009a

PR #739, opened 2016-04-18
[DROOLS-1039] fail fast when there is an invalid kie.conf

mariofusco:


ge0ffrey:
+1 (really getting some food now :)

PR #740, opened 2016-04-18
- fixing byteman tests for non-oracle-based JDK's

mrietveld:
@baldimir Better late than never.. :)


baldimir:
Hi Marco, do we have similar tests that run on IBM JDK? I know that these tests were never executed on IBM because Byteman, but I want to be sure that we have covered what these tests cover also on IBM JDK.

mrietveld:
@baldimir Thanks for picking this up -- I originally thought that it was impossible to run byteman tests on IBM JDK, but I'm seeing that that's not true. 

I'm currently trying to figure out how to apply the fix @istudens that used in Wildly/EAP: https://issues.jboss.org/browse/AS7-5914

I'll update the PR and add a message letting you know when I've done that.

baldimir:
Great, thanks Marco.

PR #741, opened 2016-04-18
- fixing byteman tests for non-oracle-based JDK's (cherry picked from…

mrietveld:
… commit dcd5ca5f72e1532df261ad83b8d7a40da33f5331)


baldimir:
My comment from master PR: Hi Marco, do we have similar tests that run on IBM JDK? I know that these tests were never executed on IBM because Byteman, but I want to be sure that we have covered what these tests cover also on IBM JDK.

PR #744, opened 2016-04-19
JBPM-5084: troubleshooting function for workitem

bxf12315:
@mswiderski 
related issue in here https://issues.jboss.org/browse/JBPM-5084.
I not sure it is reasonable, so please check it.
and i don't know how to create unit test for it , it just is draft code .

### DIFF
--- a/drools-core/src/main/java/org/drools/core/process/instance/impl/DefaultWorkItemManager.java
+++ b/drools-core/src/main/java/org/drools/core/process/instance/impl/DefaultWorkItemManager.java
@@ -112,6 +112,16 @@ public class DefaultWorkItemManager implements WorkItemManager, Externalizable {
     	}
     }
 
+    public void retryWorkItemWithParams(long workItemId,Map<String,Object> map) {
+        WorkItem workItem = workItems.get(workItemId);
+        
+        if (workItem != null) {
+            workItem.setParameters( map );
+            
+            retryWorkItem(workItem.getId());
+        }
+    }
+    
     public Set<WorkItem> getWorkItems() {
         return new HashSet<WorkItem>(workItems.values());
     }
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
@@ -54,7 +54,7 @@ public class JPAWorkItemManager implements WorkItemManager {
         Environment env = this.kruntime.getEnvironment();
         WorkItemInfo workItemInfo = new WorkItemInfo(workItem, env);
         
-        PersistenceContext context = ((PersistenceContextManager) env.get( EnvironmentName.PERSISTENCE_CONTEXT_MANAGER )).getCommandScopedPersistenceContext();
+        PersistenceContext context = getPersistenceContext();
         workItemInfo = context.persist( workItemInfo );
         
         ((WorkItemImpl) workItem).setId(workItemInfo.getId());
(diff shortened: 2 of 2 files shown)


drools-core/src/main/java/org/drools/core/process/instance/impl/DefaultWorkItemManager.java
```diff
@@ -112,6 +112,19 @@ public void retryWorkItem(long workItemId) {
     	}
     }
 
+    public void retryWorkItemWithParams(long workItemId,Map<String,Object> map) {
+        WorkItem workItem = workItems.get(workItemId);
+        
+        if (workItem != null) {
+            workItem.setParameters( map );
```
mswiderski, line 119: except this line rest of the method is a copy of retryWorkItem so maybe retryWorkItem should call this will null parameters which would then be omitted - anyway good to check if coming map is not null as we should never set null as parameters on work item

drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
```diff
@@ -81,9 +81,16 @@ public WorkItemHandler getWorkItemHandler(String name) {
...
-    	WorkItem workItem = getWorkItem(workItemId);
+    public void retryWorkItemWithParams(long workItemId, Map<String,Object> map) {
+        Environment env = this.kruntime.getEnvironment();
+        WorkItem workItem = getWorkItem(workItemId);
+        
     	if (workItem != null) {
+    	    workItem.setParameters( map );
+            WorkItemInfo workItemInfo = new WorkItemInfo(workItem, env);
+            PersistenceContext context = getPersistenceContext();
+            context.merge( workItemInfo );
```
mswiderski, line 92: why do you create new work item info here? wouldn't it be enough to just load existing and set the parameters on it after load?

mswiderski:
we certainly need tests for this and I believe regular retryWorkItem as well, though best to add it to jbpm repo as it's easier to test it when you have processes available

drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
bxf12315, line 92: I want to save update workitem into database.
Is it not necessary?

drools-core/src/main/java/org/drools/core/process/instance/impl/DefaultWorkItemManager.java
bxf12315, line 119: yes you are right
i will update it.

bxf12315:
@mswiderski  
I don't understand the line "though best to add it to jbpm repo as it's easier to test it when you have processes available".
how to add it to jbpm repo ?
I have other suggestion . we can add retryworkitemwithparams into interface  workitemmanager.
We can easily find and use it.

drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
mswiderski, line 92: the work item returned from getWorkItem should be already bound the the entity so any updates on it should be persisted, though we need test for it...

mswiderski:
test should be added to jbpm repo

bxf12315:
@mswiderski 
I was update it. 
Please check it

mswiderski:
I don't see any updates mentioned in the comments? have you pushed the updates to your remote fork?

bxf12315:
@mswiderski  Sorry, update it.

mswiderski:
this can be closed as https://github.com/droolsjbpm/drools/pull/780 comes with updates

PR #745, opened 2016-04-19
[BZ-1328380] Node Hashing with type mismatch condition results in sil…

tkobayas:
…ent misfiring
- Test case


mariofusco:
Incorported by https://github.com/droolsjbpm/drools/compare/5801072...434c6e2

PR #746, opened 2016-04-19
[BZ-1328380] fix node hashing when type coercion is involved

mariofusco:


drools-core/src/main/java/org/drools/core/util/MathUtils.java
```diff
@@ -60,4 +60,7 @@ public static BigInteger getBigInteger(Object value) {
         return ret;
     }
 
+    public static int compare(Number n1, Number n2) {
+        return getBigDecimal( n1 ).compareTo( getBigDecimal( n2 ) );
+    }
```
etirelli, line 65: Are you sure this works @mariofusco ? How does the big decimal handles differences when comparing doubles?

mariofusco:
See https://github.com/droolsjbpm/drools/pull/752

PR #748, opened 2016-04-19
[DROOLS-1126] make org.kie.scanner.ArtifactResolver public

mariofusco:


psiroky:
Not directly related to this PR, but I think we should think about separating maven-related classes (like the ArtifactResolver) into different package, or possibly, module. I believe there are uses cases which require maven-integration, but are actually not interested in the KieScanner (e.g. when you want to collect set of dependencies in kie-maven-plugin). I am not exactly sure if making new module would be better, but perhaps just creating package `org.kie.maven` would be enough.

kie-ci/src/main/java/org/kie/scanner/ArtifactResolver.java
```diff
@@ -41,15 +41,15 @@
...
+public class ArtifactResolver {
 
     private static final Logger log = LoggerFactory.getLogger(KieScanner.class);
 
     private final PomParser pomParser;
 
     private final MavenRepository mavenRepository;
 
-    ArtifactResolver() {
+    public ArtifactResolver() {
```
psiroky, line 52: Would it make sense to make the other constructors public as well?

psiroky:
+1 to merge

PR #749, opened 2016-04-19
[DROOLS-1127] allow to configure a custom ThreadFactory

mariofusco:


mariofusco:
@ge0ffrey I didn't create a KieBaseOption because it doesn't make sense to configure this for the single KieBase.

drools-core/src/main/java/org/drools/core/concurrent/ExecutorProviderImpl.java
```diff
@@ -29,14 +29,29 @@
...
         static {
+            String threadFactoryClass = System.getProperty( THREAD_FACTORY_PROPERTY );
+            ThreadFactory threadFactory;
+
+            if (threadFactoryClass == null) {
+                threadFactory = new DaemonThreadFactory();
+            } else {
+                try {
+                    threadFactory = (ThreadFactory) Class.forName( threadFactoryClass ).newInstance();
+                } catch (Exception e) {
```
ge0ffrey, line 46: Personally, I'd only catch the checked reflection exceptions thrown by those reflection methods. Although because it's **rethrown and chained**, it's not eaten, so it doesn't actually hurt (so ok to leave it as is). 

So I guess the main reasons I 'd do it is because there are checkstyle/PMD rules that state that we should never catch Throwable or Exception - not that there aren't exceptions to that rule - and not that we have any of those active (but automated QA checks of our open source code will have it - think sonar - maybe set up by a 3th party organization or academic research group - and they'll grade us blindly).

ge0ffrey:
+1 to merge

PR #750, opened 2016-04-19
[6.3.x] DROOLS-1106 - xstream xml marshaller does not work for more than one …

mswiderski:
this is as part of patch on 6.2.x product based on https://bugzilla.redhat.com/show_bug.cgi?id=1319548


psiroky:
@mswiderski I presume we need this? And it seems it missed the tag...

mswiderski:
it was merged, just manually and I forgot to close it:
https://github.com/droolsjbpm/drools/commit/fa7f87a977c86c5f37239fc60e4536df1b047b98
https://github.com/droolsjbpm/drools/commit/b546c097e3f7bd2560899e78314c0e6e99d2c1c9

so it should be in the tag... but double check make sense...

psiroky:
Correct, the commits are indeed on 6.3.x branch. Closing.

PR #752, opened 2016-04-20
[BZ-1328380] fix node hashing when type coercion is involved

mariofusco:


drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
```diff
@@ -751,7 +751,7 @@ public void setValue(final int index,
             } else {
                 this.type = OBJECT;
                 if ( !isNull ) {
-                    this.ovalue = value.getValue();
+                    this.ovalue = vtype.coerce( value.getValue() );
```
mariofusco, line 754: @etirelli This coercion based on the field type makes the specific number comparison no longer useful.

drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
```diff
@@ -860,19 +860,20 @@ public boolean equals(final Object object) {
...
+                    return this.bvalue == other.getBooleanValue();
                 case LONG :
-                    return (this.index == other.index) && (this.lvalue == other.getLongValue());
+                    return this.lvalue == other.getLongValue();
                 case DOUBLE :
-                    return (this.index == other.index) && (this.dvalue == other.getDoubleValue());
+                    return this.dvalue == other.getDoubleValue();
                 case OBJECT :
                     final Object otherValue = other.getObjectValue();
-                    if ( (this.ovalue != null) && (this.ovalue instanceof Number) && (otherValue instanceof Number) ) {
```
mariofusco, line 872: @etirelli I added many other test cases trying different combination and I think that this specific comparison for Numbers is no longer necessary due to coercion I'm doing based on the field type.

mariofusco:
@etirelli I reviewed what I did yesterday and think and closed the other PR. See my other comments.

etirelli:
@mariofusco here is my concern with the fix: BigDecimals are tricky for typical users, specially if they don't know that behind the scenes there is a coercion to BigDecimals happening. E.g.:

new BigDecimal("1").equals( new BigDecimal( "1.0" ) ) ->> this returns false

So what happens on your rules if a given Double attribute is 1.0 and the user writes the rule as:

X( doubleAttr == 1.00 ) 
X( doubleAttr == 1.0 )
X( doubleAttr == 1 )

Or all the infinite combinations that can be derived from that?

I might be misunderstanding your changes, but I do think we need to handle non BigDecimal numbers natively as integers or floating point numbers, and only use BigDecimals for explicit BigDecimal attributes.

wdyt?

etirelli:
Ok, after discussing this further with @mariofusco , +1 to merge.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/compare/5801072...434c6e2

PR #754, opened 2016-04-21
[DROOLS-1135] fix race conditions in OTN and sink propagator intialization

mariofusco:


drools-core/src/main/java/org/drools/core/reteoo/ObjectTypeNode.java
```diff
@@ -268,11 +268,15 @@ public void assertInitialFact(final InternalFactHandle factHandle,
         propagateAssert(factHandle, context, workingMemory);
     }
 
-    private void checkDirty() {
+    protected void checkDirty() {
         if (dirty) {
-            resetIdGenerator();
-            updateTupleSinkId(this, this);
-            dirty = false;
+            synchronized (this) {
```
mrietveld, line 273: @mariofusco If you're adding a `synchronized(this)` when setting `dirty = false`, do you then also want to add a `synchronized(this)` when setting `dirty = true` in the `networkUpdated` method?

drools-core/src/main/java/org/drools/core/reteoo/CompositeLeftTupleSinkAdapter.java
```diff
@@ -172,15 +172,19 @@ public BaseNode getMatchingNode(BaseNode candidate) {
...
-            }
+        LeftTupleSink[] sinks = new LeftTupleSink[this.sinks.size()];
+
+        int i = 0;
+        for ( LeftTupleSinkNode sink = this.sinks.getFirst(); sink != null; sink = sink.getNextLeftTupleSinkNode() ) {
+            sinks[i++] = sink;
         }
-        return sinkArray;
+
+        this.sinkArray = sinks;
```
mrietveld, line 186: @mariofusco (Caveat: I _definitely_ do not fully understand the logic here, given that I don't even fully understand the PHREAK algorithm :) ). 

However, it seems like you might be assuming that there's a happens-before guarantee here -- when there isn't one. This logic allows the JVM to assign the empty `sinks` variable to `this.sinkArray` and then fill it in (In other words, line 179, line 186, and _then_ lines 181-184). 

If you're trying to guarantee that `this.sinkArray` is set to the fully-filled `sinks` array, then I think you'll have to make `this.sinkArray` volatile. 

Does that make sense?

drools-core/src/main/java/org/drools/core/reteoo/ObjectTypeNode.java
mariofusco, line 273: @mrietveld After a KieBase is in use the only thing who could set that dirty flag back to true is an incremental compilation, but in that case there will be other higher level synchronization that will guarantee thread-safety.

drools-core/src/main/java/org/drools/core/reteoo/CompositeLeftTupleSinkAdapter.java
mariofusco, line 186: @mrietveld You're totally right. I'll make this.sinkArray volatile.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/compare/89cc437...c94053d

PR #758, opened 2016-04-22
[DROOLS-1139] clone EvalInvoker in EvalCondition evaluator

mariofusco:


etirelli:
+1

drools-core/src/main/java/org/drools/core/rule/GroupElement.java
```diff
@@ -302,7 +300,7 @@ protected GroupElement clone(boolean deepClone) {
         GroupElement cloned = new GroupElement();
         cloned.setType( this.getType() );
         for ( RuleConditionElement re : children ) {
-            cloned.addChild( deepClone && ( re instanceof GroupElement || re instanceof Pattern ) ? re.clone() : re );
+            cloned.addChild( deepClone ? re.clone() : re );
```
mariofusco, line 303: @etirelli I've found another failing test case. The problem was caused that it wasn't cloning the EvalCondition here. Adding a further "|| re instanceof EvalCondition" fixed the problem, but I honestly didn't understand why it shouldn't call clone() regardless of the type of the RuleConditionElement. Any clue about this? Is this fix ok or do you see any drawback?

mariofusco:
ok to test

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/8eadcaaff

PR #759, opened 2016-04-25
 change impact

mrietveld:
@mariofusco (and @etirelli or @mdproctor?) Could you take a look at this code and let me know what you think. 

Depends on: https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/143

This code is **not** ready for merging, but I'm getting pretty close to it. 

Basically, this code collects type-references (and, eventually, part-references) during rule compilation. These type-references are then used for impact-analysis in the workbench
. 

Unfortunately, I still need to do the following: 
1. The  [TypeReferenceNameCollector](https://github.com/mrietveld/drools/blob/03f85c624a39f23c84e63cb212745e394e701c63/drools-compiler/src/main/java/org/drools/compiler/commons/jci/compilers/TypeReferenceNameCollector.java), which is the (eclipse) AST visitor-implementation class used during eclipse java compilation to collect type-reference information, is not fully done. Most importantly, I need to add a _lot_ of tests for this. 
2. I need to think of a way to make processing of the dialect [AnalysisResults](https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/compiler/AnalysisResult.java) implementations optional so that it does not interfere/affect drools use when not used in the workbench. 
   - For example, see:
     - https://github.com/droolsjbpm/drools/pull/759/files#diff-c9ba9357ad0bb25c48c4b41785dc436dR319
     - https://github.com/droolsjbpm/drools/pull/759/files#diff-c9ba9357ad0bb25c48c4b41785dc436dR337
     - https://github.com/droolsjbpm/drools/pull/759/files#diff-e9a6c5c471194c2aa2bb6426ad2a7a53R519

I'm also not happy with [this use](https://github.com/droolsjbpm/drools/compare/master...mrietveld:_change-impact?expand=1#diff-c9ba9357ad0bb25c48c4b41785dc436dR170) of a Map, but I don't see another way to do [this](https://github.com/droolsjbpm/drools/compare/master...mrietveld:_change-impact?expand=1#diff-c9ba9357ad0bb25c48c4b41785dc436dR457)

And lastly, but most importantly, if you have any questions, please ask them!


drools-compiler/src/main/java/org/drools/compiler/builder/impl/ProjectCompilationResult.java
```diff
@@ -0,0 +1,124 @@
...
+            log.error("Null resource found in package " + pkgName);
+        } else {
+            String uniqueResourcePath = null;
+            if( resource.getSourcePath() != null ) {
+                uniqueResourcePath = resource.getSourcePath();
+            } else {
+                if( resource instanceof ByteArrayResource ) {
+                    uniqueResourcePath = "ByteArray@" + System.identityHashCode(resource);
+                }
+                assert resource instanceof ByteArrayResource : "Unexpected resource type without a source path: " + resource.getClass().getName();
```
baldimir, line 56: Hi Marco, shouldn't this be simply part of else statement of previous if clause with some exception thrown? With this assert it is somehow disconnected from previous logic (Personally I'm not a fan of asserts in code except tests, but that can be only my personal preference). 

Edit: Found a bit info on asserts in code. They are normally removed at runtime. See here: http://programmers.stackexchange.com/questions/137158/is-it-better-to-use-assert-or-illegalargumentexception-for-required-method-param

mrietveld, line 56: @baldimir If I did this on purpose, why do you think I did that?

baldimir, line 56: @mrietveld I don't know. Maybe debugging? But for that case loggers can be used.

mrietveld, line 56: @baldimir , the logic is currently this: 

```
if( resource.getSourcePath() != null ) {
  uniqueResourcePath = resource.getSourcePath();
} else {
  if( resource instanceof ByteArrayResource ) {
    uniqueResourcePath = "ByteArray@" + System.identityHashCode(resource);
  }
  assert resource instanceof ByteArrayResource : "Unexpected resource type without a source path: " + resource.getClass().getName();
}
assert uniqueResourcePath != null : "Unable to create unique resource path for resource!";
```

So: 
1. If there's a resource path, then we're fine, we have a unique resource path
2. otherwise it **might** be a `ByteArrayResource`, in which case, we create a unique resource path. 
3. But wait, it also might **NOT** be a `ByteArrayResource`. In that case, I want the assert to fail. 

So..... 3 is why I have the code structured the way I do! 

And as to `assert`, yes, that's **EXACTLY** why I'm using assert here.

baldimir, line 56: After discussion with Marco. This looks to me like an unhandled incorrect state. The state when resource is instance of other type than ByteArrayResource is a state that shouldn't happen, but can happen. So it should be properly handled by exceptions.

PR #765, opened 2016-04-29
Stuff generated by tests should be in target/

psiroky:
- it is not ideal to hardcode the "target/" strings in the
  tests, but it is much better than the previous state where
  the files were created in "user.dir"


drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieLoggersTest.java
```diff
@@ -168,8 +168,8 @@ public void testKieFileLogger() throws Exception {
         // create the builder
         KieSession ksession = getKieSession(dt);
 
-        String fileName = "testKieFileLogger";
-        File file = new File(fileName+".log");
+        String fileName = "target/testKieFileLogger";
```
mariofusco, line 171: Does this work on Windows? Shouldn't we use the File.separator char here?

psiroky, line 171: I haven't tried, but I would expect it to work. At least java.io.File does abstract this away. In any case, I will double-check this.

psiroky, line 171: Other tests are using the same approach. See https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java#L1610.

I've also tried to run the tests directly on Windows machine and they work as well.

PR #766, opened 2016-04-29
[BZ-1312164] guided-dtables: configure correct classloader for xstream

psiroky:
@manstis, @mariofusco please take a look. This fixes the issue I am facing, but I am not 100% sure I am setting the correct classloader. If I understand correctly, this XStream instance just needs access to classes in the guided-dtable module, so if that is the case it should be correct to use the getClass().getClasloader(). However, in case it would also need access to user (domain?) classes which are in different bundle, this might not work.

I am also trying to create simple reproducer in droolsjbpm-integration repo (as we need to run it on karaf).

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTXMLPersistence.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTXMLPersistence.java
@@ -105,6 +105,10 @@ public class GuidedDTXMLPersistence {
         xt.alias( "valueNumeric",
                   Number.class,
                   BigDecimal.class );
+
+        // this is needed for OSGi as XStream needs to be able to load classes from the guided-dtable module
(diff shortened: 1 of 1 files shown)


manstis:
You are sort of correct @psiroky; however it's likely also to need classes in [here](https://github.com/psiroky/drools/tree/64x-bz-1312164-xstream-guided-dtable/drools-workbench-models/drools-workbench-models-datamodel-api) too. Classes there are "common" to the most of the different models (Guided Rule, Guided Rule Template, Guided Decision Table).

psiroky:
Thanks for the additional info @manstis. I believe that should work as well, because guided-dtable OSGi-imports the packages from datamodel-api, and thus the classloader should have access to them. It would, however, be great if someone else could verify my hypothesis....

manstis:
@psiroky Unfortunately I'm unable to verify your hypothesis... as IDK OSGi.. hopefully @mariofusco and his (OSGi-centric) associates can provide the enlightenment you seek.

mariofusco:
+1 (but if possible try to reproduce the problem with a test case before merging)

psiroky:
I finally managed to create a good-enough reproducer, see https://github.com/droolsjbpm/droolsjbpm-integration/pull/462

PR #771, opened 2016-05-03
[DROOLS-1031] prevent to remove a parent rule during incremental compilation if one of its children is still present

mariofusco:


etirelli:
+1

drools-compiler/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/AddRemoveRulesTest.java
```diff
@@ -574,59 +573,20 @@ public void testAddRemoveWithExtends() {
...
-        fact2.put("name", "Michael");
-        fact2.put("test", 1);
-        session.insert(fact2);
-
-        assertTrue(session.fireAllRules() == 0);
+        try {
+            session.getKieBase().removeRule(packageName, "parentRule");
+            fail("A parent rule cannot be removed if one of its children is still there");
+        } catch (Exception e) {
+            // OK
```
psiroky, line 588: Would it make sense to catch more specific exception? Or maybe check the exception message? If the code throws NPE, the test will still succeed which does not seem to be correct.

PR #772, opened 2016-05-04
Fix and unify logback*.xml configs

psiroky:
- using the %line _may_ lower performance,
  but I was not able to observe any difference
  when running the drools build (mvn clean install -T10).
  I believe the line number gives useful info when trying to
  determine what is going wrong.

Basically I just unified the message patterns, set the default logging level to WARN and configured `org.kie` and `org.drools` packages to long also INFO messages.

We also need to figure out how to share the default `logback-test.xml` as it is ridiculous to have to copy&paste the config into every new module. This sounds promising, but will need to take a deeper look: http://blog.sonatype.com/2008/04/how-to-share-resources-across-projects-in-maven/

@mrietveld would mind quickly reviewing this?

### DIFF
--- a/drools-beliefs/src/test/resources/logback-test.xml
+++ b/drools-beliefs/src/test/resources/logback-test.xml
@@ -3,9 +3,7 @@
 
   <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <!-- %l lowers performance -->
-      <!--<pattern>%d [%t] %-5p %l%n  %m%n</pattern>-->
-      <pattern>%d [%t] %-5p %m%n</pattern>
+      <pattern>%date{HH:mm:ss.SSS} [%thread] %-5level %class{36}.%method:%line - %msg%n</pattern>
     </encoder>
   </appender>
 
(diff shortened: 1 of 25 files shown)


mrietveld:
@psiroky This looks good. Thanks for doing this! +1

With regards to sharing the `logback-test.xml`, -1 because of the following: 

In the past, there were issues because the logback-test.xml was being included in, among others, the `jbpm-test` jar, which we contained some of our base test classes. The problem was that the included (`jbpm-test`) `logback-test.xml` would override the locally defined `logback-test.xml` file, making it impossible to exclude messages (hibernate and bitronix messages are usually way to verbose.. ). 

I agree that it may seem like a duplication, but when I'm testing and working on the code, it's _incredibly_ helpful to lower or raise the specific logging queries (to see, for example, transaction behavior, hibernate query trees or logger.debug messages in our own code).  For that reason, I'd rather keep individual `logback-test.xml` files per jar. 

If, you still want to per se add `logback-test.xml` files to `src/test/resource` directories where there is _not_ an existing `logback-test.xml` file (and then _only_ if there is no existing one), then that's fine. However, I'm against overriding existing `logback-test.xml` files. 

Does that make sense?

psiroky:
@mrietveld thanks for the additional info, I was not aware of all the different issues. I am not sure how Logback handles duplicated config these, but I agree that we could do this only in case it would allow overriding the logging config in child modules.

What is worse the fact it would then not be easily possible to tweak the config (or am I not aware of way to do so) for debugging purposes. This is likely blocker as do use this quite often as well.

I will leave this as is for now as I don't see an easy solution which would cover all the cases we need.

PR #775, opened 2016-05-04
[DROOLS-1156] use a ManagedExecutorService when running in a J2EE container

mariofusco:


mswiderski:
nice - didn't know they finally put it into the spec - though it's only JEE7 - which might be good for us on master

drools-core/src/main/java/org/drools/core/concurrent/ExecutorProviderImpl.java
```diff
@@ -35,23 +37,36 @@
...
+            try {
+                InitialContext ctx = new InitialContext();
+                newExecutor = (java.util.concurrent.ExecutorService) ctx.lookup("java:comp/env/concurrent/ThreadPool");
+            } catch (NamingException e) {
+                // not in a J2EE container -- ignore
+            }
+
+            if (newExecutor == null) {
+                String threadFactoryClass = System.getProperty( THREAD_FACTORY_PROPERTY );
+                ThreadFactory threadFactory;
```
psiroky, line 51: Just a thought: should we first try to check the system property value? It seems the purpose of it is to override the defaults, which would no longer be possible in some cases? (e.g. when the JEE ManagedExecutorService is found).

mariofusco:
@mswiderski yeah, I guess this will work only from JEE7, but I don't see any better alternative. Any suggestion?

drools-core/src/main/java/org/drools/core/concurrent/ExecutorProviderImpl.java
mariofusco, line 51: On JEE there's no need to check that System property because the ExecutorService will be provided by the container.

psiroky, line 51: Yeah, but what if I want to override the JEE container's default? I am not sure if this a valid use case though.

mariofusco, line 51: I don't think so: if you're in a JEE container you have to use the ExecutorService provided by the container itself otherwise you'll end up with those phantom threads reported in the jira.

mswiderski:
@mariofusco yeah, that's what I think so too. There is no easy way to deal with it. For the time being we relied on EJB @Asynchronous methods to use managed threads for async operations but that is rather heavy solution. So I am really looking forward for moving completely to JEE7 as this will simplify all our async stuff

etirelli:
I think it looks fine Mario. We will need to properly test this and if necessary, do any additional fixes. +1.

PR #779, opened 2016-05-10
[DROOLS-1171] fix native java compiler

mariofusco:


ge0ffrey:
+1 looks good

ge0ffrey:
I've pulled this locally and it fixes the problem with the native compiler and it makes jdk 9 work :)

PR #781, opened 2016-05-12
BZ-1321281: add test

winklerm:
Added test covering the fix for BZ 1321281. Please note that the test does not rely on non-case-preserving OS, it simulates the original issue by attempting to load a class which generates ExceptionInInitializerError (subclass of LinkageError).

Please do not merge yet, I would like to gather feedback first.

### DIFF
--- a/drools-test-coverage/pom.xml
+++ b/drools-test-coverage/pom.xml
@@ -19,6 +19,7 @@
 
   <modules>
     <module>standalone</module>
+    <module>test-suite</module>
   </modules>
 
 </project>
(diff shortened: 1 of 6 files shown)


winklerm:
@mariofusco, @psiroky When you have time, I would be happy to hear your comments.

Meanwhile, I am waiting for feedback from GSS.

winklerm:
Confirmed the test case with GSS.

I should probably squash the commits into one, I'll do it after review so there would be no further changes.

baldimir:
+1 for me. Looks good.

psiroky:
+1

PR #783, opened 2016-05-13
DROOLS-1176 Persistence tests doesn't use transactions for handling LOBs in Postgresql Plus

baldimir:
- Added com.edb driver handling to PersistenceUtil.useTransactions() method.


psiroky:
+1

PR #784, opened 2016-05-13
DROOLS-1176 Persistence tests doesn't use transactions for handling LOBs in Postgresql Plus

baldimir:
- Added com.edb driver handling to PersistenceUtil.useTransactions() method.

Master PR: https://github.com/droolsjbpm/drools/pull/783

This is needed for testing next planned 6.4.x builds on Postgresql Plus.


psiroky:
+1, only touches tests.

PR #786, opened 2016-05-16
Filter out Java source files from compilation in KieBuilder

mbarkley:
This PR is required for LiveSpark, where we require the ability to prevent client-side classes with GWT dependencies from being compiled.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java
@@ -3,7 +3,7 @@
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mbarkley:
Related PRs:
https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/145
https://github.com/droolsjbpm/kie-wb-common/pull/359

mswiderski:
+1 

@mariofusco could you please check this too?

mariofusco:
@mswiderski see my comment here https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/145/files#r63663654

mbarkley:
This PR has been updated based on discussion [here](https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/145).

mbarkley:
jenkins, retest this

mbarkley:
The build passes on my machine. The single failing test (CepEspTest#testEventWithShortExpiration) doesn't appear to be related to the changes in this PR, and it looks like the test depends on something happening during short call to `Thread.sleep`, so I suspect this might have to do with some environment-dependent behaviour.

@mariofusco @mswiderski Any ideas?

psiroky:
@mbarkley yeah, that is a flaky test. Failing "randomly". For sure not related with changes in this PR.

(we should somehow fix/remove that test...)

mariofusco:
@mbarkley +1 I'm about to merge this.
The only improvement I could think about is using that new Predicate to filter any other kind of resources and not only java sources, but this is something I can implement on my own.

PR #787, opened 2016-05-17
[DROOLS-1174] rehash instead of remove deleted RightTuples from the RightTupleMemory before processing right updates

mariofusco:


mariofusco:
@ge0ffrey Please test this fix with optaplanner and tell me if it is ok for you.

ge0ffrey:
This doesn't seem to fix it, I get:

java.lang.IllegalStateException: Score corruption: the workingScore (-3290) is not the uncorruptedScore (-3190) after completedAction (Lily {8.9} <-> Allison {0.3}):
  The corrupted scoreDirector has 1 ConstraintMatch(s) which are in excess (and should not be there):
    org.optaplanner.examples.dinnerparty.solver/twoSameJobTypePerTable/level0/[8, SOCIALITE]=-100
  The corrupted scoreDirector has no ConstraintMatch(s) which are missing.
  Check your score constraints.

```
at org.optaplanner.core.impl.score.director.AbstractScoreDirector.assertWorkingScoreFromScratch(AbstractScoreDirector.java:418)
```

mariofusco:
@ge0ffrey I updated my PR, please give it a second try

mariofusco:
ok to test

yurloc:
I confirm that the first commit only fixed the isolated reproducer but not the original dinner party test. The second commit also fixes the original reproducer.

I'm now running complete OptaPlanner test suite locally to verify no regression was introduced.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/abdf21f2d

PR #791, opened 2016-05-20
[DROOLS-1185] fix consequence jitting when LHS contains a OR

mariofusco:


mariofusco:
@etirelli This is a 6.4.x regression (it worked on 6.3.x) caused by RT/LT unification, so I think it could be good to also backport this fix to 6.4.x

mariofusco:
ok to test

etirelli:
+1

PR #792, opened 2016-05-20
Drools 1180 Migrate QE tests upstream.

baldimir:
This is second in a series of PRs that migrate some tests from QE regression test suite upstream (there will be more).
- Migrated test classes that inherit from EnginesTest class in QE regression test suite.


baldimir:
First PR (already merged) from the series is here if somebody is curious: https://github.com/droolsjbpm/drools/pull/788

PR #794, opened 2016-05-24
DROOLS-1189 - Parser error doesn't point at the problematic line in c…

tkobayas:
…ase of extra double quote
- Unit test


mariofusco:
won't fix

PR #796, opened 2016-05-25
DROOLS-1192 - DialectUtil.normalizeRuleName() doesn't normalize multi…

tkobayas:
…byte rule names

I18nTest.testMultibyteRuleName() : Failing test case described in the JIRA
I18nTest.testMultibyteRuleNameWithWhitespace() : This can be compiled because whitespace matches NON_ALPHA_REGEX.
DialectUtilTest.testNormalizeRuleName(): Low level test case


mariofusco:
Fixed by https://github.com/droolsjbpm/drools/commit/1f85d0324

PR #798, opened 2016-05-25
[DROOLS-1192] fix rule name normalization

mariofusco:


etirelli:
+1, just might be worth replacing [^\w] by \W

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/1f85d0324

PR #800, opened 2016-05-26
[DROOLS-1193][DROOLS-1194] fixes for GAE compatibility

mariofusco:


ge0ffrey:
+1

ge0ffrey:
This should be backported to 6.4.x. I think we can use this product jira for that:
  https://issues.jboss.org/browse/RHBRMS-1812

PR #801, opened 2016-05-31
[RHBRMS-2463] logical delete expired events

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -6033,4 +6033,106 @@ private void checkNoExpiration(String lhs) {
...
+        ksession.fireAllRules();
+        assertEquals(0, counter.get());
+
+        sessionClock.advanceTime(20, TimeUnit.MILLISECONDS);
+        ksession.insert( 1 );
+        ksession.fireAllRules(); // MyEvent is expired
+
+        assertEquals(0, counter.get());
+    }
 }
```
etirelli, line 6236: Mario, is it necessary to also test the behaviour when the rule uses not/exists/accumulate/forall CEs? Or do you think current coverage is enough?

drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
```diff
@@ -522,17 +528,13 @@ private void deleteStated( RuleImpl rule, Activation activation, InternalFactHan
             // If the equality key is now empty, then remove it, as it's no longer state either
             if ( key.isEmpty() && key.getLogicalFactHandle() == null ) {
                 tms.remove( key );
-            } else if ( key.getLogicalFactHandle() != null) {
+            } else if ( key.getLogicalFactHandle() != null && propagationContext != null) {
```
etirelli: Why is it necessary to check the propagationContext here?

etirelli:
Mario, looks good, just not sure if we need the additional tests to cover the cases with not/exists/accumulate/forall.
+1 to merge.

mariofusco:
@mdproctor @etirelli I found that checking for Tuple/FactHandle expiration when evaluating constraints is useless because I'm removing the RightTuples when the event is expired. Additionally I'm now retracting the FactHandle when the handle is expired and has no further pending activations https://github.com/droolsjbpm/drools/pull/801/commits/7b16ef233dacef1a031aaa54f5f622ea62f5e368#diff-5ef9b9f27ecdca2968a52455f8557919R1124

This change also fixes the original test case reported by https://issues.jboss.org/browse/DROOLS-311 so I also unignored it https://github.com/droolsjbpm/drools/pull/801/commits/7b16ef233dacef1a031aaa54f5f622ea62f5e368#diff-b3ff35a827e4a0ba8b48149df4892eaaL2963 I think this covers the not/exists scenario as asked by Edson. Also the accumulate is already covered by other tests in CepEspTest.

I added another test trying to break this fix as suggested by Mark but I couldn't find any use case that doesn't work with this. Please let me know if I'm missing something here or if you think that there is some other test case that could potentially break what I've done.

mariofusco:
@mdproctor @etirelli I think now I'm better understanding what's happening with that last test case: the retraction of the RightTuple on event expiration is unlinking the segment/path and that's why the rule is not triggering as expected. That's an interesting side-effect though. How can I let the expiration to retract the RT with unlinking the path?

mariofusco:
@etirelli @mdproctor I changed that last test case according to Mark's suggestion. Now there's a non-expired event so the path doesn't unlink, but still it is firing only for that event and not for the expired one, so the test is still not failing.

etirelli:
Mario, looks good! +1 to merge.

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/454dfeb1d

PR #804, opened 2016-06-03
[DROOLS-1155] fix eval expressions wiring during deserialization of rules with ORs

mariofusco:


drools-core/src/main/java/org/drools/core/base/mvel/MVELEvalExpression.java
```diff
@@ -125,7 +125,18 @@ public MVELEvalExpression clone() {
...
         return clone;
     }
-    
+
+    public MVELEvalExpression clonePreservingDeclarations(MVELEvalExpression original) {
+        MVELCompilationUnit cloneUnit = unit.clone();
+        cloneUnit.setPreviousDeclarations( original.unit.getPreviousDeclarations() );
+        MVELEvalExpression clone = new MVELEvalExpression( cloneUnit,
+                                                           id );
+        // expr should be stateless, so it should be fine to share the reference
```
etirelli, line 134: Is expr thread safe? with the move to support concurrency, is this a concern?

etirelli:
+1

drools-core/src/main/java/org/drools/core/base/mvel/MVELEvalExpression.java
mariofusco, line 134: I think that what is not thread safe is the MVELCompilationUnit that is also used by the MVELEvalExpression other than mvel based constraints evaluation. I tried to convince Mark to get rid of it, but with no luck so far :)

PR #806, opened 2016-06-08
[DROOLS-1200] Multibyte bind variable name fails with java dialect

tkobayas:
- Unit test

### DIFF
--- a/drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/functional/UnicodeTest.java
+++ b/drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/functional/UnicodeTest.java
@@ -26,12 +26,14 @@ import org.kie.api.KieServices;
 import org.kie.api.command.Command;
 import org.kie.api.definition.type.FactType;
 import org.kie.api.io.Resource;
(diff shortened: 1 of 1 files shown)


baldimir:
Hi, thanks for the test. I tried it and it fails also for me. 

One question though. Would it be possible to move it somewhere else please? The Misc2Test class is pretty huge now. We have a new class for unicode tests here [1]. Maybe that can be a good fit for this test. What do you think? 

Thanks, 

Tibor

[1] https://github.com/droolsjbpm/drools/blob/master/drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/functional/UnicodeTest.java

tkobayas:
Hi Tibor,

I wasn't aware of the "UnicodeTest", thanks! I changed the commit to put it in UnicodeTest.

baldimir:
Hi Toshiya, great, thanks! :)

baldimir:
+1

mariofusco:
Test case incorporated and fixed by https://github.com/droolsjbpm/drools/commit/9781eb38089de0c5a1d644a36edef732be7e0e59

PR #808, opened 2016-06-13
[DROOLS-1186] rehash deleted RightTuples in correct order while processing right updates

mariofusco:


ge0ffrey:
+1 This fixes the incarnation 3 (and therefore all known incarations of this bug). Can you also backport this?

PR #816, opened 2016-06-23
drools-pmml: explicitly depend on EE deps to make it work on JDK9

psiroky:
- works both on JDK8 and JDK9
- the drools-pmml needs to transitively bring the jaxb
  and javax.activation deps, otherwise projects depending
  on drools-pmml (e.g. drools-scorecards) still won't work


mariofusco:
+1

PR #817, opened 2016-06-24
[DROOLS-1212] use try-with-resources to properly close streams

psiroky:
@mariofusco could you please take a look?


mariofusco:
@psiroky that's a +1 for me
My only doubt is that we could probably need to backport these fixes to 6.x branches and in that case you'll have to reimplement it because we cannot use a Java 7 feature there.

psiroky:
@mariofusco in case we would need to backport it, I would replace it with the horribly looking `finally` block :). On master we should keep using it as it is much less boiler plate.

triceo:
+1 on backporting, but make sure that there is a product JIRA tracking this.

PR #818, opened 2016-06-24
[DROOLS-1219] fix ClassObjectTypes rewiring during packages merge

mariofusco:


etirelli:
Nice catch. Good job! +1 to merge.

PR #819, opened 2016-06-25
[DROOLS-1213] restore system properties after test finishes

psiroky:


baldimir:
+1 looks good if all checks pass.

psiroky:
It may seem like there are now 2 checks ("Jenkins CI" and "Linux"), but in fact there is just one. I changed the name during the PR build, so Jenkins/GitHub got  a bit confused.

PR #820, opened 2016-06-27
add basic test coverage for column diffs

zkrejcov:
... and also improve a bit the BRL column diffs (the diffs themselves are currently not really used AFAIK).

(part n+1 of moving-tests-to-community series)


manstis:
Thanks @zkrejcov 

+1 to merge.

PR #823, opened 2016-06-27
GUVNOR-2580: Guided rule editor : Insert formulas in an OR/AND constraint

manstis:
See https://issues.jboss.org/browse/GUVNOR-2580


manstis:
See https://github.com/droolsjbpm/drools-wb/pull/220 too.

mariofusco:
+1

PR #829, opened 2016-07-08
DROOLS-1199 Memory leak in KieScanner

baldimir:
- Added test case for KieScanner memory leak issue.

### DIFF
--- a/drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/memory/KieScannerMemoryTest.java
+++ b/drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/memory/KieScannerMemoryTest.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.*;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import org.drools.compiler.TurtleTestCategory;
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
 import org.drools.core.util.FileManager;
 import org.drools.testcoverage.common.util.KieBaseTestConfiguration;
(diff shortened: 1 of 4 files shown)


drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/memory/KieScannerMemoryTest.java
```diff
@@ -0,0 +1,137 @@
...
+                logger.debug("Average memory: " + averageMemory);
+                averageMemoryFootprints.add(averageMemory);
+                averageMemory = 0;
+            }
+        }
+    }
+
+    private void waitForMillis(final long millis, final long startTimeMillis) {
+        while ((System.currentTimeMillis() - startTimeMillis) < millis) {
+            // do nothing - wait
```
winklerm, line 127: At first, this looked suspicious to me (will eat up one CPU core) but I can see you are waiting for 30 milliseconds at most, so I think it is fine in this case.

baldimir, line 127: Yes it was intentional. I sacrificed CPU for accuracy in this case.

drools-test-coverage/test-suite/src/test/java/org/drools/testcoverage/memory/KieScannerMemoryTest.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    private String getMemoryMeasurmentsString(final List<Long> memoryMeasurements) {
```
winklerm: There's a typo in method name - "Measurments".

baldimir: Fixed.

winklerm:
Looks good to me, +1 for merging.

PR #831, opened 2016-07-11
GUVNOR-2602: Guided Rule Editor: Facts created in RHS used in RHS bindings are not retained on re-opening rule

manstis:
See https://issues.jboss.org/browse/GUVNOR-2602

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
@@ -319,7 +319,8 @@ class RuleModelPersistenceHelper {
         }
         for ( ModelField modelField : modelFields ) {
             if ( modelField.getName().equals( field ) ) {
-                return modelField.getType();
+                return getSimpleFactType( modelField.getType(),
+                                          dmo );
             }
         }
         return null;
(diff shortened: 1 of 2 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
```diff
@@ -319,7 +319,8 @@ static String inferDataType( final ActionFieldList action,
         }
         for ( ModelField modelField : modelFields ) {
             if ( modelField.getName().equals( field ) ) {
-                return modelField.getType();
+                return getSimpleFactType( modelField.getType(),
```
manstis, line 322: This is the fix.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceUnmarshallingTest.java
```diff
@@ -8534,4 +8535,83 @@ public void compositeFieldConstraintWithOnePredicateAndOneLiteral() {
                       fp_cfp_sfp2.getConstraintValueType() );
     }
 
+    @Test
```
manstis, line 8538: This is the test.

manstis:
I've highlighted the fix and the test.. the rest is just tidying up existing tests (as IDEA lit up like a 1KW lightbulb with warnings in the test class).

mrietveld:
+1 when the tests complete.

PR #833, opened 2016-07-13
DROOLS-1198 NoClassDefFoundError when using str[endsWith] on a field...

tarilabs:
Fixing ClassTypeResolver issues when running on Mac/OSX.

(I guess I will review this with @mariofusco, but anyway will add details below in the PR, also for reference when discussing later with anybody this PR.)

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
+++ b/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
@@ -302,6 +302,8 @@ public class ClassTypeResolver
                 clazz = this.classLoader.loadClass( qualifiedClass );
             } catch ( final ClassNotFoundException e ) {
                 clazz = null;
+            } catch ( final NoClassDefFoundError e ) {	// potential mis-match induced by Mac/OSX
+                clazz = null;
             }
 
             // maybe its a nested class?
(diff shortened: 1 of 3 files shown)


tarilabs:
I believe this is actually an issue in drools-core when ClassTypeResolver is running on Mac/OSX, and in turn when the drools-compiler is invoking it via org.drools.compiler.rule.builder.PatternBuilder.createDeclarationForOperator(PatternBuilder.java:1593) during compilation.

Now, if I understand correctly:
1. the ClassTypeResolver's resolveType will throw (manually) a ClassNotFoundException if the name parameter does not correspond to any resolvable type.
2. the ClassTypeResolver's chain of: resolveType -> getClassFromImports -> importClass, relies heavily on null returns signifying problem.

However attached reproducer, demonstrate on Mac/OSX if a by any chance the name parameter passed to ClassTypeResolver's resolveType is case-INsensitive-equal to some existing class, everything break with a NoClassDefFoundError for a wrong name.

See:
screen shot 2016-07-13 at 09 00 19
screen shot 2016-07-13 at 11 06 31

Please note I've taken the screenshot on Mac/OSX before working on this proposed fix.

tarilabs:
Then, please notice indeed those test would pass on a non-Mac/OSX platform, say Linux:
screenshot from 2016-07-13 09-29-49

Then, I've worked on a simulator (`SimulateMacOSXClassLoader`) to achieve _analogous_ behavior of the Mac/OSX problem, on the non-Mac/OSX platform:
screenshot from 2016-07-13 11-10-29
screenshot from 2016-07-13 11-10-34

Then, code changes of `ClassTypeResolver` is my proposal to solve for the issue.

tarilabs:
Discussed and To be closed in favour of #838 .

PR #834, opened 2016-07-13
Test for DROOLS-355

mrietveld:
This is an example test for the DROOLS-355: by having it in master _before the fix/PR_ is applied, the merger can see what has changed. 

Thanks,
Marco

@mswiderski or @psiroky, would you mind reviewing and/or merging? Thanks.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/runtime/pipeline/impl/DroolsJaxbHelperTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/runtime/pipeline/impl/DroolsJaxbHelperTest.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


drools-compiler/src/test/java/org/drools/compiler/runtime/pipeline/impl/DroolsJaxbHelperTest.java
```diff
@@ -0,0 +1,110 @@
...
+        Class xsdClass = xsdObj.getClass();
+        try {
+            Method m2 = xsdClass.getMethod( "getContent" );
+            assertNotNull( m2 );
+            assertEquals( List.class, m2.getReturnType() );
+
+            assertEquals( 0, xsdClass.getFields().length );
+            java.lang.reflect.Field[] declaredFields = xsdClass.getDeclaredFields();
+            assertEquals( 1, declaredFields.length );
+            assertEquals( "content", declaredFields[0].getName() );
```
mswiderski, line 151: is the content default name for elements? I can't find it in the xsd...

mswiderski:
@mrietveld not sure I am the best one to review it as I don't know much about this feature (psd with rules). Anyway just added one question out of curiosity as this really looks interesting :)

drools-compiler/src/test/java/org/drools/compiler/runtime/pipeline/impl/DroolsJaxbHelperTest.java
mrietveld, line 151: Yes, it is. (FYI: I didn't know it either, this was also what I discovered.. ). 

However, I'll expand the tests with some more XSD generated classes. Thanks.

PR #835, opened 2016-07-13
Container is not upgraded until restarted - classloader caches kjar c…

mswiderski:
…lasses upon updateToVersion - test case

this shows the issue reported as https://issues.jboss.org/browse/RHBPMS-4048

I tried to find a solution but to no avail, so I submit just the test case to ease the investigation.
@mariofusco once you're back could you please take a look?


mariofusco:
Incorporated by https://github.com/droolsjbpm/drools/pull/843

PR #837, opened 2016-07-15
Don't report disabled tests as skipped

psiroky:
- reporting the thousands of generated tests
  as skipped skewes the overall "skipped tests"
  statistic which is bad (before the change 25k
  tests were reported as skipped)

@mariofusco, @baldimir please take a look.


psiroky:
After discussion with @baldimir, I'll rework this to support the `-DrunTurtleTests` property.

psiroky:
@baldimir please take a look. Now, by default, the turtle tests are just ignored and can be enabled using `-DrunTurtleTests` property (same as before, but using JUnit's category instead of Assume)

baldimir:
+1 Looks good. Thanks @psiroky.

PR #842, opened 2016-07-19
test case for checking classes loaded from kcontainer ClassLoader

mariofusco:


mariofusco:
This test case will be incorporated by https://github.com/droolsjbpm/drools/pull/835

PR #844, opened 2016-07-20
Don't report turtle tests as skipped

psiroky:
- reporting the thousands of generated tests
  as skipped skews the overall "skipped tests"
  statistics which is very bad
- before the change 25k tests were reported as skipped
- turtle tests can be executed using specific property, e.g.
  `mvn clean test -DrunTurtleTests`

Backport of #837.


baldimir:
+1

PR #852, opened 2016-07-28
[DROOLS-1240] allow to define an alternative JNDI name for the ManagedExecutorService

mariofusco:


drools-core/src/main/java/org/drools/core/concurrent/ExecutorProviderImpl.java
```diff
@@ -39,11 +42,16 @@
         static {
             java.util.concurrent.ExecutorService newExecutor = null;
 
+            String executorServiceName = System.getProperty( EXECUTOR_SERVICE_PROPERTY );
```
psiroky, line 45: This is a trivial thing -- but you can pass the default value as a second parameter of the `getProperty()` method. It will be used in case the property value is null.

tarilabs, line 45: @psiroky yes you are correct, but I believe in this case using that strategy, would then require a System.getProperty call again later line 52 below, in order to distinguish if the original property was set by the user or not - so that if the user deliberately set the `EXECUTOR_SERVICE_PROPERTY` and this is not available on JNDI, fail with RuntimeException. Otherwise if the property `EXECUTOR_SERVICE_PROPERTY` was not set by the user, then we know we actually tried to understand ourselves if we're in the Java EE container, and the previous fail on JNDI means we're not on the JNDI container. Please correct me if I'm wrong :)

psiroky, line 45: Yeah, you are right @tarilabs. I was reading the code too fast.

tarilabs, line 45: np always better to double-check :+1:

tarilabs:
+1 for merge, this improvement semantically performs a "convention over configuration" and should make Java EE devs more happy =) thanks

PR #853, opened 2016-07-28
[DROOLS-1241] fix recognition of type as event during incremental compilation

mariofusco:


mariofusco:
ok to test

etirelli:
Hi Mario, the code looks good to merge, +1. 

Question somewhat unrelated: what happens if we declare or annotate a super-class as an Event? I think the current code does not inherit the role, does it? Shouldn't it?

mariofusco:
@etirelli No, declaring a type as an event doesn't make its subtypes to be events too. Also the @Role annotation ( https://github.com/droolsjbpm/droolsjbpm-knowledge/blob/master/kie-api/src/main/java/org/kie/api/definition/type/Role.java ) is not flagged with @Inherited and then it is not propagated to subclasses. I don't know (and don't have a strong opinion) if this is correct and I'm open to discuss it.

PR #854, opened 2016-07-29
DROOLS-1243 ClassCastException at runtime when trying to match "from"...

tarilabs:
... and a built-in accumulate function.

Submitting reproducer for the problem.


mariofusco:
Test case modified and incorporated by https://github.com/droolsjbpm/drools/pull/856
We decided that pattern that cannot be matched due a type incompatibility shouldn't compile at all.

PR #856, opened 2016-08-01
[DROOLS-1243] prevent compilation of from patterns that cannot be matched due to type incompatibility

mariofusco:


tarilabs:
@mariofusco I've played with the branch locally by fiddling with the test case, for me is okay.

Just a minor correction, I've performed the necessary change to all the built-in accumulate functions, otherwise a rule like:

```
  $theFrom : Double() from accumulate(MyPerson( $val : age );
                                           min( $val ) )     
```

would fail with:

```
Pattern of type: '[ClassObjectType class=java.lang.Double]' on rule 'R' is not compatible with type java.lang.Number returned by accumulate function. : [Accumulate: input=[AND [[Pattern: id=null; objectType=MyPerson]] ]]
```

for a false negative.

I've raised PR against your fork, so it should be easy to merge, and update on this drools upstream repo original PR.

Kindly let me know, thank you

tarilabs:
I believe a false positive because the failing test:

```
org.drools.compiler.integrationtests.CepEspTest.testEventWithShortExpiration

expected:<0> but was:<1>
Stacktrace

java.lang.AssertionError: expected:<0> but was:<1>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:645)
    at org.junit.Assert.assertEquals(Assert.java:631)
    at org.drools.compiler.integrationtests.CepEspTest.testEventWithShortExpiration(CepEspTest.java:5609)
```

It is leveraging a Thread.sleep:

```
        ksession.insert( "test" );
        assertEquals( 1, ksession.fireAllRules() );

        Thread.sleep(2L);
        ksession.fireAllRules();
```

instead of a pseudoclock-based test; I will propose a better test later, but I would like to try for now bot interaction on github...

ok to test

Jenkins test again

tarilabs:
+1 super, it was indeed a false positive :8ball:

psiroky:
@tarilabs you are correct, that test has been "randomly" failing for quite some time now. If you make it reliable I believe lots of people will be thanking you :) (including me !)

mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/29a1a5015

PR #857, opened 2016-08-03
Improve Adventures game

mdproctor:
-Improve the invaders game, to demonstrate transitive closures via the "search" command.

-Update the game to lock the door and place the key in an envelop, which is in a draw, which is in desk.


salaboy:
+1

PR #858, opened 2016-08-03
fix tests in kie-ci to make them work when using custom maven settings

mariofusco:


psiroky:
+1 (assuming the build succeeds)

PR #860, opened 2016-08-03
[DROOLS-414] allow any kind of expression to be used as input for a query

mariofusco:


mariofusco:
Merged by https://github.com/droolsjbpm/drools/commit/987d68ab3

PR #861, opened 2016-08-04
[DROOLS-1247] prevent recursive segments creation during eager evaluation

mariofusco:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -8809,4 +8809,53 @@ public String toString() {
             return "Shift " + employee + " from " + start + " to " + end;
         }
     }
+
+    @Test(timeout = 10000L)
+    public void testSegMemInitializationWithForceEagerActivation() throws InterruptedException {
```
baldimir, line 8814: Hi Mario, would it be possible to put this test into some other class? Maybe some new class named Segment(Memory)InitializationTest, SegmentMemoryTest or I found that there is SegmentCreationTest class. I think we should stop adding new tests into Misc*Test classes. They are already pretty huge. What do you think?

mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/874

PR #862, opened 2016-08-05
RHBPMS-4171 - Exception in JBPM Executor Command leads to loss of ori…

mswiderski:
…ginal error

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
@@ -389,21 +389,21 @@ public class SingleSessionCommandService
         rollbackTransaction(t1, transactionOwner, true);
     }
 
-    private void rollbackTransaction(Exception t1,
-            boolean transactionOwner, boolean logstack) {
+    private void rollbackTransaction(Exception cause, boolean transactionOwner, boolean logstack) {
         try {
 
             if (logstack) {
-                logger.warn( "Could not commit session", t1 );
+                logger.warn( "Could not commit session", cause );
             } else {
-                logger.warn( "Could not commit session due to {}", t1.getMessage() );
+                logger.warn( "Could not commit session due to {}", cause.getMessage() );
             }
             txm.rollback( transactionOwner );
-        } catch ( Exception t2 ) {
-            logger.error( "Could not rollback",
-                    t2 );
-            throw new RuntimeException( "Could not commit session or rollback",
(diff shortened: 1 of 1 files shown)


drools-persistence-jpa/src/main/java/org/drools/persistence/SingleSessionCommandService.java
```diff
@@ -400,10 +400,10 @@ private void rollbackTransaction(Exception t1,
...
             txm.rollback( transactionOwner );
         } catch ( Exception t2 ) {
-            logger.error( "Could not rollback",
-                    t2 );
-            throw new RuntimeException( "Could not commit session or rollback",
-                    t2 );
+            // log rollback exception
+            logger.error( "Could not rollback", t2 );
+            // propagate original exception that caused the rollback
+            throw new RuntimeException( "Could not commit session or rollback", t1 );
```
psiroky: Not directly related to this PR, but the exception variable names (t1, t2) could be more descriptive :)

Also it seems that at this point we know for sure that both "could not commit session" _and_ "rollback failure" happened. Perhaps we could improve the message to clarify that? Or did I misunderstood this?

It would also be nice to explicitly say that the "caused by" exception is the original one (and not the rollback one), as that may not be 100% clear.

mswiderski: that's the whole point of this PR to throw/propagate the original exception that caused the issue and only log the one that caused rollback to fail. In both ways the immediate consequence is commit or rollback failed.

psiroky: I guess I am not 100 % understanding the code, so will take a deeper look.

What I meant with the message is that "Could not commit session or rollback" does not really say what kind exception is attached, so was thinking about changing that to:

"Could not commit session or rollback. Original exception attached as 'caused by'. See log for full rollback exception."...WDYT?

mariofusco: The message "Could not commit session or rollback" is a bit confusing also for me ( even if I am the last person in the team who should be allowed of speaking about "clear error messages" :) ). Maybe "Could neither commit nor rollback session" is clearer? That said I agree that it's better to propagate the original exception t1 there and just log t2.

mswiderski:
@psiroky @mariofusco updated message and variable names, how does it look now?

psiroky:
Looks good! Thanks @mswiderski

+1 from me

PR #864, opened 2016-08-05
[6.4.x] RHBPMS-4171 - Exception in JBPM Executor Command leads to loss of ori…

mswiderski:
…ginal error


mswiderski:
this one (6.4.x) is for one of patch requested by GSS

PR #865, opened 2016-08-05
Not closed FileInputStream Object caused "To many open files" IOException

sisivy:
In our case, this part caused endless loop(blocking system resources) in the jenkins build process(tests execution) and was returning incorrect error message. The getBytes method was not closing FileInputStream at the end of reading files and caused IOException "To many open files".


mariofusco:
Your commit is correct, thanks a lot for it. Please just fix code identation before I merge it.

triceo:
@mariofusco  When you're merging, can you please merge to 6.5.x as well?

sisivy:
We are using 6.3.0-Final. It would be nice if you can merge into 6.3.x branch as well.

PR #866, opened 2016-08-08
DROOLS-1235 MBean naming rework

tarilabs:
Depends on droolsjbpm/droolsjbpm-knowledge#150
- give KieContainer an optional user defined name
- KieContainer MBean and various refactorings.
- jargon: containerId 
- jargon: configuredReleaseId, resolvedReleaseId
- unit testing
- assuming the containerId can be "recycled" on dispose.
- introducing test for KieServices API with containerId.
- fixing other tests.
- avoid using lock and leverage ConcurrentMap without java8
- fixing objectname conventions.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
@@ -512,7 +512,7 @@ public class KieContainerImpl
         } else if (conf instanceof RuleBaseConfiguration) {
             ((RuleBaseConfiguration)conf).setClassLoader(cl);
         }
-        InternalKnowledgeBase kBase = (InternalKnowledgeBase) KnowledgeBaseFactory.newKnowledgeBase( conf );
+        InternalKnowledgeBase kBase = (InternalKnowledgeBase) KnowledgeBaseFactory.newKnowledgeBase( kBaseModel.getName(), conf );
 
         kBase.addKnowledgePackages( pkgs );
         return kBase;
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
@@ -15,12 +15,32 @@
 
 package org.drools.compiler.kie.builder.impl;
 
+import static org.drools.compiler.kie.builder.impl.AbstractKieModule.buildKnowledgePackages;
+import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.filterFileInKBase;
+import static org.drools.compiler.kie.util.CDIHelper.wireListnersAndWIHs;
+import static org.drools.core.util.ClassUtils.convertResourceToClassName;
+import static org.drools.core.util.Drools.isJndiAvailable;
+
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.management.ObjectName;
+
 import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
 import org.drools.compiler.compiler.PackageBuilderErrors;
 import org.drools.compiler.kie.util.ChangeSetBuilder;
 import org.drools.compiler.kie.util.KieJarChangeSet;
 import org.drools.compiler.kproject.models.KieBaseModelImpl;
 import org.drools.compiler.kproject.models.KieSessionModelImpl;
+import org.drools.compiler.management.KieContainerMonitor;
 import org.drools.core.RuleBaseConfiguration;
 import org.drools.core.common.InternalWorkingMemory;
 import org.drools.core.common.ProjectClassLoader;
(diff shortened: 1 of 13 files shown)


tarilabs:
failed on `JpaOptLockPersistentStatefulSessionTest.testOptimisticLockInterceptor`. False positive?

Jenkins, retest this :8ball:

tarilabs:
...and of course it's now failing on the `CepJavaTypeTest.testEventWithShortExpiration` another false positive.

Jenkins, retest this :8ball: :8ball:

etirelli:
Looks good! +1 to merge!

tarilabs:
rebased on master to include droolsjbpm/drools@7b09586 commit, in order for the other PR droolsjbpm/kie-docs#130 on the kie-docs to build succesfully - hopefully
(thanks @manstis for pointing it out)

tarilabs:
my new friend, `CepEspTest.testEventWithShortExpiration`

Jenkins, retest this :8ball:

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -104,14 +105,46 @@
...
+    
+    /**
+     * Please note: the recommended way of getting a KieContainer is relying on {@link org.kie.api.KieServices KieServices} API,
+     * for example: {@link org.kie.api.KieServices#newKieContainer(ReleaseId) KieServices.newKieContainer(...)}.
+     * The direct manual call to KieContainerImpl constructor instead would not guarantee the consistency of the supplied containerId.
+     * @deprecated Use {@link #KieContainerImpl(String, KieProject, KieRepository, ReleaseId)} to assign the container Id to the KieContainer.
+     */    
+    @Deprecated
+    public KieContainerImpl(KieProject kProject, KieRepository kr, ReleaseId containerReleaseId) {
+        this("impl"+UUID.randomUUID(), kProject, kr, containerReleaseId);
```
tarilabs, line 124: @etirelli for completeness, I believe this approach of doing internally is better, as doesn't require modification to kie-wb-common nor to optaplanner, etc. Moreover it clearly mark the constructor as deprecated. Additionally I've added explicit javadoc note to all KieContainerImpl constructors that is a direct internal call and therefore shall be used carefully. Hope is okay with you.

etirelli, line 124: @tarilabs I agree that assigning the generated ID inside the constructor is a better solution, I am just not sure if marking the constructor as "@Deprecated" is correct. Since this is am implementation class constructor, I think it is pretty obvious they should not be used by "client" code. A better alternative than using @Deprecated might be to simply provide a factory method and the client code can then use the factory?

tarilabs, line 124: I see your point @etirelli . However to my understanding the factory method you describe exists already, the `KieServices.newKieContainer`. Here is my assumption: the other projects which used this class directly,  found the KieServices not covering the use-case they really needed. So at the end I see your point, but that would translate imho in a separate improvement JIRA for the KieServices. What do you think?

etirelli, line 124: I agree, but for now, just remove the @Dreprecated annotation, as these constructors are not really deprecated.

tarilabs, line 124: okay, done :+1:

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -91,24 +97,82 @@
 
     private final KieRepository        kr;
 
+    private ReleaseId configuredReleaseId;
```
mariofusco: @tarilabs Shouldn't this be final?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -91,24 +97,82 @@
...
     private ReleaseId containerReleaseId;
 
+	private String containerId;
```
mariofusco: @tarilabs Shouldn't this be final?

drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java
```diff
@@ -224,6 +226,14 @@ public KnowledgeBaseImpl(final String id,
         }
     }
 
+    @Override
+	public void initMBeans() {
+		if (config != null && config.isMBeansEnabled() && !mbeanRegistered) {
```
mariofusco: @tarilabs Why do you need this `mbeanRegistered` guard here? Isn't this method only called when the KnowledgeBase is created by the KieContainer? In case you really need it, are you sure it shouldn't be thread-safe (and then you should have an AtomicBoolean instead of this boolean field).

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
tarilabs: indeed I missed it, thanks.

drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java
tarilabs: Yes being an internal api it is intended to be used only when called from KieContainer when creating the KnowledgeBase. Anyhow given we stumble upon use-cases where internal api are used anyway, I preferred to be guarded.

About thread safety, frankly, I didn't see too much the risk (external usage of internal api + multiple threads initializing a single KB, I believe would have been extremely unlikely).

However I see your point and for consistency updated following your suggestion of AtomicBoolean. Thanks. :+1:

tarilabs:
(added MBean interface in public Kie API on already existing package dedicated for this purpose. As per today's review)

etirelli:
+1 to merge

tarilabs:
(renamed `ReleaseIdInfo` to `GAV` as discussed)

tarilabs:
[ `CepEspTest.testEventWithShortExpiration` ]

Jenkins, test again :8ball:

tarilabs:
[ `CepEspTest.testEventWithShortExpiration` ]

Jenkins, retest this :8ball:

PR #867, opened 2016-08-08
[DROOLS-1175][DROOLS-1242] infer numeric type for sum expression in an accumulate pattern + make accumulate functions null safe

mariofusco:


ge0ffrey:
+1 to merge. Running benchmarks now to get an idea how much perf this gives us.

PR #873, opened 2016-08-09
GUVNOR-2503: Asset screen Validate-button should use incremental builder

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2503 and https://github.com/droolsjbpm/kie-wb-common/pull/421

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieFileSystemImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieFileSystemImpl.java
@@ -24,9 +24,15 @@ import org.kie.api.io.ResourceType;
 import org.kie.internal.io.ResourceTypeImpl;
 import org.kie.api.io.Resource;
 import org.kie.api.io.ResourceConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
 import java.util.Properties;
 
 import static org.drools.core.util.IoUtils.readBytesFromInputStream;
@@ -36,8 +42,11 @@ import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.RESOURCES_ROOT
 
 public class KieFileSystemImpl
         implements
-        KieFileSystem {
-    
+        KieFileSystem,
+        Serializable {
+
+    private static Logger logger = LoggerFactory.getLogger( KieFileSystemImpl.class );
+
     private final MemoryFileSystem mfs;
 
     public KieFileSystemImpl() {
@@ -89,7 +98,7 @@ public class KieFileSystemImpl
             throw new RuntimeException("Unable to write Resource: " + resource.toString(), e);
         }
     }
-    
+
     public void delete(String... paths) {
         for ( String path : paths ) {
             mfs.remove(path);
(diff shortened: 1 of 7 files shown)


karreiro:
Jenkins please retest this.

karreiro:
Jenkins, please retest this.

manstis:
@etirelli @mariofusco Could you please review this PR? The issue is that when validating a _single_ file in the Workbench (i.e. the User edits a DRL file and, before saving, validates it compiles OK etc) we used to create a new `KieBuilder` add the file being validated and compile. However since a DRL file can reference queries, functions, globals, Java classes etc in _other_ files we had to add (in some cases) ALL files to the temporary `KieBuilder` and compile it; then extract any compilation errors for the single file being validated. This could be slow for large projects. So.. to fix it we're now cloning the existing `KieFileSystem` and reusing the clone in conjunction with an incremental compilation to just compile and validate the single file.

manstis:
There are five PRs all together:-
- https://github.com/droolsjbpm/drools/pull/873
- https://github.com/droolsjbpm/guvnor/pull/346
- https://github.com/droolsjbpm/kie-wb-common/pull/421
- https://github.com/droolsjbpm/drools-wb/pull/265
- https://github.com/droolsjbpm/optaplanner-wb/pull/89

etirelli:
@manstis I will let @mariofusco review this one as he is more familiar with this code than I am. The approach sounds fine though until we get an asynchronous, standalone compilation and validation service.

manstis:
@etirelli ok. @porcelli Let me know when we have the asynchronous, standalone compilation and validation service available.. it sounds terrific! ;-)

etirelli:
LOL, @porcelli promised it by EOB today. Watch for it. :)

porcelli:
@etirelli if you volunteer to babysit son for a couple of days.. it may get done :)

mariofusco:
@manstis @Rikkola @karreiro Just to be clear this approach works only if you have to validate one (or more) brand new files. If you're using it only in this scenario this is a +1 for me. Conversely if you're also changing one of the already existing resources this no longer works because you're not removing from the builder the artifacts created by the old resource before the change (as done by incremental compilation),

karreiro:
Thanks for the review @mariofusco

Unfortunately, I'm not sure if I'm missing something.

Here the steps for validating a resource:
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/Validator.java#L46
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorBuildService.java#L48
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/validation/asset/ValidatorBuildService.java#L64-L68
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/BuildServiceImpl.java#L336-L339
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java#L412-L412
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java#L383
- https://github.com/karreiro/kie-wb-common/blob/d9cc28783d9e408527916f731fc0f7fc0076de1b/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/Builder.java#L539-L539
- https://github.com/karreiro/drools/blob/incremental-validator/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderSetImpl.java#L92
- https://github.com/karreiro/drools/blob/incremental-validator/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderSetImpl.java#L133

Probably I've misunderstood, but I think that the artifacts created by the old resource before the change are being cleaned, right?

Thanks, again :-)

etirelli:
Discussed this with @karreiro and the scenario described by @mariofusco seems to be working. Lets go ahead and merge this PR. Next week when Mario is back we can discuss if any additional fix is necessary. +1 to merge.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieFileSystemImpl.java
```diff
@@ -128,5 +133,42 @@ public KieFileSystem writeKModuleXML(String content) {
...
+    }
+
+    public KieFileSystem clone() {
+        try {
+            final ByteArrayOutputStream byteArray = writeToByteArray( this );
+            final KieFileSystem kieFileSystem = readFromByteArray( byteArray );
+
+            return kieFileSystem;
+        } catch ( IOException e ) {
+            e.printStackTrace();
```
psiroky: We should log the exception instead of using `printStackTrace()` (similar case also few lines down).

karreiro: Thanks @psiroky

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieFileSystemImpl.java
```diff
@@ -15,118 +15,162 @@
...
+        try {
+            final ByteArrayOutputStream byteArray = writeToByteArray( this );
+            final KieFileSystem kieFileSystem = readFromByteArray( byteArray );
+
+            return kieFileSystem;
+        } catch ( IOException ioe ) {
+            logger.error( "Error cloning KieFileSystemImpl: " + this.toString(), ioe );
+            return null;
+        } catch ( ClassNotFoundException cnfe ) {
+            logger.error( "Error cloning KieFileSystemImpl: " + this.toString(), cnfe );
```
psiroky: Sorry for being nosy :) -- should we fail fast in this case? Logging the exception as an error suggests the state is not recoverable. But the method actually returns "normally".

karreiro: You're totally right @psiroky ... I'm gonna use `logger.warn` here. I had the same feeling about the `looger.error` right after I pushed.

karreiro:
Jenkins, please retest this.

csadilek:
jenkins retest this.

etirelli:
@csadilek as we discussed, the tests are failing before applying the PR, so the failures are not related.

PR #877, opened 2016-08-23
DROOLS-1236 MBean aggregation of statistics mechanism

tarilabs:
Requires droolsjbpm/droolsjbpm-knowledge#154 .

@etirelli @mariofusco can we review this, please? Thx.


tarilabs:
(org.drools.compiler.integrationtests.KieContainerTest.testIncrementalCompilationSynchronization test timed out after 10000 milliseconds)

Jenkins, retest this

tarilabs:
(JpaOptLockPersistentStatefulSessionTest.testOptimisticLockInterceptor)

Jenkins, test again :8ball:

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -763,14 +773,39 @@ private KieSessionConfiguration getKieSessionConfiguration( KieSessionModel kSes
...
+            }
+            DroolsManagementAgent.getInstance().unregisterMBeansFromOwner(this);
+        }
+        
         ((InternalKieServices) KieServices.Factory.get()).clearRefToContainerId(this.containerId, this);
     }
 
+    private boolean isMBeanOptionEnabled() {
+        return MBeansOption.isEnabled( System.getProperty( MBeansOption.PROPERTY_NAME, MBeansOption.DISABLED.toString() ) );
+    }
```
etirelli, line 807: This is ok for now, but we will need to improve it later... MBeans can be activated via configuration on the kbases (instead of the system property), so we will need a way to prevent things from getting out of sync... i.e., mbeans not enabled on the kie-container but enabled on the kbases...

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -519,8 +528,9 @@ public void dispose() {
         }
 
         this.kBase.disposeStatefulSession( this );
-        if (this.kBase.getConfiguration().isMBeansEnabled()) {
-            DroolsManagementAgent.getInstance().unregisterKnowledgeSession(this);
+
+        if (this.mbeanRegistered.get()) {
+            DroolsManagementAgent.getInstance().unregisterKnowledgeSessionUnderName(mbeanRegisteredCBSKey, this);
```
etirelli, line 532: I am not sure the mbean should be unregistered on dispose()... lets discuss this, but as we talked before, the mbeans should only be disposed when the kie-container is disposed... no longer when the session/kbase is disposed.

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -252,6 +253,9 @@
 
     protected PropagationList propagationList;
 
+    private AtomicBoolean mbeanRegistered = new AtomicBoolean(false);
```
etirelli, line 256: Is this extra flag necessary?

etirelli:
@tarilabs lets discuss the registration of the mbeans, but everything else looks good.

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
tarilabs: Yep I agree, and actually this does not unregister the MBean from the JMX server.

This does unregister the _session_ from the MBean monitoring [all the session with] that name - by "name" here represented by the key containerId+kbaseId+ksessionName.

For instance [here](https://github.com/droolsjbpm/drools/pull/877/files#diff-2e5aaaa15892760aba8fecfe30550ef8R281) I still have the mbean registered on JMX, that's why I can (still) ask the mbean how many session does it still monitor. Similarly a [couple of line below](https://github.com/droolsjbpm/drools/pull/877/files#diff-2e5aaaa15892760aba8fecfe30550ef8R283), which returns 0 because all the sessions are gone, but the mbean is still on JMX. The only time when the MBean is disappeared from JMX is on KieContainer dispose as per [specific unit test part](https://github.com/droolsjbpm/drools/pull/877/files#diff-2e5aaaa15892760aba8fecfe30550ef8R184).

Hope to have well understood your comment and provided relevant feedback on it, but we can continue discussing it on quick meeting in case :+1:

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
tarilabs: As this was done similarly before on `KnowledgeBaseImpl` for DROOLS-1235, to avoid possibly inconsistent state of mbean registered on JMX - in case this method is called explicitly manually, or from outside the KieContainer.

etirelli:
Excellent, thank you for the explanations. +1 to merge!

tarilabs:
`JpaOptLockPersistentStatefulSessionTest.testOptimisticLockInterceptor`

Jenkins, test again :8ball:

tarilabs:
@etirelli , this additional commit separates MBean for Stateful Vs Stateless, as discussed also with @mariofusco . There is a "companion commit" in the referenced PR too as expected to reflect this. Could you kindly review again, please? Thx.

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
etirelli: ok

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
etirelli: ok

drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
```diff
@@ -124,21 +119,61 @@ public void registerKnowledgeBase(InternalKnowledgeBase kbase) {
...
+     * Get currently registered session monitor, eventually creating it if necessary.
+     */
+    public GenericKieSessionMonitoringImpl getKnowledgeSessionBean(CBSKey cbsKey, KieRuntimeEventManager ksession) {
+        if (mbeansRefs.get(cbsKey) != null) {
+            return (GenericKieSessionMonitoringImpl) mbeansRefs.get(cbsKey);
+        } else {
+            if (ksession instanceof StatelessKnowledgeSession) { 
+                StatelessKieSessionMonitoringImpl mbean = new StatelessKieSessionMonitoringImpl( cbsKey.kcontainerId, cbsKey.kbaseId, cbsKey.ksessionName );
+                GenericKieSessionMonitoringImpl result = (GenericKieSessionMonitoringImpl) mbeansRefs.putIfAbsent(cbsKey, mbean);
+                if (result == null) {
```
etirelli: I think you might need to wrap this whole registration routine with a lock or monitor. The problem here is 2 concurrent calls, the first adds the mbean to the mbeansRefs map, the second returns the just created mbean to the method caller, but then the first fails registration below with an exception. That means the second call returned a non-registered mbean to the calling code.

tarilabs: you are right, sorry I missed this originally. I reworked it accordingly to your suggestions.

drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
```diff
@@ -124,21 +115,60 @@ public void registerKnowledgeBase(InternalKnowledgeBase kbase) {
...
+    /**
+     * Get currently registered session monitor, eventually creating it if necessary.
+     * @return the currently registered or newly created session monitor, or null if unable to create and register it on the JMX server.
+     */
+    private GenericKieSessionMonitoringImpl getKnowledgeSessionBean(CBSKey cbsKey, KieRuntimeEventManager ksession) {
+        if (mbeansRefs.get(cbsKey) != null) {
+            return (GenericKieSessionMonitoringImpl) mbeansRefs.get(cbsKey);
+        } else {
+            if (ksession instanceof StatelessKnowledgeSession) {
+                synchronized (mbeansRefs) {
```
etirelli, line 145: @tarilabs unfortunately I don't think this is quite right yet. You will still face potential concurrency issues with this code. Inside the critical section you have to check again for pre-existence of an already registered bean or you risk overriding it.

tarilabs: My mistake indeed. I've told @mariofusco by voice I would have modified the code doing "double-check locking à-la Singleton creation" style, but then actually I've missed to do it :( sorry will correct this; thx for pointing it out!

drools-compiler/src/test/java/org/drools/compiler/integrationtests/MBeansMonitoringTest.java
```diff
@@ -166,5 +179,196 @@ public void testContainerMBeans() throws Exception {
...
+        checkTotalSessions(aggrMonitor, 1);
+
+        ksession.fireAllRules();
+        print(aggrMonitor);
+        checkAgendaTotals(aggrMonitor, 2, 1, 1);
+        checkTotalFactCount(aggrMonitor, 1);
+        tft = checkTotalFiringTimeGEQ(aggrMonitor, tft);
+        checkTotalSessions(aggrMonitor, 1);
+        
+        System.out.println("---");
```
psiroky: Please don't use `System.out.println`. I think using `DEBUG` logging would better in this case?

tarilabs: done

psiroky: thanks!

etirelli:
I think it is good to merge now. +1

PR #878, opened 2016-08-23
DROOLS-1256 ForceEagerActivationOption and "not()" LHS pattern does n…

tarilabs:
…ot invoke matchCreated listener

just the reproducer


mariofusco:
Won't fix - see https://issues.jboss.org/browse/DROOLS-1256

PR #880, opened 2016-08-23
DROOLS:1257 drools-examples traits is running on a Infinite loop Problem fixed

AvinduDinushan:
There was a problem with the drools-example traits example. This is the solution for it's second run(), traitsDon.drl file


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/881

PR #883, opened 2016-08-29
[DROOLS-1071] print context when a constraint evaluation fails

mariofusco:


baldimir:
+1. The performance impact of this is negligible. See results here: http://pastebin.com/zKS0u8WB

PR #884, opened 2016-08-29
[DROOLS-1262] log state machine transitions at trace level

mariofusco:


drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1467,8 +1466,8 @@ private void setCurrentState(ExecutionState state) {
         if (currentState == ExecutionState.DISPOSED) {
             return;
         }
-        if ( log.isDebugEnabled() ) {
-            log.debug("State was {} is now {}", currentState, state);
+        if ( log.isTraceEnabled() ) {
```
ge0ffrey, line 1469: this if loop check is unneeded: it is the first thing that log.trace does, even before the log message string concatination/variable replacement.

ge0ffrey:
+1

PR #889, opened 2016-08-31
RHBPMS-856: Condition column headers are not getting reflected when a XLS Decision Table is converted to Guided Decision Table

manstis:
See https://issues.jboss.org/browse/RHBPMS-856

This PR adds a couple of new constructors to `RuleModelVisitor` for use within `drools-wb`. Furthermore `BRLConditionColumn.setHeader(..)` and `BRLActionColumn.setHeadet(..)` overrides were removed as they are called by Errai when unmarshalling a Guided Decision Table which leads to undesirable results. I checked use of the (removed) overridden methods and we populate the header of child columns there too - so the override was not necessary anyway.


manstis:
Part of a (small) ensemble:
- https://github.com/droolsjbpm/drools/pull/889
- https://github.com/droolsjbpm/drools-wb/pull/269

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

csadilek:
+1

PR #890, opened 2016-09-01
DROOLS-1236 JMX objectnaming separation for Stateful Vs Stateless

tarilabs:
(and for alignment with RHQ/JON plug-in).

### DIFF
--- a/drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
+++ b/drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
@@ -18,6 +18,7 @@ package org.drools.core.management;
 
 import org.drools.core.common.InternalWorkingMemory;
 import org.drools.core.impl.InternalKnowledgeBase;
+import org.kie.api.builder.model.KieSessionModel;
 import org.kie.api.event.KieRuntimeEventManager;
 import org.kie.api.management.KieManagementAgentMBean;
 import org.kie.api.management.KieSessionMonitoringMBean;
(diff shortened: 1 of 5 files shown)


drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
```diff
@@ -87,11 +88,22 @@ public static ObjectName createObjectNameBy(String containerId) {
...
 	
-    public static ObjectName createObjectNameBy(String containerId, String kbaseId, String ksessionName) {
+    public static ObjectName createObjectNameBy(String containerId, String kbaseId, KieSessionModel.KieSessionType ksessionType, String ksessionName) {
         return DroolsManagementAgent.createObjectName(CONTAINER_NAME_PREFIX + ":kcontainerId="+ObjectName.quote(containerId) 
                     + ",kbaseId=" + ObjectName.quote(kbaseId) 
+                    + ",ksessionType=" + ksessionType(ksessionType)
                     + ",ksessionName=" + ObjectName.quote(ksessionName) );
     }
+    
+    private static String ksessionType(KieSessionModel.KieSessionType ksessionType) {
```
etirelli, line 98: Matteo, this is nitpicking, but when assigning strings/number/values to enums, usually the best way is to do that in the enum itself. This avoids code duplication and make sure the strings used are consistent across all modules of the system. E.g.: here we are assigning strings to ClockType: https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/ClockType.java

etirelli:
+1 to merge

drools-core/src/main/java/org/drools/core/management/DroolsManagementAgent.java
tarilabs: @etirelli Edson I know, but this time I preferred to avoid any modification of public Kie API which are likely shared across multiple drools-related projects especially the workbench. With `ReleaseId` story I understood we should NOT modify these kind of stuff if we want to backport to minor branches - in this case if we want to backport this to 6.5.x ?

etirelli: I think it is fine if you prefer to leave like this, but adding it to the public API would be fine, as it would just be assigning a string to the enum values.

PR #895, opened 2016-09-02
[6.5.x] GUVNOR-2503: Asset screen Validate-button should use incremental builder

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2503

Mirror from: https://github.com/droolsjbpm/drools/pull/873 (using JDK 7)


csadilek:
+1

PR #896, opened 2016-09-02
RHBRMS-432 : Not possible to use Stateless session in Test Scenarios

Rikkola:
https://issues.jboss.org/browse/RHBPMS-1165

### DIFF
--- a/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/ScenarioRunner4JUnit.java
+++ b/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/ScenarioRunner4JUnit.java
@@ -16,6 +16,10 @@
 
 package org.drools.workbench.models.testscenarios.backend;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
 import org.drools.workbench.models.testscenarios.shared.Scenario;
 import org.junit.internal.AssumptionViolatedException;
 import org.junit.internal.runners.model.EachTestNotifier;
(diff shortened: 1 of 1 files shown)


drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/ScenarioRunner4JUnit.java
```diff
@@ -105,7 +105,7 @@ private void runScenario(RunNotifier notifier, Scenario scenario) {
                     if (ksessionName == null) {
                         eachNotifier.addFailure(new NullPointerException("Test scenario runner could not find the default knowledge session."));
                     } else {
-                        eachNotifier.addFailure(new NullPointerException("Test scenario runner could not find a stateful knowledge session with the name \'" + ksessionName + "\'."));
+                        eachNotifier.addFailure( new NullPointerException( "Test Scenarios require Stateful KieSession to run." ) );
```
manstis, line 108: OK.. so you've made the failure clearer... but can you explain _why_ Test Scenarios cannot use Stateless Sessions (I assume it's because you need it to be stateful to query the results after `fireAllRules()`?)

Rikkola: Can't remember what the exact limitations were, but some of the selections in the test scenario menu only work for stateless. Even with stateful we can query the results once the rules stop firing ( I think ).

So in order to actually fix this ticket we would have to extend the logic for the "new item" menu items in test scenarios and then if the state is set from stateful <-> stateless check again if the scenario is valid. There is some talk about this in the ticket and the text change was what we ended going with.

With the state that the code is in test scenarios this requires quite a lot of refactoring to MVP and better test coverage. For 7.x this was the quick fix and after 7.0 we need to think if we want to keep test scenarios, we can fix and extend it or replace it with Cucumber + simple DSL system and have something that is easier to maintain.

PR #905, opened 2016-09-08
DROOLS-1270 - support fact handle operations for REST APIs on kie server

mswiderski:
@mariofusco @etirelli could you please have a look

depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/170


etirelli:
+1

mswiderski:
please hold on with merge as this affects kie remote client at the moment

mswiderski:
ok, should be now fixed, change was in the DeleteCommand as it's used by kie-remote-client (which regenerates commands for jaxb and that changes the type of fact handle to string. So I had to add that "fake" field to not break backward compatibility of DeleteCommand used by kie-remote-client

PR #907, opened 2016-09-08
Drools rule compiler reports false Java compilation errors in Eclipse Neon

bbrodt:
See https://issues.jboss.org/browse/DROOLS-1250


mariofusco:
@bbrodt This change is causing 1861 failures on Jenkins. Can you give it a second look?

mariofusco:
As discussed with @bbrodt implementing the isPackage method inside the ProjectClassLoader is tricky, slower and very likely useless.

PR #910, opened 2016-09-09
[6.5.x] GUVNOR-2585: In test scenario, expert contraint of type enum can be set but are always false

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2585

Auto-import enums from the class in order to avoid this:
enum-bug

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
+++ b/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
@@ -16,11 +16,16 @@
 
 package org.drools.core.base;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


manstis:
@Rikkola could you please review this...

karreiro:
Jenkins, please retest this.

manstis:
@Rikkola @karreiro Toni and I have talked about "auto-import" in the past. My only concern is what happens currently with - and what should now happen - with _all_ the other editors... For example does the Guided Rule Editor work OK with enums (w/o importing the enum.. if it does then what's different between that and Test Scenarios.. if it does not work; then the "auto-import" needs adding there too). We should strive for consistency (of behaviour/operation) of _all_ (Drools) editors.

Rikkola:
@manstis Guided Rule Editor should work like this. The difference here is that this feature was missing from the test scenario runner. Drools core probably does a similar auto-import.

The auto import for editors, so that the imports automatically get into the imports tab. That would really improve the usability, maybe we should try to get that done after 7.0.

+1

karreiro:
Thanks @Rikkola and @manstis :-)

manstis:
@karreiro Looks good to me - @mariofusco did you want tor review (and merge?)

mariofusco:
I agreed with @karreiro that this PR cannot be accepted as it is. The enum discovery introduced in the TypeResolver will introduce a compile time delay that will be useless in 99.9...% of cases.

PR #911, opened 2016-09-09
GUVNOR-2585: In test scenario, expert contraint of type enum can be set but are always false

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2585

Auto-import enums from the class in order to avoid this:
enum-bug

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
+++ b/drools-core/src/main/java/org/drools/core/base/ClassTypeResolver.java
@@ -16,11 +16,16 @@
 
 package org.drools.core.base;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


manstis:
@Rikkola could you please review this...

karreiro:
Jenkins, please retest this.

Rikkola:
+1

manstis:
IIRC "getDeclaredFields()" does just that - excluding inherited and delegated. You might want to look at including fields on super classes and methods with return types (that are enums).

karreiro:
Jenkins, please retest this.

karreiro:
Thanks @Rikkola and @manstis :-)

manstis:
@karreiro Looks good to me - @mariofusco did you want tor review (and merge?)

mariofusco:
I agreed with @karreiro that this PR cannot be accepted as it is. The enum discovery introduced in the TypeResolver will introduce a compile time delay that will be useless in 99.9...% of cases.

PR #916, opened 2016-09-13
DROOLS-1283 RHBRMS-2582 Memory leak using KieBuilder.build() method

baldimir:
- Added test cases reproducing the memory leak.


mariofusco:
I agreed with @baldimir that this tests are not very reliable and could also cause some false negative

PR #917, opened 2016-09-13
GUVNOR-2644: Guided Rule Template: Generates incorrect DRL when template data has intentionally missing values

manstis:
See https://issues.jboss.org/browse/GUVNOR-2644


manstis:
@Rikkola @karreiro Hi, could you please review?

The issue was a MVEL template constructed to generate the DRL:-

**Original**

```
Person( age > @{min}, age <= @max} @if(!(min == empty || max == empty)),@end name == "Fred" )
```

**New**

```
Person( age > @{min}, age <= @max} @if(!(min == empty && max == empty)),@end name == "Fred" )
```

The comma before the literal was only being inserted if both `@{min}` and `@{max}`  were not empty. If one was empty and the other not the comma was being skipped.

Rikkola:
I'll smurf this!
+1

PR #918, opened 2016-09-14
DROOLS-1283 KieContainer registered w/in KieServices iff ID is explicit.

tarilabs:
If there is no explict ID, the KieContainer is not registered w/in
KieServices however it will still be assigned a random ID for JMX
identification.


tarilabs:
Depends on droolsjbpm/droolsjbpm-knowledge#172 .

tarilabs:
```
 org.drools.persistence.session.JpaOptLockPersistentStatefulSessionTest.testOptimisticLockInterceptor
 Error Details
expected:<6> but was:<9>
```

Already seen occasionally and randomly failing,

Jenkins, test again :8ball:

PR #920, opened 2016-09-14
Make MavenEmbedderTest more isolated by using empty maven settings

mariofusco:


mariofusco:
Jenkins, test again

psiroky:
+1 as it fixes the issue (creating public method just for testing is not ideal, but I too don't see better solution right now)

PR #922, opened 2016-09-15
DROOLS-1286: Guided Decision Table: Unable to specify field type with Condition BRL fragment

manstis:
See https://issues.jboss.org/browse/DROOLS-1286


manstis:
@Rikkola Please review.

Rikkola (approved):
+1

PR #924, opened 2016-09-15
[DROOLS-1292] make initialization of inner declarations inside accumulate node thread-safe

mariofusco:


tarilabs (approved):
+1 to merge, as discussed in presence.

PR #926, opened 2016-09-16
GUVNOR-2687: Guided Decision Table: Only a single negate Pattern shown when multiple defined

manstis:
See https://issues.jboss.org/browse/GUVNOR-2687


manstis:
@Rikkola please wave your wand of approval over this PR.

Rikkola (approved):
+1

PR #927, opened 2016-09-19
[DROOLS-1297] avoid checking available PermGen space under GAE

mariofusco:


ge0ffrey:
+1 looks good to merge

Any reason why isAndroid() is on ClassUtils, but isGAE() isn't?

PR #935, opened 2016-09-23
[6.5.x] GUVNOR-2587: Running test scenario one by one work but all together messes all up

karreiro:
See: https://issues.jboss.org/browse/GUVNOR-2587
- Before:
  guvnor-2587-before
- After:
  guvnor-2587-after


karreiro:
@Rikkola Could you please review this? :-)

manstis:
Requires https://github.com/droolsjbpm/drools-wb/pull/306 too.

Rikkola (approved):
+1

PR #936, opened 2016-09-27
Extending DeleteTest test cases

cimbalek:
- added few test cases into the DeleteTest class

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
@@ -32,7 +32,6 @@ import org.kie.api.runtime.rule.FactHandle;
 import org.kie.api.runtime.rule.QueryResults;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertEquals;
 
 public class DeleteTest {
 
@@ -43,17 +42,14 @@ public class DeleteTest {
     @Before
     public void setUp() {
         KieFileSystem kfs = KieServices.Factory.get().newKieFileSystem();
-
         kfs.write(KieServices.Factory.get().getResources()
                 .newClassPathResource(DELETE_TEST_DRL, DeleteTest.class));
 
         KieBuilder kbuilder = KieServices.Factory.get().newKieBuilder(kfs);
-
         kbuilder.buildAll();
 
         List<Message> res = kbuilder.getResults().getMessages(Level.ERROR);
-
-        assertEquals(res.toString(), 0, res.size());
+        assertThat(res).isEmpty();
 
         KieBase kbase = KieServices.Factory.get()
                 .newKieContainer(kbuilder.getKieModule().getReleaseId())
@@ -73,10 +69,12 @@ public class DeleteTest {
 
         FactHandle george = ksession.insert(new Person("George", 19));
         QueryResults results = ksession.getQueryResults("informationAboutPersons");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$countOfPerson")).isEqualTo(2L);
 
         ksession.delete(george);
         results = ksession.getQueryResults("informationAboutPersons");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$countOfPerson")).isEqualTo(1L);
     }
 
@@ -84,13 +82,16 @@ public class DeleteTest {
     public void deleteFactTwiceTest() {
         FactHandle george = ksession.insert(new Person("George", 19));
         QueryResults results = ksession.getQueryResults("countPerson");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$personCount")).isEqualTo(1L);
 
         ksession.delete(george);
         results = ksession.getQueryResults("countPerson");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$personCount")).isEqualTo(0L);
 
         ksession.delete(george);
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$personCount")).isEqualTo(0L);
     }
 
@@ -106,10 +107,12 @@ public class DeleteTest {
         ksession.update(person, new Person("John", 21));
 
         QueryResults results = ksession.getQueryResults("countPerson");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$personCount")).isEqualTo(1L);
 
         ksession.delete(person);
         results = ksession.getQueryResults("countPerson");
+        assertThat(results).isNotEmpty();
         assertThat(results.iterator().next().get("$personCount")).isEqualTo(0L);
     }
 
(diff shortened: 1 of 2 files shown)


baldimir:
+1 if all checks pass.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
 
         FactHandle george = ksession.insert(new Person("George", 19));
+        QueryResults results = ksession.getQueryResults("informationAboutPersons");
+        assertEquals(2L, results.iterator().next().get("$countOfPerson"));
 
-        QueryResults results = ksession
-                .getQueryResults("informationsAboutPersons");
+        ksession.delete(george);
+        results = ksession.getQueryResults("informationAboutPersons");
+        assertEquals(1L, results.iterator().next().get("$countOfPerson"));
```
triceo: For these non-obvious operations, I'd like us to use AssertJ. They make the code more understandable, and the failures have better descriptions.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
+    @Test(expected = IllegalArgumentException.class)
+    public void deleteNullFactTest() {
+        ksession.delete(null);
+        fail("Delete null fact should have failed.");
+    }
 
-        ksession.dispose();
+    @Test
+    public void deleteUpdatedFactTest() {
+        FactHandle george = ksession.insert(new Person("George", 18));
```
triceo: Since you're changing the person's name later, maybe the variable should have been called `person`.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
+
+        ksession.update(george, new Person("John", 21));
+
+        QueryResults results = ksession.getQueryResults("countPerson");
+        assertEquals(1L, results.iterator().next().get("$personCount"));
```
triceo: Since we're updating the person's name, it'd be nice to see (and test) what happens when there's a query that actually includes the name.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
+        results = ksession.getQueryResults("informationAboutPersons");
+        assertEquals(1L, results.iterator().next().get("$countOfPerson"));
+    }
 
-        assertEquals(2L, results.iterator().next().get("$countOfPerson"));
+    @Test
+    public void deleteFactTwiceTest() {
+        FactHandle george = ksession.insert(new Person("George", 19));
+        QueryResults results = ksession.getQueryResults("countPerson");
+        assertEquals(1L, results.iterator().next().get("$personCount"));
```
triceo: Are the `Long` casts actually necessary?

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
-        results = ksession.getQueryResults("informationsAboutPersons");
+        ksession.delete(george);
+        assertEquals(0L, results.iterator().next().get("$personCount"));
+    }
 
-        assertEquals(1L, results.iterator().next().get("$countOfPerson"));
+    @Test(expected = IllegalArgumentException.class)
+    public void deleteNullFactTest() {
+        ksession.delete(null);
+        fail("Delete null fact should have failed.");
```
triceo: Since the exception is `expected`, the `fail()` call is not necessary. Keeping either one should be enough.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -52,24 +59,59 @@ public void deleteFactTest() {
...
+    @Test
+    public void deleteFactTest() {
         ksession.insert(new Person("Petr", 25));
 
         FactHandle george = ksession.insert(new Person("George", 19));
+        QueryResults results = ksession.getQueryResults("informationAboutPersons");
+        assertEquals(2L, results.iterator().next().get("$countOfPerson"));
 
-        QueryResults results = ksession
-                .getQueryResults("informationsAboutPersons");
```
triceo, line 62: Nitpicking here - the plural of "information" is "information".

triceo (changes requested):
I've provided a few suggestions to the PR.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
baldimir: Yes, accumulate returns Long in this case.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
cimbalek: I would rather include this in UpdateTest which I will be working on next.

triceo: OK.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
cimbalek: Done.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
cimbalek: Done.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
cimbalek: Done.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
```diff
@@ -73,45 +73,44 @@ public void deleteFactTest() {
 
         FactHandle george = ksession.insert(new Person("George", 19));
         QueryResults results = ksession.getQueryResults("informationAboutPersons");
-        assertEquals(2L, results.iterator().next().get("$countOfPerson"));
+        assertThat(results.iterator().next().get("$countOfPerson")).isEqualTo(2L);
```
triceo, line 73: See what options become available to you if you start with `assertThat(results)`. ;-)

cimbalek: Unfortunately none of them is really something we want or need to use in this case. But thank you for investigating this with me and making code safer.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/DeleteTest.java
cimbalek: Is this just explanation of edit or should I revert this change?

triceo (approved):

PR #937, opened 2016-09-27
[6.5.x] GUVNOR-2706: Adding Multibyte class name Java file fails with 'The public type <class name> must be defined in its own file' in business-central

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2706
- Before:
  rhbrms-2-before
- After:
  rhbrms-2-after


karreiro:
@etirelli Hi, would you mind taking a look please :-)

etirelli:
Looks good! +1

PR #939, opened 2016-09-28
[DROOLS-1308] make drl marshaller to wrap from patterns in parenthesis

mariofusco:


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -657,11 +657,18 @@ public void visitFromCompositeFactPattern( final FromCompositeFactPattern patter
             if ( pattern.getFactPattern() != null ) {
                 final LHSGeneratorContext gctx = generatorContextFactory.newChildGeneratorContext( rootContext,
                                                                                                    pattern.getFactPattern() );
+                // DROOLS-1308 - wraps from pattern in parenthesis
```
Rikkola, line 660: Personally I prefer to have these comments as unit tests or as a comment in the unit test.
The test fails if anyone changes the code so that this comment is not honored.

Rikkola (approved):
Just a small comment, but no need to change it for this PR.

+1

PR #940, opened 2016-09-29
DROOLS-1279: [GSS] Compilation of spreadsheet fails with specific condition

manstis:
See https://issues.jboss.org/browse/DROOLS-1279

`LhsBuilder` is re-used when a spreadsheet contain merged columns. It contains state for a specific constraint's `FieldType` which was being overwritten by second and subsequent merged columns. We need to store the `FieldType` per column to avoid it being overwritten.


manstis:
@mariofusco Hi Mario, you're probably the only other person who knows much about this (when I say "knows" I mean has had any reason to look at any of this code in recent years!). Would you mind reviewing please.

mariofusco:
@manstis The fix is ok for me. Just note that in case you'll backport this fix to 6.x you will have to avoid the diamond operator that you used in the LhsBuilder constructor.

PR #943, opened 2016-09-30
[DROOLS-1310] remove expired events with pending activation from the …

mariofusco:
…object store when left tuples are retracted


tarilabs (approved):
summarized on IRC, follows my +1 to merge

PR #953, opened 2016-10-07
JBPM-4646 - Case management - extended ProcessContext to allow access…

mswiderski:
… to case data and assignments

depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/176


krisv:
+1

PR #959, opened 2016-10-13
Reproducer for CEP KieSession deserialization NullPointerException with Timer and expiration

valdar:


valdar:
See DROOLS-1328: https://issues.jboss.org/browse/DROOLS-1328

mariofusco:
Test case incorporated and fixed by https://github.com/droolsjbpm/drools/pull/962

PR #964, opened 2016-10-14
-pong fixes. Redraw is now simpler.

mdproctor:
-Reversed dispose and show double buffer


tarilabs:
Testing on RHEL/Gnome:

Pong working ok:
image

Invaders 1 ok for keystroke:
image

Invaders 2 ok for draw spaceship:
image

Invaders 3 ok for movement:
image

Invaders 4 ok for bounding box:
image

Invaders 5 ok draw all:
image

Invaders 6 ok with firings:
image

From my side is good to merge and port to 6.5.x please.

PR #967, opened 2016-10-15
[DROOLS-1329] fix expirations order for events that are already expir…

mariofusco:
…ed at insertion time


etirelli:
Mario, this PR looks good to me, +1. I do have to caveat that I don't have the full picture if this solves all the pending issues. But lets go ahead with this one, and tackle any remaining issue in an additional PR.

mariofusco:
Jenkins test

mariofusco:
It seems that there is something wrong on our Jenkins now: the build job is not starting at all. Anyway I ran the whole test suite on my machine and everything looks ok.

PR #970, opened 2016-10-19
JBPM-4646 - Case management - simplified use for case role assignments

mswiderski:


mswiderski:
since PR https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/178 was merged, merging this one to resolve compilation errors - it's test related only change to comply to interface

PR #974, opened 2016-10-21
[DROOLS-1338] fix conflicting right tuples updates in not node

mariofusco:


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/975

PR #977, opened 2016-10-24
[DROOLS-1335] make KieRepository to try to retrieve a KieModule also …

mariofusco:
…from current context classloader


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/978

PR #982, opened 2016-10-25
Cconv

mdproctor:


mariofusco:
Merged by https://github.com/droolsjbpm/drools/pull/997

PR #984, opened 2016-10-27
RHBRMS-2619 Rules compilation failure depending on condition ordering

baldimir:
- Reproducer from RHBRMS-2619 issue provided by @martinweiler (Thanks!).


baldimir:
Closed because test fail from reproducer isn't actually a fail :) See comments from here https://issues.jboss.org/browse/DROOLS-1308

PR #985, opened 2016-10-28
[DROOLS-1346] always pass to the consequence the original fact handle…

mariofusco:
… even when cloned by a WindowNode


tarilabs (approved):
+1 as discussed

```
<mmortari> mfusco, so let me try to sum my understanding about pull/985: the consequence was updating the "cloned" FH (from the window), not the root FH, hence the rule which is not sharing the window, and having a link to the root-FH, was not aware and not been notified of the update? If my understanding is correct, please go ahead with the merge.
<mfusco> mmortari: more or less yes, the consequence and then the update was passed with the wrong FH, the cloned one instead of the original one
<mfusco> mmortari: doing so the WindowNode wasn't finding the original RT and then propagated it as a new insert instead of a modify, thus causing the bug
```

PR #986, opened 2016-10-28
add CompilationProblemResult interface to solve duplicate code problem

artyom-88:


mariofusco:
I really appreciate that you're avoiding code duplication but personally I believe you're doing it in the wrong way. You could obtain the same result avoiding to introduce a new interface just to reuse the 2 default methods in there with a "fake" inheritance. In particular you could move the implementations of those 2 default methods inside 2 static methods in an utility class and pass to it the object and message as arguments.

I'm closing this pull request. Please let me know if what I'm suggesting makes sense and you will send a second pull request. Thanks.

artyom-88:
Ok, i'll add an utility class with static methods

PR #987, opened 2016-10-31
RHBPMS-4413: DRL file with spaces in name fails first validation

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4413
- Before:
  a
- After:
  b

The `buildChanges` method from the `KieBuilderSetImpl` was failing because [this validation](https://github.com/droolsjbpm/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java#L988) had a false negative due to [this comparison](https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/definitions/impl/KnowledgePackageImpl.java#L715-L715).


karreiro:
Jenkins, please retest this.

karreiro:
@Rikkola @mariofusco Could you please review this one too?

Rikkola:
+1

PR #993, opened 2016-11-02
[DROOLS-1349] fix no-loop with eager evaluation

mariofusco:


tarilabs (approved):
+1, the propagation context need ensured to be set, before forcing the flush

mariofusco:
Jenkins retest

PR #995, opened 2016-11-02
add BuilderResultUtils to drools-compiler to avoid code duplicate

artyom-88:
add BuilderResultUtils to drools-compiler to avoid code duplicate with static methods.

Rework of pull-request https://github.com/droolsjbpm/drools/pull/986


mariofusco:
ok to test

PR #998, opened 2016-11-07
(6.4.x) RHBRMS-2573: separators generated according to peers

jomarko:


manstis:
Thanks @jomarko however I have chosen to cherry-pick the fix for master and 6.5.x to 6.4.x.

Please see https://github.com/droolsjbpm/drools/pull/999

PR #1002, opened 2016-11-11
Fixed imports - the class didn't compile on jdk7.

baldimir:
- This is needed for running tests with jdk7.


psiroky (approved):

psiroky:
I'll also configure the animal-sniffer-plugin to scan also test sources, not just "main" sources, to avoid these kinds of issues in future.

baldimir:
Thanks Petr!

psiroky:
Unfortunately it seems that `animal-sniffer` is not able to check test classes, see https://github.com/mojohaus/animal-sniffer/issues/5 for more info.

Someone even proposed a patch some time ago, but it probably never made it into the main sources: http://mojo.10943.n7.nabble.com/Fwd-animal-sniffer-for-maven-test-classes-td43156.html.

There is nothing we can do right now, but if I find some time I'll see if I could add this feature into the plugin.

PR #1003, opened 2016-11-14
RHBRMS-2689: Unable to build project if kbase and ksession names are identical

manstis:
See https://issues.jboss.org/browse/RHBRMS-2689

@mariofusco This is a failing Unit Test for the mentioned JIRA.


manstis:
@mariofusco has produced a fix and other unit tests.

PR #1005, opened 2016-11-15
Fix race condition

anujshahwork:
Fix race condition on assignment of imports to unsafe parserConfiguration field.

[DROOLS-1359](https://issues.jboss.org/browse/DROOLS-1359)

### DIFF
--- a/drools-core/src/main/java/org/drools/core/rule/MVELDialectRuntimeData.java
+++ b/drools-core/src/main/java/org/drools/core/rule/MVELDialectRuntimeData.java
@@ -315,10 +315,12 @@ public class MVELDialectRuntimeData
 
             }
 
-            this.parserConfiguration = new ParserConfiguration();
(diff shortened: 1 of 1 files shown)


etirelli:
@mariofusco I looked at this PR, and while it makes a point on a concurrency issue, I don't think the fix by itself is enough. It seems to me that the whole MVELDialectRuntimeData class is thread unsafe. So the first question is: does it need to be thread-safe? and if so, then we need a lot more done there, including synchronization of critical areas. Can you please take a look?

-1 to this PR as it is insufficient as is, but it does demonstrate one of things in this class that is not thread safe.

mariofusco:
I also don't see why the MVELDialectRuntimeData should be thread-safe at the moment and if this is true this fix is far from being enough.

@anujshahwork If you have the evidence of a race condition caused the MVELDialectRuntimeData please send a test case demonstrating it.

PR #1009, opened 2016-11-18
DROOLS-1174 reproducer

yurloc:
This commit adds missing test coverage for the fix made some time ago.

I believe this is a valuable addition as the test fails when the fix is reverted and there was no such failing test before the bug was fixed and no test was included with the fix ([abdf21f2d2](https://github.com/droolsjbpm/drools/commit/abdf21f2d2)).


yurloc:
Note: the test method name should be changed before merging.

mariofusco:
@yurloc Good job. Please rename the test method testReorderRightMemoryOnIndexedField and I'll merge it.

yurloc:
Jenkins, retest this. Let's give it one more try but I guess it's just unstable tests because all tests pass on my machine.

yurloc:
@mariofusco method name updated, please go ahead and merge.

PR #1014, opened 2016-11-24
removed PMML from drools-compiler and used an assembler instead

mdproctor:
removed PMML from drools-compiler and used an assembler instead.

Do not merge, as PMMLUsageDemoTest.invokePmmlWithTrait failed. 
"MockColdTrait cannot be resolved to a type"

Although I don't see how moving to an assembler would cause this. Please fix/verify then merge.


mariofusco:
outdated

PR #1016, opened 2016-12-02
GUVNOR-2745: Add bug reproducer test

jomarko:


manstis:
Thanks for the test @jomarko 

I've included it in [here](https://github.com/droolsjbpm/drools/pull/1021) along with the fix.

PR #1018, opened 2016-12-02
Fix drools-example Sudoku grid "Dump" feature

tarilabs:
Original code actually caused `fmt.close();` to close `System.out`, so not only the Dump feature would have worked only once, but any further `System.out` call after clicking the Dump button would have simply not rendered any output on the console.


tarilabs:
@mariofusco I understand ( double-checked with @etirelli ) we can backport this as it's not touching the core code but self-contained in the example code. Can you kindly review+merge, please?

PR #1020, opened 2016-12-02
Fix drools-example GameUI keylistener

tarilabs:
Avoid to Pong, Invaders end users the need to click the Canvas in order
to actually produce the key pressed events.
This way, the external JFrame registers the keylistener.
The internal JPanel and Canvas are not responsive to the user hence are
set as non-focusable.
When the external JFrame becomes visible, the keylistener is immediately
active to create the key pressed events necessary for the game/rules.

Please notice GameUI is only used by Canvas-based games: Pong, Invaders.

### DIFF
--- a/drools-examples/src/main/java/org/drools/games/GameUI.java
+++ b/drools-examples/src/main/java/org/drools/games/GameUI.java
@@ -56,8 +56,10 @@ public class GameUI extends Canvas{
         panel.add(this);
         setIgnoreRepaint(true);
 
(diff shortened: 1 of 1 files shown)


tarilabs:
Please notice GameUI is only used by Canvas-based games: Pong, Invaders.
i.e.:
image

tarilabs:
I've manually tested Pong, Invaders6, Invaders5 and on RHEL/Gnome I don't have any longer to click on the Canvas; simply as soon as the JFrame is visible I can start immediately to play by pressing the keys.

@mdproctor @mariofusco what do you think?

PR #1023, opened 2016-12-06
Revert "Fixes after JPA 2.0 -> 2.1 upgrade"

mariofusco:
Reverts droolsjbpm/drools#1019


psiroky:
I guess we don't need this anymore as the revert was already pushed to master: https://github.com/droolsjbpm/drools/commit/94b79436bec5ffbd70a03e081ef8cf2ba57b9662

PR #1026, opened 2016-12-06
[DROOLS-1377] remove reteoo implementation

mariofusco:


mariofusco:
ok to test

psiroky:
I think this definitely qualifies as a backwards-incompatible change and as such should be documented in the upgrade recipe https://github.com/droolsjbpm/drools-website/blob/master/download/upgradeRecipe/upgradeRecipe7.0.adoc

PR #1028, opened 2016-12-07
Remove drools-reteoo leftovers

psiroky:


psiroky:
Merging to fix the build. The removed method is private and not used anywhere, so it should be safe to remove.

PR #1029, opened 2016-12-08
[DROOLS-1353] define ExecutableRunner interface - fix for osgi

mswiderski:
depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/196

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
@@ -137,7 +137,7 @@ public class PersistableRunner implements SingleSessionCommandService {
 
         ((InternalKnowledgeRuntime) this.ksession).setEndOperationListener( new EndOperationListenerImpl(this.txm, this.sessionInfo ) );
 
-        this.sessionContext = RequestContext.create().with( this.ksession );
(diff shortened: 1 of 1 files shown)


mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
anyone willing to review?

psiroky:
I would very much like to, but I have unfortunately no idea how this code is supposse to work and if this is the correct classloader to use....

mswiderski:
it needs to use same class loader that has the implementation class mentioned in RequestContext interface https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/196/files#diff-555c711c48d850e1ac1855682aecf68cR38

psiroky:
You mean the same classloader as `org.drools.core.command.RequestContextImpl` uses?

mswiderski:
yes, pretty much

psiroky:
So `org.drools.core.command.RequestContextImpl.class.getClassLoader()` would work as well? If we really need the same CL, then this make it more obvious I think.

mswiderski:
osgi wise it needs to come from the same bundle so I just used class loader from drools as they are bundled and thus same bundle class loader is used.

psiroky:
OK, that makes sense. Although looking just at the diff  I would not be able to easily figure that out.

PR #1030, opened 2016-12-09
[DROOLS-1378] Option: Force reading String from a Numeric type cell i…

tkobayas:
…n SpreadSheet

Introduced a system property "drools.spreadsheet.numeric.disabled" which avoids to read a double value and relies on DataFormatter instead.

This system property is enabled only in SpreadsheetCompilerUnitTest#testDisableNumeric(). But I also tested the all drools-decisiontables test cases with this property and confirmed that only SpreadsheetCompilerUnitTest#testProcessSheetForExtremeLowNumbers() failed. It's expected because the test is intended to read a double value.

Note: the Numeric cell ("6600") in the test case XLS (Numeric.xls) was copied from customer's cell which was generated by VBA macro.

Feel free to let me know your thought on this.


manstis:
@tkobayas Hi, it would be better, IMO, if you made the flag work the same as other flags already present (for example [EscapeQuotes](https://github.com/tkobayas/drools/blob/5ced58a913755673473c2387d7f225c7d68edadc/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/DefaultRuleSheetListener.java#L76)). and [Sequential](https://github.com/tkobayas/drools/blob/5ced58a913755673473c2387d7f225c7d68edadc/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/DefaultRuleSheetListener.java#L75). There is a [test](https://github.com/tkobayas/drools/blob/5ced58a913755673473c2387d7f225c7d68edadc/drools-decisiontables/src/test/java/org/drools/decisiontable/parser/RuleWorksheetParseTest.java#L450) showing use of EscapeQuotes that will be useful to understand how it works (the flag is set in the "RuleSet" section of the decision table).

The new flag would also need inclusion in the [documentation](https://github.com/droolsjbpm/kie-docs/blob/master/docs/drools-docs/topics/UserGuide/DecisionTables-section.adoc).

manstis (changes requested):
The flag should be added to RuleSet and not a System Property. Can you please make the appropriate changes and resubmit along with a documentation PR.

tkobayas:
Thank you for the suggestion, Mike! Sure I'll make changes soon.

tkobayas:
Made changes and also sent a doc PR (https://github.com/droolsjbpm/kie-docs/pull/191).

Btw, please check with Mario if he agrees this feature request (See comments in https://issues.jboss.org/browse/DROOLS-1378).

manstis (approved):

PR #1033, opened 2016-12-12
Add reactivity support for Collection interfaces

tarilabs:
Specifically for `List` and `Set` as specialized `ReactiveList` and `ReactiveSet`, both extending an abstract `ReactiveCollection`. Include reactive support for `Iterator` and `ListIterator`.


tarilabs:
@mdproctor this is my proposal for core side - currently now working to upgrade the Maven support and some doc.

mariofusco:
ok to test

PR #1037, opened 2016-12-14
[RHBRMS-2462] allow to insert nulls inside the red-black tree index

mariofusco:


drools-core/src/main/java/org/drools/core/util/TupleRBTree.java
```diff
@@ -293,8 +301,15 @@ private void replaceNode(Node<K> oldn,
     }
 
     public Node<K> insert(K key) {
-        Node<K> insertedNode = new Node<K>( key );
```
tarilabs, line 296: as discussed this is not related to this JIRA, but a micro optimization to avoid creating the Node if it's actually not required.

tarilabs (approved):

PR #1038, opened 2016-12-14
GUVNOR-2786 : Guided Decision Tables: Generate source code based on selected hit mode

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2619

This closes tickets:
GUVNOR-2786 : Guided Decision Tables: Generate source code based on selected hit mode
GUVNOR-2766 : Guided Decision Tables: Add hit policy into dtable model

And completes the code generation for:
GUVNOR-2792 : Guided Decision Table V&V - Hit policy: Rule Order
GUVNOR-2746 : Guided Decision Table V&V - Hit policy: First hit
GUVNOR-2794 : Guided Decision Table V&V - Hit policy: Unique

@manstis and @jomarko 
Throwing this out for review. This could even be merged in. It defaults the hit policies to NONE and there is currently no way in the workbench to create anything else than NONE. I chose not to use the name "Any", at least in the code since there is also the Any hit mode that we are not implementing and is a bit different than our default free editing mode.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
@@ -91,7 +91,28 @@ public class GuidedDecisionTable52 implements HasImports,
         LIMITED_ENTRY
     }
 
+    /**
+     * Lists available dtable hit policies.
+     * <br>
+     * <br>
+     * Source code is genarated based on the selection.
+     * NONE is the default, setting no limitations for the table setup
+     * and generating nothing that the user does not add himself.
+     */
+    public enum HitPolicy {
+        NONE,
+        UNIQUE_HIT,
+        FIRST_HIT,
+        RULE_ORDER;
+
+        public static HitPolicy getDefault() {
+            return NONE;
+        }
+    }
+
     private TableFormat tableFormat = TableFormat.EXTENDED_ENTRY;
+
+    private HitPolicy hitPolicy = HitPolicy.getDefault();
 
     /**
      * First column is always row number. Second column is description.
@@ -331,6 +352,14 @@ public class GuidedDecisionTable52 implements HasImports,
 
     public void setTableFormat( final TableFormat tableFormat ) {
         this.tableFormat = tableFormat;
+    }
+
+    public HitPolicy getHitPolicy() {
+        return hitPolicy == null ? HitPolicy.getDefault(): hitPolicy;
+    }
+
+    public void setHitPolicy( final HitPolicy hitPolicy ) {
+        this.hitPolicy = hitPolicy;
     }
 
     /**
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/validation/DecisionTableValidator.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/validation/DecisionTableValidator.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.drools.workbench.models.guided.dtable.shared.validation;
+
+import java.util.Set;
+
+import org.drools.workbench.models.guided.dtable.shared.model.AttributeCol52;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+
+public class DecisionTableValidator {
+
+    private final GuidedDecisionTable52 dtable;
+
+    private final Set<String> illegalAttributes;
+
+    public DecisionTableValidator( final GuidedDecisionTable52 dtable ) {
+        this.dtable = dtable;
+        this.illegalAttributes = HitPolicyValidation.getReservedAttributes( dtable.getHitPolicy() );
+    }
+
+    public void validate() {
+        validateAttributes();
+    }
+
+    public void isValidToAdd( final AttributeCol52 attributeCol52 ) {
+
+        isValid( attributeCol52 );
+
+        for ( final AttributeCol52 existingAttributeCol52 : dtable.getAttributeCols() ) {
+            if ( attributeCol52.getAttribute()
+                    .equals( existingAttributeCol52.getAttribute() ) ) {
+                throw new DuplicateAttributeException( existingAttributeCol52.getAttribute() );
+            }
+        }
+    }
+
+    private void isValid( final AttributeCol52 attributeCol52 ) {
+        if ( illegalAttributes.contains( attributeCol52.getAttribute() ) ) {
+            throw new InvalidAttributeColumnForHitPolicyException( attributeCol52.getAttribute(),
+                                                                   dtable.getHitPolicy() );
+        }
+    }
+
+    private void validateAttributes() {
+        for ( final AttributeCol52 attributeCol52 : dtable.getAttributeCols() ) {
+            isValid( attributeCol52 );
+        }
+    }
+}
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceRuleOrderHitPolicyTest.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceRuleOrderHitPolicyTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.drools.workbench.models.guided.dtable.backend;
+
+import org.drools.workbench.models.guided.dtable.backend.util.DataUtilities;
+import org.drools.workbench.models.guided.dtable.shared.model.AttributeCol52;
+import org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.drools.workbench.models.guided.dtable.backend.TestUtil.*;
+
+public class GuidedDTDRLPersistenceRuleOrderHitPolicyTest {
+
+
+    private GuidedDecisionTable52 dtable;
+
+    @Before
+    public void setUp() throws
+                        Exception {
+        dtable = new GuidedDecisionTable52();
+        dtable.setTableName( "Rule order table" );
+
+        dtable.setHitPolicy( GuidedDecisionTable52.HitPolicy.RULE_ORDER );
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void blockUseOfSalience() {
+
+        final AttributeCol52 attributeCol52 = new AttributeCol52();
+        attributeCol52.setAttribute( "salience" );
+        attributeCol52.setDefaultValue( new DTCellValue52( "123" ) );
+
+        dtable.getAttributeCols()
+                .add( attributeCol52 );
+
+        GuidedDTDRLPersistence.getInstance()
+                .marshal( dtable );
+    }
+
+    @Test
+    public void salienceEqualsOrderOfLinesNumberRunsFromBottomToTop() {
+
+        dtable.setData( DataUtilities.makeDataLists(
+                new Object[][]{
+                        new Object[]{1, "desc-row1"},
+                        new Object[]{2, "desc-row2"},
+                        new Object[]{3, "desc-row3"}
+                } ) );
+
+        final String drl = GuidedDTDRLPersistence.getInstance()
+                .marshal( dtable );
+
+        assertContainsLinesInOrder( drl,
+                                    "rule \"Row 1 Rule order table\"",
+                                    "salience 2",
+                                    "rule \"Row 2 Rule order table\"",
+                                    "salience 1",
+                                    "rule \"Row 3 Rule order table\"",
+                                    "salience 0" );
+    }
+}
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceUniqueHitPolicyTest.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceUniqueHitPolicyTest.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.drools.workbench.models.guided.dtable.backend;
+
+import org.drools.workbench.models.guided.dtable.backend.util.DataUtilities;
+import org.drools.workbench.models.guided.dtable.shared.model.AttributeCol52;
+import org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.drools.workbench.models.guided.dtable.backend.TestUtil.*;
+
+public class GuidedDTDRLPersistenceUniqueHitPolicyTest {
+
+    private GuidedDecisionTable52 dtable;
+
+    @Before
+    public void setUp() throws
+                        Exception {
+        dtable = new GuidedDecisionTable52();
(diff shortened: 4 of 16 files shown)


drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/DecisionTableHitPolicyEnhancer.java
```diff
@@ -0,0 +1,102 @@
+package org.drools.workbench.models.guided.dtable.backend;
```
manstis: License!

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/DecisionTableHitPolicyEnhancer.java
```diff
@@ -0,0 +1,102 @@
...
+        attributeCol52.setAttribute( GuidedDecisionTable52.ACTIVATION_GROUP_ATTR );
+        dtable.getAttributeCols()
+                .add( attributeCol52 );
+
+        final int columnIndex = dtable.getExpandedColumns()
+                .indexOf( attributeCol52 );
+
+        for ( final List<DTCellValue52> row : dtable.getData() ) {
+            row.add( columnIndex,
+                     new DTCellValue52( "activation-group \"" + activationGroupType + " " + dtable.getTableName() + "\"" ) );
```
manstis: You don't need to add "activation-group" to the cell's value (unless you want ```activation-group activation-group first-hit-policy-group mytable``` as the activation groups value (which of course is quite valid; I was just commenting if this was your intent).

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
```diff
@@ -333,6 +354,14 @@ public void setTableFormat( final TableFormat tableFormat ) {
         this.tableFormat = tableFormat;
     }
 
+    public HitPolicy getHitPolicy() {
+        return hitPolicy == null ? HitPolicy.NONE : hitPolicy;
```
manstis: Better to use ```HitPolicy.getDefault()```?

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/DecisionTableHitPolicyEnhancer.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    private void attributeDoesNotExist( final String attribute ) {
+        for ( final AttributeCol52 attributeCol52 : dtable.getAttributeCols() ) {
+            if ( attributeCol52.getAttribute()
+                    .equals( attribute ) ) {
+                throw new IllegalArgumentException( "A decision table can not have both attribute: " + attribute + " and hit policy: " + dtable.getHitPolicy()
```
manstis: It might be nicer to throw a new ```@Portable``` exception that we can catch and handle better on the client-side (for example, when the User elects to "show source" and we marshall to DRL this exception might be thrown and we could use a ```CommandDrivenErrorCallback``` to trap the specific exception). Also; I think we probably need to re-use this validation when the User attempts to add an Attribute column themselves via the UI (ATM this check only happens when the DRL is generated, which would be when ```Builder``` runs - and at that point any exception will just barf the build process).

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/DecisionTableHitPolicyEnhancer.java
Rikkola: The code generated from this would actually not compile. Need to add tests that catches this.

Rikkola: The code generated from this would actually not compile. Need to add tests that catches this.

Since it would make:
`activation-group activation-group "first-hit-policy-group mytable"`

Rikkola: Actually it makes:
`activation-group "activation-group "first-hit-policy-group First hit policy table""`

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/DecisionTableHitPolicyEnhancer.java
Rikkola: Good point. We want to use the exception in client side.

Rikkola: Investigated.

`IllegalArgumentException` is `@Portable` by default. We also use it in other parts of [dtable code generation](https://github.com/droolsjbpm/drools/blob/b956ace981d022e188b2cf8c56c3aba2b6e13450/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java#L719-L719).

DRL source tab always shows the same boring "The workbench failed to generate source for this file" when it can not make the source, but if we chose to show the cause we still could with the `IllegalArgumentException`.

This still brought up two tasks that I think were worth doing ( and should get done ).
1. I separated the `Enchancer` and `Validator` to different classes for better reuse. They both have a job to do, so we should not mix them. `Enhancer` still uses `Validator` to check dtable can be enhanced.
2. Both `Validator` and `Enhancer` should be in the `shared` package. They need to be usable in both sides.

Rikkola:
@manstis Went over the notes you had and did some changes. Let me know if your happy with the solutions.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyValidator.java
```diff
@@ -0,0 +1,62 @@
...
+                // We do nothing.
+                break;
+        }
+    }
+
+    private void attributeDoesNotExist( final String attribute ) {
+        for ( final AttributeCol52 attributeCol52 : dtable.getAttributeCols() ) {
+            if ( attributeCol52.getAttribute()
+                    .equals( attribute ) ) {
+                throw new IllegalArgumentException( "A decision table can not have both attribute: " + attribute + " and hit policy: " + dtable.getHitPolicy()
```
manstis: Heh, OK, so my concern remains. 

If you raised a (for example) ```InvalidAttributeColumnForHitPolicyException``` we could i18n the message on the client by adding a command to ```CommandDrivenErrorCallback``` for ```InvalidAttributeColumnForHitPolicyException``` (just because ```IllegalArgumentException``` is used elsewhere doesn't make it right!)

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
```diff
@@ -91,8 +91,29 @@
         LIMITED_ENTRY
     }
```
manstis, line 93: Do you not think we should still override the ```add(..)```, ```addAll(..)``` (and other similar methods) on the "attribute columns" list implementation and use ```DecisionTableHitPolicyValidator``` to check someone (code or User) is not trying to add an inappropriate attribute for a given Hit Policy? Or are you planning on doing that type of validation elsewhere?

manstis: There might be one or two many uses of negatives in that sentence. I hope it makes sense!?!

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyValidator.java
Rikkola: i18n is a valid point for changing the exceptions. And I guess I do prefer more specific exceptions than illegal argument.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
Rikkola: I was just going to hide the attribute values in the add attribute popup and then throw `IllegalArgumentException` on compile time, but maybe it makes sense to also validate when attribute is added. That fails faster.
 
It would require a `do - undo` -action. I guess that fits into our plans of command based editing. For this I'll just do a simple solution targeting at adding the attribute. Not going to do the full command queue implementation.

Rikkola:
I'll add a more specific exception.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/GuidedDecisionTable52.java
manstis: @Rikkola Hehe, so I started off writing this comment asking why will we need do- undo- actions; then I thought a bit and now I think a bit more! If we add the column to the dtable and then use ```DecisionTableHitPolicyValidator``` then I see we'd need to be able to "undo" the operation if validation failed. However we could just check if adding a column of a particular type is valid before adding it to the table. You don't have any code for this atm; but something simple ```if(validator.canAddColumn( attribute ) ) { //Add it }``` and that method checks the HP and attribute.

Rikkola: Yep that is doable. Might have some other reuses too. For example the user can not add salience twice.

jomarko: Hi @Rikkola and @manstis . I would like to ask which of the options bellow will be implemented.

1. User can add manually attribute. The exception will be thrown during marshalling if the hit mode tries to add the same attribute.
2. User can add manually attributes. The exception is thrown immediately if the attribute would be added during marshalling.
3. Attributes added automatically added during marshalling are hidden from 'Add Attribute' dialogue.

I had a look also on the source code I have no suggestions.

manstis: @jomarko IMO Users should be prevented from adding Attributes in the UI required by certain Hit Policies (they should also be unable to remove them; but IDK whether Toni's electing to not show them - so they cannot be deleted; or disable their deletion). A check during the attempt to add/remove the column to/from the underlying collection could achieve this. The check during **marshalling** is, IMO, a safe-guard only and we should not need to marshall the model to DRL to validate the columns.

This is my opinion; but I'll let @Rikkola give his definitive answers :-)

Rikkola: Yes, going to do several. I imagine there are several layers on this.

1. We need to prevent adding the attributes in UI. This is so that UI users will not get confused.
2. Safe guard in marshalling. I would not be surprised if users were generating dtables and skipping UI. So for these users we fail in compile/view source.

At this stage I'm filling in the columns only when the DRL is generated. This is mainly because I am scared the implementation might change. This should change to actually adding the columns and making them disabled before final is released. Since if the code generation changes between 7.x and 8.x then the ksession functionality will change. Supporting several code generator versions will end up being almost impossible. While supporting a dtable with salience and activation-group column is easy.

Rikkola:
jenkins please retest my latest changes

Rikkola:
jenkins please retest

Rikkola:
jenkins please retest.
Rebased this can be reviewed and merged if all is ok.

Rikkola:
Or not. Getting some new exception with these changes. Hold the merge.

Rikkola:
Ok, this is good. Just had to refresh everything in Idea for SDM.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/validation/DecisionTableValidator.java
```diff
@@ -0,0 +1,66 @@
...
+import org.drools.workbench.models.guided.dtable.shared.model.AttributeCol52;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+
+public class DecisionTableValidator {
+
+    private final GuidedDecisionTable52 dtable;
+
+    private final Set<String> illegalAttributes;
+
+    public DecisionTableValidator( final GuidedDecisionTable52 dtable ) {
```
manstis, line 29: Should this be ```public``` if you're to use ```DecisionTableValidator.validate( dtable )```?

I see you also have a ```public``` instance method ```isValidToAdd(..)``` which would obviously need this constructor to be ```public``` so I wonder therefore whether it makes more sense to make ```validate(..)``` an instance method too?

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceRuleOrderHitPolicyTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private GuidedDecisionTable52 dtable;
+
+    @Before
+    public void setUp() throws
+                        Exception {
+        dtable = new GuidedDecisionTable52();
+        dtable.setTableName( "Rule order table" );
+
+        dtable.setHitPolicy( GuidedDecisionTable52.HitPolicy.RULE_ORDER );
+
```
manstis: Could you remove this flurry of empty lines?

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceUniqueHitPolicyTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.drools.workbench.models.guided.dtable.backend.TestUtil.*;
+
+public class GuidedDTDRLPersistenceUniqueHitPolicyTest {
+
+
```
manstis: Do you run code formatter? You seem to like white space - not that we have any policy around excessive use! ;-)

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/HitPolicyPersistenceTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.junit.Assert.*;
+
+public class HitPolicyPersistenceTest {
+
+    @Test
+    public void testDefault() throws
+                              Exception {
+        assertEquals( GuidedDecisionTable52.HitPolicy.NONE,
+                      new GuidedDecisionTable52().getHitPolicy() );
+
```
manstis: You just love white space... does it reflect the snow in Finland? ;-)

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/TestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class TestUtil {
+
+    public static void assertContainsLinesInOrder( final String text,
+                                                   final String... strings ) {
+
+        final Iterator<String> theseAreTheLinesYouAreLookingFor = Arrays.asList( strings )
+                .iterator();
+
+        String thisIsTheLineYouAreLookingFor = theseAreTheLinesYouAreLookingFor.next();
```
manstis, line 34: Made me think of [this](https://www.youtube.com/watch?v=532j-186xEQ)

manstis (commented):
@Rikkola Sweet. Just a couple of minor comments.

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceUniqueHitPolicyTest.java
Rikkola: Looks like the code formatter ignores empty lines for me. Checking if I need tick some box.

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/TestUtil.java
Rikkola: The force is strong with this one.

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/HitPolicyPersistenceTest.java
Rikkola: You know our work is calculated on commits and commit changes. The more lines I get in, the harder it looks like I work.

Rikkola: And not much snow here :(

triceo: Come to Brno, apparently we got all the winter this year.

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceUniqueHitPolicyTest.java
Rikkola: Looks like the formatter shrinks 2+ rows to just two rows.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/validation/DecisionTableValidator.java
Rikkola: Removed the static validate method to make using this class more clear.

Rikkola:
Removed a static `validate()` method and empty lines.

manstis (approved):
@Rikkola Looks good to me.. let me know if you want it merged (and as you say, it has no effect on anything at the moment... so I suspect you'll want it merged.. just clarifying).

PR #1039, opened 2016-12-16
DROOLS-1379: Add reproducer for cloning of OOPath

winklerm:
When two KieContainers for the same ReleaseId are created, the KieBase being built
is forced to deep clone, which fails with OOPath constraints due to UnsupportedOperationException
at org.drools.core.rule.constraint.XpathConstraint.writeExternal() [1].

[1] https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/rule/constraint/XpathConstraint.java#L138


mariofusco:
Fixed and merged by https://github.com/droolsjbpm/drools/pull/1040

PR #1046, opened 2017-01-02
[DROOLS-1390] make sure scanner state is always consistent

psiroky:
Before this change it was possible to "break" the consistency
of the scanner, in case the polling interval was very small (20ms).
It was because of the unsynchronized method run() in the ScanTask.
It was possible to schedule new run() method (by the Timer)
during the call to stop(). The run() would stop at scanNow()
(as that is synchronized), but would be executed after the stop()
method finished. So the scanner would be running one last time
even after it was stopped. This then lead to inconsistent scanner
status as the run() method changed it to RUNNING after the scanNow()
finished.

With this fix, the scanNow() + status=RUNNING is synchronized via
the scanner lock making sure it can not be called in the middle of
the stop() method.

There is no test associated with this as I was not able to figure out how to write a good test (without Thread.sleeps()). We would probably need something like Byteman to test this. If someone has an idea about to easily add a test, please let me know.

### DIFF
--- a/kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
+++ b/kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
@@ -68,7 +68,7 @@ public class KieRepositoryScannerImpl implements InternalKieScanner {
 
     private ArtifactResolver artifactResolver;
 
-    private Status status = Status.STARTING;
+    private volatile Status status = Status.STARTING;
 
     private KieScannerMBean mbean;
 
@@ -267,8 +267,14 @@ public class KieRepositoryScannerImpl implements InternalKieScanner {
 
     private class ScanTask extends TimerTask {
         public void run() {
-            scanNow();
-            status = Status.RUNNING;
+            synchronized (KieRepositoryScannerImpl.this) {
+                // don't scan if the scanner was already stopped! This would lead to inconsistent scanner behavior.
+                if (status == Status.STOPPED) {
+                    return;
+                }
+                scanNow();
+                status = Status.RUNNING;
+            }
         }
     }
 


kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
```diff
@@ -262,13 +262,28 @@ private void startScanTask(long pollingInterval) {
         status = Status.RUNNING;
         this.pollingInterval = pollingInterval;
         timer = new Timer(true);
-        timer.schedule(new ScanTask(), pollingInterval, pollingInterval);
+        timer.schedule(new ScanTask(this), pollingInterval, pollingInterval);
```
mariofusco: ScanTask is a non-static inner class, so it already has a reference to this and it is not necessary to pass it.

kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
```diff
@@ -262,13 +262,28 @@ private void startScanTask(long pollingInterval) {
...
+        private final Object scannerLock;
+
+        public ScanTask(Object scannerLock) {
+            this.scannerLock = scannerLock;
+        }
+
         public void run() {
-            scanNow();
-            status = Status.RUNNING;
+            synchronized (scannerLock) {
```
mariofusco: Avoid the scannerLock object and synchronize on KieRepositoryScannerImpl.this. I believe this will make also clearer that you're sync'ing on the same lock of the scanNow() method.

psiroky: Will do that, thanks. Originally I was thinking about passing in different lock object for testing purposes, but it's easy to create the test even in that case...

mariofusco:
If I remember correctly you have a failing test case using Thread.sleep. I know it is not ideal but I'd add it to this commit anyway, also because, even if using the sleep, it could reproduce the problem almost always.

psiroky:
The test case is likely dependent on my machine. I had to figure the proper sleep intervals by trial&error, that's why I don't want to use such a test as I don't really know how will that work on Jenkins...

PR #1047, opened 2017-01-02
[6.5.x][RHBPMS-4435] make sure scanner state is always consistent

psiroky:
Before this change it was possible to "break" the consistency
of the scanner, in case the polling interval was very small (20ms).
It was because of the unsynchronized method run() in the ScanTask.
It was possible to schedule new run() method (by the Timer)
during the call to stop(). The run() would stop at scanNow()
(as that is synchronized), but would be executed after the stop()
method finished. So the scanner would be running one last time
even after it was stopped. This then lead to inconsistent scanner
status as the run() method changed it to RUNNING after the scanNow()
finished.

With this fix, the scanNow() + status=RUNNING is synchronized via
the scanner lock making sure it can not be called in the middle of
the stop() method.


psiroky:
This is a backport of #1046.

PR #1049, opened 2017-01-02
[DROOLS-1392] fix parsing of boolean cells with poi 3.15 - extra hand…

mswiderski:
…ling for boolean cell type


mswiderski:
jenkins retest this

PR #1053, opened 2017-01-05
DROOLS-1399 Comparison of DRL changes should not consider spaces duri…

tarilabs:
…ng incremental compilation

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
@@ -24,6 +24,7 @@ import org.drools.compiler.lang.descr.GlobalDescr;
 import org.drools.compiler.lang.descr.PackageDescr;
 import org.drools.compiler.lang.descr.RuleDescr;
 import org.drools.core.io.impl.ByteArrayResource;
+import org.drools.core.util.StringUtils;
 import org.kie.api.io.ResourceType;
 import org.kie.internal.builder.ChangeType;
 import org.kie.internal.builder.ResourceChange;
@@ -70,8 +71,11 @@ public class ChangeSetBuilder {
                 byte[] ob = original.getBytes( file );
                 byte[] cb = currentJar.getBytes( file );
                 if( ! Arrays.equals( ob, cb ) ) {
-                    // parse the file to figure out the difference
-                    result.getChanges().put( file, diffResource( file, ob, cb ) );
+                    // check that: (NOT drl file) OR (NOT equalsIgnoringSpaces)
+                    if ( ! ( file.endsWith(".drl") && StringUtils.equalsIgnoreSpaces(new String(ob), new String(cb)) ) ) {
+                        // parse the file to figure out the difference
+                        result.getChanges().put( file, diffResource( file, ob, cb ) );
+                    }
                 }
             } else {
                 // file was added
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
@@ -72,7 +72,7 @@ public class ChangeSetBuilder {
                 byte[] cb = currentJar.getBytes( file );
                 if( ! Arrays.equals( ob, cb ) ) {
                     // check that: (NOT drl file) OR (NOT equalsIgnoringSpaces)
-                    if ( ! ( file.endsWith(".drl") && StringUtils.codeAwareEqualsIgnoreSpaces(new String(ob), new String(cb)) ) ) {
+                    if ( !ResourceType.DRL.matchesExtension(file) || !StringUtils.codeAwareEqualsIgnoreSpaces(new String(ob), new String(cb)) ) {
                         // parse the file to figure out the difference
                         result.getChanges().put( file, diffResource( file, ob, cb ) );
                     }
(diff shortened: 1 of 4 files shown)


tarilabs:
```
	Test Result (2 failures / +2)
org.drools.compiler.integrationtests.KieContainerTest.testIncrementalCompilationSynchronization
org.drools.compiler.integrationtests.TimerAndCalendarTest.testHaltAfterSomeTimeThenRestartButNoLongerHolding
```
Logged to 8ball-report.

Jenkins, test again :8ball:

drools-core/src/main/java/org/drools/core/util/StringUtils.java
```diff
@@ -999,31 +999,62 @@ public static int skipBlanks(String string, int start) {
...
+        if ( in2.length() == 0 ) {
+            in2 = " ";
+        }
+        
+        int idx1 = 0; boolean quoted1 = false;
+        int idx2 = 0; boolean quoted2 = false;
+        boolean equals = true;
+        while ( idx1 < in1.length() && idx2 < in2.length() ) {
+            
+            equals &= in1.charAt(idx1) == in2.charAt(idx2);
```
mariofusco: I think this will fail if one of the 2 strings starts with a space while the other doesn't. The trivial way to fix this would be move this statement after you have skipped spaces with the 2 indexes.

drools-core/src/main/java/org/drools/core/util/StringUtils.java
```diff
@@ -999,31 +999,62 @@ public static int skipBlanks(String string, int start) {
...
+            if (!equals) {
```
mariofusco: If you move the former statement at the bottom of the while loop and add 'equals' to the while condition you can get rid of this.

drools-core/src/main/java/org/drools/core/util/StringUtils.java
```diff
@@ -999,31 +999,62 @@ public static int skipBlanks(String string, int start) {
...
+            }
+            
+            if ( in1.charAt(idx1) == '"' ) {
+                if (idx1 == 0 || in1.charAt(idx1-1) != '\\') {
+                    quoted1 = !quoted1;
+                }
+            }
+            if ( in2.charAt(idx2) == '"' ) {
+                if (idx2 == 0 || in2.charAt(idx2-1) != '\\') {
+                    quoted2 = !quoted2;
```
mariofusco: These 2 if blocks look identical, maybe you can avoid repetition refactoring them in a private method returning a boolean that is the new value for quoted?

drools-core/src/main/java/org/drools/core/util/StringUtils.java
```diff
@@ -999,31 +999,62 @@ public static int skipBlanks(String string, int start) {
...
-                sb.append(charAt);
+            idx1++;
+            idx2++;
+            
+            while ( idx1 < in1.length() && !quoted1 && isWhitespace(in1.charAt(idx1)) ) {
+                idx1++;
             }
+            while ( idx2 < in2.length() && !quoted2 && isWhitespace(in2.charAt(idx2)) ) {
+                idx2++;
+            }
```
mariofusco: Same as before, encapsulate these loop in a method returning an int that is the new value for the index.

drools-core/src/main/java/org/drools/core/util/StringUtils.java
```diff
@@ -999,31 +999,62 @@ public static int skipBlanks(String string, int start) {
...
-        return sb.toString();
-    }
-    
-    /**
-     * Compares two string being equals ignoring whitespaces, but preserving whitespace between double-quotes
-     */
-    public static boolean codeAwareEqualsIgnoreSpaces(String s1, String s2) {
-        return codeAwareRemoveWhitespaceChars(s1).equals(codeAwareRemoveWhitespaceChars(s2));
+        // considered equals if equals check succeeded and both indexes reached end of respective string.
+        return equals && idx1 == in1.length() && idx2 == in2.length();
```
mariofusco: Is it necessary to also check the lengths? I'm not sure why.

drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
```diff
@@ -70,8 +71,11 @@ public KieJarChangeSet build( InternalKieModule original, InternalKieModule curr
                 byte[] ob = original.getBytes( file );
                 byte[] cb = currentJar.getBytes( file );
                 if( ! Arrays.equals( ob, cb ) ) {
-                    // parse the file to figure out the difference
-                    result.getChanges().put( file, diffResource( file, ob, cb ) );
+                    // check that: (NOT drl file) OR (NOT equalsIgnoringSpaces)
+                    if ( !ResourceType.DRL.matchesExtension(file) || !StringUtils.codeAwareEqualsIgnoreSpaces(new String(ob), new String(cb)) ) {
```
tarilabs, line 75: @etirelli at file-level comparison, check that DRL file is changed IFF is semantically changed. Other type of files, will keep the original behavior.

drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java
```diff
@@ -169,10 +173,10 @@ public String getName(GlobalDescr descr) {
                     found = true;
                     it.remove();
 
-                    // using byte[] comparison because using the descriptor equals() method
+                    // using codeAwareEqualsIgnoreSpaces comparison because using the descriptor equals() method
                     // is brittle and heavier than iterating an array
-                    if ( !segmentEquals(ob, ord.getStartCharacter(), ord.getEndCharacter(), cb, crd.getStartCharacter(), crd.getEndCharacter() ) ||
-                         (type == ResourceChange.Type.RULE && updatedRules.contains( ( (RuleDescr) crd ).getParentName() )) ) {
+                    if ( !StringUtils.codeAwareEqualsIgnoreSpaces(new String(Arrays.copyOfRange(ob, ord.getStartCharacter(), ord.getEndCharacter())),new String(Arrays.copyOfRange(cb, crd.getStartCharacter(), crd.getEndCharacter())) )
```
tarilabs, line 178: @etirelli a Rule, a Function, a Global is no longer compared on their byte[] representation, but with codeaware semantics (e.g.: ignore whitespaces).

This is also needed in case the Line 75 detect a genuine change, like a new rule is added, but at this stage another, already existing, rule comparison has been modified only in whitespace. As @mariofusco noted, it is important to check the rule is added to the list of changes IFF is now semantically changed, a whitespace does no longer imply the rule is changed.

This second check is important. Otherwise if keeping only Line 75 the behavior would have differed in the case the "whitespaced" rule was alone or in a file with a genuine-modified rule.

tarilabs: Consider for instance this scenario:
```
file_v1.drl : rule B
file_v2.drl : rule B modified only in whitespace
```
in this scenario, only Line 75 would have sufficed.

In this other scenario:
```
file_v1.drl : rule B
file_v2.drl : rule A, rule B modified only in whitespace
```
in this scenario, having only Line 75 would detect file genuinely changed (at line 75) but later on here resulting in asserting rule B as changed too (because previous way comparing byte[] of the rule B in their version). This is not what we want. Therefore, it is important to extend the code-aware semantic comparison also at the level of the Rule, Function, Global, as per this line 178.

drools-core/src/main/java/org/drools/core/util/StringUtils.java
etirelli: also, looking at the DRL6Lexer.g, DRL strings can be single or double quoted, so you have to handle single quoted strings here as well.

etirelli:
@tarilabs the overall idea seems fine, just fix the handling of single quoted strings and it should be good.

etirelli:
looks good now. +1

PR #1056, opened 2017-01-11
[DROOLS-1402] fix incremental compilation when a java class is modified

mariofusco:


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/1058

PR #1060, opened 2017-01-12
JBPM-5541 - Case file conditional event fails other cases

mswiderski:
please do not merge until approved

@krisv could you please take a look... more in jbpm part actually


mswiderski:
closing, a replacement will be submitted

PR #1061, opened 2017-01-13
[DROOLS-1406] fix second incremental compilation during fireUntilHalt

mariofusco:


mariofusco:
Replace by https://github.com/droolsjbpm/drools/pull/1062

PR #1064, opened 2017-01-13
[DROOLS-1405] avoid NPE during incremental compilation when more than…

mariofusco:
… one rule extends another rule


mariofusco:
ok to test

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1197,7 +1197,7 @@ private void sortRulesByDependency(PackageDescr packageDescr) {
         Map<String, List<RuleDescr>> children = new HashMap<String, List<RuleDescr>>();
         LinkedHashMap<String, RuleDescr> sorted = new LinkedHashMap<String, RuleDescr>();
         List<RuleDescr> queries = new ArrayList<RuleDescr>();
-        List<String> compiledRules = new ArrayList<String>();
+        Set<String> compiledRules = new HashSet<String>();
```
tarilabs, line 1200: +1

manstis:
Your test has an incorrect comment in it (says something like "add file with error... expect 1 error"). This is not the case.

PR #1065, opened 2017-01-16
RHBPMS-4435 Fixes behaviour when interval scanning is stopped when kieScanner.scanNow() is called

baldimir:
When scanNow() was called after interval scanning was started with kieScanner.start(),
the interval scanning was stopped, because "status" of kieScanner was set to STOPPED.
Adds reproducer for mentioned issue and proposed fix.

### DIFF
--- a/kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
+++ b/kie-ci/src/main/java/org/kie/scanner/KieRepositoryScannerImpl.java
@@ -282,11 +282,13 @@ public class KieRepositoryScannerImpl implements InternalKieScanner {
         if (getStatus() == Status.SHUTDOWN ) {
             throw new IllegalStateException("The scanner was already shut down and can no longer be used.");
         }
+        // Polling can be started so remember the original state.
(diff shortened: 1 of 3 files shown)


psiroky:
Looks good, thanks @baldimir.

Two questions regarding the test: (1) is there any chance to remove the Thread.sleep()?, (2) should we stop/shutdown the scanner after the test? It seems the background thread will be running until the JVM shutdown, which is not ideal.

baldimir:
After discussion with @psiroky I will add new commit with changes.

baldimir:
@psiroky, requested changes added.

psiroky:
Thanks @baldimir! This is what I had originally in mind https://github.com/droolsjbpm/drools/pull/1065/commits/ba546362b139d2e2cfce6736d48b49b2577d2efd. Please let me know what you think.

baldimir:
@psiroky thanks, that will do, I added the commit also to https://github.com/droolsjbpm/drools/pull/1066

psiroky (approved):

PR #1066, opened 2017-01-16
RHBPMS-4435 Fixes behaviour when interval scanning is stopped when kieScanner.scanNow() is called

baldimir:
When scanNow() was called after interval scanning was started with kieScanner.start(),
the interval scanning was stopped, because "status" of kieScanner was set to STOPPED.
Adds reproducer for mentioned issue and proposed fix.

(cherry picked from commit cda7563)

Master PR: https://github.com/droolsjbpm/drools/pull/1065


baldimir:
After discussion with @psiroky I will add new comment with changes.

baldimir:
@psiroky, requested changes added.

PR #1069, opened 2017-01-18
[DROOLS-936] fix dependency issues introducing duplicated classes

psiroky:
* remove com.sun.codemodel as it contains same classes as jaxb-xjc.
   jaxb-xjc is basically uberjar as it bundles other classes as well. However,
   there does not seem to be a way to get rid of it, as there is no proper
   replacement. Drools code itself should get rid of that dependency
   alltogether and use just public jaxb APIs

 * exclude commons-logging as the jars should not depend on any
   logging impl

 * remove javax.stream:stax-api as the same classes are provided by
   stax-api:stax-api (and other as well)

 * ignoring check for kie-ci-osg in drools-distribution as it is
   uberjar and the duplicated classes are expected there

Should be merged together with https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/363

### DIFF
--- a/drools-beliefs/pom.xml
+++ b/drools-beliefs/pom.xml
@@ -63,9 +63,9 @@
     </dependency>
 
     <dependency>
-      <groupId>javax.enterprise</groupId>
-      <artifactId>cdi-api</artifactId>
-      <scope>provided</scope><!-- HACK for OSGi: should be <optional>true</optional> instead -->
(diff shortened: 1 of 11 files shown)


psiroky:
Could someone please take a look and review these changes?

mariofusco:
+1

PR #1070, opened 2017-01-19
[DROOLS-1411] fix consraint using declaration bound to xpath + compil…

mariofusco:
…e-time error using more than one xpath in same pattern

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
@@ -735,11 +735,30 @@ public class PatternBuilder
 
         List<BaseDescr> initialDescrs = new ArrayList<BaseDescr>( descr.getDescrs() );
         for ( BaseDescr d : initialDescrs ) {
-            Constraint constraint = isXPathDescr(d) ?
+            boolean isXPath = isXPathDescr(d);
+            if (isXPath && pattern.hasXPath()) {
+                context.addError( new DescrBuildError( context.getParentDescr(),
+                                                       descr,
+                                                       null,
+                                                       "More than a single oopath constraint is not allowed in the same pattern" ) );
+                return constraints;
+            }
+            Constraint constraint = isXPath ?
                                     buildXPathDescr(context, patternDescr, pattern, d, mvelCtx) :
                                     buildCcdDescr(context, patternDescr, pattern, d, descr, mvelCtx);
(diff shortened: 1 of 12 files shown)


drools-compiler/src/test/java/org/drools/compiler/TestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.kie.api.KieServices;
+import org.kie.api.builder.KieFileSystem;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TestUtil {
+
+    public static void assertDrlHasCompilationError( String str, int errorNr ) {
```
tarilabs, line 27: Just thinking.. only because we need to backport, otherwise we could have used `Predicate<Results>` instead of a fixed `int errorNr` ?

drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
```diff
@@ -735,11 +735,30 @@ protected void build( final RuleBuildContext context,
...
+                                                       null,
+                                                       "More than a single oopath constraint is not allowed in the same pattern" ) );
+                return constraints;
+            }
+            Constraint constraint = isXPath ?
                                     buildXPathDescr(context, patternDescr, pattern, d, mvelCtx) :
                                     buildCcdDescr(context, patternDescr, pattern, d, descr, mvelCtx);
             if (constraint != null) {
-                constraints.add(constraint);
+                Declaration declFromXpath = getDeclarationCorrespondingToXpath( pattern, isXPath, constraint );
```
tarilabs: When we updated the method name, I think we missed to change as well the variable name `declFromXpath` to `declCorrXpath` too, here and..

drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
```diff
@@ -765,6 +784,22 @@ protected void build( final RuleBuildContext context,
         return constraints;
     }
 
+    private Declaration getDeclarationCorrespondingToXpath( Pattern pattern, boolean isXPath, Constraint constraint ) {
+        Declaration declFromXpath = null;
```
tarilabs: ..and change as well the variable name `declFromXpath` to `declCorrXpath` also here.

tarilabs:
+1

PR #1074, opened 2017-01-27
Reproducer (not fix) for DROOLS-1420 (int) ($p * 1_000.0) throws a misleading exception

ge0ffrey:
See https://issues.jboss.org/browse/DROOLS-1420


mariofusco:
Fixed by https://github.com/droolsjbpm/drools/commit/d70058bb6ea7516a0054ef5063d94d173cda479f

PR #1075, opened 2017-01-27
[RHBRMS-2739][DROOLS-1403] avoid stack overflow on kbase serialization

mariofusco:


mariofusco:
ok to test

mariofusco:
ok to test

PR #1062, opened 2017-01-13
[DROOLS-1406] fix second incremental compilation during fireUntilHalt

mariofusco:

PR #1080, opened 2017-02-02
[DROOLS-1425] avoid deadlock when an action is enqueued during a fire…

mariofusco:
…UntilHalt

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
+++ b/drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
@@ -16,6 +16,18 @@
 
 package org.drools.core.common;
 
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1152,8 +1152,10 @@ public boolean isFiring() {
 
     @Override
     public void executeTask( ExecutableEntry executable ) {
-        if (!executionStateMachine.toExecuteTask( executable )) {
-            return;
+        synchronized (propagationList) {
```
tarilabs, line 1155: As reviewed, during DROOLS-1406 (#1062) the `FireUntilHaltRestHandler` acquires the lock on propagationList and then if transitioning with `inactiveOnFireUntilHalt`  attempt to acquire the lock on stateMachine.

While original this method called `toExecuteTask` and was acquiring the lock on the stateMachine, and if to enqueue, in turn would have called `executable.enqueue()` -> `<anon>.doAdd()` -> `SynchronizedPropagationList.internalAddEntry` where internalAddEntry is synchronized method, hence attempt to acquire the lock on the propagationList. In short, acquiring the lock on the stateMachine, and then attempt to acquire the lock on the propagationList.

Potential deadlock described above.

The solution is to re-align the ordering of the lock acquired, as per this line.

tarilabs (approved):
+1

PR #1082, opened 2017-02-03
[DROOLS-1425] avoid deadlock when disposing during a fireUntilHalt

mariofusco:


mariofusco:
ok to test

drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1123,24 +1123,32 @@ public PropagationEntry handleRest(DefaultAgenda agenda, boolean isInternalFire)
         class FireUntilHaltRestHandler implements RestHandler {
             @Override
             public PropagationEntry handleRest(DefaultAgenda agenda, boolean isInternalFire) {
+                boolean deactivated = false;
```
tarilabs, line 1126: This modification does not "stack" the locks anymore (was: hold propagationList lock and then stateMachine lock) and now it does lock in the following order: stateMachine, propagationList, stateMachine, but these locks are "pipelined" not "stacked".

This has also the advantage you does not need any longer the additional `synchronized(propagationList)` on the start of the two methods `executeTask( ExecutableEntry executable )` , `dispose(InternalWorkingMemory wm)` below

tarilabs (approved):
+1

PR #1084, opened 2017-02-07
GUVNOR-2854i : Guided Decision Table V&V - Hit policy: Resolved

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2854

Code generation for Resolved Hit policy. No way to set this up in master at the moment. Just getting it in like the other hit policies.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyEnhancer.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyEnhancer.java
@@ -20,6 +20,7 @@ import java.util.List;
 import org.drools.workbench.models.guided.dtable.shared.model.AttributeCol52;
 import org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52;
 import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.drools.workbench.models.guided.dtable.shared.model.MetadataCol52;
 import org.drools.workbench.models.guided.dtable.shared.validation.DecisionTableValidator;
 
 /**
@@ -39,11 +40,14 @@ public class DecisionTableHitPolicyEnhancer {
     }
 
     private GuidedDecisionTable52 enhance() {
-
         new DecisionTableValidator( dtable ).validate();
 
         switch ( dtable.getHitPolicy() ) {
 
+            case RESOLVED_HIT:
+                addActivationGroup( "resolved-hit-policy-group" );
+                addSalienceByMetadata();
+                break;
             case UNIQUE_HIT:
                 addActivationGroup( "unique-hit-policy-group" );
                 break;
@@ -57,12 +61,96 @@ public class DecisionTableHitPolicyEnhancer {
                 break;
 
         }
-
         return dtable;
     }
 
+    private void addSalienceByMetadata() {
+
+        final int priorityColumnIndex = getResolvedHitMetadataColumnIndex();
+
+        if ( priorityColumnIndex >= 0 ) {
+            final RowPriorityResolver rowPriorityResolver = getRowPriorityResolver( priorityColumnIndex );
+            final int indexOfSalienceColumn = makeSalienceColumn();
+            final RowPriorities priorities = rowPriorityResolver.getPriorityRelations();
+
+            for ( final RowNumber rowNumber : priorities.getRowNumbers() ) {
+                dtable.getData()
+                        .get( rowNumber.getRowIndex() )
+                        .get( indexOfSalienceColumn )
+                        .setNumericValue( priorities.getSalience( rowNumber )
+                                                  .getSalience() );
+            }
+        }
+    }
+
+    private RowPriorityResolver getRowPriorityResolver( final int priorityColumnIndex ) {
+
+        final int rowNumberIndex = getRowNumberColumnIndex();
+
+        final RowPriorityResolver rowPriorityResolver = new RowPriorityResolver();
+
+        for ( final List<DTCellValue52> row : dtable.getData() ) {
+
+            final int rowNumber = row.get( rowNumberIndex )
+                    .getNumericValue()
+                    .intValue();
+
+            final int priorityOver = getPriorityOver( priorityColumnIndex,
+                                                      row );
+
+            rowPriorityResolver.set( rowNumber,
+                                     priorityOver );
+        }
+
+        return rowPriorityResolver;
+    }
+
+    private int makeSalienceColumn() {
+        final AttributeCol52 attributeCol = new AttributeCol52();
+        attributeCol.setAttribute( "salience" );
+        dtable.getAttributeCols()
+                .add( attributeCol );
+
+        int columnIndex = dtable.getExpandedColumns()
+                .indexOf( attributeCol );
+
+        for ( final List<DTCellValue52> row : dtable.getData() ) {
+            row.add( columnIndex,
+                     new DTCellValue52( 0 ) );
+        }
+
+        return columnIndex;
+    }
+
+    private int getPriorityOver( final int priorityColumnIndex,
+                                 final List<DTCellValue52> row ) {
+        final String stringValue = row.get( priorityColumnIndex )
+                .getStringValue();
+        if ( stringValue == null || stringValue.trim()
+                .equals( "" ) ) {
+            return 0;
+        } else {
+            return Integer.parseInt( stringValue );
+        }
+    }
+
+    private int getResolvedHitMetadataColumnIndex() {
+        int columnIndex = -1;
+        for ( final MetadataCol52 metadataCol52 : dtable.getMetadataCols() ) {
+            if ( GuidedDecisionTable52.HitPolicy.RESOLVED_HIT_METADATA_NAME.equals( metadataCol52.getMetadata() ) ) {
+                columnIndex = dtable.getExpandedColumns()
+                        .indexOf( metadataCol52 );
+            }
+        }
+        return columnIndex;
+    }
+
+    private int getRowNumberColumnIndex() {
+        return dtable.getExpandedColumns()
+                .indexOf( dtable.getRowNumberCol() );
+    }
+
     private void addActivationGroup( final String activationGroupType ) {
-        // Add Column
         final AttributeCol52 attributeCol52 = new AttributeCol52();
         attributeCol52.setAttribute( GuidedDecisionTable52.ACTIVATION_GROUP_ATTR );
         dtable.getAttributeCols()
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Over.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Over.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.drools.workbench.models.guided.dtable.shared.hitpolicy;
+
+/**
+ * The row number used in this class is the actual row number declared in the dtable data.
+ * Not the row index.
+ * <p>
+ * Priority is declared with Row X has priority over Row Y.
+ * This Over class contains the row number of Y.
+ */
+class Over
+        implements Comparable<Over> {
+
+    private Integer over;
+
+    public Over(final Integer value) {
+        over = value;
+    }
+
+    public Over(final RowNumber rowNumber) {
+        this(rowNumber.getRowNumber());
+    }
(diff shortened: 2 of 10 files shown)


Rikkola:
@manstis @jomarko You two might be interested of this.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyEnhancer.java
```diff
@@ -57,12 +61,96 @@ private GuidedDecisionTable52 enhance() {
...
+        final int priorityColumnIndex = getResolvedHitMetadataColumnIndex();
+
+        if ( priorityColumnIndex >= 0 ) {
+            final RowPriorityResolver rowPriorityResolver = getRowPriorityResolver( priorityColumnIndex );
+            final int indexOfSalienceColumn = makeSalienceColumn();
+            final RowPriorities priorities = rowPriorityResolver.getPriorityRelations();
+
+            for ( final RowNumber rowNumber : priorities.getRowNumbers() ) {
+                dtable.getData()
+                        .get( rowNumber.getRowNumber() - 1 )
```
manstis: Just a comment... everywhere else (go on, prove me wrong!) "row index" and "column index" are zero-based, whereas ```RowNumber``` is one-based. Any reason you want to be different? ;-)

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/RowPriorityResolver.java
```diff
@@ -0,0 +1,74 @@
...
+
+        if ( rowPriorities == null ) {
+
+            rowPriorities = new RowPriorities( priorityRelations );
+
+            for ( final RowNumber rowNumber : priorityRelations.getRowNumbers() ) {
+
+                final Over over = priorityRelations.getOver( rowNumber );
+
+                if ( over.getOver() != 0 ) {
```
manstis: I guess this would need to move to some other "magic" value if ```RowNumber``` became zero-based? ```Optional<RowNumber>``` or ```Optional<Integer>``` I assume.. anyway; it is just an observation re: one-based... Perhaps a simple JavaDoc comment saying ```RowNumber``` and ```Over``` are one-based is enough?

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistenceResolvedHitPolicyTest.java
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
```
manstis: 2017

manstis (commented):
Sweet @Rikkola job well done. Minor comment re: one-based index; otherwise convince me it's OK and can merge.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyEnhancer.java
jomarko: Row numbers are one-based in the UI, what could be more comfortable for less technical users. However I am not against zero-based row numbers. I wrote just my observation.

Rikkola: This could really be either, I don't have that strong opinion about it. Just went with the "real" row number so that a row number is indeed the actual row number, but if we start from zero in other locations then it is better to change this to start from 0 too.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/RowPriorityResolver.java
Rikkola: Well -1 is the common "no selected value" used in Java. I'll go with that.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/DecisionTableHitPolicyEnhancer.java
manstis: @jomarko @Rikkola I don't have a strong opinion either; but ```RowNumber``` is just internal for the "Decision Table enrichment". IIUC the "row number" entered by Users in the UI to override other rules would remain one-based. They're only then used internally to determine salience values... Just let me know when you want this merged.. like I say, I didn't have a strong opinion - ```getData().get( rowNumber.getRowNumber()-1)``` just seemed "different" - and "different" isn't always bad :-)

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/RowPriorityResolverTest.java
```diff
@@ -0,0 +1,193 @@
...
+        assertEquals( 1,
+                      priorities.getSalience( new RowNumber( 2 ) ).getSalience() );
+    }
+
+    @Test
+    public void simple3Row() throws
+                             Exception {
+
+        rowPriorityResolver.set( 1,
+                                 0 );
```
jomarko: Hi @Rikkola , according to method `DecisionTableHitPolicyEnhancer.java .getPriorityOver( index , row )`, is this zero saying that Row 1 has not priority over any other row? If it is the case shouldn't be the salience values like:
* Row 1 : salience 0
* Row 2 : salience 0
* Row 3 : salience 1

Am I right with this, If Row A has priority over Row B, then Row A salience is bigger than Row B salience?

Rikkola: Well zero means there is no line below the line that has priority over it (I'm sure this sentence cleared things up).

The engine sets priority by row order by default. Meaning Row 1 > Row 2 > Row 3. Now when we give row 3 priority over 2. We still need to remember that row 1 also has priority over 2. This is why the code raises the salience on the row above 2 here.

jomarko: Well, here we came to the situation, when the combining zero-based and one-based indexing causes hard understanding (from my site).  The code `rowPriorityResolver.set( 1, 0 );` causes insertion into collections `PriorityRelations.priorities` and `PriorityRelations.rows`. In these two collections I would expect just `RowNumber`s and `Over`s holding just real number of rows. It was hardly to understand for me that `RowNumber`s or `Over`s holding zero have special meaning.

However, if the zero has meaning you described, should the test that doesn't call `rowPriorityResolver.set( [any], [any] );` return salience like this (assume you have three rows in the table)?
* Row 1 : salience 2 [3]
* Row 2 : salience 1 [2]
* Row 3 : salience 0 [1]

Rikkola: Good point. I just changed the default to -1 and row numbers to row index. It gets quite hard to understand. Using the real row number is easier to read, in code we even wrap it with RowNumber, not RowIndex.

Rikkola:
@manstis Ok. I kept the actual row number for readability. I did change the getRowNumber() -1 to getRowIndex(). That still returns the same value, but is less magical.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/PriorityRelations.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private void addRowNumberOver( final Over over,
+                                   final RowNumber number ) {
+        priorities.put( number,
+                        over );
+    }
+
+    private void addOverRowNumber( final Over over,
+                                   final RowNumber number ) {
+        if ( !rows.containsValue( over ) ) {
```
jomarko: @Rikkola please use `containsKey` instead of `containsValue`. Thanks.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/PriorityRelations.java
```diff
@@ -0,0 +1,78 @@
...
+
+    Over getOver( final RowNumber rowNumber ) {
+        return priorities.get( rowNumber );
+    }
+
+    boolean containsPriorityFor( final Over over ) {
+        return rows.containsKey( over );
+    }
+
+    Set<RowNumber> getRowsThatHavePrirityOver( final Over over ) {
```
jomarko: small typo mistake here

Rikkola:
And found a bug. The merge can wait.

jomarko:
@Rikkola here is link for commit in my fork. You can find there tests which I tried locally. It seems to me that in added test cases is the priority (salience) computed correctly. Hope this helps.

https://github.com/jomarko/drools/commit/568d571db43ca11a2695c65be301383366788466

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Over.java
```diff
@@ -0,0 +1,67 @@
...
+        over = null;
+    }
+
+    public Integer getOver() {
+        return over;
+    }
+
+    @Override
+    public int compareTo( final Over o ) {
+        return over.compareTo( o.getOver() );
```
manstis: ```over``` can be ```null``` here.... can't it?!

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/RowNumber.java
```diff
@@ -0,0 +1,63 @@
...
+        return rowNumber;
+    }
+
+    public int getRowIndex() {
+        return rowNumber - 1;
+    }
+
+    @Override
+    public int compareTo( final RowNumber o ) {
+        return rowNumber.compareTo( o.getRowNumber() );
```
manstis: ```rowNumber``` can be ```null``` here.

manstis (commented):
@jomarko I found a couple of silly edge cases; but could you pass review of this PR for @Rikkola please (you spotted the original error!)

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Salience.java
```diff
@@ -0,0 +1,34 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.drools.workbench.models.guided.dtable.shared.hitpolicy;
+
+/**
+ * Wrapper for salience.
+ * Salience is used on run time to make sure the rule priority order defined in the editor is respected.
+ */
+class Salience {
```
jomarko, line 22: This class is used in similar situations like `RowNumber` and `Over`, I mean it is stored as a value in a map. I know that currently it is not used as key in the map, however it may be good to implement `equals` and `hashCode` (maybe `compareTo`) also in this class. Or, have you omitted implementation from some special reason?

jomarko (commented):
@Rikkola I really like as you prevented never-ending loops with throwing exception in cases that Row with lower row number is prioritized over Row with higher row number. I tried locally also some additional testcases and I am definitely for merging. If you are for I can create separate PR with tests which I tried locally.

I would just wait for your reaction on @manstis and my comments and I think the PR can be merged then.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Over.java
Rikkola: Best to play it safe and protect against it.

Rikkola:
@jomarko Please do the PR for the tests. We can use any coverage possible with these. We even discussed with @manstis some time ago that we should have tests that take a .gdst file, generate the DRL from it and then run the rules and verify the results are correct, but haven't yet had the time to make them.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/hitpolicy/Salience.java
Rikkola: Probably worth it to implement `equals` and `hashcode`, as if the implementation changes it is easy to forget that they need to be made. A missing `compareTo` gives an exception so I'll leave that out for now.

Rikkola:
@jomarko @manstis Ok. Changes done, let's see if the tests pass. Merging this does not break anything up stream.

jomarko:
+1 from me

PR #1090, opened 2017-02-09
[6.5.x] RHBPMS-6421 - Rollback transaction with Status.STATUS_ROLLEDBACK befo…

martinweiler:
…re transaction start

### DIFF
--- a/drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
+++ b/drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
@@ -200,7 +200,16 @@ public class JtaTransactionManager
     }
 
     public boolean begin() {
-        if ( getStatus() == TransactionManager.STATUS_NO_TRANSACTION ) {
+        // RHBPMS-4621 - transaction can be marked as rollback
+        // and still be associated with current thread
+        // See WFLY-4327
+        int status = getStatus();
+        if ( status == TransactionManager.STATUS_ROLLEDBACK ) {
+            logger.debug("Cleanup of transaction that has been rolled back previously");
+            rollback(true);
+            status = getStatus();
+        }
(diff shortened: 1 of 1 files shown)


drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
```diff
@@ -200,6 +200,13 @@ protected UserTransaction getUt() {
     }
 
     public boolean begin() {
+        // RHBPMS-6421 - transaction can be marked as rollback
+        // and still be associated with current thread
+        // See WFLY-4327
+        if ( getStatus() == TransactionManager.STATUS_ROLLEDBACK ) {
+            logger.info("Cleanup of transaction that has been rolled back previously");
+            rollback(true);
```
mswiderski: @martinweiler can we refactor this to call getStatus only once in majority of cases and then refresh it after the rollback here? That should eliminate most of the duplicated calls to getStatus for each transaction start

tkobayas: @martinweiler btw, you typo the JIRA ID RHBPMS-4621 :)

mariofusco: @martinweiler Any reason why this pull request is against branch 6.5.x? We'd need one against master first (provided that the problem you're fixing is present also there).

mswiderski:
@martinweiler could you please squash commits and update the commit message as there isa type in jira number?

martinweiler:
@mswiderski I have squashed the commits and updated the commit message.

mswiderski:
@martinweiler thank you, just got info that this can't be merged for 6.4.2 patch so will need to stay here for a while, I assigned this to me so once we have acks for it, I'll merge. In meantime will prepare PR for master

PR #1091, opened 2017-02-09
[DROOLS-1426] remove CDI extension from drools-compiler

mariofusco:


drools-examples-api/ruleunit/target/surefire-reports/TEST-org.drools.example.api.reactivekiesession.RuleUnitExampleTest.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mswiderski: is this here by accident?

mswiderski: and few more ...

mswiderski (commented):
there is a need to add drools-cdi dependency to https://github.com/droolsjbpm/jbpm/tree/master/jbpm-services/jbpm-services-cdi as it uses CDIHelper class that was moved + reorganisation of imports

mswiderski (approved):
looks good to me

drools-cdi/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+  </dependencies>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <configuration>
+            <excludeFilterFile>${project.basedir}/src/main/findbugs/findbugs-exclude.xml</excludeFilterFile>
```
psiroky, line 126: This needs to be removed as the file does not exist.

PR #1095, opened 2017-02-13
[DROOLS-1369] avoid calculating property reactive bitmasks when using…

mariofusco:
… update with property names on non-propreactive objects

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationCache.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationCache.java
@@ -15,6 +15,18 @@
 
 package org.drools.compiler.builder.impl;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
 import org.drools.compiler.compiler.PackageRegistry;
 import org.drools.compiler.compiler.TypeDeclarationError;
 import org.drools.compiler.lang.descr.Annotated;
@@ -27,21 +39,8 @@ import org.drools.core.rule.TypeDeclaration;
 import org.drools.core.util.ClassUtils;
 import org.kie.api.definition.type.Modifies;
 import org.kie.api.definition.type.Position;
-import org.kie.api.definition.type.Role;
 import org.kie.api.io.Resource;
 import org.kie.api.runtime.rule.Match;
-
-import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
 
 import static org.drools.compiler.builder.impl.TypeDeclarationConfigurator.processTypeAnnotations;
 import static org.drools.core.util.BitMaskUtil.isSet;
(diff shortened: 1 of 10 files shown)


drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
```diff
@@ -307,6 +307,16 @@ private ObjectType getObjectType(RuleBuildContext context, PatternDescr patternD
...
+        TypeDeclaration typeDeclaration = getTypeDeclaration( context, userProvidedClass );
+
+        if (typeDeclaration != null) {
+            return typeDeclaration.getRole() == Role.Type.EVENT;
+        }
+        Role role = userProvidedClass.getAnnotation( Role.class );
+        return role != null && role.value() == Role.Type.EVENT;
+    }
+
+    private TypeDeclaration getTypeDeclaration( RuleBuildContext context, Class<?> userProvidedClass ) {
```
tarilabs, line 319: extracted to private method.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertySpecificTest.java
```diff
@@ -41,16 +50,24 @@
...
-import java.util.concurrent.atomic.AtomicInteger;
-
-import static org.drools.core.reteoo.PropertySpecificUtil.*;
+import static org.drools.core.reteoo.PropertySpecificUtil.calculateNegativeMask;
+import static org.drools.core.reteoo.PropertySpecificUtil.calculatePositiveMask;
 
 public class PropertySpecificTest extends CommonTestMethodBase {
-    
+
+    public static List<String> getSettableProperties(InternalWorkingMemory workingMemory, ObjectTypeNode objectTypeNode) {
```
tarilabs, line 58: Not sure why this moved from `PropertySpecificUtil` to here, because only needed from tests?
Doesn't any other external project eg jBPM or Optaplanner use them? - just asking

drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
```diff
@@ -318,23 +318,21 @@ public void update(FactHandle handle,
...
                        String... modifiedProperties) {
         Class modifiedClass = object.getClass();
-        update( (InternalFactHandle) handle,
-                object,
-                calculatePositiveMask(asList(modifiedProperties), getSettableProperties(kBase, modifiedClass)),
-                modifiedClass,
-                null);
+
+        TypeDeclaration typeDeclaration = kBase.getOrCreateExactTypeDeclaration( modifiedClass );
+        BitMask mask = typeDeclaration.isPropertyReactive() ?
```
tarilabs, line 323: main fix.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationCache.java
```diff
@@ -263,11 +262,6 @@ private void processModifiedProps(Class<?> cls,
     private TypeDeclaration createTypeDeclarationForBean(Class<?> cls) {
         TypeDeclaration typeDeclaration = new TypeDeclaration(cls);
 
-        Role role = cls.getAnnotation(Role.class);
```
tarilabs, line 266: Moved to `org.drools.core.rule.TypeDeclaration`.

drools-core/src/main/java/org/drools/core/rule/TypeDeclaration.java
```diff
@@ -471,7 +482,7 @@ public void setNovel(boolean novel) {
 
     public List<String> getSettableProperties() {
         if ( settableProprties == null ) {
```
tarilabs, line 484: not a real problem, but I noticed this is misspelled.

tarilabs (approved):
+1 although with some non-blocking questions

tarilabs:
About the failing CI test (`PropertySpecificTest.testWrongUasgeOfWatchAnnotationOnNonPropertySpecificClass`), I would argue the test assertion should be modified from exactly 1 error to >0 errors.

```
Wrong usage of @Watch annotation on class org.drools.compiler.integrationtests.B that is not annotated as @PropertyReactive : [Rule name='R1']
Unknown property s in @Watch annotation : [Rule name='R1']
Unknown property on in @Watch annotation : [Rule name='R1']
```

drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertySpecificTest.java
mariofusco: Those methods are only used for tests. I believe they should not be used by our core code and I wanted to remove any temptation :)

PR #1098, opened 2017-02-13
[DROOLS-1438] Make property reactivity to work even when the property…

mariofusco:
… setter is in an inherited class

### DIFF
--- a/drools-core/src/main/java/org/drools/core/util/ClassUtils.java
+++ b/drools-core/src/main/java/org/drools/core/util/ClassUtils.java
@@ -16,11 +16,6 @@
 
 package org.drools.core.util;
 
-import org.drools.core.common.DroolsObjectInputStream;
-import org.drools.core.common.DroolsObjectOutputStream;
-import org.kie.api.definition.type.Modifies;
-import org.kie.internal.utils.ClassLoaderUtil;
-
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.Externalizable;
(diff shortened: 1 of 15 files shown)


mariofusco:
ok to test

drools-core/src/main/java/org/drools/core/util/ClassUtils.java
```diff
@@ -411,13 +411,13 @@ public static String getPackage(Class<?> cls) {
...
-    public static List<String> getSettableProperties(Class<?> clazz) {
-        Set<SetterInClass> props = new TreeSet<SetterInClass>();
+    public static List<String> getAccessibleProperties( Class<?> clazz ) {
+        Set<PropertyInClass> props = new TreeSet<PropertyInClass>();
         for (Method m : clazz.getMethods()) {
-            if (m.getParameterTypes().length == 1) {
-                String propName = setter2property(m.getName());
-                if (propName != null) {
-                    props.add( new SetterInClass( propName, m.getDeclaringClass() ) );
+            if (m.getParameterTypes().length == 0) {
```
tarilabs, line 417: iterating on the getters, means the 0-args methods.

drools-core/src/main/java/org/drools/core/util/ClassUtils.java
```diff
@@ -411,13 +411,13 @@ public static String getPackage(Class<?> cls) {
...
+                String propName = getter2property(m.getName());
+                if (propName != null && !propName.equals( "class" )) {
```
tarilabs, line 419: excluding the "class property", which is not really a property.

tarilabs (approved):
+1 now iterating generally on the getters for accessible properties, instead of before iterating directly on the setter.

PR #1099, opened 2017-02-14
[DROOLS-1379] Additional tests for OOPath syntax

winklerm:
Adds a couple of unit tests for OOPath syntax parser (XpathAnalysis class).


winklerm:
@mariofusco @tarilabs There are 4 failing tests; however, I think that majority of them fails because the expression which they test is not valid - my plan was to create this PR anyway so we could discuss it.

winklerm:
Fixed failing tests and handling non-oopath expressions (avoid StringIndexOutOfBoundsException and add missing 'return').

winklerm:
Please let me know if I should squash the commits.

mariofusco:
ok to test

baldimir:
I think it would be worth a try to also use some path in condition. E.g. "/address/street{something/somethingAttribute == \"Elm\"}"

PR #1102, opened 2017-02-17
RHBPMS-4621 - Rollback transaction with Status.STATUS_ROLLEDBACK befo…

mswiderski:
…re transaction start


jomarko:
Hi @mswiderski, just a question. Is this tested somewhere? Because it seems to me as change in logic that should be tested.

mswiderski:
it can only be reproduced on wildfly/eap due to special handling as mentioned in the comment. Though this is a no op in case no transaction timeout in specific environment happened

PR #1107, opened 2017-02-20
[DROOLS-1446] ruleunit implementation

mariofusco:


tarilabs (approved):
RuleUnitExecutor and guard system as dicussed

PR #1111, opened 2017-02-22
[DROOLS-1445] make property reactivity to also automatically watch va…

mariofusco:
…riables used in patterns different from the declaring one


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/1112

PR #1120, opened 2017-02-28
[DROOLS-1445] add declarations required by accumulate functions into …

mariofusco:
…accumulate node left mask

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/AccumulateNode.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/AccumulateNode.java
@@ -21,14 +21,17 @@ import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 
 import org.drools.core.RuleBaseConfiguration;
+import org.drools.core.base.ClassObjectType;
 import org.drools.core.common.BetaConstraints;
 import org.drools.core.common.InternalFactHandle;
 import org.drools.core.common.InternalWorkingMemory;
 import org.drools.core.common.Memory;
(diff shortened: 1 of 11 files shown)


drools-core/src/main/java/org/drools/core/reteoo/PropertySpecificUtil.java
```diff
@@ -111,7 +111,7 @@ public static boolean isPropertySetOnMask(BitMask mask, int index) {
         return mask.isSet(index + CUSTOM_BITS_OFFSET);
     }
 
-    public static List<String> getSettableProperties(InternalKnowledgeBase kBase, Class<?> nodeClass) {
+    public static List<String> getAccessibleProperties( InternalKnowledgeBase kBase, Class<?> nodeClass ) {
```
tarilabs, line 114: OK

drools-core/src/main/java/org/drools/core/reteoo/AccumulateNode.java
```diff
@@ -82,11 +91,31 @@ public AccumulateNode(final int id,
...
+
         hashcode = this.leftInput.hashCode() ^
                    this.rightInput.hashCode() ^
                    this.accumulate.hashCode() ^
                    this.resultBinder.hashCode() ^
                    Arrays.hashCode( this.resultConstraints );
+
+    }
+
+    private void addAccFunctionDeclarationsToLeftMask( InternalKnowledgeBase kbase, LeftTupleSource leftInput, Accumulate accumulate ) {
```
tarilabs, line 104: mask for accumulate node set its eleft mask correctly so that the LIA which is on top see correctly the map

tarilabs (approved):
+1

PR #1121, opened 2017-02-28
DROOLS-1379 Add test for OOPath in DSL

yurloc:
This is a basic test for OOPath inside DSL. I couldn't think of anything more complicated or error-prone that the variable reference and OOPath constraint. Both use curly braces and the OOPath constraint's braces must be escaped to work correctly. This is a known and documented requirement of DSL definitions.

Any other ideas?


winklerm:
Thinking about a test using an inline cast (to verify that the '#' character works correctly with DSL), e.g. /person{ #Employee, wage == 10 }.

What do you think?

yurloc:
@winklerm Good idea, done.

PR #1122, opened 2017-03-01
[DROOLS-1453] make property reactive mask to let pass any modificatio…

mariofusco:
…n when there is no pattern

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertySpecificTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertySpecificTest.java
@@ -1877,28 +1877,6 @@ public class PropertySpecificTest extends CommonTestMethodBase {
     }
 
     @Test(timeout = 5000)
-    public void testEval() throws Exception {
-        String rule = "package org.drools.compiler.integrationtests\n" +
-                "import " + PropertySpecificTest.Order.class.getCanonicalName() + "\n" +
-                "rule R1 when\n" +
-                "   $o : Order()\n" +
-                "   eval($o.getId().equals(\"1\"))" +
-                "then\n" +
-                "   modify( $o ) { setDiscounted( true ) };\n" +
-                "end\n";
-
-        KnowledgeBase kbase = loadKnowledgeBaseFromString(rule);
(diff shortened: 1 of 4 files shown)


drools-core/src/main/java/org/drools/core/reteoo/LeftTupleSource.java
```diff
@@ -250,15 +250,13 @@ protected void initDeclaredMask(BuildContext context,
...
+        // if pattern is null (e.g. for eval or query nodes) we cannot calculate the mask, so we set it all
+        if ( pattern != null && isPropertyReactive(context, objectClass) ) {
+            Collection<String> leftListenedProperties = pattern.getListenedProperties();
+            List<String> settableProperties = getAccessibleProperties( context.getKnowledgeBase(), objectClass );
+            leftDeclaredMask = calculatePositiveMask( leftListenedProperties, settableProperties );
+            leftNegativeMask = calculateNegativeMask( leftListenedProperties, settableProperties );
+            setLeftListenedProperties(leftListenedProperties);
         } else {
             // if property specific is not on, then accept all modification propagations
             leftDeclaredMask = AllSetBitMask.get();
```
tarilabs, line 262: OK

> if property specific is not on ...

and now this else branch will also trigger for "pattern is null (e.g. for eval or query nodes)"

drools-compiler/src/test/java/org/drools/compiler/integrationtests/PropertySpecificTest.java
```diff
@@ -1877,28 +1877,6 @@ public void testNoConstraint2() throws Exception {
     }
 
     @Test(timeout = 5000)
-    public void testEval() throws Exception {
```
tarilabs, line 1880: OK for remove.
The original test was actually faulty, because a rule of this kind would go inloop forever (match Order() and eval() always pass).

- Changing by adding a `no-loop` would have made the meaning of testing the `no-loop`.
- Changing by having the RHS to instead insert an object in a list, or something anyway not modifying the pattern fact, would have simply tested the `eval()` for which there are already existing tests.

tarilabs (approved):
+1

PR #1126, opened 2017-03-02
[DROOLS-1452] fix manual plexus wiring after upgrade to Maven 3.3.9

psiroky:
Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/401

### DIFF
--- a/kie-ci/pom.xml
+++ b/kie-ci/pom.xml
@@ -167,8 +167,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.sonatype.sisu</groupId>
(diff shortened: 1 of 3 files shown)


psiroky:
@mariofusco FYI this is the upgrade to Maven 3.3.9. It was not that bad. However, we should really remove the manual wiring as it is very costly to maintain. I tried to run the default Guice wiring in OSGi and it failed horribly, but it could have been just configuration issue, not sure (I haven't really had the time to figure it out).

mariofusco:
+1

PR #1128, opened 2017-03-06
[DROOLS-1356] add command to advance pseudo-clock

mariofusco:


mariofusco:
ok to test

drools-core/src/main/java/org/drools/core/command/runtime/AdvanceSessionTimeCommand.java
```diff
@@ -0,0 +1,104 @@
...
+    public Long execute(Context context ) {
+        KieSession ksession = ((RegistryContext)context).lookup( KieSession.class );
+        SessionPseudoClock sessionClock = ksession.<SessionPseudoClock>getSessionClock();
+
+        sessionClock.advanceTime( amount, unit );
+        long result = sessionClock.getCurrentTime();
+
+        ExecutionResultImpl results = ((RegistryContext)context).lookup( ExecutionResultImpl.class );
+        if ( results != null ) {
+            results.getResults().put( this.outIdentifier, result );
```
tarilabs, line 72: Sorry to insist :) but with reference to draft from here:
https://github.com/tarilabs/drools/blob/patch-pseudoclockcmds/drools-core/src/main/java/org/drools/core/command/runtime/PseudoClockAdvanceTimeCommand.java#L58-L63

I think the out identifier for this command should be optional, and if is valorized (meaning the user is interested in the out value) then we place the result in the returning results.

This is a bit different from the `GetSessionTimeCommand` where we can always expect to have the out identifier.

What do you think, please?

tarilabs (changes requested):
I think the public factory should update for the option to specify the out identifier?

drools-core/src/main/java/org/drools/core/command/impl/CommandFactoryServiceImpl.java
```diff
@@ -357,7 +357,7 @@ public Command newEnableAuditLog( String filename ) {
     }
 
     @Override
-    public Command<Long> newAdvanceSessionTime(String outIdentifier, long amount, TimeUnit unit) {
+    public Command<Long> newAdvanceSessionTime(long amount, TimeUnit unit, String outIdentifier) {
```
tarilabs: thanks :)

tarilabs (approved):
+1

PR #1129, opened 2017-03-07
fix examples after having enabled property reactivity by default

mariofusco:


drools-examples/src/main/java/org/drools/games/Run.java
```diff
@@ -15,6 +15,9 @@
 
 package org.drools.games;
 
+import org.kie.api.definition.type.ClassReactive;
+
+@ClassReactive
```
mariofusco, line 20: @tarilabs This is the only actual fix. All other changes are removing redundant @PropertyReactive annotations and replacing retract with delete.

tarilabs: as `Run` is not conform to Java Beans spec. ok.

tarilabs (approved):
+1

mariofusco:
ok to test

PR #1130, opened 2017-03-07
Refactors and adds new OOPath tests

baldimir:
- Renames XPathTest and related classes and package to OOPathTest
- Splits XPathTest to more classes
- Adds new accumulate tests
- Adds collect test
- Adds fireUntilHalt test in OOPathBindTest class
- Adds some other test cases


baldimir:
@winklerm could you please review this?

baldimir:
Failing test is reproducer for this JIRA: https://issues.jboss.org/browse/RHBRMS-2755

drools-compiler/src/test/java/org/drools/compiler/oopath/OOPathAccumulateTest.java
```diff
@@ -0,0 +1,122 @@
...
+        final Man bob = new Man( "Bob", 40 );
+        bob.addChild( new Child( "Charles", 12 ) );
+        bob.addChild( new Child( "Debbie", 8 ) );
+
+        ksession.insert( bob );
+        ksession.fireAllRules();
+
+        final Number result = (Number) ksession.getGlobal("globalVar");
+        if (result instanceof Double) {
+            assertEquals(expectedResult.doubleValue(), result.doubleValue(), 0);
```
winklerm: Please use AssertJ for assertions, if possible.

drools-compiler/src/test/java/org/drools/compiler/oopath/OOPathBindTest.java
```diff
@@ -0,0 +1,270 @@
...
+                        "then\n" +
+                        "  list.add( $age );\n" +
+                        "end\n";
+
+        final KieSession ksession = new KieHelper().addContent( drl, ResourceType.DRL )
+                .build()
+                .newKieSession();
+
+        if (fireUntilHalt) {
+            new Thread(ksession::fireUntilHalt).start();
```
winklerm: It might be better to use Executors.newSingleThreadExecutor() and get the result from the Future object at the end of the test so a potential exception won't get away unnoticed.

drools-compiler/src/test/java/org/drools/compiler/oopath/OOPathBindTest.java
```diff
@@ -0,0 +1,270 @@
...
+            assertEquals( 1, list.size() );
+        }
+        assertTrue( list.contains( 40 ) );
+    }
+
+    @Test
+    public void testBindString() {
+        final String drl =
+                "import org.drools.compiler.oopath.*;\n" +
+                        "global java.util.List list\n" +
```
winklerm: Do we want to have the DRL (or multi-line Strings in general) indented this way? I am not strongly against it, rather thinking about setting up common practice.

drools-compiler/src/test/java/org/drools/compiler/oopath/Person.java
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
 */
 
-package org.drools.compiler.xpath;
+package org.drools.compiler.oopath;
```
winklerm: I am thinking about moving the model classes into a separate package, such as _org.drools.compiler.oopath.model_, what do you think?

drools-compiler/src/test/java/org/drools/compiler/oopath/OOPathBindTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        final List<String> list = new ArrayList<>();
+        ksession.setGlobal( "list", list );
+
+        final Man bob = new Man( "Bob", 40 );
+
+        ksession.insert(bob);
+        if (fireUntilHalt) {
+            try {
+                while (list.size() < 1) {
```
winklerm: Extracting the while loop into a private method might make this part of code more readable.

winklerm (changes requested):
Looks good, only a few minor comments - please see inline.

winklerm:
@baldimir btw. thanks a lot for splitting the XPathTest and renaming XPath to OOPath!

drools-compiler/src/test/java/org/drools/compiler/oopath/OOPathBindTest.java
baldimir: We don't have a common practice for this. I think it is fine if it is readable. These indentations came from already existing tests that were moved.

baldimir:
@winklerm all proposed changes incorporated. Please review.

winklerm (approved):
Looks good to me, +1 to merge.

PR #1131, opened 2017-03-08
Reproducer for DROOLS-1470 Match.getObjects() should include acculumates

ge0ffrey:
See
  https://issues.jboss.org/browse/DROOLS-1470
Let's discuss. Maybe we need a new method getObjectsDeep() instead if we can't change getObjects() for backwards compatibility reasons.


mariofusco:
Replaced by https://github.com/droolsjbpm/drools/pull/1138

PR #1132, opened 2017-03-08
[DROOLS-1227] fix events expiration inference

mariofusco:


mariofusco:
@etirelli There were a few failing tests, but I reviewed all of them and I'm pretty sure that those tests just relied on the presence of the bug, so I corrected the wrong assertions. Please double check what I did and let me know if you want to further discuss this.

drools-core/src/main/java/org/drools/core/time/TemporalDependencyMatrix.java
```diff
@@ -53,20 +53,20 @@ public void setEvents(List<Pattern> events) {
...
         int index = events.indexOf( pattern );
-        for( Interval interval : matrix[index] ) {
-           expiration = Math.max( expiration, interval.getUpperBound() );
+
+        Interval[] intervals = matrix[index];
+        for (int i = 0; i < intervals.length; i++) {
+            if (i != index) { // skip values on the diagonal
+                expiration = Math.max( expiration, intervals[i].getUpperBound() );
+                if (expiration == Long.MAX_VALUE) {
+                    return expiration;
```
tarilabs: IFF expiration has already become infinite, simply return infinite == keep forever

drools-compiler/src/test/java/org/drools/compiler/integrationtests/CepEspTest.java
```diff
@@ -3778,15 +3778,15 @@ public void AfterOperatorInCEPQueryTest() {
...
 
         results = ksession.getQueryResults("EventsBeforeNineteenSeconds");
 
-        assertEquals( 2, results.size() );
+        assertEquals( 0, results.size() );
 
         results = ksession.getQueryResults("EventsBeforeHundredSeconds");
 
-        assertEquals( 3, results.size() );
+        assertEquals( 1, results.size() );
```
tarilabs, line 3789: I agree results of this query accordingly to doc definitions, should be `0 0 1`

drools-compiler/src/test/java/org/drools/compiler/integrationtests/StreamsTest.java
```diff
@@ -567,7 +567,7 @@ public void testEventExpirationValue() throws Exception {
         ObjectType stot = new ClassObjectType( StockTick.class );
         for (ObjectTypeNode otn : otns) {
             if (otn.getObjectType().isAssignableFrom( stot )) {
-                assertEquals( -1,
+                assertEquals( Long.MAX_VALUE,
```
tarilabs: I am wondering if given the expiration offset before had the convention that `-1` signified "infinite", there could be an hard-coded check for `-1` when the expiration is being considered for the actual facts?
Or possibly it was doing the check that the offset needed to be `>= 0` ?

tarilabs (commented):
I think indeed @etirelli should double check

drools-core/src/main/java/org/drools/core/time/TemporalDependencyMatrix.java
etirelli: Mario, I have the same concern as @tarilabs. -1 offset meaning infinity is used in other places, including when scheduling the expiration call if I am not mistaken.

mariofusco:
@etirelli @tarilabs What happened in reality is that the expiration job wasn't triggered anyway because it avoids to schedule it both when the expiration offset was negative and when it is equal to Long.MAX_VALUE [1]. However I agree that -1 was used with meaning of "never expires" throughout the code so I restored it and for better clarity put it in a constant value.

[1] https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PropagationEntry.java#L170

drools-core/src/main/java/org/drools/core/rule/TypeDeclaration.java
```diff
@@ -46,6 +46,8 @@
     Externalizable,
     Comparable<TypeDeclaration> {
 
+    public static final long NEVER_EXPIRES = -1;
```
etirelli, line 49: Nice, finally someone had the guts to create a constant! :)

etirelli:
+1

etirelli (approved):

PR #1134, opened 2017-03-08
DROOLS-1424 Variance and standard deviation accumulate functions

ge0ffrey:


drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3303,4 +3303,99 @@ public void testTypedMaxOnAccumulate() {
...
+                "  accumulate(\n" +
+                "    Cheese($price : price);\n" +
+                "    $result : variance((double) $price)\n" +
+                "  )\n" +
+                "then\n" +
+                "  list.add($result);\n" +
+                "end";
+
+        KieBase kieBase = new KieHelper().addContent(drl, ResourceType.DRL).build();
+        Function<int[], Double> cheeseInsertsFunction = (prices) -> {
```
mariofusco: I don't understand why you're using a Function here. It looks like only a fancy way to invoke a method. Can't replace it with a plain private method?

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3303,4 +3303,99 @@ public void testTypedMaxOnAccumulate() {
...
+
+    @Test
+    public void testVarianceDouble() {
+        String drl =
+                "import org.drools.compiler.Cheese\n" +
+                "global java.util.List list;\n" +
+                "rule R when\n" +
+                "  accumulate(\n" +
+                "    Cheese($price : price);\n" +
+                "    $result : variance((double) $price)\n" +
```
mariofusco: Is this cast to double necessary? I believe it should work even without it.

drools-core/src/main/java/org/drools/core/base/accumulators/VarianceAccumulateFunction.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public void init(VarianceData data) {
+        data.count = 0;
+        data.mean = 0;
+        data.squaredSum = 0.0;
+    }
+
+    @Override
+    public void accumulate(VarianceData data, Object value) {
+        double x = ((Number) value).doubleValue();
```
mariofusco, line 60: Is double enough here? Wouldn't be better to use BigDecimal, or at least having another implementation using BigDecimal?

drools-core/src/main/java/org/drools/core/base/accumulators/VarianceAccumulateFunction.java
```diff
@@ -0,0 +1,96 @@
...
+        double higherDelta = x - data.mean;
+        data.mean = data.mean * data.count / (data.count - 1.0) - x / (data.count -1.0);
+        double lowerDelta = x - data.mean;
+        data.count--;
+        data.squaredSum -= lowerDelta * higherDelta;
+    }
+
+    @Override
+    public Object getResult(VarianceData data) {
+        return new Double(data.squaredSum / data.count);
```
mariofusco: new Double() is redundant here

drools-core/src/main/java/org/drools/core/base/accumulators/StandardDeviationAccumulateFunction.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.Serializable;
+
+/**
+ * An implementation of an accumulator capable of calculating standard deviation.
+ */
+public class StandardDeviationAccumulateFunction extends VarianceAccumulateFunction {
+
+    @Override
+    public Object getResult(VarianceData data) {
+        return new Double(Math.sqrt(data.squaredSum / data.count));
```
mariofusco: new Double is redundant here

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3303,4 +3303,99 @@ public void testTypedMaxOnAccumulate() {
...
+
+    @Test
+    public void testStandardDeviationDouble() {
+        String drl =
+                "import org.drools.compiler.Cheese\n" +
+                "global java.util.List list;\n" +
+                "rule R when\n" +
+                "  accumulate(\n" +
+                "    Cheese($price : price);\n" +
+                "    $result : standardDeviation((double) $price)\n" +
```
mariofusco: See above: cast shouldn't be necessary

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3303,4 +3303,99 @@ public void testTypedMaxOnAccumulate() {
...
+                "  )\n" +
+                "then\n" +
+                "  list.add($result);\n" +
+                "end";
+
+        KieBase kieBase = new KieHelper().addContent(drl, ResourceType.DRL).build();
+        Function<int[], Double> cheeseInsertsFunction = (prices) -> {
+            KieSession ksession = kieBase.newKieSession();
```
mariofusco: See above: replace Function with private method

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
ge0ffrey: yes, will do

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
ge0ffrey: yes, but then it's implicit. I prefer such casts to be explicit, personally, because a + b / 2 differs if a is an int or double. I'll remove it anyway.

drools-core/src/main/java/org/drools/core/base/accumulators/VarianceAccumulateFunction.java
ge0ffrey: Yes, we need a BigDecimal implementation too, in time. Problem is that brings in BigDecimal's scale complexity and especially the lack of Math.sqrt(BigDecimal) in Java is a pita. So I left that out of scope for now.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
ge0ffrey: +1, will do

drools-core/src/main/java/org/drools/core/base/accumulators/VarianceAccumulateFunction.java
ge0ffrey: ok, will do (I am not fan of autoboxing normally, I've seen it cause bugs too many times).

ge0ffrey:
All feedback done, should be ready to merge

PR #1135, opened 2017-03-08
[RHBPMS-4634] handle pom type dependencies in kie-ci

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/kproject/ReleaseIdImpl.java
```diff
@@ -78,7 +82,9 @@ public void setVersion(String version) {
 
     @Override
     public String toString() {
-        return groupId + ":" + artifactId + ":" + version;
+        return type == null || type.equals( "jar" ) ?
```
tarilabs, line 85: looks ok

drools-compiler/src/main/java/org/drools/compiler/kproject/ReleaseIdImpl.java
```diff
@@ -44,12 +45,15 @@ public ReleaseIdImpl(String releaseId) {
         this.version = split[2];
     }
 
-    public ReleaseIdImpl(String groupId,
-            String artifactId,
-            String version) {
+    public ReleaseIdImpl(String groupId, String artifactId, String version) {
+        this(groupId, artifactId, version, "jar");
```
tarilabs, line 49: looks ok default jar.

drools-compiler/src/main/java/org/drools/compiler/kproject/ReleaseIdImpl.java
```diff
@@ -142,28 +148,28 @@ private static ReleaseId getReleaseIdFromProperties(Properties props, String pat
...
-        if (artifactId != null ? !artifactId.equals(that.artifactId) : that.artifactId != null)
-            return false;
-        if (groupId != null ? !groupId.equals(that.groupId) : that.groupId != null)
-            return false;
-        if (version != null ? !version.equals(that.version) : that.version != null)
-            return false;
-
-        return true;
+    private boolean equalsIgnoreNull(Object o1, Object o2) {
+        return o1 != null ? o1.equals(o2) : o2 != null;
```
tarilabs: If the o1 is diff than null, o1 must equals o2.
If the o1 is null, return "o2 is _diff than_ null" ? I think this second part is wrong?

tarilabs (changes requested):
Can you kindly check `equalsIgnoreNull` comment, please?

drools-compiler/src/main/java/org/drools/compiler/kproject/ReleaseIdImpl.java
mariofusco: @tarilabs good catch, thanks

kie-ci/src/test/java/org/kie/scanner/KieModuleBuilderTest.java
```diff
@@ -119,4 +138,74 @@ protected String createDRLWithImport( String ruleName ) {
...
+        File pomFile = fileManager.newFile("pom.xml");
+        fileManager.write(pomFile, pom);
+
+        InternalKieModule kJar1 = createKieJarWithPomDependencies( ks, releaseIdNoDep, pom );
+        MavenRepository repository = getMavenRepository();
+        repository.installArtifact(releaseIdNoDep, kJar1, pomFile);
+
+        InternalKieModule kJar2 = createKieJarWithDependencies( ks, releaseIdWithDep, true, "rule1", releaseIdNoDep);
+        KieContainer kieContainer2 = ks.newKieContainer( releaseIdWithDep );
+        KieSession kieSession = kieContainer2.newKieSession();
```
tarilabs, line 161: I assume original problem was throwing Exception, hence the test does not need any assertions.

tarilabs (approved):
+1

kie-ci/src/test/java/org/kie/scanner/KieModuleBuilderTest.java
mariofusco: Yes, not sure what I could assert here. Maybe that ksession is not null?

PR #1137, opened 2017-03-10
OOPath tests with CEP - second part

winklerm:
Added more tests for using OOPath with CEP:
- the remaining temporal operators,
- event expiration and duration.


baldimir (approved):
Looks good. +1.

PR #1140, opened 2017-03-11
DROOLS-1474 Reproducer for getObjectsDeep() should include exists elements too

ge0ffrey:
See https://issues.jboss.org/browse/DROOLS-1474


mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/1143

PR #1142, opened 2017-03-13
RHBRMS-2759: Unable to validate a Guided Rule with DSL if a date field is used

manstis:
See https://issues.jboss.org/browse/RHBRMS-2759


manstis:
@Rikkola Hi, could you please review when you get a moment of idleness!

Rikkola (approved):
Hehe, looks good, but the diff was a mile long. ?w=1 saved me :)

manstis:
@Rikkola Yes, ```?w=1``` has saved me whittling away an afternoon looking at formatting too! :-)

PR #1144, opened 2017-03-14
exclude classes generated by ANTLR from jacoco instrumentation

jsoltes:
This commit fixes build failing with active code-coverage profile, which was introduced by recent pull-request: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/381. The classes causing failure of the jacoco plugin were excluded from the instrumentation.

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -188,8 +188,27 @@
             <excludeFilterFile>${project.basedir}/src/main/findbugs/findbugs-exclude.xml</excludeFilterFile>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.jacoco</groupId>
+          <artifactId>jacoco-maven-plugin</artifactId>
+          <executions>
+            <execution>
+              <id>default-instrument</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -97,6 +97,24 @@
...
+              </goals>
+              <configuration>
+                <excludes>
+                  <exclude>**/DRL5Lexer.class</exclude>
+                  <exclude>**/DRL6Lexer.class</exclude>
+                </excludes>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
```
psiroky: Would it make sense to put this directly into `drools-compiler`'s POM? Or is this needed in other modules as well?

jsoltes: This influences only drools-compiler, so I can move it there.

jsoltes:
@psiroky I have successfuly moved it to the pluginManagement of the drools-compiler project, can you please have another look?

psiroky (approved):

psiroky:
Thanks @jsoltes!

PR #1152, opened 2017-03-15
[RHBPMS-4649] avoid NPE when evaluating a Boolean expression returnin…

mariofusco:
…g null

### DIFF
--- a/drools-compiler/src/main/resources/org/drools/compiler/rule/builder/dialect/java/javaRule.mvel
+++ b/drools-compiler/src/main/resources/org/drools/compiler/rule/builder/dialect/java/javaRule.mvel
@@ -10,14 +10,16 @@ public static org.drools.core.spi.FieldValue @{methodName}(@foreach{declr : decl
 public static boolean @{methodName}(@foreach{declr : declarations} @{declr.typeName} @{declr.identifier} @end{","} @if{localDeclarations != empty && declarations != empty},@end{}
                                                       @foreach{declr : localDeclarations } @{declr.typeName} @{declr.identifier} @end{","} @if{globals != empty && (localDeclarations != empty || declarations != empty)},@end{}
                                                       @foreach{type : globalTypes, identifier : globals} @{type} @{identifier} @end{","} ) throws java.lang.Exception {
-    return ( @{text} );
+    Boolean $predicate$result$ = ( @{text} );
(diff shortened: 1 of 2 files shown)


drools-compiler/src/main/resources/org/drools/compiler/rule/builder/dialect/java/javaRule.mvel
```diff
@@ -10,14 +10,16 @@ public static org.drools.core.spi.FieldValue @{methodName}(@foreach{declr : decl
 public static boolean @{methodName}(@foreach{declr : declarations} @{declr.typeName} @{declr.identifier} @end{","} @if{localDeclarations != empty && declarations != empty},@end{}
                                                       @foreach{declr : localDeclarations } @{declr.typeName} @{declr.identifier} @end{","} @if{globals != empty && (localDeclarations != empty || declarations != empty)},@end{}
                                                       @foreach{type : globalTypes, identifier : globals} @{type} @{identifier} @end{","} ) throws java.lang.Exception {
-    return ( @{text} );
+    Boolean $predicate$result$ = ( @{text} );
+    return $predicate$result$ == null ? false : $predicate$result$.booleanValue();
```
tarilabs, line 14: check predicate represented by `@{text}`, return false if null, otherwise its preodicate value ok

drools-compiler/src/main/resources/org/drools/compiler/rule/builder/dialect/java/javaRule.mvel
```diff
@@ -10,14 +10,16 @@ public static org.drools.core.spi.FieldValue @{methodName}(@foreach{declr : decl
...
 }
 @end{}
 
 @declare{"evalMethod"}
 public static boolean @{methodName}(@foreach{declr : declarations} @{declr.typeName} @{declr.identifier} @end{","} @if{globals != empty && declarations != empty},@end{}
                                     @foreach{type : globalTypes, identifier : globals} @{type} @{identifier} @end{","} ) throws java.lang.Exception {
-    return ( @{text} );
+    Boolean $eval$result$ = ( @{text} );
```
tarilabs, line 21: check `@{text}` which being called in eval is a predicate, return false if null, otherwise its preodicate value ok

tarilabs (approved):

mariofusco:
ok to test

PR #1155, opened 2017-03-20
[DROOLS-1489] fix ReactiveFromNode serialization

mariofusco:


drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufOutputMarshaller.java
```diff
@@ -371,7 +371,8 @@ private static void writeNodeMemories(MarshallerWriteContext context,
                         _node = writeRIANodeMemory( baseNode.getId(), baseNode, memories );
                         break;
                     }
-                    case NodeTypeEnums.FromNode: {
+                    case NodeTypeEnums.FromNode:
+                    case NodeTypeEnums.ReactiveFromNode: {
```
tarilabs, line 375: ok

tarilabs (approved):
+1

PR #1158, opened 2017-03-22
[DROOLS-1379] Oopath test persistence jpa

cimbalek:


drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
```diff
@@ -99,521 +99,459 @@ public void tearDown() throws Exception {
...
         ksession.fireAllRules();
-        
+
         list = (List<?>) ksession.getGlobal("list");
-        
-        assertEquals( 3,
-                list.size() );
-        
+
+        assertEquals(3, list.size());
```
triceo: `AssertJ` is your friend. `Assertions.assertThat(list).hasSize(3);`
Let the test fail and see for yourself the vast difference in the failure message if you do it like that.

drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
```diff
@@ -99,521 +99,459 @@ public void tearDown() throws Exception {
...
+        final KieBase kbase = new KieHelper().addContent(str, ResourceType.DRL).build();
+
+        final KieSession ksession = KieServices.get().getStoreServices().newKieSession(kbase, null, env);
+        final List<?> list = new ArrayList<>();
+
+        ksession.setGlobal("list", list);
+
+        ksession.insert(1);
+        ksession.insert(2);
+        ksession.insert(3);
```
triceo: Is the 3 here the same 3 as below? If so, perhaps it should be a constant.

drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
```diff
@@ -99,521 +99,459 @@ public void tearDown() throws Exception {
...
+                "\n";
+
+        final KieBase kbase = new KieHelper().addContent(str, ResourceType.DRL).build();
+
+        final KieSession ksession = KieServices.get().getStoreServices().newKieSession(kbase, null, env);
+        final List<?> list = new ArrayList<>();
+
+        ksession.setGlobal("list", list);
+
+        ksession.insert(1);
```
triceo: What about `IntStream.rangeClosed(1, 3).forEach(i -> ksession.insert(i));`?
Let's not repeat ourselves.

baldimir: I think this would be an overkill for this case. If it is as it is I see at first sight that there are three facts inserted in the test. I don't like using streams for such small ranges. It adds a bit of obfuscation to the code. However I would not argue against extracting this into some method like "insertThreeFacts()" or something, because it is used at different places in the test class. But as of me, I will leave this for @cimbalek and your consideration. I don't consider this a blocker for review considering the whole test class.

baldimir (changes requested):
Please replace the assertions with AssertJ as @triceo mentioned. No other comments from me.

drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
cimbalek: I agree with @baldimir. It's just three numbers and it's pretty straightforward and visible at the first sight what's happening here

triceo: I'm not going to block the PR for this, but I'd like to suggest you think about this in a different way. What you're doing is pure imperative - I need three things, so I'll add three things.

However, the `Stream` implementation actually tells you more things - it says "take these three numbers" and then it says "do this with all them". It separates logic from data, which is why I consider it better.

baldimir (approved):

drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/JpaPersistentStatefulSessionTest.java
cimbalek: Replaced this by the method using `Stream`. I hope this satisfies both parties :)

triceo (approved):

PR #1174, opened 2017-03-28
Remove Byteman from drools-compiler

baldimir:
Tests that used Byteman are updated so they don't use it anymore. 
- Byteman was used for thread synchronization in tests. I replaced this synchronization with help of new CyclicBarrier instances added to threads according to the Byteman rules that were used in the tests.
- This removal enables us to run these tests also on IBM JDKs, because Byteman in combination with JUnit wasn't able to run on these JDKs. 

@mariofusco @psiroky could you please review this?


drools-compiler/src/test/java/org/drools/compiler/kie/builder/impl/KieModuleRepoTest.java
```diff
@@ -228,130 +204,107 @@ public void removeStoreArtifactMapTest() throws Exception {
...
+        };
 
         // 2. remove and redploy
+        final Thread removeThread = new Thread(removeRunnable);
         removeThread.setName("remove");
         removeThread.start();
 
+        final Thread redeployThread = new Thread(redeployRunnable);
         redeployThread.setName("store");
         redeployThread.start();
```
triceo: Are you not stopping these? If so, perhaps it'd be better if the runnables were submitted to an executor and then shutting down the executor at the end.

baldimir: threadsFinishedBarrier should make the test fail on timeout if something gets stuck and then JUnit can kill the threads. But to be 100% safe I added ExecutorService to each of the tests I updated.

baldimir:
jenkins retest this

baldimir:
Failed test from checks wasn't related to this PR.

triceo (approved):

PR #1175, opened 2017-03-28
[DROOLS-1501] add test modifying a subclass over a window

mariofusco:


drools-core/src/main/java/org/drools/core/reteoo/EntryPointNode.java
```diff
@@ -260,7 +261,7 @@ public static void removeRightTuplesMatchingOTN( PropagationContext pctx, Intern
         // remove any right tuples that matches the current OTN before continue the modify on the next OTN cache entry
         RightTuple rightTuple = modifyPreviousTuples.peekRightTuple(partition);
         while ( rightTuple != null &&
-                ((BetaNode) rightTuple.getTupleSink()).getObjectTypeNode() == node ) {
+                ((BaseNode) rightTuple.getTupleSink()).getObjectTypeNode() == node ) {
```
tarilabs, line 264: common parent super-class where to call the method now.

drools-core/src/main/java/org/drools/core/common/BaseNode.java
```diff
@@ -162,6 +162,8 @@ protected abstract boolean doRemove(RuleRemovalContext context,
      */
     public abstract boolean isInUse();
 
+    public abstract ObjectTypeNode getObjectTypeNode();
```
tarilabs, line 165: common parent super-class declares now this method, as abstract, it is used by `EntryPointNode`

tarilabs (approved):
+1

PR #1182, opened 2017-03-30
[DROOLS-4666] fix workbench model marshalling of a rule using forall

mariofusco:


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -2593,11 +2594,17 @@ private CompositeFactPattern parseExistentialElementDescr(final RuleModel m,
...
-                        new CompositeFactPattern(CompositeFactPattern.COMPOSITE_TYPE_EXISTS);
+        CompositeFactPattern comp;
+        if (conditionalDescr instanceof NotDescr) {
+            comp = new CompositeFactPattern(CompositeFactPattern.COMPOSITE_TYPE_NOT);
+        } else if (conditionalDescr instanceof OrDescr) {
+            comp = new CompositeFactPattern(CompositeFactPattern.COMPOSITE_TYPE_OR);
+        } else if (conditionalDescr instanceof ExistsDescr) {
+            comp = new CompositeFactPattern(CompositeFactPattern.COMPOSITE_TYPE_EXISTS);
+        } else {
+            throw new IllegalArgumentException( "Unknown conditional descr type: " + conditionalDescr );
```
manstis, line 2605: In an ideal world, if the type of ```conditionalDescr``` was not supported in the UI model (and hence not supported by UI elements) I'd like to replace the single ```ConditionalElementDescr``` with a ```FreeFormLine``` rather than raising the exception which causes the whole rule to fallback to ```FreeFormLine``` (text).

For example, consider:-
```
Person( name == "mario" )
forall( Person( age > 25 ) )
```
This was authored by the following steps:
1. User created "Pattern" and added the constraint using a custom UI widget
2. The User added a "Free Form Line" for the "forall" as it is not supported by a corresponding UI widget
3. User saves

When the User re-opens the rule; at the moment the "forall" is _incorrectly_ replaced with an "exists" UI model element and UI widget. The fix leads to the whole rule being represented in the editor as "Free Form Lines" (text) and the original "Pattern" widget is lost.

So whilst this ensures the rule, when re-opened, is syntactically identical it looses the UI representation of elements that are supported by the UI (i.e. the "Pattern"). Is there a way to recover the DRL relating to a "Forall" descriptor? if so we could catch an exception in the code that called ```parseExistentialElementDescr(..)``` and replace only the single descriptor.

Otherwise, this is probably all we can do until the UI is rewritten.

manstis:
@mariofusco I added a comment I'd like your opinion on. I want to prevent a reciprocol JIRA reporting that "supported UI elements" appear as "text" when re-opening a rule containing "forall". I can guarantee one will be raised. If the effort-reward to fall back to text for a single descriptor is not worth it then we can only advise they need to wait for a new editor...

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
mariofusco: @manstis I see what you mean but if you step in with the debugger you'll find that it is nearly impossible to recreate the original text of the forall pattern starting from the ForallDescr object. The only way I see to do so is probably getting that text from the original drl using another regular expression, but I find this extremely error-prone and I'm afraid this could put us in bigger troubles. Please let me know how to proceed about this, but I'd vote to merge this pull request as it is.

manstis: @mariofusco I knew that; IDK if there was an easy way to unwind the descriptor to the source DRL... anyway I agree with you; I was just doing "due diligence" as I'll need to comment the JIRA to the effect that _when_ an unsupported element is encountered we fallback the whole rule to DRL.

manstis (approved):

PR #1183, opened 2017-03-30
1. To use extension elements inside inputData elements in order to pr…

akoufoudakis:
**_Please, review the following code._**
1. To use extension elements inside inputData elements in order to provide “external” information: GEO location, time and location tracking of a decision making (when and were specific decision was taken). This is the primary scope of the use.

2. In case extension element represents a complex type customer prefers to get back an instance of the complex type, assuming the complex type is defined as a class by the development team on the customer side. However, the customer also understands that it can be more complex and considers the possibility to get the complex type as an XML snippet. The parsing and turning the XML snippet into the complex type, defined by the customer.
3. The possibility of getting the list of extension elements is expected from other DMN elements, which extend from DMNElement, according to the DMN specification and meta model diagram by OMG. In other words, the usage of extension elements is not limited to be a part of input data elements only.
4. The client is not expecting “authoring” of the extension elements.
Since extension elements are a part of DMN specification, this is an attempt to make it possible for drools to register custom extension elements in a generic way.

### DIFF
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/marshalling/v1_1/DMNExtensionElementRegister.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/marshalling/v1_1/DMNExtensionElementRegister.java
@@ -14,23 +14,13 @@
  * limitations under the License.
  */
 
-package org.kie.dmn.api.core;
+package org.kie.dmn.api.marshalling.v1_1;
 
-import org.kie.api.io.Resource;
-import org.kie.dmn.api.marshalling.v1_1.DMNExtensionElementRegister;
-import org.kie.dmn.model.v1_1.Definitions;
+import com.thoughtworks.xstream.XStream;
 
-import java.io.Reader;
 
-public interface DMNCompiler {
+public interface DMNExtensionElementRegister {
 
-    DMNModel compile( Resource resource );
+    public void registerExtensionConverters(XStream xstream);
 
-    DMNModel compile( Resource resource, DMNExtensionElementRegister extensionElementRegister );
-
-    DMNModel compile( Reader source );
-
-    DMNModel compile( Reader source, DMNExtensionElementRegister extensionElementRegister );
-
-    DMNModel compile(Definitions dmndefs);
 }
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/DMNMarshallerFactory.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/DMNMarshallerFactory.java
@@ -16,6 +16,7 @@
 
 package org.kie.dmn.backend.marshalling.v1_1;
 
+import org.kie.dmn.api.marshalling.v1_1.DMNExtensionElementRegister;
 import org.kie.dmn.backend.marshalling.v1_1.xstream.XStreamMarshaller;
 import org.kie.dmn.api.marshalling.v1_1.DMNMarshaller;
 
@@ -24,5 +25,10 @@ public class DMNMarshallerFactory {
     public static DMNMarshaller newDefaultMarshaller() {
         return new XStreamMarshaller();
     }
+
+    public static DMNMarshaller newMarshallerWithExtensions(DMNExtensionElementRegister extensionElementRegister) {
+        return new XStreamMarshaller(extensionElementRegister);
+    }
+
 
 }
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
@@ -40,6 +40,7 @@ import com.thoughtworks.xstream.io.xml.AbstractPullReader;
 import com.thoughtworks.xstream.io.xml.QNameMap;
 import com.thoughtworks.xstream.io.xml.StaxDriver;
 import com.thoughtworks.xstream.io.xml.StaxWriter;
+import org.kie.dmn.api.marshalling.v1_1.DMNExtensionElementRegister;
 import org.kie.dmn.api.marshalling.v1_1.DMNMarshaller;
 import org.kie.dmn.backend.marshalling.CustomStaxReader;
 import org.kie.dmn.backend.marshalling.CustomStaxWriter;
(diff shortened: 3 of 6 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -41,6 +42,8 @@
 public class DMNAssemblerService implements KieAssemblerService {
 
     private static final Logger logger = LoggerFactory.getLogger( DMNAssemblerService.class );
+    private static final String DMN_EXTENSION_REGISTER = "ork.kie.dmn.extension.element.register";
```
etirelli, line 45: there is a typo in the property: "ork" instead of "org". Also, I suggested we use property name prefix: "org.kie.dmn.marshaller.extension.". Please note that this is a prefix, and every "app" could use one or more extensions by adding an ID at the end: "org.kie.dmn.marshaller.extension.foo", "org.kie.dmn.marshaller.extension.bar", etc

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNCompiler.java
```diff
@@ -25,7 +26,11 @@
 
     DMNModel compile( Resource resource );
 
+    DMNModel compile( Resource resource, DMNExtensionElementRegister extensionElementRegister );
```
etirelli, line 29: I think we need to change this slightly and wrap the DMNExtensionElementRegister in a DMNCompilerConfiguration object. This will prevent us from braking the interface in the future when more options are added.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/marshalling/v1_1/DMNExtensionElementRegister.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.dmn.api.marshalling.v1_1;
+
+import com.thoughtworks.xstream.XStream;
+
+
+public interface DMNExtensionElementRegister {
```
etirelli, line 22: I would just make the name slightly shorter, but this is just personal preference: DMNExtensionRegister

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/DMNMarshallerFactory.java
```diff
@@ -25,4 +26,9 @@ public static DMNMarshaller newDefaultMarshaller() {
         return new XStreamMarshaller();
     }
 
+    public static DMNMarshaller newMarshallerWithExtensions(DMNExtensionElementRegister extensionElementRegister) {
```
etirelli, line 29: We need to support multiple extension registers, so maybe a signature like:

public static DMNMarshaller newMarshaller( DMNExtensionRegister... registers )

or

public static DMNMarshaller newMarshaller( List<DMNExtensionRegister> registers )

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
```diff
@@ -87,6 +88,8 @@
         implements DMNMarshaller {
 
     private static Logger logger = LoggerFactory.getLogger( XStreamMarshaller.class );
+    private DMNExtensionElementRegister extensionElementRegister;
```
etirelli, line 91: needs to be a list, as mentioned above.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
```diff
@@ -108,6 +111,15 @@ public StaxWriter createStaxWriter(XMLStreamWriter out, boolean writeStartEndDoc
         staxDriver.setRepairingNamespace(false);
     }
 
+    public XStreamMarshaller() {
+
+    }
+
+    public XStreamMarshaller (DMNExtensionElementRegister extensionElementRegister) {
```
etirelli, line 118: Needs to be a list.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -52,7 +55,19 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
             throws Exception {
 
         DMNCompiler dmnCompiler = DMNFactory.newCompiler();
-        DMNModel model = dmnCompiler.compile( resource );
+        String extensionRegisterClassName = ( (KnowledgeBuilderImpl) kbuilder ).getBuilderConfiguration().getChainedProperties().getProperty(DMN_EXTENSION_REGISTER,
```
etirelli, line 58: change to handle multiple registers as mentioned above and create a DMNCompilerConfiguration with it.

etirelli:
@akoufoudakis the overall idea is correct. Please take a look at my comments, please, and we need a few tests for this feature.

Thank you!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -52,7 +55,19 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
...
+                null);
+
+        DMNModel model = null;
+
+        if(extensionRegisterClassName != null && !extensionRegisterClassName.isEmpty()) {
+            DMNExtensionElementRegister extensionElementRegister =
+                    (DMNExtensionElementRegister)Class.forName(extensionRegisterClassName).newInstance();
```
tarilabs, line 65: I'm not 100% sure, I'm wondering if it would be better here to use the kbuilder classloader, something like:
```
( (KnowledgeBuilderImpl) kbuilder ).getRootClassLoader().loadClass( <name> ).newInstance();
```
@etirelli what do you think?

etirelli: Good point @tarilabs ! I agree. We might even have to try both, but certainly the kbuilder classloader first.

akoufoudakis:
Closing the pull request, since the review is complete. Thanks to reviewers.

PR #1187, opened 2017-04-03
[DO NOT MERGE] remove timeout to let jenkins hanging on MultithreadTe…

mariofusco:
…st.testConcurrentFireAndDispose

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
@@ -18,6 +18,20 @@
 
 package org.drools.compiler.integrationtests;
 
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
 import org.drools.compiler.CommonTestMethodBase;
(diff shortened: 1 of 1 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
```diff
@@ -815,7 +815,7 @@ public void run() {
         }
     }
 
-    @Test(timeout = 10000)
+    @Test//(timeout = 10000)
```
psiroky, line 818: Is this a good idea? It means if the test hangs, the whole build will timeout.

Could we perhaps plug into the jUnit and print the thread stack trances in case the tests timeouts? Something like https://github.com/windie/netty/commit/f0db2b1e1938bf5d7f3d6cbd0090512952b1ef17

tarilabs: For JUnit, I was led to believe the timeout callback is invoked after the timeout; this at least accordingly to documentation, but it might have been out of date then. I'll check this.

This PR was to test with manual ThreadDump using `/probe`, but appears I don't have authorization to access it on the PR builds?

tarilabs:
Jenkins test again
(trying to replicate the issues)

drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
psiroky: Ah, sorry, I haven't noticed the [DO NOT MERGE] sign. If its just for reproducing, then it is fine.

About the ThreadDump - the machines have the firewall enabled, so the remote connection will probably not work (at least not without further changes).

mariofusco:
ok to test

drools-compiler/src/test/java/org/drools/compiler/integrationtests/MultithreadTest.java
tarilabs: @psiroky I'm trying to address this with kiegroup/drools#1197

PR #1188, opened 2017-04-03
DROOLS-1505 Model cannot compile if an item definition is using anoth…

tarilabs:
…er item definition that is define after it


tarilabs:
@etirelli can you review, please?

etirelli:
@tarilabs I think the solution you propose looks fine as long as there are no cycles in the references. I wonder if we need to allow reference cycles? 

In any case, lets go ahead and merge this solution, and then if necessary to allow reference cycles we implement a new one. 

+1

etirelli (approved):

PR #1189, opened 2017-04-03
DROOLS-1508: prevent NPE when the dmn compiler fails to resolve a typeRef

etirelli:
@tarilabs this is in addition to your fix for the item definitions compilation


tarilabs (approved):
+1

PR #1190, opened 2017-04-04
DMN Validator tests part 1

baldimir:
First part of validator tests review:
- Validation rules and tests reorganized into packages (still some reorg left)
- Added new test cases. 

I tried to keep separate things in separate commits for review purposes. Next on todo list for this is to review validation of DTables, Expression extending classes, type references + some remaining less important model classes + some final global review if there isn't anything else missing. 

@etirelli @tarilabs could you please review this? Thanks.


kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-auth-req.drl
```diff
@@ -0,0 +1,60 @@
...
+end
+
+rule AUTHREQ_DEP_REQ_INPUT_NOT_INPUT
+when
+  $authReq : AuthorityRequirement(requiredInput != null,
+                                  $elemRef: requiredInput,
+                                  $targetId: rightOfHash( requiredInput.href))
+  DMNElement(id == $targetId, !(this instanceof InputData))
+then
+  reporter.report(DMNMessage.Severity.ERROR, $elemRef , Msg.REQ_DEP_INVALID_TYPE, $targetId, $authReq.getIdentifierString());
```
tarilabs, line 59: Because this message is the same, and because requiredInput, requiredDecision, requiredAuthority are all of type `DMNElementReference` I believe these 3 rules can be collapsed into one, using similar approach as per original DRL.

But as mentioned, nothing against having them explicited - just noted.

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-business-context.drl
```diff
@@ -0,0 +1,57 @@
...
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
+import org.kie.dmn.core.util.Msg;
+
+import function org.kie.dmn.validation.ValidatorUtil.rightOfHash;
+
+global org.kie.dmn.validation.MessageReporter reporter;
+
+rule ORG_UNIT_DECISION_MADE_WRONG_TYPE
+when
+    $d: OrganizationUnit(decisionMade != null,
```
tarilabs: can't recall by static looking into it, but is this null check really required?

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-business-context.drl
```diff
@@ -0,0 +1,57 @@
...
+                         $elemRef: /decisionMade)
+    $element: DMNElement(id == rightOfHash($elemRef.href), !(this instanceof Decision))
+then
+    reporter.report(DMNMessage.Severity.ERROR, $elemRef , Msg.REQ_DEP_INVALID_TYPE, $element.getId(), $d.getIdentifierString());
+end
+
+rule ORG_UNIT_DECISION_OWNED_WRONG_TYPE
+when
+    $d: OrganizationUnit(decisionOwned != null,
```
tarilabs: same question about the null check used in OOPath also here.

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-business-context.drl
```diff
@@ -0,0 +1,57 @@
...
+                         $elemRef: /decisionOwned)
+    $element: DMNElement(id == rightOfHash($elemRef.href), !(this instanceof Decision))
+then
+    reporter.report(DMNMessage.Severity.ERROR, $elemRef , Msg.REQ_DEP_INVALID_TYPE, $element.getId(), $d.getIdentifierString());
+end
+
+rule PERF_INDICATOR_IMP_DECISION_WRONG_TYPE
+when
+    $d: PerformanceIndicator(impactingDecision != null,
```
tarilabs: ... and here

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-context.drl
```diff
@@ -0,0 +1,54 @@
...
+import org.kie.dmn.core.util.Msg;
+
+import function org.kie.dmn.validation.ValidatorUtil.rightOfHash;
+
+global org.kie.dmn.validation.MessageReporter reporter;
+
+rule CONTEXT_DUP_ENTRY
+when
+  $ce1 : ContextEntry( parent instanceof Context, variable != null )
+  $ce2 : ContextEntry( variable != null, variable!.name == $ce1.variable!.name, parent == $ce1.parent,
```
tarilabs: I think these null check on `variable` field can be avoided given is later used with safe accessor `!.` ?

kie-dmn/kie-dmn-model/src/main/java/org/kie/dmn/model/v1_1/DMNModelInstrumentedBase.java
```diff
@@ -44,13 +44,13 @@ public String getIdentifierString() {
         }
     }
 
-    public DMNModelInstrumentedBase getParentDRGElement() {
-        if( this instanceof DRGElement ) {
-            return this;
-        } else if( this instanceof ItemDefinition && parent != null && parent instanceof Definitions ) {
-            return this;
+    public DMNElement getParentDRDElement() {
```
tarilabs: I know there is not 100% correspondence with "DRGElement" as in the original method name, and the `DRGElement` class (type), but I believe method was originally `getParentDRGElement()` to signify it is the parent in the DRG and not the DRD.

tarilabs (commented):
There are just some open questions about:

- `getParentDRGElement()` method name change
- Opportunities to collapse multiple rules into the same one, given the RHS is the same
- Not 100% sure some of the null check are really required, for instance when used in combination with OOPath or safe accessing `!.`.

... but nothing mandatory from my side.

Besides, thank you very much for the additional checks and the refactoring work.

kie-dmn/kie-dmn-model/src/main/java/org/kie/dmn/model/v1_1/DMNModelInstrumentedBase.java
etirelli: @tarilabs I discussed this with @baldimir before and the agreement was that the intent of the method is to return the top level element within the model (i.e. within <definitions>), so this change is fine.

@baldimir I just think that the return type will have to be DMNModelInstrumentedBase because that is the class that contains the getIdentifierString() method. DMNElement is a generated class, so we shouldn't add any method into it.

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/DMNValidatorImpl.java
```diff
@@ -77,7 +77,19 @@ public DMNValidatorImpl() {
...
+                ks.getResources().newClassPathResource("dmn-validation-rules-business-context.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-context.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-decision.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-dmnelementref.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-dtable.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-info-req.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-inputdata.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-know-req.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-know-source.drl", getClass() ),
+                ks.getResources().newClassPathResource("dmn-validation-rules-typeref.drl", getClass() ));
```
etirelli, line 92: @baldimir I really like the split you did here for the rules, we just need to double check it doesn't increase build time too much. But I like it! +1

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-business-context.drl
etirelli: I think the question here is should this rule be using OOPath at all? with a simple attribute assignment, shouldn't the pattern just be:

OrganizationUnit( decisionMade != null, $elemRef : decisionMade )

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-context.drl
etirelli: As long as the variable!=null on line 34 is kept, then yes, we can probably remove the one on line 35. I might have added the check myself, don't remember.

etirelli:
@baldimir I think the only thing we must check/fix for this PR is the return type of the getParentDRDElement() as per my comment. The other comments are nice to have things/fixes, you might want to consider.

kie-dmn/kie-dmn-model/src/main/java/org/kie/dmn/model/v1_1/DMNModelInstrumentedBase.java
baldimir: Agree. Changed that. Will push it with other fixes.

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-auth-req.drl
baldimir: I kept it separated, because each one needs to be checked for different type. One is KnowledgeSource, one is Decision and the last one is InputData.

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-business-context.drl
baldimir: It needs to be OOPath because we need to check if each member of decisionMade list is a Decision.

baldimir: And yes Matteo, the null check is not necessary there. I will remove it and similar ones.

kie-dmn/kie-dmn-validation/src/main/resources/dmn-validation-rules-context.drl
baldimir: Removed that. Will push with other fixes.

baldimir:
I made some changes according to comments. Please check the last commit and replies to comments. @etirelli I will check the build perf of the KieContainer with more rule files. Will keep you updated.

baldimir:
@etirelli, I "quickbenchmarked" the DMNValidatorImpl constructor and here are the results: 

Nonsplitted from master (so one rulefile, without this PR (less rules) ): 
DMNValidatorCreationBenchmark.testValidatorCreation 263.986 ± 5.424  ms/op

Splitted from this PR: 
DMNValidatorCreationBenchmark.testValidatorCreation  352.199 ± 7.623  ms/op

So the diff is around 90ms. So yes it is 90ms slower. Is this acceptable tradeoff for categorization of the rules? I think we will get to splitting the file in the future because the number of validation rules will grow in time, but maybe it could be split to few less files. What do you think?

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/ValidatorUtil.java
```diff
@@ -0,0 +1,46 @@
...
+    public static ItemDefinition getRootItemDef(final ItemDefinition id) {
+        ItemDefinition root = id;
+        while (!(root.getParent() instanceof Definitions)) {
+            root = (ItemDefinition) root.getParent();
+        }
+        return root;
+    }
+
+    public static String formatMessages(final List<DMNMessage> messages) {
+        return messages.stream().map(Object::toString).collect( Collectors.joining( "\n" ) );
```
triceo: If these are user-facing messages, shouldn't this be turned into platform-specific newlines?

baldimir: Good point. Done.

baldimir:
Fail not related to this PR.

baldimir:
jenkins retest this

etirelli:
> So the diff is around 90ms. So yes it is 90ms slower. Is this acceptable tradeoff for categorization of the rules? I think we will get to splitting the file in the future because the number of validation rules will grow in time, but maybe it could be split to few less files. What do you think?

I think for now it is acceptable. We will need to work on this perf to improve it, but it will have to be done at the core level I think. 

Thank you for benchmarking it.

etirelli (approved):
Excellent, thank you!

kie-dmn/kie-dmn-validation/src/test/java/org/kie/dmn/validation/ValidatorTypeRefTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.kie.dmn.api.core.DMNMessageType;
+
+public class ValidatorTypeRefTest extends AbstractValidatorTest {
+
+    @Test
+    public void testTYPEREF_NO_FEEL_TYPE() {
+        final List<DMNMessage> validate = validator.validate(
+                getReader("typeref/TYPEREF_NO_FEEL_TYPE.dmn"),
+                VALIDATE_SCHEMA, VALIDATE_MODEL, VALIDATE_COMPILATION);
+        assertThat(ValidatorUtil.formatMessages(validate), validate.size(), is(1));
```
yurloc, line 38: How about replacing this with AssertJ's `assertThat(validate).hasSize(1)`. It's easier to read and you don't need the `formatMessages()` utility method.

kie-dmn/kie-dmn-validation/src/test/java/org/kie/dmn/validation/ValidatorTypeRefTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertTrue(validate.stream().anyMatch(p -> p.getMessageType().equals(DMNMessageType.TYPE_REF_NOT_FOUND)));
```
yurloc, line 39: If moving to AssertJ, this can be written as:
```java
assertThat(validate).extracting(p -> p.getMessageType()).contains(DMNMessageType.TYPE_REF_NOT_FOUND);
```

yurloc (approved):

etirelli:
Jenkins test again

PR #1192, opened 2017-04-04
DROOLS-1477 DMN Extension Elements

akoufoudakis:
_**Please, review the following code. Update after the first review.**_

    To use extension elements inside inputData elements in order to provide “external” information: GEO location, time and location tracking of a decision making (when and were specific decision was taken). This is the primary scope of the use.

    In case extension element represents a complex type customer prefers to get back an instance of the complex type, assuming the complex type is defined as a class by the development team on the customer side. However, the customer also understands that it can be more complex and considers the possibility to get the complex type as an XML snippet. The parsing and turning the XML snippet into the complex type, defined by the customer.

    The possibility of getting the list of extension elements is expected from other DMN elements, which extend from DMNElement, according to the DMN specification and meta model diagram by OMG. In other words, the usage of extension elements is not limited to be a part of input data elements only.

    The client is not expecting “authoring” of the extension elements.
    Since extension elements are a part of DMN specification, this is an attempt to make it possible for drools to register custom extension elements in a generic way.


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/api/DMNFactory.java
```diff
@@ -29,4 +30,8 @@ public static DMNContext newContext() {
 
     public static DMNCompiler newCompiler() { return new DMNCompilerImpl(); }
 
+    public static DMNCompiler newCompilerWithConfiguration(DMNCompilerConfiguration dmnCompilerConfig) {
```
etirelli: We usually keep the method name the same, just overload it with the parameter:

```java
public static DMNCompiler newCompiler( DMNCompilerConfiguration config )
```

Also, can you please create a factory method for the config?

```java
public static DMNCompilerConfiguration newCompilerConfiguration()
```

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,30 @@ public ResourceType getResourceType() {
...
             throws Exception {
+        Map<String, String> extensionProperties = new HashMap<String, String>();
+        ( (KnowledgeBuilderImpl) kbuilder ).getBuilderConfiguration().getChainedProperties()
+                .mapStartsWith(extensionProperties, DMN_EXTENSION_REGISTER_PREFIX, false);
+        DMNCompiler dmnCompiler = null;
+        if(!extensionProperties.isEmpty()) {
+            List<DMNExtensionRegister> extensionRegisters = new ArrayList<DMNExtensionRegister>();
+            for(Map.Entry<String, String> extensionProperty : extensionProperties.entrySet()) {
+                String extRegClassName = extensionProperty.getValue();
+                DMNExtensionRegister extRegister = (DMNExtensionRegister)( (KnowledgeBuilderImpl) kbuilder ).getRootClassLoader()
```
etirelli: You will need to wrap this in a try/catch to avoid blowing up in case one of the classes can't be loaded.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,30 @@ public ResourceType getResourceType() {
...
+                        .loadClass(extRegClassName).newInstance();
+                extensionRegisters.add(extRegister);
+            }
+            DMNCompilerConfiguration compilerConfig = new DMNCompilerConfigurationImpl();
```
etirelli: ... then use the factory method here to instantiate the compiler configuration.

etirelli:
@akoufoudakis this is excellent, thank you! Just take a look at the minor changes in my comments and add the tests and we are good to go!

etirelli:
@akoufoudakis did you had time to create some tests for this feature?

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/DMNElementConverter.java
```diff
@@ -37,6 +38,9 @@ public DMNElementConverter(XStream xstream) {
     protected void assignChildElement(Object parent, String nodeName, Object child) {
         if ( DESCRIPTION.equals( nodeName ) && child instanceof String ) {
             ((DMNElement) parent).setDescription( (String) child );
+        } else if(EXTENSION_ELEMENTS.equals(nodeName)
+                && child instanceof DMNElement.ExtensionElements) {
+            ((DMNElement)parent).setExtensionElements((DMNElement.ExtensionElements)child);
```
tarilabs, line 43: as discussed, yes.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/ExtensionElementsConverter.java
```diff
@@ -25,30 +26,43 @@
...
-            reader.moveUp();
+        if(extensionRegisters.size() == 0) {
+            while (reader.hasMoreChildren()) {
+                reader.moveDown();
+                String nodeName = reader.getNodeName();
+                // skipping nodeName
+                reader.moveUp();
+            }
+        } else {
+            parseElements( reader, context, obj );
```
tarilabs, line 65: as suggested, yes.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/ExtensionElementsConverter.java
```diff
@@ -73,4 +87,10 @@ public boolean canConvert(Class clazz) {
         return clazz.equals( ExtensionElements.class );
     }
 
+    @Override
+    public void assignChildElement(Object parent, String nodeName, Object child) {
+        ExtensionElements id = (ExtensionElements)parent;
+        id.getAny().add(child);
```
tarilabs, line 93: correct, as suggested.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
```diff
@@ -335,10 +350,17 @@ private XStream newXStream() {
         xStream.registerConverter(new QNameConverter());
         xStream.registerConverter(new DMNListConverter( xStream ) );
         xStream.registerConverter(new ElementCollectionConverter( xStream ) );
-        xStream.registerConverter(new ExtensionElementsConverter( xStream ) );
+        xStream.registerConverter(new ExtensionElementsConverter( xStream, extensionRegisters ) );
```
tarilabs, line 352: correct.

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/compiler/extensions/dakkapel_uitvoering_en_conversie.xml
```diff
@@ -0,0 +1,1550 @@
...
+<definitions xmlns="http://www.omg.org/spec/DMN/20151101/dmn.xsd"
+	xmlns:biodi="http://bpmn.io/schema/dmn/biodi/1.0" id="definitions_0bexc14"
+	name="Dakkapel" namespace="http://camunda.org/schema/1.0/dmn"
+	xmlns:xhtml="http://www.w3.org/1999/xhtml" 
+	xmlns:content="http://www.example.org/Content"
+	xmlns:uitv="http://www.example.org/Uitvoeringsregels" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.omg.org/spec/DMN/20151101/dmn.xsd XSD/DMN.xsd
+             					 http://www.example.org/Uitvoeringsregels XSD/Uitvoeringsregel.xsd">
+	<!-- <extensionElements>
+		<uitv:uitvoeringsregels>
```
tarilabs: I think the question about this file - in general - is that seems to be related to a specific customer business logic. I'm not sure if it's meant to be kept internal and a good idea to have it in the upstream codebase. Possibly an "anonymized" file can be created or a simplified one?

akoufoudakis: We have a couple of simple files in our test cases: drools/kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/0001-filter.dmn. Can I add a couple of extension elements to a couple of InputData elements?

tarilabs: @akoufoudakis I think _better_ to add a new file, even just a copy and augment the copy, leaving the original sample files unchanged.
First, because this is a new feature which is unrelated with the existing files.
Second, I believe this would be a better approach because the current existing files are used in the unit test and I would leave them unchanged to more easily detect regression (shall a regression occur).

akoufoudakis: Sure... Of course. Do we care about complexity or string values will be fine?

kie-dmn/kie-dmn-core/src/test/java/nl/rws/dso/inception/backend/extension/ConversieregelRefRegister.java
```diff
@@ -0,0 +1,28 @@
...
+
+package nl.rws.dso.inception.backend.extension;
+
+import com.thoughtworks.xstream.XStream;
+import org.kie.dmn.api.marshalling.v1_1.DMNExtensionRegister;
+
+public class ConversieregelRefRegister implements DMNExtensionRegister {
+
+    public void registerExtensionConverters(XStream xStream) {
+        xStream.alias("conversieregelRef", String.class);
```
tarilabs: Not 100% what was the real intent here.
In the test, there appears to be an annotated `ConversieregelRef` class.
Here, `conversieregelRef` is aliased directly to a `java.lang.String`.
I think it depends what a "Conversieregel" is meant to be in the use-case demonstrated by this test:

1. a simple String, but then it cannot be differentiated inside a potential collection of heterogeneous extensionelements
2. a proper business domain object as represented by the `ConversieregelRef` class

If case n1, then there is no need for the `ConversieregelRef` class.
If the case n2, then here was possibly meant to be `processAnnotations` and the test case
https://github.com/kiegroup/drools/pull/1192/files#diff-a3b9b0a42509b9e6a4138c8b76cbe099R132
need to be aligned accordingly.

Hope this helps?

akoufoudakis: Yes... Anyway, it will be gone, because I will use a different file and different extension elements.

drools-persistence/drools-persistence-jpa/src/test/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<persistence version="2.1" xmlns="http://xmlns.jcp.org/xml/ns/persistence" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd">
+	<persistence-unit name="drools-persistence-jpa">
+	</persistence-unit>
+</persistence>
```
etirelli: I am confused. Why is this file part of this PR? I don't see the relationship with the extension elements support...

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,34 @@ public ResourceType getResourceType() {
...
+        } else {
+            dmnCompiler = DMNFactory.newCompiler();
+        }
+
+
+        DMNModel model = dmnCompiler.compile(resource);
 
-        DMNCompiler dmnCompiler = DMNFactory.newCompiler();
-        DMNModel model = dmnCompiler.compile( resource );
+        model = dmnCompiler.compile(resource);
```
etirelli: this can throw a null pointer exception. We need to decide if in case of errors loading extension classes we still create a compiler or we abort. In either case, this has to be handled here.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,34 @@ public ResourceType getResourceType() {
...
+                    String extRegClassName = extensionProperty.getValue();
+                    DMNExtensionRegister extRegister = (DMNExtensionRegister) ((KnowledgeBuilderImpl) kbuilder).getRootClassLoader()
+                            .loadClass(extRegClassName).newInstance();
+                    extensionRegisters.add(extRegister);
+                }
+                DMNCompilerConfiguration compilerConfig = DMNFactory.newCompilerConfiguration();
+                compilerConfig.addExtensions(extensionRegisters);
+                dmnCompiler = DMNFactory.newCompiler(compilerConfig);
+            } catch(ClassNotFoundException e) {
+                logger.error( "Trying to load a non-existing extension element register {}", e.getLocalizedMessage());
```
etirelli: In case of errors, we should also add a DMNKnowledgeBuilderError into the builder I believe.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+public class DMNExtensionRegisterTesting {
+
+    private void setProperties() {
```
etirelli: this method should be annotated with @Before instead of calling it from tests, otherwise and error might propagate between tests.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private Map<String, String> loadExtensionProperties() {
+        ChainedProperties props = new ChainedProperties("extension.properties", this.getClass().getClassLoader(), true);
+        Map<String, String> extensionClassNames = new HashMap<String, String>();
+        props.mapStartsWith(extensionClassNames, "org.kie.dmn.marshaller.extension.", false);
+        return extensionClassNames;
+    }
+
+    private void clearProperties() {
```
etirelli: this method should be annotated with @After instead of calling it from tests, otherwise and error might propagate between tests.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,34 @@ public ResourceType getResourceType() {
     @Override
     public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration)
             throws Exception {
+        Map<String, String> extensionProperties = new HashMap<String, String>();
+        ( (KnowledgeBuilderImpl) kbuilder ).getBuilderConfiguration().getChainedProperties()
+                .mapStartsWith(extensionProperties, DMN_EXTENSION_REGISTER_PREFIX, false);
+        DMNCompiler dmnCompiler = null;
+        if(!extensionProperties.isEmpty()) {
```
etirelli: Isn't this code reloading the extension registers for each resource in the kjar? I believe we need to move the creation of the compilerConfig (and possibly the compiler itself) into the class constructor and cache it. This way we can reuse it for each resource. Before we were not caching the compiler because the creation was very lightweight, but scanning the classpath to load classes every time is a really heavy operation.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -73,7 +82,14 @@ public DMNModel compile(Resource resource) {
     @Override
     public DMNModel compile(Reader source) {
         try {
-            Definitions dmndefs = DMNMarshallerFactory.newDefaultMarshaller().unmarshal( source );
+            Definitions dmndefs = null;
+            if(dmnCompilerConfig != null) {
+                if (dmnCompilerConfig.getRegisteredExtensions().isEmpty()) {
```
etirelli: this code seems wrong. If the dmnCompilerConfig is !=null but the list of extensions is not empty, it will not unmarshall the source.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+        } catch(ClassNotFoundException e) {
+            System.out.println( "Trying to load a non-existing extension element register " + e.getLocalizedMessage());
+        } catch(Exception e) {
+            System.out.println("Other exception");
+        }
+        return compilerConfig;
+    }
+
+    @Test
+    public void checkClassesLoaded() {
```
etirelli: Unfortunately this test is testing the test code (method loadDMNCompilerConfig()). Instead, I believe the loadDMNCompilerConfig() method should be completely removed from here and the test should instantiate the DMNAssemblerService and test that.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+        DMNCompilerConfiguration compilerConfig = loadDMNCompilerConfig();
+        assertEquals(2, compilerConfig.getRegisteredExtensions().size());
+        assertTrue(compilerConfig.getRegisteredExtensions().get(0) instanceof DMNExtensionRegister);
+        assertEquals(FirstNameDescriptionRegister.class, compilerConfig.getRegisteredExtensions().get(0).getClass());
+        clearProperties();
+    }
+
+    @Test
+    public void testFirstNameConverter() {
```
etirelli: similar to above, this method is testing the test code. We need to create the kjar (in memory as other tests do) and test that the model is properly loaded.

etirelli: If we need to support straight unmarshalling with models (without compilation), then we should move the code from the loadDMNCompilerConfig() into the DMNMarshallerFactory (or a helper class), so that it can be used elsewhere.

etirelli (changes requested):
Please see my comments.

etirelli:
Is there a JIRA for this work? I don't see it referenced in the commit messages.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+        ChainedProperties props = new ChainedProperties("extension.properties", this.getClass().getClassLoader(), true);
+        Map<String, String> extensionClassNames = new HashMap<String, String>();
+        props.mapStartsWith(extensionClassNames, "org.kie.dmn.marshaller.extension.", false);
+        return extensionClassNames;
+    }
+
+    private void clearProperties() {
+        System.clearProperty("foo");
+        System.clearProperty("org.kie.dmn.marshaller.extension.firstname");
+        System.clearProperty("org.kie.dmn.marshaller.extension.lastname");
```
triceo: Ideally, the property names would be externalized into a constant.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -50,9 +56,34 @@ public ResourceType getResourceType() {
     @Override
     public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration)
             throws Exception {
+        Map<String, String> extensionProperties = new HashMap<String, String>();
```
triceo: We're on Java 8 - `new HashMap<>()` will suffice.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
```diff
@@ -108,6 +113,16 @@ public StaxWriter createStaxWriter(XMLStreamWriter out, boolean writeStartEndDoc
         staxDriver.setRepairingNamespace(false);
     }
 
+    public XStreamMarshaller() {
+
+    }
+
+    public XStreamMarshaller (List<DMNExtensionRegister> extensionRegisters) {
+        this.extensionRegisters = new ArrayList<DMNExtensionRegister>();
+        this.extensionRegisters.addAll(extensionRegisters);
```
triceo: `new ArrayList<>(extensionRegisters)` will do the same thing, with the additional benefit of properly sizing the list.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+            for (Map.Entry<String, String> extensionProperty : extensionClassNames.entrySet()) {
+                String extRegClassName = extensionProperty.getValue();
+                DMNExtensionRegister extRegister = (DMNExtensionRegister) ((KnowledgeBuilderImpl) kbuilder).getRootClassLoader()
+                        .loadClass(extRegClassName).newInstance();
+                extensionRegisters.add(extRegister);
+            }
+            compilerConfig = DMNFactory.newCompilerConfiguration();
+            compilerConfig.addExtensions(extensionRegisters);
+        } catch(ClassNotFoundException e) {
+            System.out.println( "Trying to load a non-existing extension element register " + e.getLocalizedMessage());
```
triceo: Let's not use `System.out` - slf4j logging is preferred.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNCompilerConfiguration.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
```
yurloc: It's 2017.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTesting.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+public class DMNExtensionRegisterTesting {
```
yurloc: Let's rename this class to `DMNExtensionRegisterTest` to keep consistency.

For example, when one needs to search for a pattern in all test files, it's possible to do `grep --include=*Test.java foo` if all tests are named by this convention.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/ExtensionElementsConverter.java
```diff
@@ -25,30 +26,43 @@
...
 
-    /**
-     * Currently ignoring all extensionElements.
-     * Please note overriding {@link DMNBaseConverter#unmarshal(HierarchicalStreamReader, UnmarshallingContext)} in order to just skip all child elements of this extensionElements.
-     */
+    public ExtensionElementsConverter(XStream xStream, List<DMNExtensionRegister> extensionRegisters) {
+        super(xStream);
+        this.extensionRegisters = new ArrayList<DMNExtensionRegister>();
+        if(extensionRegisters != null && extensionRegisters.size() > 0)
+            this.extensionRegisters.addAll(extensionRegisters);
```
tarilabs: I'm reworking this code in the context of the class, also to avoid potential NPE if the extensionRegisters field might have not been initialized, from the no-args constructor.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
tarilabs: I think it's more dangerous the potential NPE if the extensionRegisters field might have not been initialized, from the no-args constructor; I'm reworking the code in this perspective.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: I think we discussed this previously and the decision was the logger-only, but I will consider this comment for the code rework.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: looks like line 86 is a dup of line 84 making the compilation twice? Will take into account into code rework,

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: I believe it cannot be done in the constructor as the `KnowledgeBuilder` is required to understand which are the classes to be loaded as extension, but I will take into account the caching requirement for the code rework.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNCompilerConfiguration.java
tarilabs: ok.

drools-persistence/drools-persistence-jpa/src/test/resources/META-INF/persistence.xml
tarilabs: erased, ok.

tarilabs:
Jenkins, test this please.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderConfigurationImpl.java
```diff
@@ -303,6 +303,11 @@ public void setProperty(String name,
             } catch (IllegalArgumentException e) {
                 log.warn("Invalid value " + value + " for option " + LanguageLevelOption.PROPERTY_NAME);
             }
+        } else {
+            // if the property from the kmodule was not intercepted above, just add it to the chained properties.
+            Properties additionalProperty = new Properties();
+            additionalProperty.setProperty(name, value);
+            chainedProperties.addProperties(additionalProperty);
```
tarilabs, line 310: @etirelli please notice as discussed this required the `KnowledgeBuilderConfiguration(Impl)` to stop ignoring not-known properties from the kmodule and inserting them inside the ChainedProperties.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTest.java
```diff
@@ -0,0 +1,175 @@
...
+            assertTrue(id2elements != null);
+            assertEquals(1, id2elements.getAny().size());
+            LastNameDescription lastNameDescription = (LastNameDescription) id2elements.getAny().get(0);
+            assertTrue(lastNameDescription.getContent().equals("Last name in latin characters"));
+        } catch (Exception e) {
+            LOG.error("{}", e.getLocalizedMessage(), e);
+            throw e;
+        } finally {
+            System.clearProperty("org.kie.dmn.marshaller.extension.firstname");
+            System.clearProperty("org.kie.dmn.marshaller.extension.lastname");
```
tarilabs, line 84: System properties is only used in this test.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTest.java
```diff
@@ -0,0 +1,175 @@
...
+        assertEquals("Last Name", inputData2.getName());
+        DMNElement.ExtensionElements id2elements = inputData2.getExtensionElements();
+        assertTrue(id2elements != null);
+        assertEquals(1, id2elements.getAny().size());
+        LastNameDescription lastNameDescription = (LastNameDescription) id2elements.getAny().get(0);
+        assertTrue(lastNameDescription.getContent().equals("Last name in latin characters"));
+    }
+    
+    @Test
+    public void testUsingKModuleProperty_WrongClasses() {
```
tarilabs, line 137: added test for demonstrate wrong configuration.

tarilabs:
Jenkins, test this please

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderConfigurationImpl.java
etirelli: ok

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/extensions/DMNExtensionRegisterTest.java
etirelli: ok

etirelli (approved):

PR #1196, opened 2017-04-06
DROOLS-1513 Evaluation result Type error check and reporting

tarilabs:


tarilabs:
Please note commit d0c712c is being reverted as part of this PR.

tarilabs:
Build 1009 failure is unrelated.
```
Test Result (1 failure / +1)
org.drools.compiler.integrationtests.MultithreadTest.testConcurrentFireAndDispose
```

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BaseDMNTypeImpl.java
```diff
@@ -128,7 +128,7 @@ public String toString() {
     @Override
     public boolean isInstanceOf(Object o) {
         if ( o == null ) {
-            return true;
+            return true; // null is instance of any type
```
psiroky: Isn't this counter intuitive? Looking at how Java itself handles null in `isInstanceOf` (http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.20.2): `At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and...`

tarilabs: Good question.

For Java I think it's a bit different because there is the concept of primitives Vs references which is impacting in the case.

But on second thought I'm not sure I implemented here what I really wanted, checking..

tarilabs: So if I go by the spec:

> Note that type is not mapped to the domain, and null is not the name of a type, and
**null** is not an instance of any type.

But then, `UNKNOWN` is the type of null, and this is contradicting the spec by the current master implementation.

Also if null is not an instance of any type, in the test case here: https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/0019-flight-rebooking.dmn#L261-L268
It would fail.

I think I need some wise @etirelli feedback on this one too, please :)

tarilabs:
To recap:

- I believe type check on Decision, BKM, Context should suffice without extending to other nodes. Although at present BKM type check is not really possible as the return type of a function is not inferred just yet.
- not yet convinced about null check, as pointed out in the comments.

etirelli:
+1

etirelli (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BaseDMNTypeImpl.java
etirelli: @psiroky @tarilabs this is a bit counter-intuitive and the specification unfortunately has some ambiguities. For the use case that we want to handle here, null is a valid value for any type, and so the instance of should work as defined by @tarilabs in the implementation.

tarilabs: I have done more work on this with #1214 .
Thanks for the feedback.

PR #1197, opened 2017-04-10
Perform thread dump from JUnit Listener when test timeout

tarilabs:


tarilabs:
First I'll demonstrate the thread dump with a deliberately failing deadlock test.

Following, I'll place another commit to remove the deliberately failing test and keep only the JUnit test listener modification.

tarilabs:
@mariofusco can you check the thread dump as generated now by the test listener the format contains relevant information; in the hope this can be an additional tool when a test timeout occurs only on CI, please?
Link: https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/drools-pullrequests/1012/artifact/drools-compiler/target/testStatusListener.11578rhel7large1957localdomain.305444842453.log/*view*/
halfway through the file
In this case the timeout is caused by the deliberate deadlock in the temporary test

@psiroky what do you think, please?

psiroky:
@tarilabs looks good, thanks! Thinking out loud - would it make sense to store the thread dumps in separate files, with the test name being part of the filename? It may just not be necessary though, since we won't (hopefully) need to diagnose the thread dumps often.

tarilabs:
@psiroky yes it's a good point and I though about it; I just preferred to use the current already existing log file for which proved to be quite reliable - I wanted also to minimize any additional IO operations, to have the best guarantee the dump will happen without errors, so relying on the already existing log files which mention the test sequence and used already by the listener, I think it's the safest approach.

PR #1203, opened 2017-04-13
[DROOLS-1523] defer retraction of tuples generated by expired events …

mariofusco:
…if there are setFocus action in the propagation queue


tarilabs (approved):
as discussed, agenda group entries should delay/defer the actual expiration of the tuples, as per test case demonstrated. +1

PR #1206, opened 2017-04-13
  [DROOLS-1523] defer retraction of tuples generated by expired events if there are setFocus action in the propagation queue

mariofusco:
…if there are setFocus action in the propagation queue (#1203)

[DROOLS-1523] defer retraction of tuples generated by expired events if there are setFocus action in the propagation queue
(cherry picked from commit d959f206857b434ff9651bb49196ab43435e335f)


mariofusco:
ok to test

mariofusco:
ok to test

PR #1207, opened 2017-04-13
[DROOLS-1524] expire events even if they are not used in any rule

mariofusco:


drools-core/src/main/java/org/drools/core/reteoo/ClassObjectTypeConf.java
```diff
@@ -82,9 +85,20 @@ public ClassObjectTypeConf(final EntryPointId entryPoint,
...
         this.typeDecl = kBase.getTypeDeclaration( clazz );
-        isEvent = typeDecl != null && typeDecl.getRole() == Role.Type.EVENT;
-        if (isEvent) {
-            expirationOffset = typeDecl.getExpirationOffset();
+        if (typeDecl != null) {
+            isEvent = typeDecl.getRole() == Role.Type.EVENT;
+            if (isEvent) {
+                expirationOffset = typeDecl.getExpirationOffset();
+            }
+        } else {
```
tarilabs, line 93: if class is not used in the DRL (neither as a `declare`)

tarilabs (approved):
+1. Example use case: use drools session just to keep last x hours of event. In this case "ttl" is not using a `declare` but annotation directly on the Pojo

PR #1214, opened 2017-04-18
DROOLS-1531 Fix instanceof accordingly to FEEL spec table 49 and Inst…

tarilabs:
…anceOfNode

Also just reference reminder that in Java:
```
null instanceof Object 
Object.class.isInstance(null)
```
are both `false` but:
```
Object value = null;
```
is valid.

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
@@ -64,4 +64,27 @@ public class MapBackedType
         }
         return true;
     }
+
+    @Override
+    public boolean isAssignableValue(Object value) {
+        if ( value == null ) {
+            return true;
+        }
+        Map<String, Object> instance = null;
+        try {
+            instance = (Map<String, Object>) value;
+        } catch ( Exception e ) {
+            return false;
+        }
+        for ( Entry<String, Type> f : fields.entrySet() ) {
+            if ( instance.get(f.getKey()) == null ) {
+                return false;
+            } else {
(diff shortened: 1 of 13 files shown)


tarilabs:
@etirelli kindly review when suitable please (not urgent)

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
```diff
@@ -64,4 +64,27 @@ public boolean isInstanceOf(Object o) {
...
+
+    @Override
+    public boolean isAssignableValue(Object value) {
+        if ( value == null ) {
+            return true;
+        }
+        Map<String, Object> instance = null;
+        try {
+            instance = (Map<String, Object>) value;
+        } catch ( Exception e ) {
```
etirelli, line 76: @tarilabs usually raising exceptions is expensive. Wouldn't be better to simply check for instanceOf map?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
```diff
@@ -64,4 +64,27 @@ public boolean isInstanceOf(Object o) {
...
+            return false;
+        }
+        for ( Entry<String, Type> f : fields.entrySet() ) {
+            if ( instance.get(f.getKey()) == null ) {
```
etirelli, line 80: really minor, but it would be good to avoid the double instance.get() call.

etirelli:
@tarilabs I will merge as is, but it would be good if you can check my comment about possibly unnecessarily raising the exception.

etirelli (approved):

PR #1215, opened 2017-04-18
[DROOLS-1532] fix removal of a LeftTuple from its FactHandle in case …

mariofusco:
…of rule deletion during incremental compilation

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/AddRemoveRule.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/AddRemoveRule.java
@@ -1064,7 +1064,6 @@ public class AddRemoveRule {
                 return;
             }
 
-            boolean   isRootLt = removingLt.getLeftParent() == null; // is the LT for the LIAN, if so we need to process the FH too.
             InternalFactHandle fh = removingLt.getFactHandle();
 
             // This is the first LT in a peer chain. Only this LT is hooked into the left and right parent LT and RT and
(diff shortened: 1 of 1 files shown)


drools-core/src/main/java/org/drools/core/phreak/AddRemoveRule.java
```diff
@@ -1116,10 +1114,9 @@ private static void deleteLeftTuple(LeftTuple removingLt, LeftTuple removingLt2,
                 if (leftParent.getLastChild() == removingLt) {
                     leftParent.setLastChild(nextPeerLt);
                 }
-            } else if ( isRootLt ) {
```
tarilabs, line 1119: removed spurious check - was always equivalent with else of the above branch without any check

drools-core/src/main/java/org/drools/core/phreak/AddRemoveRule.java
```diff
@@ -1116,10 +1114,9 @@ private static void deleteLeftTuple(LeftTuple removingLt, LeftTuple removingLt2,
...
-                if (fh.getFirstLeftTuple() == removingLt) {
-                    fh.setFirstLeftTuple(nextPeerLt);
-                }
```
tarilabs, line 1122: was missing further statements, with jira 1024

drools-core/src/main/java/org/drools/core/phreak/AddRemoveRule.java
```diff
@@ -1116,10 +1114,9 @@ private static void deleteLeftTuple(LeftTuple removingLt, LeftTuple removingLt2,
...
+            } else {
+                // is the LT for the LIAN, if so we need to process the FH too
+                fh.removeLeftTuple(removingLt);
```
tarilabs, line 1119: this method is delegated internally with all the required work.

tarilabs (approved):
+1

PR #1217, opened 2017-04-20
[DROOLS-1533] avoid need to explicitly bind a datasource to the execu…

mariofusco:
…tor when it is already programmatically injected into the ruleunit

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/RuleUnitTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/RuleUnitTest.java
@@ -22,6 +22,7 @@ import java.util.concurrent.Semaphore;
 
 import org.drools.compiler.Person;
 import org.drools.compiler.util.debug.DebugList;
+import org.drools.core.impl.InternalRuleUnitExecutor;
 import org.drools.core.ruleunit.RuleUnitFactory;
 import org.junit.Test;
 import org.kie.api.KieBase;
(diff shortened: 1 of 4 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/RuleUnitTest.java
```diff
@@ -104,13 +105,45 @@ public void testBindDataSource() throws Exception {
...
+
+        KieBase kbase = new KieHelper().addContent( drl1, ResourceType.DRL ).build();
+        RuleUnitExecutor executor = RuleUnitExecutor.create().bind( kbase );
+
+        DataSource<Person> persons = DataSource.create( new Person( "Mario", 42 ),
+                                                        new Person( "Marilena", 44 ),
+                                                        new Person( "Sofia", 4 ) );
+
+        // explicitly create unit
+        assertEquals(2, executor.run( new AdultUnit(persons) ) );
```
tarilabs, line 140: DataSource is assigned via RuleUnit constructor to ruleunit field, without the need of injection binding.

drools-core/src/main/java/org/drools/core/ruleunit/RuleUnitFactory.java
```diff
@@ -35,53 +38,66 @@ public RuleUnitFactory bindVariable( String name, Object dataSource ) {
...
+                if (existingValue == null || (existingValue instanceof Number && ( (Number) existingValue ).intValue() == 0)) {
+                    try {
                         field.set( ruleUnit, var );
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException( "Unable to inject field " + fieldName + " into " + ruleUnitClass.getName(), e );
                     }
-                } catch (IllegalAccessException e) {
-                    throw new RuntimeException( "Unable to inject field " + fieldName + " into " + ruleUnitClass.getName(), e );
+                }
+            } else if (var == null && DataSource.class.isAssignableFrom( field.getType() )) {
```
tarilabs, line 82: was-missing case, if the field DataSource of the ruleunit was already assigned for instance using constructor, then it just need to be "bindDataSource" for having the working memory injected inside the DataSource

tarilabs (approved):
+1

PR #1218, opened 2017-04-21
[DROOLS-1225] add options to specify the salience range of rules in a…

mariofusco:
… Decision Table Sheet


drools-decisiontables/src/main/java/org/drools/decisiontable/parser/DefaultRuleSheetListener.java
```diff
@@ -397,19 +403,16 @@ protected void postInitRuleTable(int row,
...
+    private boolean getFlagValue(String name, boolean defaultValue) {
+        return RuleSheetParserUtil.isStringMeaningTrue( getProperties().getSingleProperty( name, "" + defaultValue ) );
     }
 
-    private boolean getNumericDisabledFlag() {
-        final String numFlag = getProperties().getSingleProperty( NUMERIC_DISABLED_FLAG, "false" );
-        return RuleSheetParserUtil.isStringMeaningTrue( numFlag );
+    private int getNumericValue(String name, int defaultValue) {
+        try {
+            return Integer.parseInt( getProperties().getSingleProperty( name, "" + defaultValue ) );
```
manstis, line 412: If the User supplies a value ```x < Integer.MIN_VALUE || x > Integer.MAX_VALUE``` you're going to default; however it could lead the User to be confused as to why the generated salience values were not as expected. I'd be tempted to at least log the substitution of their _provided_ value with the _default_. TBH I thought you'd eluded to throwing a RTE when we were discussing on IRC.

drools-decisiontables/src/main/java/org/drools/decisiontable/parser/DefaultRuleSheetListener.java
```diff
@@ -681,7 +684,10 @@ private void nextDataCell(final int row,
     private Rule createNewRuleForRow(final int row, final String headCell, final String ruleCell ) {
         Integer salience = null;
         if ( this._currentSequentialFlag ) {
-            salience = new Integer( Rule.calcSalience( row ) );
+            salience = _currentSalience--;
+            if (salience < _minSalienceTag) {
+                throw new DecisionTableParseException( "Row number is above the max" );
```
manstis: I think the message is a little confusing.. Stating something more aligned with the issue would be better; for example ```"Salience less than the minimum specified on row " + row + "."```

manstis (commented):
Looks fine; other than a couple of small comments..

manstis (approved):

PR #1223, opened 2017-04-26
DROOLS-1531 Fix instanceof accordingly to FEEL spec table 49 (optimizations)

tarilabs:
@etirelli please consider this PR for the 2 remaining comments which were noted here: https://github.com/kiegroup/drools/pull/1214#pullrequestreview-34157657

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
@@ -47,19 +47,14 @@ public class MapBackedType
 
     @Override
     public boolean isInstanceOf(Object o) {
-        Map<String, Object> instance = null;
-        try {
-            instance = (Map<String, Object>) o;
-        } catch ( Exception e ) {
+        if ( o == null || !(o instanceof Map) ) {
             return false;
         }
+        Map<?, ?> instance = (Map<?, ?>) o;
         for ( Entry<String, Type> f : fields.entrySet() ) {
-            if ( instance.get(f.getKey()) == null ) {
+            Object instanceValueForKey = instance.get(f.getKey());
+            if ( instance.get(f.getKey()) == null || !f.getValue().isInstanceOf(instanceValueForKey) ) {
                 return false;
-            } else {
-                if ( !f.getValue().isInstanceOf(instance.get(f.getKey())) ) {
-                    return false;
-                }
             }
         }
         return true;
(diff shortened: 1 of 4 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
```diff
@@ -47,19 +47,14 @@ public MapBackedType addField(String name, Type type) {
...
-            instance = (Map<String, Object>) o;
-        } catch ( Exception e ) {
+        if ( o == null || !(o instanceof Map) ) {
             return false;
         }
+        Map<?, ?> instance = (Map<?, ?>) o;
         for ( Entry<String, Type> f : fields.entrySet() ) {
-            if ( instance.get(f.getKey()) == null ) {
+            Object instanceValueForKey = instance.get(f.getKey());
+            if ( instance.get(f.getKey()) == null || !f.getValue().isInstanceOf(instanceValueForKey) ) {
```
etirelli: @tarilabs is this correct? There are two cases, when a map backed instance does not have an a given attribute, and when it has the attribute but the value is null.
For instance, a Person instance might require the "Last Name" attribute to be considered a person, but the attribute might be set to null. It is still a person if the last name is null, but in duck typing it is not a person if the last name attribute does not exist.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java
```diff
@@ -70,19 +65,14 @@ public boolean isAssignableValue(Object value) {
...
-            instance = (Map<String, Object>) value;
-        } catch ( Exception e ) {
+        if ( !(value instanceof Map) ) {
             return false;
         }
+        Map<?, ?> instance = (Map<?, ?>) value;
         for ( Entry<String, Type> f : fields.entrySet() ) {
-            if ( instance.get(f.getKey()) == null ) {
+            Object instanceValueForKey = instance.get(f.getKey());
+            if ( instanceValueForKey == null || !f.getValue().isAssignableValue(instanceValueForKey) ) {
```
etirelli: same here

tarilabs:
@etirelli you are absolutely right, I'm just a little lost as I thought this is what I did originally with the two separate if checks, possibly I lost it even before during some refactoring.. checking.

tarilabs:
@etirelli I've updated following your comments, and the new test manifested an issue which was present on master even before (https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/MapBackedType.java#L30) , now corrected.

etirelli (approved):

etirelli:
Looks good @tarilabs ! Thanks!

PR #1224, opened 2017-04-26
[DROOLS-1537] avoid to keep a reference of the KnowledgeHelper inside…

mariofusco:
… the RuleExecutor

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/marshalling/MarshallingTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/marshalling/MarshallingTest.java
@@ -82,6 +82,7 @@ import org.kie.api.conf.DeclarativeAgendaOption;
 import org.kie.api.conf.EqualityBehaviorOption;
 import org.kie.api.conf.EventProcessingOption;
 import org.kie.api.io.ResourceType;
+import org.kie.api.marshalling.KieMarshallers;
 import org.kie.api.marshalling.Marshaller;
 import org.kie.api.marshalling.ObjectMarshallingStrategy;
 import org.kie.api.marshalling.ObjectMarshallingStrategyAcceptor;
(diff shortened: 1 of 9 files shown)


drools-core/src/main/java/org/drools/core/phreak/RuleExecutor.java
```diff
@@ -424,6 +421,7 @@ private void fireActivationEvent(InternalWorkingMemory wm, InternalAgenda agenda
 
     private void innerFireActivation( InternalWorkingMemory wm, InternalAgenda agenda, Activation activation, Consequence consequence ) {
         try {
+            KnowledgeHelper knowledgeHelper = agenda.getKnowledgeHelper();
```
tarilabs, line 424: the `knowledgeHelper` is simply retrieved from the agenda, instead of being an attribute of this class.

tarilabs (approved):
+1

drools-compiler/src/test/java/org/drools/compiler/integrationtests/marshalling/MarshallingTest.java
```diff
@@ -2864,12 +2865,17 @@ public void testMarshallWithTimedRuleRealClock() {
     }
 
     public static KieSession marshallAndUnmarshall(KieBase kbase, KieSession ksession, KieSessionConfiguration sessionConfig) {
+        return marshallAndUnmarshall(kbase, kbase, ksession, sessionConfig);
+    }
+
+    public static KieSession marshallAndUnmarshall(KieBase kbase1, KieBase kbase2, KieSession ksession, KieSessionConfiguration sessionConfig) {
```
tarilabs, line 2871: Changes and these methods are clear in the scope of the updated test, good, +1.

However I would suggest either of:

1. adding a javadoc comment we expect kbase1 and kbase2 parameter to be "equivalent" (because deserializing against a different kbase I understand is not a valid use-case)
2. if there is a way, possibly add a check in code

I'd settle for option n1.

PR #1226, opened 2017-04-26
GUVNOR-2980: Test scenarios do not show correct results

manstis:
See https://issues.jboss.org/browse/GUVNOR-2980

As far as I could tell; the issue was that the Rule names _expected_ to fire was "fully qualified names" whereas the information regarding which rules had _actually_ fired was "simple names". I've changed ```TestingEventListener``` to record "fully qualified names". 

I've also tested the rule filtering (see [here](https://github.com/manstis/drools/blob/fc84d7db0fe10cf583100737feb1b619ed8227c2/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java#L58)) and confirm it works as expected with "fully qualified names".

### DIFF
--- a/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java
+++ b/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java
@@ -58,7 +58,7 @@ public class TestingEventListener
                 if ( ruleNames.isEmpty() ) {
                     return true;
                 }
-                String ruleName = activation.getRule().getName();
+                String ruleName = getFullyQualifiedRuleName(activation.getRule());
 
                 if ( inclusive ) {
                     return ruleNames.contains( ruleName );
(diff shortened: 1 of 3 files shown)


manstis:
@Rikkola Please cast your expert judgement!

@etirelli This is the fix for the blocker; backport to 7.0.x on @Rikkola approval.

jomarko:
@manstis thanks for this PR, it seems good to me, anyway, this could solve also this problem: https://issues.jboss.org/browse/GUVNOR-2840 Am I right?

manstis:
Jenkins please retest this.

manstis:
@jomarko I assume it will fix that too; but will give it a try...

manstis:
@jomarko I can confirm with this fix https://issues.jboss.org/browse/GUVNOR-2840 is also fixed.

jomarko:
@manstis thanks, good job!

Rikkola (approved):
Heh, nice to see we had existing tests for this area.

etirelli:
@manstis thanks!

PR #1229, opened 2017-04-27
[DROOLS-1536] allow to cache compiler subcomponents into the Knowledg…

mariofusco:
…eBuilderImpl

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
@@ -16,6 +16,11 @@
 
 package org.kie.dmn.core.assembler;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
 import org.drools.compiler.compiler.PackageRegistry;
 import org.drools.compiler.lang.descr.PackageDescr;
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -56,13 +56,13 @@ public ResourceType getResourceType() {
     @Override
     public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration)
             throws Exception {
-        DMNCompiler dmnCompiler = getCompiler(kbuilder);
+        KnowledgeBuilderImpl kbuilderImpl = (KnowledgeBuilderImpl) kbuilder;
+        DMNCompiler dmnCompiler = kbuilderImpl.getCachedOrCreate( DMN_COMPILER_CACHE_KEY, () -> getCompiler( kbuilderImpl ) );
```
tarilabs, line 60: good for me, thanks! cc/ @etirelli

tarilabs (approved):
+1

etirelli:
@mariofusco this is not how I would have done it, but looks good to me! :)

etirelli:
@mariofusco BTW, thank you for doing this so quick... it is *really* important for the DMN compiler.

mariofusco:
@etirelli What did you have in mind? Let me know if you have any suggestion to improve this.

etirelli:
@mariofusco I think this should work fine, as I believe the kbuilder is scoped per kbase in the kjar, right? My usual pattern would be to add a "createContext()" method to the Assembler interface and then pass back the created context as parameter to each method call. Your solution is simpler and works fine as long as we guarantee the kbuilder lifecycle.

mariofusco:
@etirelli I was doing what you're also suggesting, but yes then I realized that we wanted something with same lifecycle of the kbuilder so I could just add the cache directly inside it. This also allowed to avoid the need of an additional argument in the assembler interface.

PR #1234, opened 2017-05-01
[DROOLS-1542] Root cause Exception is hidden in MVELExprAnalyzer.anal…

tkobayas:
…yzeExpression()

- unit test and candidate fix

The test method is marked as @Ignore because it's not actually a failing test (no assertion) and it pollutes stdout. Just to explain the use case.

Please have a look, thanks!


mariofusco:
Replaced and fixed by https://github.com/kiegroup/drools/pull/1250

PR #1235, opened 2017-05-02
DROOLS-1543 DMN ItemDefinition dependencies ordering

tarilabs:
Replace Java-api Comparator implementation
with approximation of topological sort.


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/ItemDefinitionDependenciesSorter.java
```diff
@@ -0,0 +1,108 @@
...
+                    return true;
+                }
+            } else if ( ic.getTypeRef().equals(qname2) ) {
+                return true;
+            }
+        }
+        return false;
+    }
+    
+    public static void displayDependencies(List<ItemDefinition> ins, String namespaceURI) {
```
etirelli, line 91: for "test" methods I usually prefer to return a string, instead of printing to sysout... but we can clean it after.

etirelli (approved):

PR #1236, opened 2017-05-02
DROOLS-1544: Incremental Compilation with GDST ResourceType fails

manstis:
Unit Test for https://issues.jboss.org/browse/DROOLS-1544


mariofusco:
Replaced and fixed by https://github.com/kiegroup/drools/pull/1247

PR #1237, opened 2017-05-02
Drools 1535 - improving error reporting for unknown variables

etirelli:
@tarilabs take a look at this one please


kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/unknown_variable2.dmn
```diff
@@ -0,0 +1,1152 @@
...
+      <semantic:requiredInput href="#_3c232867-aad6-4e35-a85b-bb32fb940502"/>
+    </semantic:informationRequirement>
+    <semantic:informationRequirement>
+      <semantic:requiredInput href="#_0f54224c-caa5-48ea-af01-4bf89cc18db3"/>
+    </semantic:informationRequirement>
+    <semantic:context id="_1dc484cc-c1cf-4856-9e05-a6300d7b5b95">
+      <semantic:contextEntry>
+        <semantic:variable id="_74675865-f7d9-4e6c-a4bf-2e927ae99205" name="Reserves Months" typeRef="feel:number"/>
+        <semantic:literalExpression id="_0b919e93-fafe-4ad7-9f6c-85afa597fac4">
+          <semantic:text>(Borrower.liquidAssetsAmt - Down Payment Amount)/Loan Payment</semantic:text>
```
tarilabs, line 358: @etirelli just to verify, the problem is that this `liquidAssetsAmt` is starting with a lower case rather than uppercase and should be reported as unknown variable, while if it was starting with capital `L` it would have been fine? (I'm proceeding on this with mentioned assumption, kindly let me know if otherwise).

etirelli: That is correct @tarilabs. You will see that for the other case I added a validation in the NameRefNode.evaluate() method, but this is not enough. We will need to be able to do proper resolution for the QualifiedNameRefNode as well and fixe the NameRefNode.evaluate() method to support it.

tarilabs: I've actually tried to address the issue in another way... but this uncovered a number of other little problems here and there :) which I've also fixed, with a last remaining issue which I'm not really sure how to adress and maybe is just better to discuss later this week by voice. I'll try to submit a PR against your branch later today, so you can check what I'm trying to do.

etirelli: @tarilabs ok, there is another thing we need to do: I only added validation to the decision table input expression. We need to add the same for the output to provide the users with compile time feedback on errors. Can you also look into that?

etirelli:
@tarilabs, this is the previous PR, just adding your fixes as well. Can you please review/approve?

tarilabs (approved):
Commit 313b884 appeared correctly, this should conclude this PR, ok +1

PR #1239, opened 2017-05-03
GUVNOR-2900: [Guided Decision Table] User is not prompted to save or discard changes

manstis:
See https://issues.jboss.org/browse/GUVNOR-2900

This PR adds automatically generated ```equals()``` and ```hashCode()``` methods to model classes. I allowed for any ```Object``` descendent field to have a possible ```null``` value (as, to be honest, it's safer to assume any could be ```null``` than fully analyse usage)


jomarko:
Thanks for looking @manstis . I thought according to previous similar issues that the problem will be hashCode & equals related.

manstis:
@karreiro Would you mind taking a look please (it's all code generated by IDEA).

karreiro (approved):
+1 Thanks @manstis

PR #1246, opened 2017-05-08
GUVNOR-2900: [Guided Decision Table] User is not prompted to save or discard changes

manstis:
Further to https://github.com/kiegroup/drools/pull/1239 ```equals()``` should delegate to super-classes


manstis:
@jomarko @karreiro Could you please review.

karreiro (approved):
+1 Thanks @manstis

PR #1247, opened 2017-05-08
[DROOLS-1544] ensure that ReturnValueRestrictionDescr has a reference to the resource from which it has been created

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/lang/descr/ReturnValueRestrictionDescr.java
```diff
@@ -28,36 +26,14 @@
...
+    public ReturnValueRestrictionDescr(String evaluator,
+                                       RelationalExprDescr relDescr,
+                                       Object content) {
         super( evaluator,
-               isNegated,
-               parameterText );
+               relDescr.isNegated(),
+               relDescr.getParametersText() );
         this.content = content;
+        setResource( relDescr.getResource() );
```
tarilabs, line 36: ok, +1

tarilabs (approved):

mariofusco:
ok to test

PR #1250, opened 2017-05-08
[DROOLS-1542] rewrite mvel error message in cause original message is null

mariofusco:


tarilabs:
```
org.drools.compiler.integrationtests.Misc2Test.testStackOverflowInMvel

java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.drools.compiler.integrationtests.Misc2Test.testStackOverflowInMvel(Misc2Test.java:9189)
```

drools-compiler/src/main/java/org/drools/compiler/rule/builder/dialect/mvel/MVELExprAnalyzer.java
```diff
@@ -130,6 +131,10 @@ public static MVELAnalysisResult analyzeExpression(final PackageBuildContext con
                                        parserContext1 );
         } catch ( Exception e ) {
             BaseDescr base = (context instanceof RuleBuildContext) ? ((RuleBuildContext)context).getRuleDescr() : context.getParentDescr();
+            if ( e instanceof CompileException && e.getCause() != null && e.getMessage().startsWith( "[Error: null]" )) {
+                // rewrite error message in cause original message is null
+                e = new CompileException(e.getCause().toString(), ( (CompileException) e ).getExpr(), ( (CompileException) e ).getCursor(), e.getCause());
+            }
```
tarilabs, line 137: ok, +1

tarilabs (approved):

PR #1252, opened 2017-05-08
DROOLS-1548 Fix bug with filter selection operator

tarilabs:


tarilabs:
```
Test Result (1 failure / +1)
org.drools.compiler.integrationtests.Misc2Test.testStackOverflowInMvel
```

Jenkins, test again :8ball:

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/QualifiedNameNode.java
```diff
@@ -53,7 +55,17 @@ public Object evaluate(EvaluationContext ctx) {
             if ( current != null ) {
                 for ( int i = 1; i < parts.size(); i++ ) {
                     String n = parts.get( i ).getText();
-                    current = EvalHelper.getValue( current, EvalHelper.normalizeVariableName( n ) );
+                    if ( current instanceof Collection ) {
```
etirelli, line 58: @tarilabs this looks fine, but I wonder if we don't need to also change the parser? If the attribute contains spaces in its name, does the current code works? E.g.: NBA Teams.Number of Victories

tarilabs: @etirelli it doesn't need change in parser as it works correctly thanks to a previous modification, here: https://github.com/kiegroup/drools/blame/master/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4#L242

And my understanding is we can't "resolve the scope" because the expression preceding the dot `.` it can be indeed an expression not just a variable name, hence the only way for now is to keep the dynamic resolution.

I will be adding a commit to demonstrate the compiler doesn't need modification to be consistent in this PR, won't hurt anyway :)

etirelli: Excellent, thank you!

etirelli (approved):

PR #1255, opened 2017-05-09
[DROOLS-1559] fix concurrent dispose

mariofusco:


drools-core/src/main/java/org/drools/core/common/DefaultAgenda.java
```diff
@@ -1334,7 +1334,7 @@ public void reset() {
 
         public boolean toFireAllRules() {
             synchronized (stateMachineLock) {
-                if (isFiring()) {
+                if (currentState.isFiring() || !currentState.isAlive()) {
```
tarilabs, line 1337: +1, as discussed this should avoid the CI random failure of `MultithreadTest.testConcurrentFireAndDispose` :8ball:

tarilabs (approved):

PR #1257, opened 2017-05-09
[DROOLS-1560] simplify and improve performances of ChainedProperties …

mariofusco:
…discovery


drools-core/src/main/resources/META-INF/drools.default.rulebase.conf
```diff
@@ -1,22 +0,0 @@
...
-drools.shareAlphaNodes = true
-drools.shareBetaNodes = true
-drools.alphaMemory = false
-drools.alphaNodeHashingThreshold = 3
-drools.compositeKeyDepth = 3
-drools.indexLeftBetaMemory = true
-drools.indexRightBetaMemory = true
-drools.equalityBehavior = IDENTITY
-drools.logicalOverride = DISCARD
-drools.executorService = org.drools.core.concurrent.DefaultExecutorService
```
tarilabs, line 16: All is good just wondering if https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/concurrent/ExecutorProviderImpl.java#L58 need any aligment following dropping `executorService` from the default properties
See also #852

I do NOT think so, but I preferred to be sure.

tarilabs (commented):
+1

General replacement of several `.conf` scattered different files with a general `ChainedProperties.createChainedProperties( getClassLoader() )`

tarilabs:
Although proposed changes looks good to me, you may prefer a couple additional eyes on this one, possibly @etirelli can give a second look, please? :)

etirelli (approved):

etirelli:
looks fine to me. +1

PR #1258, opened 2017-05-10
DROOLS-1550: adding support to boxed function definitions 

etirelli:
and also fixing the usage of FEEL function definitions in BKM nodes


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -208,23 +210,143 @@ private DMNExpressionEvaluator compileContext(DMNCompilerContext ctx, DMNModelIm
...
 
-        ctx.enterFrame();
-        try {
-            DMNExpressionEvaluatorInvokerFunction func = new DMNExpressionEvaluatorInvokerFunction( node.getName(), funcDef );
-            for ( InformationItem p : funcDef.getFormalParameter() ) {
-                DMNCompilerHelper.checkVariableName( model, p, p.getName() );
-                DMNType dmnType = compiler.resolveTypeRef( model, node, p, p, p.getTypeRef() );
-                func.addParameter( p.getName(), dmnType );
-                ctx.setVariable( p.getName(), dmnType );
+        String kindStr = funcDef.getOtherAttributes().get( FunctionDefinition.KIND_QNAME );
```
etirelli, line 213: @tarilabs we need to find a way to simplify this whole file. Maybe split it into multiple classes, each responsible to compile a different element, plus simplify the methods and consolidate the error handling. In any case, lets do it as a separate PR.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -208,23 +210,143 @@ private DMNExpressionEvaluator compileContext(DMNCompilerContext ctx, DMNModelIm
...
+                DMNFunctionDefinitionEvaluator func = new DMNFunctionDefinitionEvaluator( node.getName(), funcDef );
+                for ( InformationItem p : funcDef.getFormalParameter() ) {
+                    DMNCompilerHelper.checkVariableName( model, p, p.getName() );
+                    DMNType dmnType = compiler.resolveTypeRef( model, node, p, p, p.getTypeRef() );
+                    func.addParameter( p.getName(), dmnType );
+                    ctx.setVariable( p.getName(), dmnType );
+                }
+
+                DMNExpressionEvaluator eval = compileExpression( ctx, model, node, functionName, funcDef.getExpression() );
+                if( eval instanceof DMNLiteralExpressionEvaluator && ((DMNLiteralExpressionEvaluator)eval).isFunctionDefinition() ) {
```
etirelli, line 240: Basically, what this code does is: if the user declared a FEEL function (e.g., function(x,y) x+y), it eliminates the indirection, as the BKM is already a function definition in itself.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -208,23 +210,143 @@ private DMNExpressionEvaluator compileContext(DMNCompilerContext ctx, DMNModelIm
...
+                            method = stripQuotes( ((LiteralExpression) ce.getExpression()).getText().trim() );
+                        }
+                    }
+                }
+                if( clazz != null && method != null ) {
+                    String params = funcDef.getFormalParameter().stream().map( p -> p.getName() ).collect( Collectors.joining(",") );
+                    String expr = String.format( "function(%s) external { java: { class: \"%s\", method signature: \"%s\" }}", params, clazz, method );
+
+                    try {
+                        FEELFunction feelFunction = feel.evaluateFunctionDef( ctx, expr, model, funcDef,
```
etirelli, line 281: Same here, generates the function based on the context and evaluates it, eliminating the indirection

etirelli:
@tarilabs can you please review?

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/java_function_context.dmn
```diff
@@ -0,0 +1,67 @@
...
+          <semantic:literalExpression id="_83bc9702-62c7-4b8f-ba1a-e2a694cae9e8">
+            <semantic:text>"cos(double)"</semantic:text>
+          </semantic:literalExpression>
+        </semantic:contextEntry>
+      </semantic:context>
+    </semantic:encapsulatedLogic>
+    <semantic:variable id="_ad5ae2f9-f1bc-41e7-9f92-8b19ba462259" name="Cosinus1"/>
+  </semantic:businessKnowledgeModel>
+  <semantic:businessKnowledgeModel id="_9de2ce2e-82dd-44b9-9bc7-9e9fc3db7cac" name="Cosinus2" triso:displayName="Cosinus2">
+    <semantic:encapsulatedLogic xmlns:drools="http://www.drools.org/kie/dmn/1.1" id="_a72a7aff-48c3-4806-83ca-fc1f1fe34320">
```
tarilabs, line 30: unused namespace local definition, not a problem, just mentioned.

tarilabs (approved):
+1, looks good.

I have one remaining doubt what would happen if the function is defined inside a Decision, combined with the "indirection" described below, but given I expect this NOT to be common use case (if a realistic use-case at all) I'd suggest to move ahead with this PR.

etirelli:
@tarilabs , the indirection only occurs on BKM nodes because they define a function by themselves. So if their body also define a function, it is a function definition of a function definition. Decisions and Contexts would be immediately evaluated without any indirection, but you are right, I need to add a test just in case.

PR #1259, opened 2017-05-10
GUVNOR-3069: Guided Rule messing up operators during round trip marshalling

manstis:
See https://issues.jboss.org/browse/GUVNOR-3069

The unmarshalling code had a couple of problems:-

1. It incorrectly constructed a ```CompositeFieldConstraint``` even when the operators differed
2. It incorrectly split the content of "eval" into separate components even though they were part of the "eval"


manstis:

PR #1261, opened 2017-05-10
DROOLS-1550: Adding OSGi import into the pom file

etirelli:


etirelli:
@mfusco can you please review?

kie-dmn/kie-dmn-model/pom.xml
```diff
@@ -32,6 +32,7 @@
             <Import-Package>
               javax.xml,
               javax.xml.stream,
+              javax.xml.namespace,
```
tarilabs, line 35: +1, for `javax.xml.namespace.QName`

tarilabs (approved):

kie-dmn/kie-dmn-model/pom.xml
psiroky: It may be a good idea to use `*`. The plugin then automatically adds all the packages that are being imported/referenced.

PR #1263, opened 2017-05-12
RHBRMS-2815: add reproducer

winklerm:
Add reproducer for RHBRMS-2815

When a fact is inserted into an entry-point of a persistent KieSession and the session is then disposed and reloaded from the database, the fact is not present in the working memory.

When inserting the fact NOT into an entry-point, but into the KieSession directly, it is present in the working memory after the session is reloaded.


mariofusco:
Test case incorporated and fixed by https://github.com/kiegroup/drools/pull/1267

PR #1266, opened 2017-05-15
DROOLS-1550: downgrading severity and raising warnings in case a PMML…

etirelli:
… function definition is present in the model


etirelli:
@tarilabs can you please review? 

In case the model contains PMML function definitions we should only raise warnings at compile time, not errors. At runtime, if a PMML function is executed, then we raise an error and return null.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -329,6 +329,15 @@ private DMNExpressionEvaluator compileFunctionDefinition(DMNCompilerContext ctx,
                                        Msg.FUNC_DEF_BODY_NOT_CONTEXT,
                                        node.getIdentifierString() );
             }
+        } else if( kind.equals( FunctionDefinition.Kind.PMML ) ) {
```
tarilabs, line 332: ok

tarilabs (approved):

PR #1267, opened 2017-05-15
[RHBRMS-2815] fix insertion from entry-point in persistable ksession

mariofusco:


drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/session/ReloadSessionTest.java
```diff
@@ -195,4 +206,40 @@ public void testListenersAfterSessionReload() {
         assertEquals(1, ksession.getRuleRuntimeEventListeners().size());
         assertEquals(1, ksession.getAgendaEventListeners().size());
     }
+
+    @Test
+    public void testInsert() {
+        final Environment env = createEnvironment();
+        final KieBase kbase = initializeKnowledgeBase( RULE_WITH_EP );
```
tarilabs, line 213: Given the entry point is not used in this test, wasn't the original intention to use `simpleRule` instead ?

tarilabs: I don't believe would make a huge difference because anyway there is no fires and no assertion on rule fired, just mentioning as I've noticed.

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
```diff
@@ -376,7 +372,21 @@ public void endOperation(InternalKnowledgeRuntime kruntime) {
     }
 
     public RequestContext createContext() {
-        return RequestContext.create(ksession.getClass().getClassLoader()).with( this.ksession );
+        RequestContext context = RequestContext.create(ksession.getClass().getClassLoader()).with( this.ksession );
+        context.set( EntryPointCreator.class.getName(), new CommandBasedEntryPointCreator(runner) );
```
tarilabs, line 376: ensure new context has the 
```
context.set( EntryPointCreator.class.getName(), new CommandBasedEntryPointCreator(runner) );
```

tarilabs (approved):

PR #1268, opened 2017-05-16
[DROOLS-1564] change oopath syntax

mariofusco:


drools-compiler/src/main/resources/org/drools/compiler/lang/DRL6Expressions.g
```diff
@@ -566,12 +566,12 @@ xpathPrimary returns [BaseDescr result]
     ;
 
 xpathChunk returns [BaseDescr result]
-    : (xpathSeparator ID)=> xpathSeparator ID (DOT ID)* (LEFT_SQUARE DECIMAL RIGHT_SQUARE)? (LEFT_CURLY xpathExpressionList RIGHT_CURLY)?
+    : (xpathSeparator ID)=> xpathSeparator ID (DOT ID)* (HASH ID)? (LEFT_SQUARE xpathExpressionList RIGHT_SQUARE)?
```
tarilabs, line 569: - Using square brakets
- Inline cast comes out of it

tarilabs (commented):
+1, although to later evaluate possibility for 

```
/chunk[conditions...][index]
```

PR #1269, opened 2017-05-16
DROOLS-1563 Align InputData values check using BaseDMNTypeImpl.isAssi…

tarilabs:
…gnableValue

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
@@ -55,9 +55,11 @@ public class DMNFEELHelper
         EvaluationContextImpl ctx = new EvaluationContextImpl( null );
         try {
             ctx.enterFrame();
-            // need to set the values for in context variables...
-            for ( Map.Entry<String,Object> entry : dmnContext.getAll().entrySet() ) {
-                ctx.setValue( entry.getKey(), entry.getValue() );
+            if ( dmnContext != null ) {
+                // need to set the values for in context variables...
+                for ( Map.Entry<String,Object> entry : dmnContext.getAll().entrySet() ) {
+                    ctx.setValue( entry.getKey(), entry.getValue() );
+                }
             }
     
             for ( UnaryTest t : unaryTests ) {
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BaseDMNTypeImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BaseDMNTypeImpl.java
@@ -21,11 +21,14 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNType;
+import org.kie.dmn.core.compiler.DMNFEELHelper;
 import org.kie.dmn.feel.lang.Type;
 import org.kie.dmn.feel.lang.impl.MapBackedType;
 import org.kie.dmn.feel.lang.types.BuiltInType;
 import org.kie.dmn.feel.runtime.UnaryTest;
+import org.kie.dmn.feel.util.EvalHelper;
 
 public abstract class BaseDMNTypeImpl
         implements DMNType {
@@ -134,7 +137,7 @@ public abstract class BaseDMNTypeImpl
         if ( isCollection() && o instanceof Collection ) {
             Collection<Object> elements = (Collection) o;
             for ( Object e : elements ) {
-                if ( !internalIsInstanceOf(e) ) {
+                if ( !internalIsInstanceOf(e) || !valueMatchesInUnaryTests(e) ) {
                     return false;
                 }
             }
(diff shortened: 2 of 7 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
```diff
@@ -55,9 +55,11 @@ public static boolean valueMatchesInUnaryTests(List<UnaryTest> unaryTests, Objec
         EvaluationContextImpl ctx = new EvaluationContextImpl( null );
         try {
             ctx.enterFrame();
-            // need to set the values for in context variables...
-            for ( Map.Entry<String,Object> entry : dmnContext.getAll().entrySet() ) {
-                ctx.setValue( entry.getKey(), entry.getValue() );
+            if ( dmnContext != null ) {
+                // need to set the values for in context variables...
+                for ( Map.Entry<String,Object> entry : dmnContext.getAll().entrySet() ) {
+                    ctx.setValue( entry.getKey(), entry.getValue() );
```
tarilabs, line 61: as before, but avoid NPE for calls coming with null DMNContext (below)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BaseDMNTypeImpl.java
```diff
@@ -143,9 +146,17 @@ public boolean isInstanceOf(Object o) {
         // .. normal case, or collection of 1 element: singleton list
         // spec defines that "a=[a]", i.e., singleton collections should be treated as the single element
         // and vice-versa
-        return internalIsInstanceOf(o);
+        return internalIsInstanceOf(o) && valueMatchesInUnaryTests(o);
     }
     
+    private boolean valueMatchesInUnaryTests(Object o) {
```
tarilabs, line 152: new method because DMNType may impose domain range restriction, which was not really checked before.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/CompositeTypeImpl.java
```diff
@@ -82,41 +85,73 @@ public CompositeTypeImpl clone() {
...
+                if ( !instance.containsKey(f.getKey()) ) {
+                    return false; // It must have key named 'f.getKey()' like a Duck.
+                } else {
+                    if ( !f.getValue().isInstanceOf(instance.get(f.getKey())) ) {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        } else {
```
tarilabs, line 100: Either is a "MapBackedType" or either must provide generic accessor methods (for "JavaBackedType")

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/CompositeTypeImpl.java
```diff
@@ -82,41 +85,73 @@ public CompositeTypeImpl clone() {
...
+                if ( !instance.containsKey(f.getKey()) ) {
+                    return false; // It must have key named 'f.getKey()' like a Duck.
+                } else {
+                    if ( !f.getValue().isAssignableValue(instance.get(f.getKey())) ) {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        } else {
```
tarilabs, line 136: similarly as above

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -369,42 +369,9 @@ private boolean checkDependencyValueIsValid(DMNNode dep, DMNContext context) {
...
-                try {
-                    value = EvalHelper.getValue(v, f.getKey());
-                } catch (IllegalAccessException | InvocationTargetException e) {
-                    return false;
-                }
-                if ( !checkValueIsValid((BaseDMNTypeImpl) f.getValue(), value, context) ) {
-                    return false;
-                }
-            }
+            return dmnType.isAssignableValue( context.get( dep.getName() ) );
```
tarilabs, line 372: much simplified.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -512,11 +512,12 @@ public void testDateAndTime() {
         context.set( "Year", 1999 );
         context.set( "Month", 11 );
         context.set( "Day", 22 );
-        context.set( "oneHour", "PT1H" );
-        context.set( "durationString", "P13DT2H14S" );
+        context.set( "oneHour", Duration.parse( "PT1H" ) ); // <variable name="oneHour" typeRef="feel:days and time duration"/>
```
tarilabs, line 515: test input value was wrong.

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/two_decision_tables.dmn
```diff
@@ -119,6 +119,6 @@
     <semantic:variable id="_15ed0551-f2f8-462d-83be-0a5bbaab13d1" name="Decision Logic 1"/>
   </semantic:businessKnowledgeModel>
   <semantic:inputData id="_bd0c3ecc-845b-45d2-b87f-dfcb3bc68a30" name="Number">
-    <semantic:variable id="_6c3a2c35-42a6-4a1f-a524-962585f099c1" name="Number" typeRef="feel:string"/>
+    <semantic:variable id="_6c3a2c35-42a6-4a1f-a524-962585f099c1" name="Number" typeRef="feel:number"/>
```
tarilabs, line 122: Judging by the name given to this inputdata, how it's used in the decision table, and what is actually passed as input, it was wrong - corrected.

tarilabs (commented):
To be noticed that, consistently to other part of the implementation

- the user could deliberately set an inputdata in the context as null, and this would be a valid check against the inputdata type. Although this is a deliberate set. Please notice this is NOT about missing to set an inputdata value in the context at all, in such case a REQ_NOT_FOUND error is still thrown.
- inputdata type check now obey the same rules as per the remainer of the evaluation type check and specifically the "special case" of single-value-list-as-singleton is a valid case. (During DROOLS-1557 this case was deliberately omitted, at that time it was discussed to impose being a Collection https://github.com/kiegroup/drools/commit/b4a6408a916e210e188b41571d2bd17146cda1a5#diff-5dabeaeee0f1e3b8755a13e40ba3eea9R399 )

etirelli (approved):

etirelli:
Looks good @tarilabs +1

PR #1270, opened 2017-05-17
[DROOLS-1562] avoid NPE when trying to upgrade KieContainer to a non …

mariofusco:
…existing ReleaseId

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
@@ -56,6 +56,7 @@ import org.kie.api.KieBaseConfiguration;
 import org.kie.api.KieServices;
 import org.kie.api.builder.KieModule;
 import org.kie.api.builder.KieRepository;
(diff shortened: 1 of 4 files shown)


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -200,7 +201,12 @@ public ReleaseId getContainerReleaseId() {
     public Results updateToVersion(ReleaseId newReleaseId) {
         checkNotClasspathKieProject();
         Results results = update(((KieModuleKieProject) kProject).getInternalKieModule(), newReleaseId);
-        containerReleaseId = newReleaseId;
+        if (results != null) {
+            containerReleaseId = newReleaseId;
+        } else {
+            results = new ResultsImpl();
+            ( (ResultsImpl) results ).addMessage( Message.Level.ERROR, null, "Cannot find KieModule with ReleaseId: " + newReleaseId );
```
tarilabs, line 208: ok

tarilabs (approved):

PR #1272, opened 2017-05-18
DROOLS-1565 Implement a KieScanner listener

tarilabs:


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieScanner.java
```diff
@@ -25,10 +25,6 @@
 
 public interface InternalKieScanner extends KieScanner {
 
-    enum Status {
```
tarilabs, line 28: Moved up to `KieScanner`.

kie-ci/pom.xml
```diff
@@ -150,6 +150,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
tarilabs, line 155: This was a transitive dependency, now needed also directly.

tarilabs:
Depends on kiegroup/droolsjbpm-knowledge#232 .

kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java
```diff
@@ -224,6 +244,175 @@ public void testKScannerWithRange() throws Exception {
...
+        assertThat( events, CoreMatchers.hasItem( new KieScannerStatusChangeEventImpl(KieScanner.Status.SCANNING) ) );
+        assertThat( events, CoreMatchers.not( CoreMatchers.hasItem( new KieScannerStatusChangeEventImpl(KieScanner.Status.UPDATING) )) );
+        events.clear();
+        
+        repository.installArtifact(releaseId1, kJar1, createKPom(fileManager, releaseId1));
+
+        KieSession ksession = kieContainer.newKieSession("KSession1");
+        checkKSession(ksession, "rule1", "rule2");
+
+        // Create faulty KJAR for
```
tkobayas: Just to understand the test case... This is not a faulty KJAR, right?

tkobayas (approved):
Thank you @tarilabs , that looks great. A client can get the result from the event listener.

kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java
tarilabs: Well spotted, I must inadvertently placed this comment in the wrong place. Thank you!

PR #1278, opened 2017-05-19
DROOLS-1569 Enrich DMN InputData node type information

tarilabs:
Depends on #1269 for DROOLS-1563 fix.

### DIFF
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNType.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNType.java
@@ -16,6 +16,7 @@
 
 package org.kie.dmn.api.core;
 
+import java.util.List;
 import java.util.Map;
 
 public interface DMNType
@@ -50,4 +51,6 @@ public interface DMNType
      * @return if value can be assigned to the type represented by this type. If the parameter is null, returns true. 
      */
     boolean isAssignableValue(Object value);
+
+    List<DMNUnaryTest> getAllowedValues();
 }
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNUnaryTest.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNUnaryTest.java
@@ -14,14 +14,8 @@
  * limitations under the License.
  */
 
-package org.kie.dmn.feel.runtime;
+package org.kie.dmn.api.core;
 
-import java.util.function.BiFunction;
-
-import org.kie.dmn.api.core.DMNUnaryTest;
-import org.kie.dmn.feel.lang.EvaluationContext;
-
-@FunctionalInterface
-public interface UnaryTest extends DMNUnaryTest, BiFunction<EvaluationContext, Object, Boolean> {
+public interface DMNUnaryTest {
 
 }
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
@@ -334,10 +334,13 @@ public class DMNCompilerImpl
                 if( topLevel || allowedValuesStr != null || itemDef.isIsCollection() != type.isCollection() ) {
 
                     // we have to clone this type definition into a new one
+                    BaseDMNTypeImpl baseType = type;
                     type = type.clone();
 
+                    type.setBaseType( baseType );
                     type.setNamespace( dmnModel.getNamespace() );
                     type.setName( itemDef.getName() );
+                    type.setAllowedValues(null);
                     if ( allowedValuesStr != null ) {
                         List<UnaryTest> av = feel.evaluateUnaryTests(
                                 ctx,
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
@@ -45,16 +45,16 @@ public class SimpleTypeImpl
     }
 
     public BaseDMNTypeImpl clone() {
-        return new SimpleTypeImpl( getNamespace(), getName(), getId(), isCollection(), getAllowedValues(), getBaseType(), getFeelType() );
+        return new SimpleTypeImpl( getNamespace(), getName(), getId(), isCollection(), getAllowedValuesFEEL(), getBaseType(), getFeelType() );
     }
 
     @Override
     protected boolean internalIsInstanceOf(Object o) {
-        return getFeelType().isInstanceOf(o);
+        return getBaseType() != null ? getBaseType().isInstanceOf(o) : getFeelType().isInstanceOf(o);
     }
 
     @Override
     protected boolean internalIsAssignableValue(Object o) {
-        return getFeelType().isAssignableValue(o);
+        return getBaseType() != null ? getBaseType().isAssignableValue(o) : getFeelType().isAssignableValue (o);
     }
 }
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/UnaryTestImpl.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/UnaryTestImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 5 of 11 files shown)


kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNType.java
```diff
@@ -50,4 +51,6 @@
      * @return if value can be assigned to the type represented by this type. If the parameter is null, returns true. 
      */
     boolean isAssignableValue(Object value);
+
+    List<DMNUnaryTest> getAllowedValues();
```
tarilabs, line 55: the new method.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNUnaryTest.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.dmn.api.core;
+
+public interface DMNUnaryTest {
```
tarilabs, line 19: Marker interface (because ExecutionContext is at FEEL layer, so I cannot "drag" UnaryTest up to the DMN layer from the FEEL layer)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -334,10 +334,13 @@ private DMNType buildTypeDef(DMNCompilerContext ctx, DMNFEELHelper feel, DMNMode
                 if( topLevel || allowedValuesStr != null || itemDef.isIsCollection() != type.isCollection() ) {
 
                     // we have to clone this type definition into a new one
+                    BaseDMNTypeImpl baseType = type;
                     type = type.clone();
 
+                    type.setBaseType( baseType );
                     type.setNamespace( dmnModel.getNamespace() );
                     type.setName( itemDef.getName() );
+                    type.setAllowedValues(null);
```
tarilabs, line 343: cloning actually clones the allowedValues as well, but that is wrong in this case.

tarilabs:
Please notice only commit a4b1f6a is actually of this PR, the other commit is inherited from another open outstanding PR.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
etirelli: Is it wrong? I wonder if there are further details on the spec about it.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
```diff
@@ -45,16 +45,16 @@ public boolean isComposite() {
...
 
     public BaseDMNTypeImpl clone() {
-        return new SimpleTypeImpl( getNamespace(), getName(), getId(), isCollection(), getAllowedValues(), getBaseType(), getFeelType() );
+        return new SimpleTypeImpl( getNamespace(), getName(), getId(), isCollection(), getAllowedValuesFEEL(), getBaseType(), getFeelType() );
     }
 
     @Override
     protected boolean internalIsInstanceOf(Object o) {
-        return getFeelType().isInstanceOf(o);
+        return getBaseType() != null ? getBaseType().isInstanceOf(o) : getFeelType().isInstanceOf(o);
```
etirelli, line 53: I am confused by this. The current type might change the properties of a base type, like defining allowed values. Is simply delegating to the base type `isInstanceOf()` method correct? don't we need to check the current type isInstanceOf()?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
```diff
@@ -45,16 +45,16 @@ public boolean isComposite() {
...
     }
 
     @Override
     protected boolean internalIsAssignableValue(Object o) {
-        return getFeelType().isAssignableValue(o);
+        return getBaseType() != null ? getBaseType().isAssignableValue(o) : getFeelType().isAssignableValue (o);
```
etirelli, line 58: same as above

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/UnaryTestImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.dmn.feel.runtime;
+
+import org.kie.dmn.feel.lang.EvaluationContext;
+
+public class UnaryTestImpl implements UnaryTest {
```
etirelli, line 21: Ok, I like this approach.

etirelli:
@tarilabs I think overall it looks good. Can you please take a look at the 2 questions I posted?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
tarilabs: I'll try to detail further by example.

- You have an ItemDefinition for "tMembership" based on feel:string, and allowed values "Gold", "Silver".

- Then the compiler start to build another ItemDefinition for "tMembershipList", not a composite type a simple type based on namespace:tMembership, and `isCollection=true`.

- It would reach the clone call at line 338 here.

Because of the call to clone(), the cloned SimpleType also received the allowedValues.
So it's like saying the tMembershipList ItemDefinition had the `<allowedValue>` tag itself, but on the original XML it was not the case, only the referenced tMembership had it.

Anyway I take your point, I will double check on the specs for this on Monday

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
tarilabs: As per DROOLS-1563 which this PR is based, that is managed at the `BaseDMNTypeImpl`, see here https://github.com/kiegroup/drools/pull/1278/files/a4b1f6a6350722f9ec4a3c20f54efd2aa15a641a#diff-5fab9456cc90d41195694cfad48e7955R158

in other words, the isInstance must guarantee first is an instance and also in the restriced domain by allowed values.

I take your point however to check what happens if I have a "layered" definition of simple types, like
tMembership based on feel:string having as allowed values "Gold", "Silver"
and
tMembershipGold based on tMembership having as allowed values "Gold"

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
etirelli: ok, lets move forward as it is here, and check for further changes that might be necessary later.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/SimpleTypeImpl.java
etirelli: ok, lets move forward with this as is and discuss further changes later.

etirelli (approved):

etirelli:
@tarilabs can you please rebase this branch to remove the conflicts?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/UnaryTestImpl.java
tarilabs: It's @mariofusco helpful feedback :) as I needed also his suggestion if there was any better way than the "generic erasure" trick used here: https://github.com/kiegroup/drools/pull/1278/files/a4b1f6a6350722f9ec4a3c20f54efd2aa15a641a#diff-5fab9456cc90d41195694cfad48e7955R102

tarilabs:
rebased. Before merging, kindly double check if `DMNUnaryTest` interface in DMN-API is okay being an empty marker-interface, basically, please?
Thanks!

etirelli:
The problem is that as a functional interface, we can't really add multiple methods to the DMNUnaryTest interface, can we? it would be nice to have a getText() that is just an explicit method instead of relying on toString(), but we can change that later.

tarilabs:
@etirelli didn't think about using default method for it, but I verified locally it works, i.e.:

```
public interface DMNUnaryTest {

    default String getText() {
        return toString();
    }
}
```

at this point `UnaryTestImpl` may redefine the getText() to just return the text (although currently there's no diff with toString().

Sorry I didn't think about this earlier, I'll place it in one of the next PRs.

PR #1284, opened 2017-05-25
DROOLS-1580 FEEL for loop expression is not type aware

tarilabs:


etirelli:
@tarilabs excellent. Long term we really need full type resolution, but this will work fine for the short term!

etirelli (approved):

PR #1285, opened 2017-05-25
[DROOLS-1581] avoid unnecessary class loader lookup during eval invok…

mariofusco:
…er bytecode generation


tarilabs (approved):

tarilabs (approved):
In this case is not good to lookup, because the searched class does not exist yet, and when it can be simply inferred/calculated.

PR #1294, opened 2017-05-31
GUVNOR-2815: DefaultValue offers items compatible with field data type

jomarko:


drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/oracle/DataType.java
```diff
@@ -56,30 +59,57 @@
         NUMERIC_LONG,
         NUMERIC_SHORT,
         DATE,
-        BOOLEAN
+        BOOLEAN;
+
+        public boolean isValidValue(String value) {
+            try {
+                if (this == NUMERIC_BIGDECIMAL) {
```
triceo: `isValidValue()` anc `isNumeric()` could be abstract on the `enum`, and therefore each instance could have its own implementation. Arguably, it could be nicer code, avoiding these monster `if`s and the potential bug when a new type is added without updating the `if`s

Although I could also see how that kind of code could be considered worse, due to all the `@Override`s. :-) So just something to consider.

jomarko: I admit it is not the best solution. I will try to come up with nicer one.

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/oracle/DataType.java
```diff
@@ -45,41 +52,157 @@
...
+                } catch (NumberFormatException nfe) {
+                    return false;
+                }
+            }
+        },
+        DATE {
+            @Override
+            public boolean isValidValue(String value) {
+                try {
+                    SimpleDateFormat format = new SimpleDateFormat(ApplicationPreferences.getDroolsDateFormat());
```
jomarko: Hi @manstis , please should I check on this place also by format stored in `ApplicationPreferences.getDroolsDateTimeFormat()`?

jomarko:
jenkins please retest this

drools-workbench-models/drools-workbench-models-datamodel-api/pom.xml
```diff
@@ -69,10 +69,34 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.kie.workbench.services</groupId>
```
manstis: @jomarko We choose to keep ```drools-workbench-models``` completely devoid of Errai or GWT etc dependencies (i.e. anything considered "client-side"). The inclusion of the "client-side" dependencies here could be leading to the problem with ```kie-maven-plugin``` compilation that appears to have discovered some Annotation Processors for client-side code. ```kie-maven-plugin``` depends on these "model" JARs to generate DRL from (for example) a Guided Decision Table XML file. You might need to check what transitive dependencies you now bring in and exclude them...

jomarko:
jenkins please retest this

drools-workbench-models/drools-workbench-models-datamodel-api/pom.xml
jomarko: Thanks for explanation @manstis. However I don't like this 'exclusion' workaround, because whenever someone add dependency into 'kie-wb-common-services-api' it could broke build of whole workbench if he doesn't add exclusion here. I think I have two options:
- I will close this PR and do changes just in drools-wb
- Dependencies were added just because of the `Date` so I can return always true for `Date.isValidValue` and check more comprehensive check in drools-wb.

What do you think?

manstis: @jomarko Personally, I find keeping this PR open with ```Date.isValidValue(String)``` always returning ```true``` to be more problematic (or more likely to be problematic in the future) than the ```pom.xml``` exclusions :-( 

I think we need to either find a way to "inject" a validator for ```Date```, e.g. see usage of  ```org.drools.workbench.models.datamodel.oracle.DateConverter``` (and the pain it brings); or (more simply) just close this PR and keep the validation closer to the Guided Decision Tables' ```ColumnUtilities```. 

WDYT?

jomarko:
I decided for closing this PR.

PR #1295, opened 2017-06-01
[JBPM-6038] OptimisticLockRetryInterceptor doesn't retry 3 times

tkobayas:
This PR doesn't contain a unit test because JpaOptLockPersistentStatefulSessionTest in master doesn't seem to work now (DROOLS-1393). Instead, I attached a unit test diff for branch 6.5.x to https://issues.jboss.org/browse/JBPM-6038

If you think this PR should contain the unit test (even though I cannot verify), please let me know, I will add it and revise this commit.

### DIFF
--- a/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/OptimisticLockRetryInterceptor.java
+++ b/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/OptimisticLockRetryInterceptor.java
@@ -89,11 +89,11 @@ public class OptimisticLockRetryInterceptor extends AbstractInterceptor {
     }
 
     protected RequestContext internalExecute( Executable executable, RequestContext ctx ) {
-        int attempt = 1;
+        int attempt = 0;
         long sleepTime = delay;
         RuntimeException originException = null;
 
-        while (attempt <= retries) {
+        while (true) {
             if (attempt > 1) {
                 logger.trace("retrying (attempt {})...", attempt);
             }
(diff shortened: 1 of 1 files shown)


mariofusco:
@tkobayas The break you added is ok, so it doesn't increment the attempt variable again and doesn't uselessly sleep one more time before exiting, but by adding that break you could also change the loop condition with a simple while(true). I'm less convinced regarding the +1, in my opinion the number of retries should also include the first attempt. In other words I believe it should 3 attempts in total is it does now and not 4. But I admit this is more a semantic issue and I'm not strongly opinionated about this. What do you think?

@mswiderski What is your opinion?

tkobayas:
@mariofusco Thank you for the review.

> you could also change the loop condition with a simple while(true)

Sure. I'll change it.

> I'm less convinced regarding the +1, in my opinion the number of retries should also include the first attempt.

Hmm, you meant if OptimisticLockRetryInterceptor.setRetries(1), it should not retry (= 1 attempt)?

mswiderski:
In general, I would say that retries should be counted from the time it failed, so max 4 times - one initial and then three retry attempts.

Looking at the fix wouldn't it be easier to change the initial attmpt variable to 0 here https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/OptimisticLockRetryInterceptor.java#L92

That would be clear and the rest can be as is - maybe include the if to avoid useless sleep...

mariofusco:
@mswiderski  @tkobayas ok, I agree.

@tkobayas so I'd change this

1. change the initial value of attempt variable to 0
2. change the while condition to just while(true) { ... }
3. remove the +1 and move the break after attempt++

Do you agree?

tkobayas:
Hi, please have a look at the revised commit.

@mariofusco

> 1.) change the initial value of attempt variable to 0
> 3.) remove the +1 and move the break after attempt++

If I apply both, again it results in 3 attempts (= 2 retries). For the step 3, it should be "remove the +1" OR "move the break after attempt++".

In the revised commit, I chose the latter "move the break after attempt++". Then I use "(attempt > retries)" which is almost equivalent to "attempt == retries + 1" (but for a clearer condition). This results in 4 attempts (= 3 retries). Could you please confirm?

PR #1296, opened 2017-06-01
[DROOLS-1592] remove referencing of left tuple from propagation context

mariofusco:


tarilabs (approved):
Substitutes the LeftTuple with the only actual information needed, the RTN. ok.

dsukhoroslov:
Hi @mariofusco,
Any chance to have this fix in 6.5.0.x too? We currently trying to move on 7.3.x version, but it is not easy at all. Many rules which were working properly in 6.5.0 just does not work in 7.2.0 and 7.3.0..

Thanks, Denis

mariofusco:
Hi @dsukhoroslov,
the 6.x series has to be considered closed, we won't deploy another community release out of it. Which problems are you having with drools 7?

dsukhoroslov:
Ok, I see. My colleague just opened an issue: https://issues.jboss.org/browse/DROOLS-1737

PR #1293, opened 2017-05-30
DROOLS-1607 Internal refactor removing KnowledgeBase in preparation for DROOLS-1583

tarilabs:
step1

Passes just src main compilation with :
```
$ mvn clean compile -pl \!drools-examples
```

requires kiegroup/droolsjbpm-knowledge#237

PR #1298, opened 2017-06-02
DROOLS-1594 MiscTest class removal

baldimir:
- Tests moved to separate new or existing test classes
- MiscTest class deleted 

Please don't mind commits messages, just squash when everything is all right.


mariofusco:
@baldimir Great initiative but bad timing :( I'm afraid this pull request will badly conflict with this @tarilabs 's refactor https://github.com/kiegroup/drools/pull/1293

@tarilabs what do you think? Should I merge this PR? Will you be able to reapply your refactor after it?

baldimir:
I will merge the conflicting files also if asteroids will fall (I mean I will solve the conflicts if this is merged after Matteo's PR) :) @mariofusco @tarilabs we can discuss this next week. I see that there are changes related to KnowledgeBase => KieBase and StatefulKnowledgeSession => KieSession changes. I did the same in this PR. I tried to change all references to old classes with the new ones. So maybe there wouldn't be much aligning.

tarilabs:
@baldimir let us know if indeed it should be easier for you to realign the "original" MiscTest in case versions differs (in fact, on my PR it should already be matching with the baseline which will be expected on your PR, I think)
Otherwise, given the number of commits, I think the opposite is true - that is, we should match this PR first (which has semantic changes), and realign my PR later (which is just snake oil)
What do you think?

PR #1299, opened 2017-06-02
GUVNOR-3134: Test Scenarios fail after upgrade to 7.1 because rule expectation in 6.x doesn't use FQN

manstis:
See https://issues.jboss.org/browse/GUVNOR-3134

### DIFF
--- a/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java
+++ b/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java
@@ -44,52 +44,53 @@ public class TestingEventListener
         implements
         AgendaEventListener {
 
-    final Map<String, Integer> firingCounts = new HashMap<String, Integer>( 100 );
+    final Map<String, Integer> firingCounts = new HashMap<String, Integer>(100);
 
     long totalFires;
 
     public TestingEventListener() {
     }
 
-    public AgendaFilter getAgendaFilter( final HashSet<String> ruleNames,
-                                         final boolean inclusive ) {
+    public AgendaFilter getAgendaFilter(final HashSet<String> ruleNames,
+                                        final boolean inclusive) {
         return new AgendaFilter() {
-            public boolean accept( Match activation ) {
-                if ( ruleNames.isEmpty() ) {
+            public boolean accept(Match activation) {
+                if (ruleNames.isEmpty()) {
                     return true;
                 }
-                String ruleName = getFullyQualifiedRuleName(activation.getRule());
+                String ruleName = activation.getRule().getName();
 
-                if ( inclusive ) {
-                    return ruleNames.contains( ruleName );
+                if (inclusive) {
+                    return ruleNames.contains(ruleName);
                 } else {
-                    return !ruleNames.contains( ruleName );
+                    return !ruleNames.contains(ruleName);
                 }
             }
         };
     }
 
-    public void afterMatchFired( AfterMatchFiredEvent event ) {
-        recordFiring( event.getMatch().getRule() );
+    public void afterMatchFired(AfterMatchFiredEvent event) {
+        recordFiring(event.getMatch().getRule());
     }
 
-    private void recordFiring( Rule rule ) {
-        record( rule, this.firingCounts );
+    private void recordFiring(Rule rule) {
+        record(rule,
+               this.firingCounts);
     }
 
-    public void agendaGroupPopped( AgendaGroupPoppedEvent event ) {
+    public void agendaGroupPopped(AgendaGroupPoppedEvent event) {
     }
 
-    public void agendaGroupPushed( AgendaGroupPushedEvent event ) {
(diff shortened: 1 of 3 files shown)


drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/backend/TestingEventListener.java
```diff
@@ -44,67 +45,75 @@
...
 
     private String getFullyQualifiedRuleName(final Rule rule) {
-        String packageName = rule.getPackageName();
-        return (packageName.isEmpty()?rule.getName():packageName+"."+rule.getName());
+        final String ruleName = rule.getName();
+        final String packageName = rule.getPackageName();
+        if (!(packageName.isEmpty() || packageName.equals(defaultPkgName))) {
+            return packageName + "." + ruleName;
+        }
+        return ruleName;
```
yurloc: @manstis If this is the only non-whitespace change, do you think it will solve the backward compatibility issue with tests scenarios that expect rule firings using the non-FQN names? By looking at the code, I'm afraid this only solves the problem with default java package vs Drools "artificial" `defaultpkg`.

I'm trying to build workbench including this fix to do thorough testing...

yurloc (changes requested):
Currently the fix solves the problem with rules and test scenarios located in the `<default>` (unnamed) package. The change is also backwards compatible.

What needs to be added in order to resolve [GUVNOR-3134](https://issues.jboss.org/browse/GUVNOR-3134) is backward compatibility measures that will make sure that 6.x test scenarios won't fail due to the fact that they refer to expected rules using their simple names while actual fired rules are now reported using their FQN (including the package).

It would also be great if the pull request didn't contain the actual fix mixed with whitespace formatting. It doesn't make the code review easier.

manstis:
@yurloc Hi.. what a muddle I find myself in :-(

The original fix for https://issues.jboss.org/browse/GUVNOR-2980 would, with hindsight, appear to be rushed without proper consideration for it's consequences. The original issue was caused by a change in (Lucene) indexing that now stores the fully qualified rule name. I am however now leaning towards reverting  the changes for GUVNOR-2980 and instead have the UI (once more) use "simple rule names". The more I've looked at this the more it seems the best option (as the "Rule Selector" in the UI is limited to rules in the same package as the Test Scenario). 

@jomarko and I discussed the use of "simple Rule Names" vs "fully qualified Rule Names" before GUVNOR-2980 was "fixed" and we both, at the time, though FQRN was best. However given the inherent mess we get with backwards compatibility and the limited scope of the "Rule Selector" I think reverting back to (and fixing) use of "simple Rule Names" is best for everyone.

@yurloc @jomarko WDYT?

yurloc:
@manstis Yes, I would revert to the UI only working with simple rule names. Especially if the Rule Selector only offers rules from the current package, there is no need to use FQRNs. In that case I'd also pay attention to UI consistency of the test scenario editor, so the FQRN should not appear niether in Audit log, nor in the Show rules fired widget. I believe that the package where the current test scenario is located is easily accessible so it should be possible to reliably strip off the package from rules presented in the UI, even if, for example the `TestingEventListener`, returns FQRNs.

manstis:
@yurloc @jomarko This PR now reverts the change made for [GUVNOR-2980](https://issues.jboss.org/browse/GUVNOR-2980). The fix for GUVNOR-2980 is now effectively 100% in ```drools-wb```. See https://github.com/kiegroup/drools-wb/pull/522.

yurloc (approved):
+1. Note: not using the FQN can lead to "false positives" where the test scenario can pass when a rule with the same name from different package fires. But this is not a regression so we may handle this in a different ticket.

manstis:
Thanks @yurloc. To be honest I'm not sure the "false positive" can actually be orchestrated with the Test Scenario Editor as it only allows selection of "expected rule firing" from rules in the same package. So, although the Audit Log may show an activation of a rule in a different package you cannot assert that it fired as part of the test.

yurloc:
@manstis I just tried that with 6.5.x. Yes, you can only expect rules in the same package. But if you're in `com.my.pkg1` and expect `rule1` to fire and it doesn't fire, and if there exists rule `com.my.pkg2.rule1` which fires, the scenario will _pass_ (which is incorrect).

But as I said, this is not a regression related to any of GUVNOR-{2840,2980,3134}. It works like that in 6.x as well, so let's handle that separately (if it needs to be handled at all - it's only a minor issue).

jomarko (approved):
Sorry for my FQN proposal.

manstis:
@jomarko No need to apologise. We both thought it a good idea.

manstis:
Part of an ensemble:

- https://github.com/kiegroup/drools/pull/1299
- https://github.com/kiegroup/kie-wb-common/pull/923
- https://github.com/kiegroup/drools-wb/pull/522

manstis:
Jenkins please retest this.

manstis:
@yurloc @jomarko OK.. all fixed. Rules with "```.```" in their name can be used; and I only show/use "Simple Rule Names" so 6.x legacy Test Scenarios should continue to function unaffected. Please re-review the ensemble.

yurloc:
All parts of the ensemble approved. Please go ahead and merge.

PR #1304, opened 2017-06-05
GUVNOR-2864: [Guided Decision Table] Values are reset during column edit

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2864

sasasasa

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * This represents a DSL sentence.
@@ -44,28 +45,28 @@ public class DSLSentence
      */
     public String toString() {
         getDefinition();
-        StringBuilder result = new StringBuilder( definition );
-        int variableStart = definition.indexOf( "{" );
-        while ( variableStart >= 0 ) {
-            int variableEnd = getIndexForEndOfVariable( result.toString(),
-                                                        variableStart );
-            String variable = result.substring( variableStart + 1,
-                                                variableEnd );
-            int variableNameEnd = variable.indexOf( ":" );
-            if ( variableNameEnd == -1 ) {
+        StringBuilder result = new StringBuilder(definition);
+        int variableStart = definition.indexOf("{");
+        while (variableStart >= 0) {
+            int variableEnd = getIndexForEndOfVariable(result.toString(),
+                                                       variableStart);
+            String variable = result.substring(variableStart + 1,
+                                               variableEnd);
+            int variableNameEnd = variable.indexOf(":");
+            if (variableNameEnd == -1) {
                 variableNameEnd = variable.length();
             }
-            String variableName = variable.substring( 0,
-                                                      variableNameEnd );
-            result.replace( variableStart + 1,
-                            variableEnd,
-                            variableName );
+            String variableName = variable.substring(0,
+                                                     variableNameEnd);
+            result.replace(variableStart + 1,
+                           variableEnd,
+                           variableName);
             variableEnd = variableEnd - variable.length() + variableName.length();
-            variableStart = result.indexOf( "{",
-                                            variableEnd );
+            variableStart = result.indexOf("{",
+                                           variableEnd);
         }
-        return result.toString().replace( "\\n",
-                                          "\n" );
+        return result.toString().replace("\\n",
+                                         "\n");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


karreiro:
@manstis @jomarko Could you please take a look? :-)

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java
```diff
@@ -105,75 +105,75 @@ public String interpolate() {
...
-            for ( DSLVariableValue value : getValues() ) {
-                newOne.getValues().add( value );
-            }
+        final DSLSentence copy = new DSLSentence();
+
+        copy.drl = getDrl();
+        copy.definition = getDefinition();
+
+        if (getValues() != null) {
+            copy.values = new ArrayList<>(getValues());
```
karreiro: The real change is here.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/DSLSentenceTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+        final DSLSentence original = makeDslSentence(definition,
+                                                     drl);
+        final DSLSentence copy = original.copy();
+
+        assertEquals(copy.getDefinition(),
+                     original.getDefinition());
+        assertEquals(copy.getDrl(),
+                     original.getDrl());
+        assertEquals(copy.getValues().size(),
```
jomarko: Please don't assert here just that sizes equals because this test will pass also if the size is zero. The `getValues()` should return one `DSLVariableValue` - city. Please could you enhance the assertion?

jomarko (changes requested):

manstis (commented):
I agree with @jomarko please check the ```DSLVariableValue``` too.

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java
manstis: We're not cloning the ```DSLVariableValue``` objects here; just cloning the list of the _original_ references. Perhaps simply calling ```copy.getValues()``` is best to create a new ```List``` of new ```DSLVariableValue``` objects?

karreiro: @manstis I'm not sure if I get the point correctly. Replacing:
```java
copy.values = new ArrayList<>(getValues());
```
by
```java
copy.values = copy.getValues();
```
would call [this method](https://github.com/karreiro/drools/blob/1fbca3cab3384b7574b908a59ce77df56dec0aad/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java#L207-L223). In that case, the value would be always `"city"` and I'm copying the `values` field to avoid that.

We could replace the line above with the following line, too:
```java
copy.values = getValues();
```
However, I tried to keep the intention of the previous behavior of this method, which was to "clone" the List (since it was adding the elements from the original `DSLSentence` in a new List instance).

Am I missing something? :-)

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/DSLSentenceTest.java
karreiro: Definitely. Thanks @jomarko :-)

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java
manstis: Hi @karreiro I'm happy to be wrong :-)

I was questioning whether "copy" should perform a "deep copy" and create new instances of ```DSLVariableValue```.

```copy.values = new ArrayList<>(getValues());``` calls (obviously!) ```getValues()``` which, if ```values``` is ```null``` calls ```parseDefinition()``` (as you state). If however ```values``` is not ```null``` then it returns existing instances of ```DSLVariableValue```.  

Replacing ```copy.values = new ArrayList<>(getValues())``` with ```copy.getValues()``` (no set or anything, just call the method) would ensure the "copy" gets new instances of ```DSLVariableValue``` that should (AFAIK!) be equal (in content) to the original ```DSLVariableValue``` instances.

Comment (karreiro) [reply drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/DSLSentence.java]: Ohhh! Ok, @manstis! Thanks for the explanation. :-)

**Comment by karreiro** [comment]
I think that everything is fixed now, @manstis. Could you give me your endorsement? :-)

File: drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/DSLSentenceTest.java [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+        assertEquals(original.getDefinition(),
+                     copy.getDefinition());
+        assertEquals(original.getDrl(),
+                     copy.getDrl());
+        assertEquals(original.getValues(),
+                     copy.getValues());
+        assertEquals(firstValue(original),
+                     firstValue(copy));
+        assertNotSame(firstValue(original),
+                     firstValue(copy));
```
manstis: Formatting (sorry)

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/DSLVariableValueTest.java
```diff
@@ -0,0 +1,35 @@
...
+import static org.junit.Assert.*;
+
+public class DSLVariableValueTest {
+
+    @Test
+    public void testCopy() {
+        final DSLVariableValue original = new DSLVariableValue("value");
+        final DSLVariableValue copy = original.copy();
+
+        assertEquals(original,
```
manstis: You could possibly check values too (but I guess, if the ```equals()``` method is broken we're really screwed!)

manstis (commented):
@karreiro Ah... so you opted for _that_ solution!

karreiro:
@manstis Thank you! Fixed ;-)

manstis (approved):
Thanks @karreiro

manstis:
Jenkins please retest this.

manstis:
@jomarko Just needs your approval please.

jomarko (approved):
Thanks for the PR.

PR #1308, opened 2017-06-07
GUVNOR-2864: [Guided Decision Table] Values are reset during column edit

manstis:
Override 'copy' for DSLComplexVariableValue

See https://issues.jboss.org/browse/GUVNOR-2864


karreiro (approved):
Thank you @manstis. I was investigating the same problem.

ederign (approved):

PR #1309, opened 2017-06-07
Expose the KieModuleMetaInfo to use it in the kie maven plugin

desmax74:
Exposed the KieModuleMetaInfo to use it in the kie maven plugin when invoked by the Maven incremental compiler


mariofusco:
@desmax74 I understand that it could be useful to expose the KieModuleMetaInfo but probably it could be done in a slightly different way. For instance the write method could return it instead of void. Moreover maybe in your case you don't need to write anything inside that trgMfs (targetMemoryFileSystem) variable, so it should be allowed to pass a null in? Can you show me where do you need to use that KieModuleMetaInfo? Do you already have another PR invoking it?

desmax74:
@mariofusco after the compilation the KieModuleMetaInfo will be added into a map in the BuildMojo of the kie maven plugin https://github.com/desmax74/droolsjbpm-integration/blob/incrementalBuild/kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java#L167

then readed from a different classloader in the Maven Incremental Compiler 

https://github.com/desmax74/kie-wb-common/blob/incrementalBuild/kie-wb-common-services/kie-wb-common-services-backend/src/main/java/org/kie/workbench/common/services/backend/builder/compiler/nio/impl/NIODefaultMavenCompiler.java#L131

I'm open to other different ways to retrieve the metadata

mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/1311

PR #1311, opened 2017-06-08
[DROOLS-1611] expose the KieModuleMetaInfo

mariofusco:


desmax74 (approved):
Ok from my side, I can retrieve the KieMetaInfo without using a ResourceStore

PR #1313, opened 2017-06-09
Maven incremental Build

desmax74:
Changes to serialize the classes when exchanged between two different classloader


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieModule.java
```diff
@@ -95,9 +95,9 @@
 
     private final Map<String, Results> resultsCache = new HashMap<String, Results>();
 
-    protected final ReleaseId releaseId;
+    protected  ReleaseId releaseId;
 
-    private final KieModuleModel kModuleModel;
+    private  KieModuleModel kModuleModel;
```
mariofusco: @desmax74 If you want to do this I'm afraid you'll have to also make the KieModuleModel (and all other classes and interfaces related to it) also Serializable.

mariofusco:
ok to test

mariofusco:
Invalid

desmax74:
Outdated

PR #1314, opened 2017-06-09
[DROOLS-1582] split kie-ci

mariofusco:


mariofusco:
depends on https://github.com/AppFormer/uberfire/pull/740

PR #1316, opened 2017-06-12
DROOLS-1568 Improve error messages when using boxed java functions

tarilabs:
Example of error messages now returned:

```
Error compiling Java function 'Cosinus1' on node 'Cosinus1': Error resolving external function as defined by: function(number) external { java: { class: "java.lang.Mathhh", method signature: "cos(double)" }}
```

```
Error compiling Java function 'Cosinus2' on node 'Cosinus2': Error resolving external function as defined by: function( number ) external { java : { class : "java.lang.Math", method signature : "cosss(double)" } }
```


tarilabs:
@etirelli for your review. I'm basically wrapping the inner FEEL message up to the DMN layer, there was already a specific message for this kind of situation.

etirelli (approved):

PR #1321, opened 2017-06-14
[DROOLS-1582] fix osgi after move of maven integration into appformer

mariofusco:


kie-ci-osgi/pom.xml
```diff
@@ -117,10 +117,10 @@
               javax.net.ssl,
               javax.security.auth.x500,
               org.apache.commons.logging,
+              org.appformer.maven.support,
               org.drools.compiler.kie.builder.impl,
               org.drools.compiler.kproject,
               org.drools.compiler.kproject.models,
-              org.drools.compiler.kproject.xml,
```
psiroky, line 123: Is this related to the maven module changes?

mariofusco: Yes, the DependencyFilter was in that package but now it has been moved in org.appformer.maven.support https://github.com/AppFormer/uberfire/blob/master/uberfire-maven-utils/uberfire-maven-support/src/main/java/org/appformer/maven/support/DependencyFilter.java

psiroky (approved):

PR #1322, opened 2017-06-14
[DROOLS-1615] make KieModule Serializable

mariofusco:


desmax74:
The tests are now ok with the Serialization of the objects readed from hte map shared with the Kie Maven Plugin

desmax74 (approved):
The Serialization of the Objects readed from the map shared with teh Kie Maven Plugin works

PR #1324, opened 2017-06-16
DROOLS-1616 Revise coverage of DMN input tests 

baldimir:
Add input tests + move some tests that are in wrong class


kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/DMNModel.java
```diff
@@ -56,10 +53,6 @@
 
     Set<BusinessKnowledgeModelNode> getBusinessKnowledgeModels();
 
-    Set<InputDataNode> getRequiredInputsForBusinessKnowledgeModelName(String bkmName );
```
etirelli, line 59: Good catch! I totally missed these methods were here!

etirelli (approved):

PR #1326, opened 2017-06-19
DROOLS-1618: fix sort function to accept a null "precedes" parameter

etirelli:
When the "precedes" parameter is not available on a sort() call, just use natural order, delegating to the second method.


etirelli:
This PR has no test because the existing test already covers this use case, but only fails on IBM JDK.

baldimir:
The fix works. I gave a run to DMN tests with IBM JDK and the tests pass for this case.

PR #1327, opened 2017-06-19
DROOLS-1546: improving FEEL tests to check for error messages…

etirelli:
and prevent regressions. Fixed a couple bugs found while checking messages.


etirelli:
@tarilabs @baldimir we were ignoring FEEL error messages in the tests so far. This PR changes the tests to check for the existence and severity of errors.

Also, as I predicted, there were a couple bugs we didn't notice before because we were ignoring the messages. This PR also fixes these bugs.

Finally, there were a couple places on the parser that require dynamic resolution, but we were not enabling it. Fixed it.

tarilabs (approved):

baldimir (approved):

PR #1329, opened 2017-06-20
[DROOLS-1619] Compile error on a multibyte-name variable as a positio…

tkobayas:
…nal query parameter

- Unit test


tkobayas:
Merged a fix from Matteo so this PR is updated.

Please review, thanks!

tarilabs:
/cc @mariofusco 

for reference of regex:
`\p{L}` a Unicode Letter
`\p{N}` a Unicode Number (digit)

PR #1330, opened 2017-06-20
DROOLS-1595 Fix the FEEL "in" operator to work with simple values and…

tarilabs:
… variables

@etirelli I'm not entirely convinced about this, as implementing the fix felt a bit easier than I expected :) so can you kindly review considering this, please? On other hand I tried think other ways, but this still looked the most correct fix to be applied. Kindly let me know, thanks!


etirelli:
@tarilabs looks good, although I am surprised as well that the change did not create ambiguities in the grammar. Good job. 
Can you please just fix the conflict on the test file? this is due to the change I did before, adding a 3rd parameter for each test: the severity of the failure (if any). Just add null as the third parameter for successful tests.

etirelli (approved):

tarilabs:
please wait the online resolve is not good

tarilabs:
Rebased.

PR #1332, opened 2017-06-20
DROOLS-1621 Fix DMN validation when checking referenced types on IBM JDK

baldimir:
- Added explicit toString calls to patterns using QName in DMN validation rules


etirelli:
@baldimir so, did this fix the problem on IBM JDK? If so, it looks fine to me.

etirelli (approved):

etirelli:
@baldimir Thank you for the fix!

PR #1333, opened 2017-06-20
DROOLS-1623: confirm there is space after template/package/import tokens

HelloChrisLowe:


mariofusco:
@ChristianLowe Your fix looks ok but it would be great to also have a test case reproducing the problem before merging this pull request. Also be aware that all pull requests have to be sent against the master branch. I cannot merge any PR sent to a branch other than master, so I'm obliged to close this. Please let me know if you'll send a second PR against master or you want me to merge both your fix and the test case reproducing the problem there.

PR #1337, opened 2017-06-26
[DROOLS-1633] prevent race condition in jitting process when enforcin…

mariofusco:
…g immediate constraint jitting


martinweiler (approved):
I tested this fix with the complex reproducer from the support ticket a couple of times, and can confirm that all test runs were successful.

etirelli (approved):

PR #1341, opened 2017-06-30
[DROOLS-1377] remove reteoo implementation

mariofusco:


drools-core/src/main/java/org/drools/core/reteoo/SubnetworkTuple.java
```diff
@@ -171,4 +172,9 @@ public void clearStaged() {
         super.clearStaged();
         stagedOnRight = false;
     }
+
+    @Override
+    public void retractTuple( PropagationContext context, InternalWorkingMemory workingMemory ) {
+        throw new UnsupportedOperationException();
```
tarilabs, line 178: Summary of explanation

the method was defined in Tuple and implement both in BaseLeftTuple and RightTupleImpl
but actually it makes sense only on the right now
so moved the declaration of that method from Tuple to RightTule
and removed the impl from BaseLeftTuple.

SubnetworkTuple is a bit of corner case because it extends left but also implements right
but anyway it is never called for that tuple
so gets the empty impl

tarilabs (approved):

PR #1347, opened 2017-07-03
DROOLS-1549 internal improvements

tarilabs:
Avoid unnecessary nulls, casts, etc from kiegroup/drools#1344 .

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateFunction.java
@@ -45,12 +45,10 @@ public class DateFunction
         } catch (DateTimeException e) {
             // try to parse it as a date time and extract the date component
             // NOTE: this is an extension to the standard
-            Object r = BuiltInFunctions.getFunction( DateTimeFunction.class ).invoke( val ).cata( BuiltInType.justNull(), Function.identity() );
-            if( r != null && r instanceof TemporalAccessor ) {
-                return invoke( (TemporalAccessor) r );
-            } else {
-                return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "from", "date-parsing exception", e));
-            }
+            return BuiltInFunctions.getFunction( DateTimeFunction.class ).invoke( val )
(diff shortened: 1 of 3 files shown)


etirelli (approved):

etirelli:
I still have trouble properly reading this expression you wrote, :) but I will trust you it has the same semantics...

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateFunction.java
```diff
@@ -45,12 +45,10 @@ public DateFunction() {
...
             // try to parse it as a date time and extract the date component
             // NOTE: this is an extension to the standard
-            Object r = BuiltInFunctions.getFunction( DateTimeFunction.class ).invoke( val ).cata( BuiltInType.justNull(), Function.identity() );
-            if( r != null && r instanceof TemporalAccessor ) {
-                return invoke( (TemporalAccessor) r );
-            } else {
-                return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "from", "date-parsing exception", e));
-            }
+            return BuiltInFunctions.getFunction( DateTimeFunction.class ).invoke( val )
+                .cata( overrideLeft -> FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "from", "date-parsing exception", e)),
```
tarilabs, line 49: ^ if it's a left, i.e.: an error, like in your original code you actually report the exception from the current catch branch,

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateFunction.java
```diff
@@ -45,12 +45,10 @@ public DateFunction() {
...
+                       r -> invoke( r )
```
tarilabs, line 50: ^ if it's a result, it will be a result of type `TemporalAccessor`. Hence just transform the result by calling the `invoke()` as per original code.

tarilabs (commented):
@etirelli 
:)
I think the only thing which is a bit different in this case, we used "cata" for reducing to a domain object, eg a String or BigDecimal.
In this case I leverage "cata" to transform the FEELFnResult in another FEELFnResult ;)

PR #1348, opened 2017-07-03
DROOLS-1599 Support Signavio Multi Instance Decision node

tarilabs:


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -90,6 +93,37 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
     }
 
     private DMNCompiler getCompiler(KnowledgeBuilderImpl kbuilderImpl) {
+        Map<String, String> dmnProfileProperties = new HashMap<>();
+        kbuilderImpl.getBuilderConfiguration().getChainedProperties()
+                .mapStartsWith(dmnProfileProperties, DMN_PROFILE_PREFIX, false);
+        if( !dmnProfileProperties.isEmpty() ) {
```
tarilabs, line 99: if a profile exists, take priority and it's only the one applied (to avoid conflicts with just dmn marshaller extensions).

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerHelper.java
```diff
@@ -47,4 +49,15 @@ public static boolean checkVariableName(DMNModelImpl model, NamedElement element
         return true;
     }
 
+    public static void reportMissingVariable(DMNModelImpl model, DRGElement node, DMNModelInstrumentedBase source, Msg.Message1 message ) {
```
tarilabs, line 52: moved from DMNCompilerImpl here,

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -58,6 +58,13 @@
     private final DMNEvaluatorCompiler evaluatorCompiler;
     private final DMNFEELHelper feel;
     private DMNCompilerConfiguration dmnCompilerConfig;
+    private Deque<DRGElementCompiler> drgCompilers = new LinkedList<>();
+    {
+        drgCompilers.add( new InputDataCompiler() );
+        drgCompilers.add( new BusinessKnowledgeModelCompiler() );
+        drgCompilers.add( new DecisionCompiler() );
+        drgCompilers.add( new KnowledgeSourceCompiler() ); // keep last as it's a void compiler
```
tarilabs, line 66: default DRGElementCompiler.

kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/MultiInstanceDecisionLogic.java
```diff
@@ -79,28 +79,49 @@ public String toString() {
...
+            MultiInstanceDecisionLogic midl = getMIDL(node).get();
+            
+            // set the evaluator accordingly to Signavio logic.
+            di.setEvaluator(new MultiInstanceDecisionNodeEvaluator(midl, model, di));
+            
+            // Remove the top level decision and its dependencies, from the DMN Model (Decision|BKM|InputData) indexes
+            // Remember that as the dependencies will be removed from indexes, are no longer available at evalutation from the DMNExpressionEvaluator
+            // hence the MultiInstanceDecisionNodeEvaluator will need to cache anything which is accessed by the DMN Model (Decision|BKM|InputData) indexes 
+            DecisionNodeImpl topLevelDecision = (DecisionNodeImpl) model.getDecisionById(midl.topLevelDecisionId);
+            recurseNodeToRemoveItAndDepsFromModelIndex(topLevelDecision, model);
```
tarilabs, line 116: I could avoid having to "retract" a DMNMessage for error.

Unfortunately, we still have to "undo" the removal from the DMN Model (Decision|BKM|InputData) indexes, as that is mandatory step1 and previous step in the evaluator compilation + assignment.

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/SignavioTest.java
```diff
@@ -34,7 +50,7 @@ public void test() {
         final KieFileSystem kfs = ks.newKieFileSystem();
         
         KieModuleModel kmm = ks.newKieModuleModel();
-        kmm.setConfigurationProperty("org.kie.dmn.marshaller.extension.MultiInstanceDecisionLogic", "org.kie.dmn.signavio.MultiInstanceDecisionLogicRegister");
+        kmm.setConfigurationProperty("org.kie.dmn.profile.signavio", "org.kie.dmn.signavio.KieDMNSignavioProfile");
```
tarilabs: changed to have only profile for all required settings, as discussed.

tarilabs:
```
17:15:48 [ERROR] COMPILATION ERROR : 
17:15:48 [INFO] -------------------------------------------------------------
17:15:48 [ERROR] /home/jenkins/workspace/drools-pullrequests/upstream-repos/droolsjbpm-knowledge/kie-api/src/main/java/org/kie/api/builder/ReleaseId.java:[23,63] cannot find symbol
17:15:48   symbol:   class ReleaseId
17:15:48   location: package org.appformer.maven.support
```

Jenkins, test again.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -90,6 +93,37 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
...
+            try {
+                List<DMNProfile> dmnProfiles = new ArrayList<>();
```
baldimir, line 101: This could be extracted to a separate method. Something like `loadProfiles`.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -90,6 +93,37 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
...
+                for (Map.Entry<String, String> dmnProfileProperty : dmnProfileProperties.entrySet()) {
+                    DMNProfile dmnProfile = (DMNProfile) kbuilderImpl.getRootClassLoader()
+                        .loadClass( dmnProfileProperty.getValue() ).newInstance();
+                    dmnProfiles.add(dmnProfile);
+                }
+
+                DMNCompilerConfiguration compilerConfig = DMNFactory.newCompilerConfiguration();
```
baldimir, line 108: This could be extracted to a separate method. Something like `getCompilerConfig(List<DMNProfile> profiles)`

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -90,6 +93,37 @@ public void addResource(KnowledgeBuilder kbuilder, Resource resource, ResourceTy
...
+                for ( DMNProfile dmnProfile : dmnProfiles ) {
+                    compilerConfig.addExtensions( dmnProfile.getExtensionRegisters() );                    
+                }
+                
+                DMNCompilerImpl compiler = (DMNCompilerImpl) DMNFactory.newCompiler(compilerConfig);
```
baldimir, line 113: Same comment as before. Could be extracted to a separate method. Something like `getCompiler(List<DMNProfile> profiles)`

baldimir (changes requested):
Just stylistic comments.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: are you sure? I would need to pass to the proposed separate method the KnowledgeBuilderImpl and the `dmnProfileProperties` which was cached a few lines above to determine if DMN Profile are to be considered.
I tried but the externalized method looks like:
```
    private List<DMNProfile> loadDMNProfiles(KnowledgeBuilderImpl kbuilderImpl, Map<String, String> dmnProfileProperties) throws Exception {
        List<DMNProfile> dmnProfiles = new ArrayList<>();
        for (Map.Entry<String, String> dmnProfileProperty : dmnProfileProperties.entrySet()) {
            DMNProfile dmnProfile = (DMNProfile) kbuilderImpl.getRootClassLoader()
                .loadClass( dmnProfileProperty.getValue() ).newInstance();
            dmnProfiles.add(dmnProfile);
        }
        return dmnProfiles;
    }
```

so in order to refactor out the cycling on the `dmnProfileProperties.entrySet()` I need to create a separate method with a kind of odd signature.. I don't think it gains significant advantages. What do you think?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: It would also need the configuration passed as parameter, which need to be created with the same logic of the profiles. Given frankly is a kind of "experimental" and internal API and I don't expect this proposed method and the one proposed above being reused elsewhere, it's just moving out from the linear ~10 of loading in the assembler, in separate private methods with "references" (not indipendent calls, there are parameters which "threads" the private methods calls),
With this considered, I don't think I would be gaining significant advantages, but kindly let me know if I missed something else, please?

baldimir (approved):
After discussion with @tarilabs, approving.

etirelli (approved):

PR #1352, opened 2017-07-04
[DROOLS-1647] fix binding of a DataSource not originally created from…

mariofusco:
… the executor


drools-core/src/main/java/org/drools/core/datasources/CursoredDataSource.java
```diff
@@ -143,7 +143,8 @@ public void delete(FactHandle fh) {
     }
 
     @Override
-    public void bind(RuleUnit unit, EntryPoint ep) {
+    public void bind(RuleUnit unit, WorkingMemoryEntryPoint ep) {
+        setWorkingMemory( ep.getInternalWorkingMemory() );
```
tarilabs, line 147: ok

tarilabs (approved):

PR #1354, opened 2017-07-04
[DROOLS-1629] avoid Mixing of ObjectMarshallingStrategies of the same class using…

ngs-mtech:
… NamedObjectMarshallingStrategy

Solving issue DROOLS-1629

Added more logging to improve evidence of the issue if its happening.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.core.marshalling;
+
+
+import org.kie.api.marshalling.ObjectMarshallingStrategy;
+
+public interface NamedObjectMarshallingStrategy extends ObjectMarshallingStrategy {
+
+	public String getName();
+}
--- a/drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
@@ -19,7 +19,17 @@ package org.drools.core.marshalling;
 
 import org.kie.api.marshalling.ObjectMarshallingStrategy;
 
+/**
+ * In order to allow multiple marshalling strategies of the same class, this interface must be used
+ * to avoid class name collision.
+ */
 public interface NamedObjectMarshallingStrategy extends ObjectMarshallingStrategy {
 
+	/**
+	 * Ensure that the name returned by this method has something related to the exact objects it handles.
+	 * Use this name to avoid collisions of the same class implementing marshalling strategy but handling different objects 
+	 * or targets (Different persistence units, Different Document Content storages, etc ).
+	 * @return The unique name in the project environment for the marshalling strategy represented
+	 */
 	public String getName();
 }
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/IdentityPlaceholderResolverStrategy.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/IdentityPlaceholderResolverStrategy.java
@@ -16,9 +16,6 @@
 
 package org.drools.core.marshalling.impl;
 
-import org.kie.api.marshalling.ObjectMarshallingStrategy;
-import org.kie.api.marshalling.ObjectMarshallingStrategyAcceptor;
-
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
@@ -26,13 +23,17 @@ import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
 
+import org.drools.core.marshalling.NamedObjectMarshallingStrategy;
+import org.kie.api.marshalling.ObjectMarshallingStrategyAcceptor;
+
 public class IdentityPlaceholderResolverStrategy
     implements
-    ObjectMarshallingStrategy {
+    NamedObjectMarshallingStrategy {
 
     private Map<Integer, Object> ids;
     private Map<Object, Integer> objects;
 
+    private String name = IdentityPlaceholderResolverStrategy.class.getName();
     private ObjectMarshallingStrategyAcceptor acceptor;
     
     public IdentityPlaceholderResolverStrategy(ObjectMarshallingStrategyAcceptor acceptor) {
@@ -44,6 +45,10 @@ public class IdentityPlaceholderResolverStrategy
     public IdentityPlaceholderResolverStrategy(ObjectMarshallingStrategyAcceptor acceptor, Map<Integer, Object> ids) {
         this(acceptor);
         setIds(ids);
+    }
+    
+    public String getName(){
+    	return this.name;
     }
 
     public Object read(ObjectInputStream os) throws IOException,
@@ -113,8 +118,11 @@ public class IdentityPlaceholderResolverStrategy
     public void setIds(Map<Integer, Object> ids) {
         this.ids = ids;
         objects.clear();
+        StringBuilder nameIds = new StringBuilder();
         for (Map.Entry<Integer, Object> idEntry : ids.entrySet()) {
             objects.put(idEntry.getValue(), idEntry.getKey());
+            nameIds.append( idEntry.getKey() ).append( "," );
         }
+        this.name = IdentityPlaceholderResolverStrategy.class.getName() + nameIds.toString();
     }
 }
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
@@ -16,15 +16,38 @@
 
 package org.drools.core.marshalling.impl;
 
+import java.util.HashSet;
+import java.util.Set;
+
+import org.drools.core.marshalling.NamedObjectMarshallingStrategy;
 import org.drools.core.util.StringUtils;
 import org.kie.api.marshalling.ObjectMarshallingStrategy;
 import org.kie.api.marshalling.ObjectMarshallingStrategyStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ObjectMarshallingStrategyStoreImpl implements ObjectMarshallingStrategyStore {
-    private ObjectMarshallingStrategy[] strategiesList;
+	private static final Logger logger = LoggerFactory.getLogger(ObjectMarshallingStrategyStoreImpl.class);
+
+	private ObjectMarshallingStrategy[] strategiesList;
     
     public ObjectMarshallingStrategyStoreImpl(ObjectMarshallingStrategy[] strategiesList) {
         this.strategiesList = strategiesList;
+        Set<String> names = new HashSet<String>();
+        for( ObjectMarshallingStrategy strategy : strategiesList ){
+        	String name;
+        	if( strategy instanceof NamedObjectMarshallingStrategy ){
+        		name = ((NamedObjectMarshallingStrategy)strategy).getName();
+        	}else{
+        		name = strategy.getClass().getName();
+        	}
+        	if( names.contains( name ) ){
+        		logger.warn( "Multiple ObjectMarshallingStrategies with the same name:" + name + " strange behaviour could occurr");
+        	}else{
+        		names.add( name );
+        	}
+        }
+        names.clear();
     }
    
     // Old marshalling algorithm methods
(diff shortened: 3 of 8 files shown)


mariofusco:
ok to test

ngs-mtech:
I can´t see what have happened!

mariofusco:
Jenkins is reporting the following compilation error:

[ERROR] /home/jenkins/workspace/drools-pullrequests/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java:[46,8] org.drools.persistence.jpa.marshaller.JPAPlaceholderResolverStrategy is not abstract and does not override abstract method getName() in org.drools.core.marshalling.NamedObjectMarshallingStrategy

Will you fix it or should I give a look?

ngs-mtech:
Sorry Mario, here you have the fix.

Regards.

mariofusco:
We believe that what you need can be easily achieved without this pull request as demonstrated here https://gist.github.com/DuncanDoyle/f8cc93ba0559fefd8e1c64b1db43f643

I also tried the solution proposed by @DuncanDoyle and it works for me. Please let me know if we're overlooking something. If this also work for you I'll close this pull request and reject the corresponding jira.

ngs-mtech:
Hi Mario,
      
      I think I didnt explain the problem we have clear enough. We have
      a marshalling strategy (a common one for the whole collection of
      modules in our project), that implements some extra features that
      we found interesting regarding variable persistence.
      
      The problem comes from the fact that, when using a marshalling
      strategy in a kie module,  it gets registered in the list together
      with the rest (as expected), but when a marshalling strategy is
      employed in dependant modules, it gets registered multiple times
      (same class, different parameters). From this approach, while
      unmarshalling, the system cannot guess which is the correct
      strategy that was used for the marshalling.
      
      In the case you did understand this, suggesting a multiple class
      implementation with same code looks not only poor, but unpractical
      approach, as it would require a redeployment of the
      workbench/server war every time a new module needs to use that
      strategy. Remember that this class must be available at war
      classloader level (not module level).
      
      Taking into account that your solution (test update) is not
      solving the problem but hiding it, by making the test that makes
      it "visible" work, we do believe that the one we 'pulled' is, in
      our humble opinion, much more robust, flexible and practical.  
      
      Hope this lets the problem we found more clear and makes you
      understand that what you propose is not an acceptable solution,
      neither for our customers, neither for anyone in the community. 
      
      Best Regards,

mswiderski:
unless I am missing something but what I think is the intended implementation of the strategies is to hide such complexity in the impl instead of making the framework to deal with it. What I am trying to say is that you should have this logic required to manage different modules. 

With that said I'd recommend to use single strategy that includes the logic to figure out what should actually be unmarshalled in what way. Similar implementation was done for case management and its case file marshalling that has single strategy but can handler different types of data - https://github.com/kiegroup/jbpm/blob/master/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/marshalling/CaseFileInstanceMarshallingStrategy.java

ngs-mtech:
Hello Maciej
      
      I am able to agree with the design approach that you propose for
      Marshalling strategies but what we are talking about here is a bug
      [https://issues.jboss.org/browse/DROOLS-1629],
      a bug that says the marshalling strategy you use to marshall an
      object is different from the one used to unmarshall, as the
      marshalling task uses the method accept(Object ) to fetch the
      correct marshalling strategy and the deserialization uses the
      class name to fetch the target.
      
      This is a fact that is happening regardless of the design approach
      of the strategies you comment on, which, in fact, is the same as
      my chief architect proposed.
      
      It is important to emphasize, as you know, from another reported
      incidence, that KieWorkbench autoregisters the
      JPAPlaceholderMarshallingStrategy each time the persistence.xml is
      modified.
      
      When one module depends from another, the situation of having two
      ObjectMarshallinStrategies (JPAPlaceholderMarshallingStrategy)
      implemented with the same class is imposed by the environment.
      
      If you find a better solution for this error, we can apply it, but
      doing nothing is not a solution to a reproducible error.
      
      In fact the code indicated in the CaseManagement suffers from the
      same error.
      
      Regards

mswiderski:
then I am really missing something big here...

> Avoid Mixing of ObjectMarshallingStrategies of the same class

is the title of this PR so to address this problem - use single strategy and not many of the same class.

With that said can we go back few steps and clearly define what's the issue. Could you please provide single test case that illustrates the bug?

ngs-mtech:
Hi Maciej,

We already have done a [unit test]( https://github.com/ngs-mtech/drools/blob/99e74e6bb98907661c9fe039b2ea5e50f371157a/drools-core/src/test/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreTest.java) exposing the issue, using the same ObjectMarshallingStrategy twice in array, is the one that DuncanDoyle has changed to show the workarround "use single strategy and not many of the same class",  by declaring different subclasses hiding the real problem.

The problem is not a development affair, is a real situation imposed by the platform that occurs while using dependant modules having each one their own marshalling strategies defined.

It will take me some time to create a kie workbench repository demostrating the issue while using a **dependant module**.

In this scenario, both modules will have their custom persistence unit and a declaration in their deployment descriptor of their own JPAPlaceholderResolverStrategy( "<module>", classLoader) defined. 

Later, in the runtime environment of the dependant module, all the marshalling strategies (the ones from base module and the ones from the dependant module) are created and registered in the execution environment, so after that there are instances of the same class but from different modules registered. 

multiple-objectmarshallingstrategies-on-context-with-same-class


In this scenario, while doing the marshalling/unmarshalling of session variables, depending on the index of the ObjectMarshallingStrategy in the environment, the task of marshalling or unmarshalling will work or fail.

Regards

mswiderski:
So is the double marshaller in the runtime a side effect of using defendant kjars that have same marshaller already defined? 
Assuming you register them via deployment descriptor then there a option to define how to deal with merging of multiple descriptors where one of them is to override all so only the top level will be used.

ngs-mtech:
Yes Maciej, this is another way to skip the bug, but it is not a solution.

I can't undestand why we need to apply a workaround having a pull request with a working solution that solves the issue. We have implemented the solution taking into account legacy implementations, so is not a breaking approach.

Where is the failure?

mswiderski:
and how is this protecting the case where you have two kjars that declare the same strategy (with same name). They will be still registered, isn't it? So we might end up with exact same issue...

in case they are slightly differently configured then the issue is still there, right? Or the assumption is that then the name should take into account the config that could differ to distinguish it...

drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface NamedObjectMarshallingStrategy extends ObjectMarshallingStrategy {
+
+	/**
+	 * Ensure that the name returned by this method has something related to the exact objects it handles.
+	 * Use this name to avoid collisions of the same class implementing marshalling strategy but handling different objects 
+	 * or targets (Different persistence units, Different Document Content storages, etc ).
+	 * @return The unique name in the project environment for the marshalling strategy represented
+	 */
+	public String getName();
```
mswiderski, line 34: I don't think there is a need for new interface, we should make use of Java 8 features and extend the org.kie.api.marshalling.ObjectMarshallingStrategy with new method getName that is a default method which returns class.getName().

drools-core/src/main/java/org/drools/core/marshalling/impl/IdentityPlaceholderResolverStrategy.java
```diff
@@ -113,8 +118,11 @@ public Context createContext() {
     public void setIds(Map<Integer, Object> ids) {
         this.ids = ids;
         objects.clear();
+        StringBuilder nameIds = new StringBuilder();
         for (Map.Entry<Integer, Object> idEntry : ids.entrySet()) {
             objects.put(idEntry.getValue(), idEntry.getKey());
+            nameIds.append( idEntry.getKey() ).append( "," );
         }
+        this.name = IdentityPlaceholderResolverStrategy.class.getName() + nameIds.toString();
```
mswiderski, line 126: wouldn't this be a breaking compatibility change in case ids where given? If so then we should not add it here and then let users extend it if needed

drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
```diff
@@ -16,15 +16,38 @@
...
     public ObjectMarshallingStrategyStoreImpl(ObjectMarshallingStrategy[] strategiesList) {
         this.strategiesList = strategiesList;
+        Set<String> names = new HashSet<String>();
+        for( ObjectMarshallingStrategy strategy : strategiesList ){
+        	String name;
+        	if( strategy instanceof NamedObjectMarshallingStrategy ){
+        		name = ((NamedObjectMarshallingStrategy)strategy).getName();
+        	}else{
+        		name = strategy.getClass().getName();
+        	}
```
mswiderski, line 43: this if/else is not needed when we use default method on org.kie.api.marshalling.ObjectMarshallingStrategy

drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
```diff
@@ -16,15 +16,38 @@
...
+        	if( names.contains( name ) ){
+        		logger.warn( "Multiple ObjectMarshallingStrategies with the same name:" + name + " strange behaviour could occurr");
```
mswiderski, line 45: if we want to add this then we should fail if there are duplicates instead of logging warning that strange behaviour can occur as that is not really helping and can lead to issues you faced and might not be easy to recover

drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
```diff
@@ -60,9 +83,22 @@ public ObjectMarshallingStrategy getStrategyObject(String strategyClassName) {
         }
         ObjectMarshallingStrategy objectMarshallingStrategy = null; 
         for( int i = 0; i < this.strategiesList.length; ++i ) { 
-           if( strategiesList[i].getClass().getName().equals(strategyClassName) ) {
-               return strategiesList[i];
-           }
+        	if( strategiesList[i] instanceof NamedObjectMarshallingStrategy
```
mswiderski, line 86: same comment here, there is no need for this block anymore once we have default getName on org.kie.api.marshalling.ObjectMarshallingStrategy, s we should always check strategy.getName

drools-core/src/main/java/org/drools/core/marshalling/impl/PersisterHelper.java
```diff
@@ -239,9 +241,13 @@ public static void writeRuntimeDefinedClasses( MarshallerWriteContext context,
...
         for( Entry<ObjectMarshallingStrategy,Integer> entry : context.usedStrategies.entrySet() ) {
-            Builder _strat = ProtobufMessages.Header.StrategyIndex.newBuilder()
+        	String name = entry.getKey().getClass().getName() ;
+        	if( entry.getKey() instanceof NamedObjectMarshallingStrategy ){
+        		name = ((NamedObjectMarshallingStrategy)entry.getKey()).getName();
+        	}
+			Builder _strat = ProtobufMessages.Header.StrategyIndex.newBuilder()
                                      .setId( entry.getValue().intValue() )
-                                     .setName( entry.getKey().getClass().getName() );
+                                     .setName( name );
```
mswiderski, line 250: same here, it should only be entry.getKey().getName()

drools-core/src/main/java/org/drools/core/marshalling/impl/SerializablePlaceholderResolverStrategy.java
```diff
@@ -22,21 +22,26 @@
...
 import org.kie.api.marshalling.ObjectMarshallingStrategyAcceptor;
 
 public class SerializablePlaceholderResolverStrategy
     implements
-    ObjectMarshallingStrategy {
+    NamedObjectMarshallingStrategy {
 
     private int index;
-    
+    private String name = SerializablePlaceholderResolverStrategy.class.getName();
```
mswiderski, line 34: not needed when there is default method on org.kie.api.marshalling.ObjectMarshallingStrategy

drools-core/src/test/java/org/drools/core/marshalling/impl/NamedCustomMarshallingStrategy.java
```diff
@@ -0,0 +1,22 @@
+package org.drools.core.marshalling.impl;
```
mswiderski, line 1: missing license header

drools-core/src/test/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreTest.java
```diff
@@ -0,0 +1,208 @@
+package org.drools.core.marshalling.impl;
```
mswiderski, line 1: missing license header

mswiderski (changes requested):

drools-core/src/main/java/org/drools/core/marshalling/NamedObjectMarshallingStrategy.java
ngs-mtech: Changing an interface is very hard to support legacy implementations but your proposal of default implementation of the getName method in interface is a nice approach to the solution. 
The problem is that the ObjectMarshallingStrategy belongs to kie-api project that is outside drools project. This means a different pull request, don't it?

drools-core/src/main/java/org/drools/core/marshalling/impl/IdentityPlaceholderResolverStrategy.java
ngs-mtech: I think the behavior will not change unless you add new ones during the execution, right? Is this the case?

ngs-mtech (commented):
I will implement your suggestions

PR #1362, opened 2017-07-07
GUVNOR-3260: Guided Rule Editor doesn't render RDRL with abbreviated combined relation condition correctly

manstis:
See https://issues.jboss.org/browse/GUVNOR-3260

Handling of some "connective constraint" scenarios was wrong.

PR #1363, opened 2017-07-07
[6.5.x] RHBRMS-2854 (GUVNOR-3260): [GSS] (6.4.z) Guided Rule Editor doesn't render RDRL with abbreviated combined relation condition correctly

manstis:
See https://issues.jboss.org/browse/RHBRMS-2854

This is the corollary of https://github.com/kiegroup/drools/pull/1362 for 6.5.x.

(cherry picked from commit 73fc9690f6524ac2279d5bd05e637af6203be4c8)


Rikkola (approved):
If we add few more test `RuleModelDRLPersistenceUnmarshallingTest` will be over 10 000 lines ;-)
Would split that, but since this is a maintenance branch not really worth it. Let's keep it in mind next time we need to touch this in master.

PR #1364, opened 2017-07-07
[DROOLS-1629] Allow multiple instances of ObjectMarshallingStrategy of the same class

ngs-mtech:
With this implementation, dependant on `[[DROOLS-1629] Allow backward compatibility without breaking interface](https://github.com/kiegroup/droolsjbpm-knowledge/pull/249)`, the bug [DROOLS-1629] the issue should be fixed.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
@@ -16,15 +16,33 @@
 
 package org.drools.core.marshalling.impl;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import org.drools.core.util.StringUtils;
 import org.kie.api.marshalling.ObjectMarshallingStrategy;
 import org.kie.api.marshalling.ObjectMarshallingStrategyStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ObjectMarshallingStrategyStoreImpl implements ObjectMarshallingStrategyStore {
-    private ObjectMarshallingStrategy[] strategiesList;
+
+	private ObjectMarshallingStrategy[] strategiesList;
     
     public ObjectMarshallingStrategyStoreImpl(ObjectMarshallingStrategy[] strategiesList) {
         this.strategiesList = strategiesList;
+        Set<String> names = new HashSet<String>();
+        //Validate received set
+        for( ObjectMarshallingStrategy strategy : strategiesList ){
+        	String name = strategy.getName();
+        	
+        	if( names.contains( name ) ){
+        		throw new RuntimeException( "Multiple ObjectMarshallingStrategies with the same name found in environment:" + name );
+        	}else{
+        		names.add( name );
+        	}
+        }
+        names.clear();
     }
    
     // Old marshalling algorithm methods
@@ -51,20 +69,20 @@ public class ObjectMarshallingStrategyStoreImpl implements ObjectMarshallingStra
     /* (non-Javadoc)
      * @see org.kie.api.marshalling.impl.ObjectMarshallingStrategyStore#getStrategyObject(java.lang.String)
      */
-    public ObjectMarshallingStrategy getStrategyObject(String strategyClassName) {
-        if( StringUtils.isEmpty(strategyClassName) ) { 
+    public ObjectMarshallingStrategy getStrategyObject(String strategyName) {
+        if( StringUtils.isEmpty(strategyName) ) { 
             return null;
         }
-        if (strategyClassName.startsWith("org.drools.marshalling.impl")) {
-            strategyClassName = strategyClassName.replaceFirst("org.drools.marshalling.impl", "org.drools.core.marshalling.impl");
+        if (strategyName.startsWith("org.drools.marshalling.impl")) {
+            strategyName = strategyName.replaceFirst("org.drools.marshalling.impl", "org.drools.core.marshalling.impl");
         }
-        ObjectMarshallingStrategy objectMarshallingStrategy = null; 
+        
         for( int i = 0; i < this.strategiesList.length; ++i ) { 
-           if( strategiesList[i].getClass().getName().equals(strategyClassName) ) {
-               return strategiesList[i];
-           }
+        	 if( strategiesList[i].getName().equals(strategyName) ) {
+            		return strategiesList[i];
+            }
         }
-        return objectMarshallingStrategy;
+        throw new RuntimeException( "Unable to find PlaceholderResolverStrategy for name : " + strategyName );
     }
     
     /* (non-Javadoc)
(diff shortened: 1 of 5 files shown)


ngs-mtech:
Its required to pull before https://github.com/kiegroup/droolsjbpm-knowledge/pull/249 in order to allow this one!

mswiderski:
ok to test

mswiderski:
same here, could you apply the formatting rules as the diff looks really bad - no idea what are the changes, and please squash the commits

drools-core/src/main/java/org/drools/core/marshalling/impl/SerializablePlaceholderResolverStrategy.java
```diff
@@ -30,13 +30,13 @@
...
     private int index;
-    
     private ObjectMarshallingStrategyAcceptor acceptor;
     
     public SerializablePlaceholderResolverStrategy(ObjectMarshallingStrategyAcceptor acceptor) {
         this.acceptor = acceptor;
     }
     
+    
     public int getIndex() {
```
mswiderski: Changes in this file are not needed as they are just spaces

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
```diff
@@ -200,30 +213,22 @@ private static Serializable callIdMethod(Object target, String methodName) throw
...
-                    for (int i = 0; i < methods.length; i++) {
-                        Method method = methods[i];
-                        Id id = method.getAnnotation(Id.class);
-                        if (id != null) {
-                            return true;
-                        }
-                    }
-        } while ((varClass = varClass.getSuperclass()) != null );
-        
+        EntityManager em = getEntityManager();
```
mswiderski: Why not using entity manager factory to get the metamodel?

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
```diff
@@ -200,30 +213,22 @@ private static Serializable callIdMethod(Object target, String methodName) throw
...
+    	if( em != null ){
+    		Set<EntityType<?>> entities = em.getMetamodel().getEntities();
```
mswiderski: Instead iterating over list of entities we coul use the metamodel.entity method https://docs.oracle.com/javaee/7/api/javax/persistence/metamodel/Metamodel.html#entity-java.lang.Class- and catch IllegalArgumentException in case it is not an entity

drools-core/src/main/java/org/drools/core/marshalling/impl/ObjectMarshallingStrategyStoreImpl.java
```diff
@@ -1,82 +1,100 @@
...
- * limitations under the License.
- */
-
-package org.drools.core.marshalling.impl;
-
-import org.drools.core.util.StringUtils;
-import org.kie.api.marshalling.ObjectMarshallingStrategy;
-import org.kie.api.marshalling.ObjectMarshallingStrategyStore;
-
-public class ObjectMarshallingStrategyStoreImpl implements ObjectMarshallingStrategyStore {
```
mswiderski: This class still seems to have wrong formatting as it shows to be completely replaced and thus I can't see what has actually changed

mswiderski (changes requested):
Overall looks good, just few comments to address

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
ngs-mtech: Throwing an exception is something very expensive, the building of a callstack is a pain for the VM, 
Let me perform another approach.

ngs-mtech (commented):
All your proposals implemented, some changes added to JPAPlaceholderResolverStrategy

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
```diff
@@ -45,36 +49,61 @@
...
+    public JPAPlaceholderResolverStrategy(String name, String persistenceUnit, ClassLoader cl) {
+    	this( persistenceUnit, cl );
+    	this.name = name;
+    }
+    
+    public String getName(){
+    	return this.name;
+    }
+    
+    private void initializeManagedClasses(){
```
mswiderski, line 94: I like this approach! Nice work

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
```diff
@@ -45,36 +49,61 @@
...
 
             this.emf = Persistence.createEntityManagerFactory(persistenceUnit);
+            initializeManagedClasses();
             this.closeEmf = true;
 
         } finally {
             Thread.currentThread().setContextClassLoader(tccl);
         }
         this.classLoader = cl;
+        this.name += ":" + persistenceUnit;
```
mswiderski: I believe this is breaking backward compatibility as it will register itself with different name then it was persisted before this change so I think we should make it via additional cunstructor so we won't break existing sessions

mswiderski (changes requested):
Last concern to address for jpa strategy and I think we are good with it

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
mswiderski: Just noticed there is this additional constructor in place do it is a matter of removing this append of pu to name by default, wdyt?

mswiderski:
If I am not mistaken this already solves this issue JBPM-6175

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/marshaller/JPAPlaceholderResolverStrategy.java
ngs-mtech: Ok, I understand your approach.

ngs-mtech:
I hope so!.

mswiderski (approved):
looks good to me

@mariofusco @DuncanDoyle any comments?

DuncanDoyle:
No comments, LGTM.

mariofusco:
@mswiderski This is ok also for me. Let me know if I can merge it.

PR #1365, opened 2017-07-11
[DROOLS-1656] fix missing fire when deleting from a shared subnetwork…

mariofusco:
… with mulitple sinks


baldimir (approved):
Looks good to me. +1

PR #1369, opened 2017-07-13
[DROOLS-1655] unregister statful ksession from the kcontainer when it…

mariofusco:
… is disposed

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/MemoryLeakTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/MemoryLeakTest.java
@@ -208,7 +208,7 @@ public class MemoryLeakTest {
 
         ksession.dispose();
 
-        checkReachability( ksession, Person.class::isInstance, false );
+        checkReachability( kContainer, Person.class::isInstance, false );
         checkReachability( kBase, ksession::equals, false );
         checkReachability( kContainer, ksession::equals, false );
     }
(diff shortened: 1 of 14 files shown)


mariofusco:
ok to test

drools-compiler/src/test/java/org/drools/compiler/MemoryLeakTest.java
```diff
@@ -0,0 +1,264 @@
...
+import org.kie.api.KieBase;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieContainer;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.rule.FactHandle;
+import org.kie.internal.utils.KieHelper;
+
+import static org.junit.Assert.*;
+
+public class MemoryLeakTest {
```
tarilabs, line 52: I understand refactored here from Misc2Test

drools-compiler/src/test/java/org/drools/compiler/integrationtests/MBeansMonitoringTest.java
```diff
@@ -368,10 +363,10 @@ public void testAggregatedAndDispose() throws Exception {
         checkStatsByRule(aggrMonitor,"ND" ,2,2,0);
         checkStatsByRule(aggrMonitor,"ND2",2,0,2);
         
-        ksession2.dispose();
+        ksession.dispose();
         checkTotalSessions(aggrMonitor, 1);
         checkTotalFactCount(aggrMonitor, 1);
-        ksession.dispose();
+        ksession2.dispose();
```
tarilabs, line 369: I don't think switching them was needed (I tested original test still passes), but anyway.

PR #1373, opened 2017-07-18
PLANNER-823: ScoreHolder support does not work in Guided Decision Tables

mcimbora:
Part of:
https://github.com/kiegroup/drools/pull/1373
https://github.com/kiegroup/drools-wb/pull/550
https://github.com/kiegroup/optaplanner-wb/pull/188

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/util/GuidedDTBRDRLPersistence.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/util/GuidedDTBRDRLPersistence.java
@@ -35,8 +35,10 @@ import org.drools.workbench.models.datamodel.rule.FieldConstraint;
 import org.drools.workbench.models.datamodel.rule.FieldNatureType;
 import org.drools.workbench.models.datamodel.rule.FreeFormLine;
 import org.drools.workbench.models.datamodel.rule.FromCollectCompositeFactPattern;
+import org.drools.workbench.models.datamodel.rule.IAction;
 import org.drools.workbench.models.datamodel.rule.IFactPattern;
 import org.drools.workbench.models.datamodel.rule.SingleFieldConstraint;
+import org.drools.workbench.models.datamodel.rule.TemplateAware;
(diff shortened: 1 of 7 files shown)


drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/util/GuidedDTBRDRLPersistence.java
```diff
@@ -320,5 +322,15 @@ public void visitFreeFormLine(final FreeFormLine ffl) {
             fflClone.setText(interpolatedResult.toString());
             super.visitFreeFormLine(fflClone);
         }
+
+        @Override
+        protected IAction preProcessIActionForExtensions(final IAction iAction) {
+            if (iAction instanceof TemplateAware) {
+                TemplateAware clone = ((TemplateAware) iAction).cloneTemplateAware();
```
manstis, line 329: I don't understand why you need to clone a ```TemplateAware``` object.. it doesn't appear to retain state or anything?!?!

mcimbora: @manstis We use the same approach for FreeFormLine (see https://github.com/kiegroup/drools/pull/1373/files/f62232f63986d27ce2ce91ea098ff347c37450a7#diff-2927d560b80d1e5fbc24b58ab68e1d79R320). The problem is that if we modify the original object, then "Source" operation in the DT editor fails, as the variables are replaced by their values and the object is no longer recognized as the one having variables https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java#L258. The code tries to process it as a boolean cell value, which is incorrect.

manstis: Thanks for the explanation @mcimbora

manstis (approved):

PR #1377, opened 2017-07-21
DROOLS-1672 "not" in variable name doesn't work

baldimir:
- Adds reproducer


etirelli:
As I reported in the ticket, this problem is not fixable at the moment as the grammar is ambiguous. We need to wait for DMN 1.2 to define a context free grammar or we need an alternative to dynamic naming resolution.

baldimir:
Closing for now.

PR #1378, opened 2017-07-23
DROOLS-1666: changing return value for decision tables with hit policy C and C#

etirelli:
changing return value for decision tables with hit policy C and C# when no rule matches are found


etirelli:
@tarilabs @baldimir can you please review?

baldimir (approved):
Looks good. +1

tarilabs (approved):

PR #1380, opened 2017-07-24
Remove recursion from testLeakAfterSessionDispose

baldimir:
@mariofusco please review this.


mariofusco:
@baldimir I appreciate you removed the recursion and probably your version (even if a bit more complex) is better than mine since it at least prevents the stack overflow. Anyway either the 2 seconds time out is too short or more likely the problem is still there and now it is just causing this timeout. Let's discuss this.

baldimir:
jenkins retest this

PR #1384, opened 2017-07-25
DROOLS-1677 - not /someOoPath not supported

bobmcwhirter:
Adjust parser for appropriate look-ahead around DIV and un-labeled
OOPath constraints.


mariofusco:
ok to test

PR #1387, opened 2017-07-25
[DROOLS-1649] ServiceRegistry refactor

mariofusco:


drools-core/src/main/java/org/drools/core/metadata/ToManyPropertyLiteral.java
```diff
@@ -88,7 +88,7 @@ public boolean isManyValued() {
 
     @Override
     public OneValuedMetaProperty<T,List<R>> asFunctionalProperty() {
-        return (OneValuedMetaProperty<T,List<R>>) this;
+        return (OneValuedMetaProperty<T,List<R>>) null;
```
baldimir: This looks suspiciously. Shouldn't this return this?

baldimir (changes requested):
Just one comment worth doublechecking. Otherwise looks good.

PR #1391, opened 2017-07-27
[DROOLS-1676] invoke FEEL function

akoufoudakis:
We have the following use case.
There are two DMN models (files). One of them is a "central" file, which contains a top level decision.
The second DMN model (file) contains a "local" decision, which is, in its turn, a required decision for the top level decision, declared in the first file.
Due to incompleteness of DMN specification for DMN Import we agreed to try to implement a new FEEL function, which would evaluate the decision by name and namespace.

There are still few assumptions, which we have to verify, however, this pull request checks whether the whole "approach" is appropriate. Additionally, we had to modify some existing to be able to pass an instance of DMNRuntime.


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/CallDecisionFunction.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class CallDecisionFunction extends BaseFEELFunction {
+
+    public CallDecisionFunction() {
+        super("call decision");
+    }
+
+
+    public FEELFnResult<DMNResult> invoke(@ParameterName("ctx") EvaluationContext ctx, @ParameterName("namespace") String namespace, @ParameterName("modelName") String modelName,
+                                          @ParameterName("decisionName") String decisionName, @ParameterName("dmnContext") DMNContext dmnContext) {
```
etirelli: minor thing, but the @ParameterNames should use FEEL "practices", so use spaces instead of camel case. For instance: @ParameterName("decision name"), etc.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/CallDecisionFunction.java
```diff
@@ -0,0 +1,41 @@
...
+        DMNRuntime dmnRuntime = ctx.getDMNRuntime();
+        DMNModel dmnModel = dmnRuntime.getModel(namespace, modelName);
+        DMNResult requiredDecisionResult = dmnRuntime.evaluateDecisionByName(dmnModel, decisionName, dmnContext);
+        return FEELFnResult.ofResult(requiredDecisionResult);
```
etirelli: Needs to check for errors before returning the value, and in case of success, return only the value of the decision invoked, not the whole DMNResult. E.g.:

```
return FEELFnResult.ofResult(requiredDecisionResult.getContext().get(decisionName));
```

the code above is just a snippet, but the actual code needs to check for nulls, errors, etc.

etirelli:
The overall idea is correct. You will need to change the code to call the appropriate constructor for the EvaluationContextImpl to inject the DMNRuntime into it.
We will also need appropriate tests.
Good job!

akoufoudakis:
About the constructor... Do we have to replace all invocations of the old constructor with one parameter to the new one with two parameters? Or should it be only done only for specific invocations? I would say for all. What do you think about it?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNInvocationEvaluator.java
```diff
@@ -136,7 +133,8 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult d
                 }
             }
 
-            EvaluationContextImpl ctx = new EvaluationContextImpl( feel.getEventsManager() );
+            //EvaluationContextImpl ctx = new EvaluationContextImpl( feel.getEventsManager() );
+            EvaluationContextImpl ctx = new EvaluationContextImpl(feel.getEventsManager(), (DMNRuntime)eventManager);
```
etirelli: We need to review this cast. It is doable, but not ideal.

akoufoudakis: Sure... Do we want to hide cast inside the constructor implementation? @tarilabs mentioned that instead we could add DMNRuntime to DMNRuntimeEventManager, but it would "break" the interface...

tarilabs: I actually proposed to add `DMNRuntime` as a field inside `DMNRuntimeEventManager` because it is, as per discussions and if I understood correctly the points, acting as a "evaluation context" class.

I mentioned I would like `DMNRuntimeEventManager` be renamed to something like `DMNRuntimeEvaluationContext` but that would break an interface we have exposed in -api.
Or we could have `DMNRuntimeEvaluationContext` as a separate class/interface in core, although doing this instead would break the `evaluate()` method signature of `DMNExpressionEvaluator`...

Bottom line my understanding is this method signature:
https://github.com/kiegroup/drools/blob/9a48e63077d7a1bc0b36596bf77100de6752e0ae/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/api/DMNExpressionEvaluator.java#L36
somehow should allow to receive a "DMN evaluation context" comprised of the Event Manager and now the DMNRuntime, but I'm not sure which is the most affordable break to achieve so.

Unless we simply add method in `DMNRuntimeEventManager` to return the DMNRuntime, wouldn't break anything in -api I think; but this is like saying everything which is an "event manager" must be able to provide back its "parent" DMNRuntime; I'm speaking just from an API/interface semantic point of view. If this is acceptable, it is the non-breaking change which could avoid this internal cast.

tarilabs:
Commit 083f853 implements indeed the last idea proposed, for me is good thank you. We shall wait for @etirelli  to validate is a valid approach.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNInvocationEvaluator.java
```diff
@@ -136,7 +133,8 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult d
                 }
             }
 
-            EvaluationContextImpl ctx = new EvaluationContextImpl( feel.getEventsManager() );
+            //EvaluationContextImpl ctx = new EvaluationContextImpl( feel.getEventsManager() );
```
etirelli: minor thing, but as a general rule, we should remove old code, instead of commenting it.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/CallDecisionFunction.java
```diff
@@ -0,0 +1,41 @@
...
+import org.kie.dmn.feel.lang.EvaluationContext;
+
+public class CallDecisionFunction extends BaseFEELFunction {
+
+    public CallDecisionFunction() {
+        super("call decision");
+    }
+
+
+    public FEELFnResult<DMNResult> invoke(@ParameterName("ctx") EvaluationContext ctx, @ParameterName("namespace") String namespace, @ParameterName("modelName") String modelName,
```
etirelli: The result type of the invoke should be FEELFnResult<Object>. See comment below about the reason.

etirelli:
@akoufoudakis All the changes look good! I think we are ready to merge, just missing tests. Can you add them, please?

tarilabs:
Jenkins, test again.

tarilabs:
When the test is run by itself, it passes.
When the test is run as part of the suite, the number of error doesn't match.
I believe is a problem of the running order of decisions, and something is not working; investigating.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -377,6 +377,7 @@ private boolean evaluateDecision(DMNContext context, DMNResultImpl result, Decis
                     dr.setEvaluationStatus( DMNDecisionResult.DecisionEvaluationStatus.SUCCEEDED );
                 } else {
                     dr.setEvaluationStatus( DMNDecisionResult.DecisionEvaluationStatus.FAILED );
+                    return false;
```
tarilabs, line 380: @etirelli I believe during CI it emerged a bug which was related to the runtime order by which dependent decisions are executed.

In this case, Line 378 implies this branch is NOT a success (in contrast to line 338)
It would have set the status to failed.

But.
It would have jumped to the finally (line 395), and then proceeded to line 394 returning a `true`.

This is wrong, in all other cases if status != SUCCESS it is returning false.
Even more problematic because when decision calls other decision, they rely on the boolean return type of this method, for instance line 289.

the introduced `return false` at the highlighted line should solve for this bug which emerged.

tarilabs:
Jenkins, test again

tarilabs:
Jenkins, test again

tarilabs:
To recap.

A "Caller" model,
image

where decision "Calling outside" is calling the externalized decision evaluation:
image

In the normal/good test case, the parametrized serves to call the following "Calling" model:
image

@etirelli can you please check the boxed invocation, the parameter names are correct, specifically the last parameter "invocation context" is the desired one, and the introduced FEEL function name "call decision" the same. Thanks

etirelli:
@tarilabs looks good to me! The only thing I am considering is if we should name the function "invoke" instead of "call decision"? It gives us more flexibility in terms of semantics (we can extend it later to invoke multiple decisions instead) and also keeps consistency with the parameter name "invocation context" (although we could abbreviate it to simply "parameters" instead). Wdyt?

akoufoudakis:
@etirelli Parameters looks more concise and make more sense. As for the invoke function... DMN standard calls many things invocation. So, we can be consistent and rename it... Shall I do it now?

tarilabs:
good for me

akoufoudakis:
Jenkins, test again

tarilabs:
Jenkins, test again

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -1322,5 +1325,62 @@ public void testArtificialAttributes() {
...
+        {
+            DMNResult dmnResult = runtime.evaluateAll( dmnModel, context );
+            assertThat( DMNRuntimeUtil.formatMessages( dmnResult.getMessages() ), dmnResult.hasErrors(), is( false ) );
+            
+            DMNContext result = dmnResult.getContext();
+            assertThat( result.get( "Final decision" ), is( "The final decision is: Hello, John Doe your number once double is equal to: 6" ) );
+        }
+        
+        // Error if wrong namespace/name
+        {
```
baldimir: Please split these cases into separate test methods. If we have these together like this and e.g. the first case fail, we don't know if the other ones are working, because the assert failure will end the test method.

baldimir (changes requested):
Please see my comment, otherwise looks good.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
tarilabs: in this case, that's what I want. Invoking an external decision must first work when "all is good" but also must throw errors when errors. Can I leave it as is ?

etirelli: @tarilabs I think @baldimir is right here, each of these cases should be an individual test to make any problem immediately visible... just split into multiple test methods.

akoufoudakis: @tarilabs @etirelli If it is just splitting three cases (one successful, two failures) into three separate test methods, can I quickly do it?

akoufoudakis: @baldimir @tarilabs @etirelli Committed the changes. Please, check. If it is fine, run the Jenkins build, in case it is needed.

tarilabs: I think the question here is if instead of cloning the context it is fine to just have it in-place --with the right or wrong data-- for each test.

akoufoudakis: @tarilabs @etirelli Ooops. Haven't seen the one above. Have already committed the changes, suggested by @tarilabs.

baldimir (approved):
Looks good now, +1.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
baldimir: @tarilabs yes, it is enough. If we want to test the clone() context operation, it should be in its own test.

akoufoudakis:
@baldimir Thanks for your comments previously. Should we run checks? I don't have access rights to do it. Can anyone else initiate the checks?

baldimir:
Jenkins test again

akoufoudakis:
@baldimir Thanks for running the check. "All checks have passed".

etirelli (approved):
Looks good.

PR #1393, opened 2017-07-28
Make repository overrideable for KieServices.

kibertoad:
Currently default KieRepositoryImpl is hardcoded within KieServices and there is no way to use custom implementation. This PR makes it possible to set a repository override.

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieServicesImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieServicesImpl.java
@@ -68,8 +68,14 @@ public class KieServicesImpl implements InternalKieServices {
     
     private final ConcurrentMap<String, KieContainer> kContainers = new ConcurrentHashMap<String, KieContainer>();
 
+    private KieRepository repository;
+
     public KieRepository getRepository() {
-        return KieRepositoryImpl.INSTANCE;
+        return (repository == null) ? KieRepositoryImpl.INSTANCE : repository;
+    }
+
+    public void setRepository(KieRepository repository) {
+        this.repository = repository;
     }
 
     /**


kibertoad:
@mariofusco Can this small PR be reviewed, please?.. If it lacks something, I'm fully prepared to work on it further.

mariofusco:
@kibertoad The KieRepositoryImpl is a singleton and something internal to drools that has not be designed to be replaced with a different implementation. Can you please clarify why you have this need?

kibertoad:
@mariofusco We want to use our own dependency management infrastructure for updating rules that is not Maven repository based; currently this is absolutely impossible.

mariofusco:
@kibertoad Do you already have that alternative KieRepo implementation? Can you show it to me?

kibertoad:
@mariofusco Sure. It's not yet 100% final, as we'll also need some sort of automatic version purging and think about how to implement currently unsupported operations, but the idea is basically this:

public class InMemoryKieRepository implements KieRepository {
  private final Map<ReleaseId, KieModule> modules = new ConcurrentHashMap<>();

  @Override
  public ReleaseId getDefaultReleaseId() {
    return null;
  }

  @Override
  public void addKieModule(KieModule kModule) {
    modules.put(kModule.getReleaseId(), kModule);
  }

  @Override
  public KieModule addKieModule(Resource resource, Resource... dependencies) {
    throw new UnsupportedOperationException();
  }

  @Override
  public KieModule getKieModule(ReleaseId releaseId) {
    return modules.get(releaseId);
  }

  @Override
  public KieModule removeKieModule(ReleaseId releaseId) {
    return modules.remove(releaseId);
  }
}

baldimir:
I think that this will need much more work if it needs to be supported in the code. Our code is dependent on Maven in many APIs, so this solution looks a bit like a hack. The idea of having some sort of in memory repository is good, but I think this must be thought through. I would rather put it in a JIRA named something like "Support custom kjar lifecycles". Which will mean that there will be some pluggable classes that can support different lifecycle providers than Maven. But that's just my 5 cents :) What do you think?

kibertoad:
@baldimir Considering Liskov substitution principle, should this alternate implementation implement interface properly, why wouldn't it work with simple cases of kieContainer.updateToVersion()? Obviously it is not going to work with the Scanner, but that's beyond the scope of this PR, there's #1398 for that.

mariofusco:
@kibertoad We understand the need of making the KieRepository (and other services) implementation pluggable by users, but this has to be done through the ServiceRegistry that I've recently refactored ( see https://issues.jboss.org/browse/DROOLS-1649 ). For this reason I will close this pull request, but I also opened a ticket on our jira ( https://issues.jboss.org/browse/DROOLS-1690 ) to keep track of this requirement and make this possible asap. I'd appreciate if you could better clarify your use case and what you're attempting to achieve as a comment to that jira ticket.

PR #1398, opened 2017-07-28
Extract AbstractKieRepositoryScanner as an extension point

kibertoad:
This allows to easily implement alternative, non-Maven repository based scanners.


mariofusco:
@kibertoad Related to what I wrote here https://github.com/kiegroup/drools/pull/1393 and for similar reasons I am about to also close this pull request. I also have to add that I'm not totally getting the final goal of the proposed change. If I'm understanding correctly you want a KieScanner which is independent from maven and then you encapsulated some of the features provided by the original implementation in an abstract class. However also this abstract class also depends on maven (it imports org.eclipse.aether.artifact.Artifact for instance) so I don't see how you could use it without depending on maven too. Please also clarify this in the jira ticket that I linked in my comment to your other pull request.

PR #1405, opened 2017-08-02
[DROOLS-1688] allow alpha sharing when an hashed constraint requires …

mariofusco:
…type coercion on a field of type String


baldimir (approved):
+1. I tested the fix also with customer's reproducer and it fixes their problem.

PR #1407, opened 2017-08-04
DROOLS-1695 When unmarshalling attributePrefix as null on IBM JDK

tarilabs:

PR #1408, opened 2017-08-04
DROOLS-1695 When unmarshalling attributePrefix as null on IBM JDK

tarilabs:
forwardport to master from 72x of #1407


tarilabs:
@baldimir thanks again for your reports; I checked and indeed when on IBM JDK the default/silent prefix is returned as `null` from the xml infra, instead of respecting the JDK contract in the Javadoc for `XMLConstants.DEFAULT_NS_PREFIX`.

For example, an element with two attribute as:
```
<myelement a="avalue" drools:kind="kindvalue">
```
on JDK would unmarshal the attributes prefix as:
` XMLConstants.DEFAULT_NS_PREFIX` for the first `a` attribute
`drools` for the second `kind` attribute

while on IBM JDK is returning:
`null` for the first, and `drools` for the second.

tarilabs:
Jenkins test again

baldimir:
+1, looks good. Thanks.

baldimir (approved):

tarilabs:
Jenkins test again

tarilabs:
`9305 tests run, 442 skipped, 0 failed.`

tarilabs:
Downstream is an unrelated failure:
```
Test Result (1 failure / +1)
org.jbpm.kie.services.test.FilteredKModuleDeploymentServiceTest.testMultipleRemotableInPojoJar
```

PR #1410, opened 2017-08-07
[DROOLS-1696] immediately propagate event's expiration when equal to current timestamp

baldimir:
@winklerm This is the issue, we discussed on Friday.


baldimir:
Please merge just when reviewed by @mariofusco .

PR #1416, opened 2017-08-09
[DROOLS-1700] Avoid forking a useless thread issueing a FireUntilHalt…

mariofusco:
…Command when the engine is already firing


mswiderski:
out of curiosity, is there a way to sop fire until halt using commands? Thinking about KIE Server use case... what would be the halt there? kie container dispose?

mariofusco:
@mswiderski There's an HaltCommand for that: https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/command/runtime/rule/HaltCommand.java

mswiderski:
all clear, thanks @mariofusco

tarilabs (approved):

mariofusco:

PR #1423, opened 2017-08-18
DROOLS-1706 Allow marshalling of DMN Extension Elements

tarilabs:
Please note it is necessary for the marshaller to know (at runtime) the composition of the namespaces, which is stored in the in-memory object Definitions, in order to assign the namespaces and prefixes correctly.

### DIFF
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/ExtensionElementsConverter.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/ExtensionElementsConverter.java
@@ -22,6 +22,7 @@ import org.kie.dmn.model.v1_1.DMNModelInstrumentedBase;
 import org.kie.dmn.model.v1_1.DMNElement.ExtensionElements;
 
 import com.thoughtworks.xstream.XStream;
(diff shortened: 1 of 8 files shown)


etirelli (approved):

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/marshalling/v1_1/DMNExtensionRegister.java
```diff
@@ -17,10 +17,14 @@
...
 
 import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.io.xml.QNameMap;
 
 
 public interface DMNExtensionRegister {
 
     public void registerExtensionConverters(XStream xstream);
 
+    default void beforeMarshal(Object o, QNameMap qmap) {
```
tarilabs, line 27: One last thing, I realized if we don't like this method here too much, we could have a 2 level interface as usual (public vs an internal one) and I move this method there.
Doesn't change however the logic, for which I need to have access to the marshalled object just before marshalling in order to setup correctly at runtime the QNameMap

etirelli: ok, looks fine

etirelli (approved):

baldimir (approved):
+1

PR #1424, opened 2017-08-21
DROOLS-1704 - NPE when updating container with java class

mswiderski:


mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/1427

PR #1425, opened 2017-08-21
[DROOLS-1707] retrieve unchanged type declarations from existing kieb…

mariofusco:
…ase during incremental compilation

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
@@ -69,7 +69,6 @@ import org.drools.compiler.compiler.PackageBuilderErrors;
 import org.drools.compiler.compiler.PackageBuilderResults;
 import org.drools.compiler.compiler.PackageRegistry;
 import org.drools.compiler.compiler.ParserError;
-import org.drools.compiler.compiler.ProcessBuilder;
 import org.drools.compiler.compiler.ProcessBuilderFactory;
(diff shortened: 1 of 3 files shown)


drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1773,6 +1772,15 @@ private void processFunctions(PackageRegistry pkgRegistry,
     }
 
     public TypeDeclaration getAndRegisterTypeDeclaration(Class<?> cls, String packageName) {
+        if (kBase != null) {
```
tarilabs, line 1775: ok

tarilabs (approved):

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1773,6 +1772,15 @@ private void processFunctions(PackageRegistry pkgRegistry,
...
+            InternalKnowledgePackage pkg = kBase.getPackage( packageName );
+            if ( pkg != null ) {
+                TypeDeclaration typeDeclaration = pkg.getTypeDeclaration( cls );
+                if ( typeDeclaration != null ) {
+                    return typeDeclaration;
+                }
+            }
+        }
```
yurloc, line 1783: @mariofusco I wonder why the `typeDeclaration` is returned without registering it through `getAndRegisterTypeDeclaration()`. Is that OK?

PR #1427, opened 2017-08-21
[DROOLS-1704] avoid NPE trying to remove no longer existing Java clas…

mariofusco:
…s during incremental compilation

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
@@ -366,8 +366,10 @@ public class KieContainerImpl
         if (modifyingUsedClass) {
             // invalidate accessors for old class
             for (Class<?> cls : modifiedClasses) {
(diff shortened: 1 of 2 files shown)


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -366,8 +366,10 @@ private boolean applyResourceChanges(InternalKieModule currentKM, InternalKieMod
         if (modifyingUsedClass) {
             // invalidate accessors for old class
             for (Class<?> cls : modifiedClasses) {
-                ( (InternalKnowledgePackage) kBase.getKiePackage( cls.getPackage().getName() ) )
-                        .getClassFieldAccessorStore().removeClass( cls );
+                InternalKnowledgePackage kpackage = ( (InternalKnowledgePackage) kBase.getKiePackage( cls.getPackage().getName() ) );
+                if (kpackage != null) {
```
tarilabs, line 370: ok

tarilabs (approved):

PR #1429, opened 2017-08-22
Process variable is a Hibernate Proxy object

izodev:
For some cases my process variables were marshalled by the `Javassistlazyinitializer'` names which ends with **"_$$_jvst91a_55"**, thus when searching for the `canonicalName `it will throw a `org.hibernate.MappingException`: Unknown entity.


mswiderski:
what lead to variable being marshalled as proxy? Please open jira and describe how to reproduce the problem + a test case for the change is required.

baldimir:
-1, please do as Maciej proprosed.

izodev:
Sorry, I will do it as soon as I can.

PR #1439, opened 2017-09-01
GUVNOR-3451 : GRE DSL: Multiline code blocks break the editor

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3451

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -29,13 +29,16 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


jomarko (commented):
@Rikkola Hi, code changes looks good to me, I am just wondering if the test `testDSLExpansionLHS` shouldn't be updated in the same way? WDYT?

Rikkola:
```
String drl = "rule \"rule1\"\n"
                + "when\n"
                + "The credit rating is AA\n"
                + "then\n"
+ "end\n";
```

The code there does not have brackets, just DSL. In this case I would be in favor of keeping the test simple. 
Thought even if I just would make a new test for this situation, but ended up editing an old one with a comment. That is not necessary the best approach.

manstis:
Personally I think a new test is preferential over modifying an existing test.

Rikkola:
I was battling myself if I should make a new test. So now that others feel like there should be one, I'll make one. 
Probably better to isolate the problems into as simple tests as possible covering a the simple situation. Should give better protection for the future.

Rikkola:
Ok. New test added that targets the bug.

jomarko (approved):
I filed issue to create more testcases https://issues.jboss.org/browse/GUVNOR-3458, approving this PR.

manstis (approved):

PR #1443, opened 2017-09-04
[DROOLS-1720] allow to use KieScanner in combination with explicit up…

mariofusco:
…dateToVersion


tarilabs (commented):
if downgrade is meant to be a supported scenario, +1.
Maybe you want a review of this by @etirelli as well.

PR #1449, opened 2017-09-06
[DROOLS-1723] throw compile time error for misspelled sliding window …

mariofusco:
…predicate

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
@@ -169,10 +169,7 @@ public class PatternBuilder
         }
 
         if ( patternDescr.getObjectType() == null || patternDescr.getObjectType().equals( "" ) ) {
-            context.addError(new DescrBuildError(context.getParentDescr(),
-                                                 patternDescr,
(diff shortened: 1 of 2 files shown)


drools-compiler/src/main/java/org/drools/compiler/rule/builder/PatternBuilder.java
```diff
@@ -380,25 +373,33 @@ private void processSource( RuleBuildContext context, PatternDescr patternDescr,
...
-                    pattern.addBehavior( window );
-                } else if ( Behavior.BehaviorType.LENGTH_WINDOW.matches( behaviorDescr.getSubType() ) ) {
-                    SlidingLengthWindow window = new SlidingLengthWindow( Integer.valueOf( behaviorDescr.getParameters().get( 0 ) ) );
+                Behavior window = createWindow( behaviorDescr );
+                if (window != null) {
                     pattern.addBehavior( window );
+                    context.setNeedStreamMode();
+                } else {
+                    registerDescrBuildError( context, patternDescr,
+                                             "Unknown window type '" + behaviorDescr.getSubType() + "'" );
```
tarilabs, line 382: ok

tarilabs (approved):

PR #1466, opened 2017-09-12
GUVNOR-3482: Assets cannot be open after version of an asset is switched in big repository

manstis:
See https://issues.jboss.org/browse/GUVNOR-3482

I've moved synchronisation to the _mutating_ methods and not the accessors.

It seems to fix the reported problem.. but IDK a great deal about custom class loaders.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
+++ b/drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
@@ -206,7 +206,7 @@ public class ProjectClassLoader extends ClassLoader {
         return clazz;
     }
 
-    public synchronized Class<?> defineClass(String name, byte[] bytecode) {
+    public Class<?> defineClass(String name, byte[] bytecode) {
         return defineClass(name, convertClassToResourcePath(name), bytecode);
     }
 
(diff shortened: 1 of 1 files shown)


tomasdavidorg:

manstis:

drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
```diff
@@ -208,7 +206,7 @@ public static ProjectClassLoader createProjectClassLoader(ClassLoader parent, Ma
         return clazz;
     }
 
-    public Class<?> defineClass(String name, byte[] bytecode) {
+    public synchronized Class<?> defineClass(String name, byte[] bytecode) {
```
mariofusco: @manstis can this synchronized be removed? this method is only calling the defineClass overload which is already sync'ed.

manstis: @mariofusco It could, I guess... I added it since it is a ```public``` method on a ```public``` class and hence _could_ be called directly by someone at sometime from some place (and not just from the overloaded method). Happy to remove, but wanted to explain my reasoning.

manstis: @manstis Note to self.. yes L#209 can have ```synchronized``` removed. It's covered by L#213

manstis:
@mariofusco ```synchronized``` removed as discussed.

PR #1471, opened 2017-09-14
[DROOLS-1675] Wrong FEEL external function definition throws NPE

kurobako:
https://issues.jboss.org/browse/DROOLS-1675

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/FEELParser.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/FEELParser.java
@@ -183,12 +183,19 @@ public class FEELParser {
                         offendingSymbol );
             }
         } else if( "in".equals( token.getText() ) ) {
-            return new SyntaxErrorEvent( FEELEvent.Severity.ERROR,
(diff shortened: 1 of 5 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/FEELParser.java
```diff
@@ -183,12 +183,19 @@ private static SyntaxErrorEvent generateInvalidVariableError(Object offendingSym
...
-                                              line,
-                                              charPositionInLine,
-                                              offendingSymbol );
+            return new SyntaxErrorEvent(FEELEvent.Severity.ERROR,
+                                       Msg.createMessage(Msg.INVALID_VARIABLE_NAME, "keyword", token.getText()),
+                                       e,
+                                       line,
+                                       charPositionInLine,
+                                       offendingSymbol);
+        } else if( "}".equals( token.getText() ) ) {
```
etirelli: Unfortunately this condition is not enough to determine the proper error message. You would also need to check this was done as part of a context, otherwise a use just writing a random variable name that includes "}" will raise this error even if it is not a context.
I will take a look at how to do that.

etirelli:
Everything looks good, except for the error handling. Lets try to improve that before merging.

PR #1480, opened 2017-09-20
[7.3.x] GUVNOR-3479 : [GSS-RFE] Add support for missing keywords like "not matches" condition in Guided Rule

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3479

Same fix that master had for uberfire.

Part of a bundle:
https://github.com/kiegroup/drools/pull/1480
https://github.com/kiegroup/kie-wb-common/pull/1173
https://github.com/kiegroup/drools-wb/pull/626

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/oracle/OperatorsOracle.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/oracle/OperatorsOracle.java
@@ -35,7 +35,7 @@ public class OperatorsOracle {
 
     public static final String[] COMPARABLE_CONNECTIVES = new String[]{ "|| ==", "|| !=", "&& !=", "&& >", "&& <", "|| >", "|| <", "&& >=", "&& <=", "|| <=", "|| >=" };
 
-    public static final String[] STRING_OPERATORS = new String[]{ "==", "!=", "<", ">", "<=", ">=", "matches", "soundslike", "== null", "!= null" };
+    public static final String[] STRING_OPERATORS = new String[]{ "==", "!=", "<", ">", "<=", ">=", "matches", "not matches", "soundslike", "not soundslike", "== null", "!= null" };
 
     public static final String[] STRING_CONNECTIVES = new String[]{ "|| ==", "|| !=", "&& !=", "&& >", "&& <", "|| >", "|| <", "&& >=", "&& <=", "|| <=", "|| >=", "&& matches", "|| matches" };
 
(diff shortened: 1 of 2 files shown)


Rikkola:
@jomarko Here you are.

manstis (approved):

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/oracle/OperatorsOracle.java
```diff
@@ -35,7 +35,7 @@
 
     public static final String[] COMPARABLE_CONNECTIVES = new String[]{ "|| ==", "|| !=", "&& !=", "&& >", "&& <", "|| >", "|| <", "&& >=", "&& <=", "|| <=", "|| >=" };
 
-    public static final String[] STRING_OPERATORS = new String[]{ "==", "!=", "<", ">", "<=", ">=", "matches", "soundslike", "== null", "!= null" };
+    public static final String[] STRING_OPERATORS = new String[]{ "==", "!=", "<", ">", "<=", ">=", "matches", "not matches", "soundslike", "not soundslike", "== null", "!= null" };
 
     public static final String[] STRING_CONNECTIVES = new String[]{ "|| ==", "|| !=", "&& !=", "&& >", "&& <", "|| >", "|| <", "&& >=", "&& <=", "|| <=", "|| >=", "&& matches", "|| matches" };
```
jomarko, line 40: Shouldn't be some operators added also into this STRING_CONNECTIVES? Just asking because it already contains "&& matches" and "|| matches"

Rikkola: Good question. Not not sure if "&& not matches" compiles, but why wouldn't it I guess. Quick to test.

manstis: Can we merge this PR "as" is (since it's a cherry-pick of what has happened on ```master```)?

manstis:
Jenkins please retest this.

jomarko (approved):
Sure Michael, lets merge this to get same thing as on master. But I am still interested in my question.

manstis:
It's worth remembering that the Drools guys consider "connectives" to be dead.. (and ideally we should strip their very existence from the Workbench). @etirelli am I right, or am I right?

etirelli:
@manstis between the two alternatives, I will pick the second: "you are right"! ;)

jomarko:
It there would be voting for getting rid of those connectives, I would vote for that option, as it means simplification of the guided rule editor.

PR #1483, opened 2017-09-20
[DROOLS-1739] ensure creation of the segment memory of a subnetwork w…

mariofusco:
…hen the main segment memory is created from a prototype

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentUtilities.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentUtilities.java
@@ -76,6 +76,9 @@ public class SegmentUtilities {
 
         smem = restoreSegmentFromPrototype(wm, segmentRoot, nodeTypesInSegment);
         if ( smem != null ) {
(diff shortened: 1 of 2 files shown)


drools-core/src/main/java/org/drools/core/phreak/SegmentUtilities.java
```diff
@@ -76,6 +76,9 @@ public static SegmentMemory createSegmentMemory(LeftTupleSource tupleSource,
 
         smem = restoreSegmentFromPrototype(wm, segmentRoot, nodeTypesInSegment);
         if ( smem != null ) {
+            if (NodeTypeEnums.isBetaNode(segmentRoot) && ( (BetaNode) segmentRoot ).isRightInputIsRiaNode()) {
+                createRiaSegmentMemory( (BetaNode) segmentRoot, wm );
```
tarilabs, line 80: uses the externalized

tarilabs (approved):

PR #1498, opened 2017-09-26
GUVNOR-3334: Test dsl date marshalling

jomarko:
@manstis If you remember, long time ago I mentioned this one.


manstis (approved):
@jomarko I don't remember! However I do like the notion of retiring at 50 :-)

PR #1508, opened 2017-10-04
[DROOLS-1667] FEEL expression fails: 1 + if true then 1 else 2

kurobako:
https://issues.jboss.org/browse/DROOLS-1667

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
+++ b/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
@@ -51,9 +51,6 @@ expression
 // #2
 textualExpression
     : functionDefinition
-    | forExpression
-    | ifExpression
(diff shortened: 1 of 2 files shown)


etirelli:
@kurobako can you please add a few tests for the quantified expressions and for expressions as well? E.g.:
quantified: if every x in [ 1, 2, 3 ] satisfies x < 5 then "foo" else "bar"
for: count( for x in [1, 2, 3] returns x+1 )

etirelli (approved):

etirelli:
@kurobako sorry, one more thing, in the FEEL grammar, you added the parser rules into primary, but I don't see it removing from the previous place they were at the top of the file... don't we need to remove them from there?

etirelli (approved):

psiroky:
Could we make sure these PRs are properly tested by Jenkins before merging?

(Looks like we should add @kurobako to our org. unit so that we don't need to approve every single PR from him, correct?)

etirelli:
@psiroky Sorry, when I merged all "icons" were green. What did I miss?

I will add @kurobako to the org.

PR #1513, opened 2017-10-05
[DROOLS-1730] Cyclic information requirement generates a stackoverflow

kurobako:
https://issues.jboss.org/browse/DROOLS-1730


etirelli (approved):

etirelli:
This looks good for cycle detection at runtime. We should open a ticket to add cycle detection on the validator as well.

PR #1543, opened 2017-10-16
[DROOLS-1764] use (un)trusted xstream instances from XStreamUtils

mariofusco:


kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
```diff
@@ -222,7 +179,7 @@ public static String formatXml(String xml){
      }
     
     private XStream newXStream() {
-        XStream xStream = createXStream( staxDriver, Definitions.class.getClassLoader() );
+        XStream xStream = createNonTrustingXStream( staxDriver, Definitions.class.getClassLoader() );
```
tarilabs, line 182: I suppose this means for DMN we need to create some callbacks in order to allow for the end-user to provide whitelisting

tarilabs (approved):
...and I assume in general the whitelisting API will also be required for Drools in general ?

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/XStreamMarshaller.java
mariofusco: Yes, this is what you told me you were already doing. Is this correct?

tarilabs: unfortunately not, I've said that is what we _should_ be doing :)

PR #1545, opened 2017-10-18
Use XStreamUtils.createTrustingXStream(..) helper

manstis:
See BSIG emails regarding XStream.


jomarko (approved):
I assume you picked just drools-workbench-models, am I right?

manstis:
@jomarko Yes.. I'm just fixing my stuff... others should pick up other uses..

I'm doing ```guvnor``` and ```kie-wb-common``` too.

Rikkola (approved):

PR #1546, opened 2017-10-19
GUVNOR-3565

desmax74:
Changed visibility from private to public of the generateKieModuleMetaInfo method to read the TypeMetaInfos in the Kie-maven-plugin


desmax74:
@mariofusco

PR #1550, opened 2017-10-20
[DROOLS-1767] FEEL: add compile-time cycle detection

kurobako:
https://issues.jboss.org/browse/DROOLS-1767

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
@@ -46,9 +46,11 @@ import java.io.IOException;
 import java.io.Reader;
 import java.util.Collections;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Map;
 
 
 public class DMNCompilerImpl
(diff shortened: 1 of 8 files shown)


kurobako:
jenkins test this

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -199,8 +201,44 @@ private void processDrgElements(DMNCompilerContext ctx, DMNFEELHelper feel, DMNM
...
     }
 
+    private void detectCycles( DMNModelImpl model ) {
+        /*
+        Boolean.TRUE = node is either safe or already reported for having a cyclic dependency
+        Boolean.FALSE = node is being checked at the moment
+         */
+        final Map<DecisionNodeImpl, Boolean> registry = new HashMap<>();
+        for ( DecisionNode decision : model.getDecisions() ) {
+            assert decision instanceof DecisionNodeImpl;
```
baldimir: I'm not sure if we want such asserts in the code. They are not enabled by default in JVM. I think that if a decision should be just of DecisionNodeImpl type, the code should handle the incorrect types. What do you think?

baldimir (changes requested):
Please see my comment on the java assert, otherwise, looks ok.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
kurobako: that's fine, we can remove asserts, it's just a code discipline I am used to.

now, with incorrect types - there is only one implementation and we have lots of code that just tries to cast DecisionNode to DecisionNodeImpl since interface doesn't have many operations like getDependencies and such. so it is ugly but safe (yet)
I think this is a call for a future refactoring maybe, though very low priority I guess. I think I will ask Edson next Monday about this, but for the time being this cast is as good as it gets

baldimir (approved):
Thanks, looks fine now for me.

etirelli (approved):

PR #1553, opened 2017-10-24
[7.4.x] GUVNOR-3432: Test constraints that evaluates some formula

jomarko:
This PR asserts, if constraint that evaluates some formula, can bind the used fact field to a variable.

@manstis please have a look.


manstis (approved):

manstis:
Thanks @jomarko 

@mariofusco Any objections to this being merged (it's tests that assert a query we had in the workbench)? I know 7.4.x is "locked down" but this is just a couple more tests.

jomarko:
@mariofusco same here, should I close this? or is there chance to merge?

PR #1568, opened 2017-10-31
DROOLS-2072: Evaluate GDT that uses DSL in runtime

jomarko:
@manstis please have a look when you have a moment.


manstis (approved):
@jomarko I am liking your _realistic_ tests. To be honest I am pleasantly surprised they work.. I've never created "meaningful" decision tables whenever I test the editor :-)

manstis:
@jomarko Please let me know when green and I'll merge.

jomarko:
@manstis thank you for merge, yes, it is harder part of the task to think of the scenario. writing the test is the easier part.

PR #1571, opened 2017-10-31
DROOLS-1598 Support additional Signavio FEEL functions

tarilabs:
First 3 commits should be your original ones.

Following 4 commits should correspond to each of my PRs continuing the work

Last commit is to align to the changed function signature of base FEEL `ConcatenateFunction` which changed in -feel for its return type (a generic was added and the javacompiler complained for the signavio function relying on the base one).

With this, should realign to the original status to which we agreed.

I have not included anything which might have come later from your side, I'm thinking about the module thing or additional testings?


tarilabs:
@etirelli I am now checking with the additional commits you appended to the original branch/PR. Was a SecondDiffFunction missing all along ?

etirelli (approved):

etirelli (approved):

kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/feel/runtime/functions/AvgFunction.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "list" ) List list) {
+        return BuiltInFunctions.getFunction(MeanFunction.class).invoke(list);
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName("list") Number single) {
+        return BuiltInFunctions.getFunction(MeanFunction.class).invoke(single);
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName("n") Object[] list) {
```
jomarko, line 43: Just question, is this "n" correct, could someone explain what it stands for?

jomarko:
One more question, there is a lot of added functions, which are not covered with unit tests. What is the plan here please? Is there some agreement that QE (@baldimir) will add? If I remember correctly he was doing something like that for standard FEEL functions (not signavio).

kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/feel/runtime/functions/AvgFunction.java
etirelli: This is a convention commonly used in the DMN core functions to name list elements. It means that the function can be called using parameters named "n<number>", e.g:

avg( n1: 10, n2 : 20, n3 : 30 )

I guess "n" stands for the nth-element or something, but this was not defined by us. Several function in the DMN spec use this convention.

etirelli:
@jomarko , all functions should have unit tests. Here is the file: https://github.com/kiegroup/drools/pull/1571/files#diff-81932fe06dc546dbf9b7267ff24eddc1

Are we missing something?

kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/feel/runtime/functions/AvgFunction.java
jomarko: thanks

jomarko:
@etirelli thanks for pointing the test, not sure what was wrong, the main thing is they are present.

PR #1574, opened 2017-11-01
DROOLS-1903 : GRE DSL: Multiline code blocks break the editor

Rikkola:
https://issues.jboss.org/browse/DROOLS-1903

@manstis @jomarko New try with this ticket. Could you two please review this.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
@@ -113,6 +113,72 @@ public class RuleModelDRLPersistenceTest extends BaseRuleModelTest {
         }
     }
 
+    /*
+     * https://issues.jboss.org/browse/DROOLS-1903
+     */
+    @Test
+    public void testDSLWithFrom() {
+        final RuleModel m = new RuleModel();
+        m.name = "testDSLWithFrom";
+        m.lhs = new IPattern[2];
+        m.rhs = new IAction[0];
+
+        final FromCollectCompositeFactPattern from = new FromCollectCompositeFactPattern();
+
+        final FromCompositeFactPattern rightPattern = new FromCompositeFactPattern();
+        final FactPattern person = new FactPattern("Person");
+        person.setBoundName("$person");
+
+        m.lhs[0] = person;
+
+        rightPattern.setFactPattern(new FactPattern("Person"));
+        rightPattern.setExpression(new ExpressionFormLine(new ExpressionVariable(person.getBoundName(),
+                                                                                 person.getFactType())));
+
+        from.setRightPattern(rightPattern);
+
+        from.setFactPattern(new FactPattern("java.util.List"));
+
+        m.lhs[1] = from;
+
+        final RuleModel modelSpy = spy(m);
+        doReturn(true).when(modelSpy).hasDSLSentences();
+
+        final String drl = ruleModelPersistence.marshal(modelSpy);
+        final RuleModel tempModel = spy(ruleModelPersistence.unmarshalUsingDSL(drl,
+                                                                               null,
+                                                                               dmo,
+                                                                               ""));
+        doReturn(true).when(tempModel).hasDSLSentences();
+
+        final String newDrl = ruleModelPersistence.marshal(tempModel);
+
+        final RuleModel newModel = ruleModelPersistence.unmarshalUsingDSL(newDrl,
+                                                                          null,
+                                                                          dmo,
+                                                                          "");
+
+        assertTrue(newModel.lhs[0] instanceof FactPattern);
+        assertTrue(newModel.lhs[1] instanceof FromCollectCompositeFactPattern);
+
+        final String[] rows = newDrl.split("\n");
+        boolean foundWhen = false;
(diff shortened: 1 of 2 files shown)


drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
```diff
@@ -113,6 +113,55 @@ private void checkMarshallingUsingDsl(String expected,
         }
     }
 
+    /*
+     * https://issues.jboss.org/browse/GUVNOR-3451
+     */
+    @Test
+    public void testDSLWithFrom() {
+        final RuleModel m = new RuleModel();
```
manstis: Perhaps it's just me.. but given the issue was ">" in the DRL and given this test purports to use DSLs I'd have liked to have seen DRL and DSLs in the test rather than mocking the use of a DSL.. Are you confident this test touches the affected code?! Happy to refer final judgement to our QE supremo @jomarko

jomarko: Building locally and checking the test, I will let you know.

Rikkola: The issue is that when you save the code it corrupts the file. So the next time you open it, you see the > in the editor.
The DRL underneath needs to have the >, but the way it saved the from before was broken. The > was added into middle of a line.

Rikkola: But we can also form this same model from DRL. We actually do it here when we take the model to DRL and back to model again.

manstis: @Rikkola Thanks for your explanations.

manstis (approved):

jomarko:
@manstis @Rikkola 
- the fix works, great. 
- the test is enough, because no real DSL sentence used in attached reproducer, just checked DSL checkbox during asset creation, great
- However the modeled rule can not be validated, seems like problem with parsing the brackets, I am investigating a little bit more

jomarko:
Discussed with @Rikkola on IRC. 
There was problem with guided rule validation if the Use DSL checkbox was ticked during asset creation. The guided rule without ticked this checkbox works fine.
screenshot from 2017-11-02 13-53-36

Rikkola:
@jomarko Well spotted. You found a bug behind the bug ;-)

Rikkola:
Updated the PR with the fix for the other issue @jomarko found.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
```diff
@@ -113,6 +113,72 @@ private void checkMarshallingUsingDsl(String expected,
...
+        final String[] rows = newDrl.split("\n");
+        boolean foundWhen = false;
+        for (final String row : rows) {
+
+            if (row.trim().startsWith("then")) {
+                break;
+            }
+
+            if (foundWhen && !row.trim().isEmpty() && !row.trim().startsWith(">")) {
+                fail("Each row inside the LHS needs to start with >. The row with the issue: " + row);
```
jomarko, line 173: Hi @Rikkola, thanks for update, question here, Just LHS? not both LHS, RHS?

jomarko:
@Rikkola I can confirm that the fix works now, please, just have a look on my comment above before my approve.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
```diff
@@ -113,6 +113,72 @@ private void checkMarshallingUsingDsl(String expected,
         }
     }
 
+    /*
+     * https://issues.jboss.org/browse/GUVNOR-3451
```
jomarko: Would you mind use https://issues.jboss.org/browse/DROOLS-1903 ?

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
Rikkola: Could be both, but this issue just failed on collect in LHS.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
Rikkola: Sure

jomarko (approved):

PR #1576, opened 2017-11-02
DROOLS-2100: adding split() function

etirelli:
/cc @tarilabs @baldimir


tarilabs (approved):

tarilabs:
Just a general comment: is this expected going to be included in the next version of the spec? Because while I see the usefulness of "extended" functions like `now` or `modulo` which are very basic, this one seems a rather more complex one (in other words, I would not expect BAs to know what a regex is :) )

etirelli:
@tarilabs we will propose split() for DMN 1.2, but there are other functions already using regex in the spec, for instance matches().

Also, for the most cases, users will want to split a string given a delimiter like ",":

split( "Oranges,Bananas,Apples", "," )

To do this kind of thing, no real knowledge of regex is necessary.

tarilabs:
> To do this kind of thing, no real knowledge of regex is necessary.

@etirelli ... until they try to split on the dot :) [speaking from personal experience]

Anyway okay, makes more sense if already proposed for 1.2.

PR #1583, opened 2017-11-06
DROOLS-1775 fix JavaBackendTypeTest.testPerson fails on ConcurrentModificationExcptn

tarilabs:
Looking at the code, and the stacktrace reported by @baldimir , it would look the problem is originating by concurrent call of

```
JavaBackedType.of()
```

Which internally performs some memoization using the cache Map called `cache`. What would happen during specific race condition, two separate thread would cause the call to `computeIfAbsent` line61 concurrently, which in the case of simple HashMap would cause the ConcurrentModException.

Replacing with ConcurrentHashMap for fast reads and only lock on write.

I tried to include a reproducer, [for instance this one](https://gist.github.com/tarilabs/b44b2451070aa1eebb79234d13c63dde), but I was really unable to have it demonstrate the problem while running and the approach would not be deterministic in reproducing the suspected race condition.

So without going for the need of a byteman or similar test, I would propose for this modification to be merged based on the simple code analysis which can be done statically by looking at the code, and summarized above.

Thanks

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/JavaBackedType.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/JavaBackedType.java
@@ -2,7 +2,14 @@ package org.kie.dmn.feel.lang.impl;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


tarilabs:
```
20:38:03 Could not update commit status of the Pull Request on GitHub.
20:38:03 java.io.InterruptedIOException
20:38:03 	at org.kohsuke.github.RateLimitHandler$1.onError(RateLimitHandler.java:41)
20:38:03 	at org.kohsuke.github.Requester.handleApiError(Requester.java:702)
20:38:03 	at org.kohsuke.github.Requester._to(Requester.java:293)
20:38:03 	at org.kohsuke.github.Requester.to(Requester.java:234)
20:38:03 	at org.kohsuke.github.GHRepository.createCommitStatus(GHRepository.java:1075)
20:38:03 	at org.jenkinsci.plugins.ghprb.extensions.status.GhprbSimpleStatus.createCommitStatus(GhprbSimpleStatus.java:272)
20:38:03 	at org.jenkinsci.plugins.ghprb.extensions.status.GhprbSimpleStatus.onBuildComplete(GhprbSimpleStatus.java:231)
20:38:03 	at org.jenkinsci.plugins.ghprb.GhprbBuilds.onCompleted(GhprbBuilds.java:192)
20:38:03 	at org.jenkinsci.plugins.ghprb.GhprbBuildListener.onCompleted(GhprbBuildListener.java:26)
20:38:03 	at hudson.model.listeners.RunListener.fireCompleted(RunListener.java:211)
20:38:03 	at hudson.model.Run.execute(Run.java:1773)
20:38:03 	at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
20:38:03 	at hudson.model.ResourceController.execute(ResourceController.java:98)
20:38:03 	at hudson.model.Executor.run(Executor.java:404)
20:38:03 Caused by: org.kohsuke.github.HttpException: {"message":"API rate limit exceeded for kie-ci."
```

etirelli (approved):
the previous code was clearly problematic. Those collections need to be thread safe.

PR #1585, opened 2017-11-07
DROOLS-2111 Fix race condition with builderCache when running on JDK9

baldimir:
Fix for this: https://issues.jboss.org/browse/DROOLS-2111

@etirelli @tarilabs could you please review?


tarilabs:
@mariofusco can you kindly take a look in my place, when suitable, please?
I find strange the KnowledgeBuilderImpl is actually invoked in parallel.
Additionally, a ConcurrentModException in this case is a symptom (imho) in this specific area that something which was meant to be indeed executed sequentially, was by mistake performed in parallel.

@baldimir 
I think I recall now. In this specific case was done to cache a ~DMNCompiler which was expensive to build. The proposed fix would only tackle the symptom but not the real cause, which is we don't want (if was the case) to instantiate by mistake the ~DMNCompiler twice or more -- a cache read for still-null when in parallel was being written should have waited for actually the first ~DMNCompiler to be written inside this cache. I hope I was able to explain at high level what I suspect might be happening.

baldimir:
Closing this because of new findings. Please see the JIRA (https://issues.jboss.org/browse/DROOLS-2111)

PR #1587, opened 2017-11-08
DROOLS-2111 Fix a case where mapping function modifies the map

baldimir:
In JDK 9 it is now not allowed to use mapping function that changes the map with HashMap.computeIfAbsent() method. 

@mariofusco @etirelli @tarilabs could you please review?

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
@@ -2429,7 +2429,7 @@ public class KnowledgeBuilderImpl implements KnowledgeBuilder {
         final Map<String, Object> builderCache = getBuilderCache();
         final T cachedValue = (T) builderCache.get(key);
         if (cachedValue == null) {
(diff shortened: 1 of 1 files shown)


drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -2426,7 +2426,15 @@ private void normalizeStrictAnnotation(TypeResolver typeResolver, AnnotationDesc
     }
 
     public <T> T getCachedOrCreate(String key, Supplier<T> creator) {
-        return (T) getBuilderCache().computeIfAbsent(key, k -> creator.get());
+        final Map<String, Object> builderCache = getBuilderCache();
+        final T cachedValue = (T) builderCache.get(key);
+        if (cachedValue == null) {
+            final T newValue = creator.get();
```
tarilabs, line 2432: +1 is preserved lazy init

tarilabs (approved):

baldimir:
jenkins retest this

etirelli (approved):

PR #1589, opened 2017-11-08
Fix LambdaDataProvider when declaration is actually a global

tarilabs:
Do not take from tuple but indeed access global from WM.

@mariofusco what do you think?

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaDataProvider.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaDataProvider.java
@@ -16,6 +16,10 @@
 
 package org.drools.modelcompiler.constraints;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+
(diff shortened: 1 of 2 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaDataProvider.java
```diff
@@ -53,7 +53,7 @@ public Object createContext() {
 
     @Override
     public Iterator getResults( Tuple tuple, InternalWorkingMemory wm, PropagationContext ctx, Object providerContext ) {
-        Object obj = tuple.get( declaration ).getObject();
+        Object obj = declaration.getExtractor().isGlobal() ? declaration.getExtractor().getValue(wm, declaration.getIdentifier()) : tuple.get(declaration).getObject();
```
tarilabs, line 56: if extractor is global, now use `declaration.getExtractor().getValue(wm, declaration.getIdentifier())`
else
continue to use `tuple.get( declaration ).getObject()` as before this change.

PR #1597, opened 2017-11-09
DMN modules: Make util classes final and not instantiable

baldimir:
@etirelli @tarilabs could you please review? It's just minor changes in util classes that I made during coverage check.


tarilabs (approved):

etirelli:
I think we made a mistake using static methods for the factory methods, as it should be possible to override them with subclasses. 

That is how the service methods in the kie-api work. 

Now, I am not sure if we should make the factories final, or that will only make it hard to refactor them in the future without creating more backward incompatibilities. 

so, definitively +1 for making util classes final and factory classes uninstantiable, not so certain about making factory classes final. 

@mariofusco @mswiderski you guys have any opinion on this?

etirelli (approved):

PR #1615, opened 2017-11-20
Don't ignore nThreads parameter

yurloc:
This looks like a copy-paste error. The `Pool.SIZE` constant is a default value that is used when calling `newFixedThreadPool()` withouth `nThreads` parameter:
```java
    public ExecutorService newFixedThreadPool() {
        return newFixedThreadPool(Pool.SIZE);
    }
```


baldimir (approved):
+1, when checks pass.

PR #1618, opened 2017-11-21
DROOLS-2147 Fix Type test error

tarilabs:
/cc @etirelli @baldimir


etirelli (approved):

etirelli:
Looks good! 

Question, does this work for nested lists? like lists of lists of lists of something?

etirelli:
@tarilabs can you also please trim the string message in that error? when printing the actual value of the object, lets say print only the first 50 characters and then print a '...' at the end if there is more than 50 characters... wdyt?

tarilabs:
Message trimmed "Apache Camel"-style.

etirelli:
@tarilabs sorry, I merged another PR that generated a conflict. Can you please rebase this PR?

PR #1621, opened 2017-11-21
Fix FEEL date and time() handling of the first parameter `date`

tarilabs:
FEEL Spec Table 58 "date is a date or date time [...]
creates a date time from the given date (ignoring any time component)"

[that means ignoring any TZ from `date` parameter, too]

/cc @etirelli @baldimir @kurobako

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
@@ -16,15 +16,19 @@
 
 package org.kie.dmn.feel.runtime.functions;
 
-import java.time.*;
+import java.time.DateTimeException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
-import java.time.temporal.Temporal;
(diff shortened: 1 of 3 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
```diff
@@ -50,12 +54,17 @@ public DateAndTimeFunction() {
...
 
-    public FEELFnResult<TemporalAccessor> invoke(@ParameterName( "date" ) Temporal date, @ParameterName( "time" ) Temporal time) {
+    public FEELFnResult<TemporalAccessor> invoke(@ParameterName( "date" ) TemporalAccessor date, @ParameterName( "time" ) TemporalAccessor time) {
         if ( date == null ) {
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "date", "cannot be null"));
         }
         if ( !(date instanceof LocalDate) ) {
-            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "date", "must be an instance of LocalDate"));
+            // FEEL Spec Table 58 "date is a date or date time [...] creates a date time from the given date (ignoring any time component)" [that means ignoring any TZ from `date` parameter, too]
+            date = BuiltInFunctions.getFunction(DateFunction.class).invoke(date).getOrElse(null);
```
etirelli: I think this works @tarilabs , but specially in terms of performance cost, I wonder if we shouldn't just use the LocalDate.from( date ) static method. It returns the local date of a temporal accessor, ignoring time components...

etirelli (approved):

PR #1627, opened 2017-11-29
DROOLS-2151 DMN default disable Runtime type-check, ...

tarilabs:
... provide flag to explicitly enable it.

/cc @etirelli @baldimir

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
@@ -26,10 +26,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.drools.core.ClockType;
 import org.drools.core.definitions.InternalKnowledgePackage;
 import org.kie.api.internal.io.ResourceTypePackage;
 import org.kie.api.io.ResourceType;
 import org.kie.api.runtime.KieRuntime;
+import org.kie.api.runtime.conf.ClockTypeOption;
 import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNDecisionResult;
 import org.kie.dmn.api.core.DMNMessage;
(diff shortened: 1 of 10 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -469,6 +444,16 @@ private String getIdentifier(DMNNode node) {
         return node.getName() != null ? node.getName() : node.getId();
     }
 
+    private boolean performRuntimeTypeCheck(DMNModel model) {
+        return overrideRuntimeTypeCheck || ((DMNModelImpl) model).isRuntimeTypeCheck();
+    }
+
+    public final <T extends DMNOption> void setOption(T option) {
```
tarilabs, line 451: @etirelli : can you kindly let me know if this method should be exposed in the `DMNRuntime` public API and in that case I would create as well the companion `getOption()` method.

tarilabs:
As discussed this disables-by-default any typecheck during evaluation time.

Type check can be requested in the `kmodule.xml`. In this case it would be "saved as a preference" in the DMNModel and once at runtime the DMNRuntime will see the model is preferred with-enabled type check and will respect the preference (from the `kmodule.xml`).
Otherwise, it can be manually set by an override flag directly on the `DMNRuntime`.

Please notice the test utility class which created the DMNRuntime in the -core tests create a type-checking-enabled runtime, to avoid all previous non-regression tests to explode.

baldimir:
It could be useful to have this "settable" also from a global environmental property, so users can set type check globally "enabled" without the need to put the type check setting into each kjar. What do you think?

tarilabs:
@baldimir : I believe what you are describing should influence during compile-time, because kmodule.xml properties inherits from global env properties. I will add a test for it to check.

Although personally I prefer for this to _not_ influence runtime overrides, as it becomes more tricky during debug/understand behavior in deployed environments.

baldimir (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
etirelli: @tarilabs I don't think so. To be honest, I think we should have a separate interface (e.g. DMNRuntimeConfiguration) that would aggregate the configuration methods and then the DMNRuntime API would only expose the getConfiguration/setConfiguration() methods, not individual ones. But for now, what you did here is enough.

etirelli (approved):

PR #1629, opened 2017-11-30
DMN: Clarify error messages for wrong "name" attributes

baldimir:
The reason behind this change is that the error messages talk about name attribute which can be present not just for variables. 

+ fix name attribute in unrelated tests

@etirelli @tarilabs could you please review? I didn't create JIRA for this because it is a very minor change (but I can create one when needed).


tarilabs (approved):
Ok.
Although this would imply from FEEL perspective if a referenced variable is named wrongly, the error message would no longer explicitly say it's a "variable", simply a "wrong name". But okay for the greater good of DMN layer.

PR #1635, opened 2017-12-04
Pmml

mariofusco:


mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/1636

PR #1638, opened 2017-12-05
DROOLS-2161 Recursive function definitions in context entries are raising compile time errors

kurobako:
https://issues.jboss.org/browse/DROOLS-2161


baldimir (approved):
+1

PR #1641, opened 2017-12-07
DROOLS-2168 Extend coverage for concurrency scenarios

baldimir:
+ Add new cases (thanks @mackaypeter!) 
+ Some refactoring around existing cases

@mariofusco could you please take a look?


baldimir:
jenkins retest this

PR #1647, opened 2017-12-12
DROOLS-2169 Fix test failing on IBM JDK

baldimir:
- Generates the KeyStore file in the test instead of using a pre-created file. 
- Separates a test case of loading password from non-existant keystore.

@mswiderski @rstancel could you please review? Especially the assert, because I changed it - the key store file creation saves a key using DES (which should be supported on all JDKs), therefore it gets encoded.


mswiderski (commented):
it looks good to me, though I will let @rstancel to have final word on it. Thanks @baldimir

rstancel:
It looks fine and is working. @mswiderski approving

mswiderski (approved):

PR #1656, opened 2017-12-18
DROOLS-2188 Remove "expected = RuntimeException" from tests

baldimir:
- Replace the expected attribute with exception catch and assert over it's message
- Remove BranchTest which is a duplicate of NamedConsequencesTest
 
@mariofusco @lucamolteni could you please review? I did the changes to tests again, because although if I find the changes from before in my old computer tomorrow, I couldn't be sure in which state they are. So better to fix it again.


drools-compiler/src/test/java/org/drools/compiler/integrationtests/KieBuilderTest.java
```diff
@@ -159,12 +160,15 @@ public void testInvalidXsdTargetNamespace() {
...
-
-        ks.newKieContainer( km.getReleaseId() );
+        try {
+            final KieModule km = createAndDeployJar(ks,
+                                                    kmodule,
+                                                    releaseId1,
+                                                    r1);
+            Assertions.fail("Test should throw a validation exception!");
+        } catch (IllegalStateException ex) {
+            Assertions.assertThat(ex).hasMessageContaining("XSD validation failed");
```
triceo: May I suggest the following instead?

    Assertions.assertThatThrownBy(() -> yourCode())
        .isIntanceOf(SomeException.class)
        .hasMessageContaining(...)

No `try-catch`, no need to `Assertions.fail()`.

baldimir: Will do. Thanks for suggestion.

baldimir:
@triceo, the exception handling with AssertJ is added.

triceo (approved):

PR #1664, opened 2017-12-20
Fix for random nullpointer exceptions while serializing drools sessio…

manjunath-sp:
…n with sliding windows

Please refer to my post drools usage google groups. https://groups.google.com/forum/#!topic/drools-usage/Mhg6ivkJ5Oo

Adding the null check seems to have fixed the issue.


mariofusco:
Adding a null check in similar situations is almost always a bad idea: it doesn't fix the problem but just hides it and possibly propagates it somewhere else where it will be even harder to investigate and solve. In practice that null check is fixing the symptom and not the cause. Conversely my suggestion is to please provide a reproducer of this issue and I'll give a look at it.

manjunath-sp:
I'm not able to reproduce this issue consistently. Unfortunately, I am not in a position to share our rule base but I will try to create a reproducer.

manjunath-sp:
Please see the reproducer for this issue. 
https://github.com/manjunath-sp/drools-marshaller-tester/tree/master/drools.marshaller.tester

PR #1669, opened 2017-12-28
DROOLS-2200 fix NPE when structured DT output field is null

tarilabs:
/cc @etirelli


etirelli (approved):

etirelli:
thank you @tarilabs !

PR #1674, opened 2018-01-03
[DROOLS-2195] avoid CCE when changing a declared type during an incre…

mareknovotny:
https://issues.jboss.org/browse/RHBRMS-3053

cherry-picking from upstream https://github.com/kiegroup/drools/commit/d4fad626514d5cfeb54ce071d2bbe1c34148f52a


baldimir (approved):
+1, same changes as in master PR.

PR #1681, opened 2018-01-04
DROOLS-2213 Add reproducer for problem with recreating KieBase

baldimir:
@mariofusco this is the reproducer for issue https://issues.jboss.org/browse/DROOLS-2213


mariofusco:
reproduced and fixed by https://github.com/kiegroup/drools/commit/faf095c618438153d3de8c6327a27b4da3fb9dea

PR #1689, opened 2018-01-10
RHDM-243 Rule engine freezes when evaluating multiple PMML scorecards…

lanceleverich:
… (master)

* Added new unit test for scorecards - This also provides an example of how to use PMML with rule units
* Updated test_scorecard.pmml so that it provides reason codes for occupations other than SKYDIVER and ASTRONAUT
* Removed some System.out.println() calls

### DIFF
--- a/kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/ScorecardTest.java
+++ b/kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/ScorecardTest.java
@@ -56,17 +56,108 @@ public class ScorecardTest extends DroolsAbstractPMMLTest {
     private static final String source1 = "org/kie/pmml/pmml_4_2/test_scorecard.pmml";
     private static final String source2 = "org/kie/pmml/pmml_4_2/test_scorecardOut.pmml";
 
+    @Test
+    public void testMultipleInputData() throws Exception {
+        RuleUnitExecutor executor[] = new RuleUnitExecutor[3];
+        PMMLRequestData requestData[] = new PMMLRequestData[3];
+        PMML4Result resultHolder[] = new PMML4Result[3];
+        Resource res = ResourceFactory.newClassPathResource(source1);
+        kbase = new KieHelper().addResource(res, ResourceType.PMML).build();
+        
+        executor[0] = RuleUnitExecutor.create().bind(kbase);
+        executor[1] = RuleUnitExecutor.create().bind(kbase);
+        executor[2] = RuleUnitExecutor.create().bind(kbase);
+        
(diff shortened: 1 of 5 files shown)


kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/ScorecardTest.java
```diff
@@ -56,17 +56,108 @@
...
+            results[y].insert(resultHolder[y]);
+        }
+        
+        for (int z = 0; z < 3; z++) {
+            executor[z].run(unitClass);
+        }
+        
+        for (int p = 0; p < 3; p++) {
+            checkResult(resultHolder[p],expectedScores[p],expectedResults[p]);
+            results[p].forEach(r -> { System.out.println(r);});
```
jiripetrlik, line 127: Are these hard coded outputs necessary. Would it be possible to use some kind of logging framework instead? Something like slf4j etc.?

kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/ScorecardTest.java
```diff
@@ -78,7 +169,7 @@ public void testScorecard() throws Exception {
         data.insert(requestData);
         resultData.insert(resultHolder);
         executor.run(unitClass);
-
+System.out.println(resultHolder);
```
jiripetrlik, line 172: Can we avoid this direct output?

kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/ScorecardTest.java
```diff
@@ -56,17 +56,108 @@
...
+        requestData[2] = createRequest("125","Sample Score", 10.0, "STUDENT", "TN", false);
+
+        for (int x = 0; x < 3; x++) {
+            requests[x] = executor[x].newDataSource("request");
+            results[x] = executor[x].newDataSource("results");
+            pmmlDatas[x] = executor[x].newDataSource("pmmlData");
+            resultHolder[x] = new PMML4Result(requestData[x].getCorrelationId());
+        }
+        List<String> possiblePackages = calculatePossiblePackageNames("Sample Score", "org.drools.scorecards.example");
+        Class<? extends RuleUnit> unitClass = getStartingRuleUnit("RuleUnitIndicator",(InternalKnowledgeBase)kbase,possiblePackages);
```
jiripetrlik, line 109: It looks like quite complex logic to use PMML models. Would it be possible to wrap this logic in some kind of facade and provide simplified interface for user?

PR #1693, opened 2018-01-12
RHBRMS-3057: Test case

martinweiler:


martinweiler:
This PR is obsolete as @mariofusco already has applied the fix for RHDM-254.

PR #1705, opened 2018-01-19
DROOLS-2252: Increase coverage of marshaling from accumulate

jomarko:
Dependent on #1700 , @manstis please have a look.


manstis:
@jomarko I don't mind the additional check, so feel free to rebase and resubmit; however ```assertDSLEscaping(..)``` checks every line in the ```WHEN``` part is starts with ```>```

jomarko:
@manstis I agree that `assertDSLEscaping` asserts every line starts with '>' but I also think that `assertDSLEscaping` pass, if there are no lines, so I think my test enhancement makes sense, WDYT?

manstis:
@jomarko Like I said; I don't mind the additional check - and you're right it is an enhancement :-)

jomarko:
@manstis This PR is ready for merge.

manstis (approved):

PR #1707, opened 2018-01-22
RHDM-44 Remove drools-verifier test from regression test suite

ryanzhang:
Clean up the drools-verifier reference in regression testsuite.


ryanzhang:
Jenkins please retest this.

PR #1708, opened 2018-01-22
RHDM-44 Remove drools-verifier test from regression test suite

ryanzhang:
This is for removing the drools-verifier reference from drools regression testsuite.


ryanzhang:
Jenkins please retest this.

baldimir:
jenkins retest this

baldimir:
The module drools-verifier is still part of Drools repository, so -1 for this. If we at last really decide that we want to remov this test, the changes here remove just DRL, not the test itself, which is located in test class AbstractCompositeRestrictionTest which should be also removed in such case. 

Or the test class could be moved to drools-verifier module tests, if we don't want it in the test-coverage module.

PR #1709, opened 2018-01-22
[RHBRMS-3062] avoid node sharing when binding different properties

mariofusco:


mariofusco:
This was just a workaround, fixing the symptom, not the cause and not in all possible cases.

PR #1711, opened 2018-01-22
DROOLS-2264 An option to enable strict conformance mode

kurobako:
https://issues.jboss.org/browse/DROOLS-2264

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNAssemblerTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNAssemblerTest.java
@@ -2,13 +2,24 @@ package org.kie.dmn.core;
 
 import static org.junit.Assert.*;
 
+import org.junit.After;
 import org.junit.Test;
 import org.kie.api.KieServices;
 import org.kie.api.builder.KieFileSystem;
 import org.kie.api.builder.Message.Level;
 import org.kie.api.builder.Results;
+import org.kie.api.builder.model.KieModuleModel;
+import org.kie.api.runtime.KieContainer;
+import org.kie.dmn.api.core.DMNContext;
+import org.kie.dmn.api.core.DMNModel;
+import org.kie.dmn.api.core.DMNResult;
+import org.kie.dmn.api.core.DMNRuntime;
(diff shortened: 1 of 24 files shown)


kurobako:
jenkins retest this

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNAssemblerTest.java
```diff
@@ -28,4 +39,48 @@ public void testDuplicateModel() {
...
+        DMNModel model = runtime.getModel( "http://www.trisotech.com/dmn/definitions/_ecf4ea54-2abc-4e2f-a101-4fe14e356a46", "strictMode" );
+        DMNContext ctx = runtime.newContext();
+        ctx.set( "timestring", "2016-12-20T14:30:22z" );
+        DMNResult result = runtime.evaluateAll(model, ctx);
+        assertEquals( DateTimeFormatter.ISO_TIME.parse( "14:30:22z", OffsetTime::from ), result.getDecisionResultByName( "time" ).getResult() );
+    }
+
+    @Test
+    public void testStrictMode() {
+        System.setProperty("org.kie.dmn.strictConformance", "true");
```
tarilabs, line 55: Just for completeness, I believe it is always a good practice to wrap this kind of statement in a try-catch, so to have the "clearing" in a finally block. This way we ensure the clean-up code is executed, without relying on potential different behavior from the test framework.

tarilabs (commented):
thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNAssemblerTest.java
kurobako: I dare to disagree -- these are special tools with clear purpose while try-finally adds noise to what test is actually doing. Though I think I should change annotation from @Before to @After: "All @After methods are guaranteed to run even if a Before or Test method throws an exception".

kurobako: updated pr, changed annotation from @before to @after

etirelli (approved):

PR #1716, opened 2018-01-23
[6.5.x] DROOLS-2263 : Unexpected results in GDST when using enumerations with commas

Rikkola:
https://issues.jboss.org/browse/DROOLS-2263

This PR has a friend:
https://github.com/kiegroup/drools-wb/pull/767

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -60,6 +60,7 @@ import org.drools.workbench.models.commons.backend.rule.context.LHSGeneratorCont
 import org.drools.workbench.models.commons.backend.rule.context.LHSGeneratorContextFactory;
 import org.drools.workbench.models.commons.backend.rule.context.RHSGeneratorContext;
 import org.drools.workbench.models.commons.backend.rule.context.RHSGeneratorContextFactory;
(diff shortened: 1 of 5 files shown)


manstis (approved):
@Rikkola Looks good to me (and nice to see a different approach to splitting the values that may potentially contain commas - my way would have been quite different!)

manstis:
@Rikkola Presumably the fixes in your 2 PRs (for ```6.5.x```) will need replicating in ```7.5.x``` and ```master```?

Rikkola:
@manstis Yes. Also realized I need to move this to the -api package to be able to use it both client and server side. So moving this a bit.

Rikkola:
@manstis Ok, moved the Splitter to -api for better reusability and killed the duplicate in drools-wb

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -1286,17 +1287,11 @@ protected void populateValueList(final StringBuilder buf,
                 workingValue = workingValue.substring(0,
                                                       workingValue.length() - 1);
             }
-            final String[] values = workingValue.split(",");
+
+            final String[] values = ListSplitter.split("\"",workingValue);
```
jomarko: Just formatting, please split method arguments by one space.

jomarko (approved):
Nice coverage in PR. Thanks @Rikkola, just one formatting issue.

Rikkola:
Ok. Had to add the trim variable to ListSplitter for backward+forward compatibility.

jomarko (approved):

PR #1726, opened 2018-01-26
code fix related to PLANNER-865

mareknovotny:
There were line removal which was not followed by removal of https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L1433 and now it assigns the same variable to itself.

I found it by accident in findbugs report https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/droolsjbpm-build-bootstrap-downstream-pullrequests-7.5.x/5/findbugsResult/module.1185363731/source.746398736072648405/#1423

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -1411,7 +1411,7 @@ public class RuleModelDRLPersistenceImpl
         private Map<String, FieldConstraint> bindingsFields;
         protected DRLConstraintValueBuilder constraintValueBuilder;
         protected RHSGeneratorContextFactory generatorContextFactory;
-        protected Collection<RuleModelIActionPersistenceExtension> extensions;
+        //        protected Collection<RuleModelIActionPersistenceExtension> extensions;
 
         protected final RHSGeneratorContext rootContext;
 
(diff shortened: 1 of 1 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -1430,7 +1430,6 @@ public RHSActionVisitor(final boolean isDSLEnhanced,
             this.bindingsFields = bindingsFields;
             this.constraintValueBuilder = constraintValueBuilder;
             this.generatorContextFactory = generatorContextFactory;
-            this.extensions = extensions;
```
manstis, line 1433: Do we not also want to remove the instance variable ```extensions```?

manstis: https://github.com/mareknovotny/drools/blob/57a537307dcfb3a6920d7874de9a9fbf3a72c050/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L1414

mareknovotny: I don't know, but it could be the case if it is not used anywhere, sorry i just followed only the findbugs warning. I will try to search usage, and update if it is not anywhere used.

mareknovotny: so no more used anywhere and I removed the field definition too. Thanks for catching that.

PR #1730, opened 2018-01-29
DROOLS-2279 Realign extension registers test to master using DMNProfile

tarilabs:
/cc @baldimir @kurobako @etirelli

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
@@ -33,7 +33,6 @@ import org.kie.api.io.ResourceType;
 import org.kie.dmn.api.core.DMNCompiler;
 import org.kie.dmn.api.core.DMNCompilerConfiguration;
 import org.kie.dmn.api.core.DMNModel;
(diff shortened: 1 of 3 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -148,28 +146,6 @@ private DMNCompiler getCompiler(KnowledgeBuilderImpl kbuilderImpl) {
             return compiler;
         }
 
-        Map<String, String> extensionProperties = new HashMap<>();
-        kbuilderImpl.getBuilderConfiguration().getChainedProperties()
-                .mapStartsWith(extensionProperties, DMN_EXTENSION_REGISTER_PREFIX, false);
-        if( !extensionProperties.isEmpty() ) {
```
tarilabs, line 154: "dead code" elimination.
Given there is now always at least 1 Profile active (the default Kie DMN Profile), code will normally return from line 148.
The only time it would get through here would be when running TCK test, but there is no need for extension registers.

In turn, extension register mechanism as discussed no longer supported and any extension register will need to go through `DMNProfile` now.

kurobako:
@tarilabs thanks!

baldimir (approved):

etirelli (approved):

PR #1747, opened 2018-02-05
DROOLS-2274 support '?' variable in decision tables

kurobako:
https://issues.jboss.org/browse/DROOLS-2274

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
@@ -374,12 +374,14 @@ public class DMNEvaluatorCompiler {
 
     private DMNExpressionEvaluator compileDecisionTable(DMNCompilerContext ctx, DMNModelImpl model, DMNBaseNode node, String dtName, DecisionTable dt) {
         java.util.List<DTInputClause> inputs = new ArrayList<>();
+        List<DMNType> inputTypes = new ArrayList<>();
         int index = 0;
         for ( InputClause ic : dt.getInput() ) {
             index++;
             String inputExpressionText = ic.getInputExpression().getText();
             String inputValuesText =  Optional.ofNullable( ic.getInputValues() ).map( UnaryTests::getText).orElse( null);
             java.util.List<UnaryTest> inputValues = null;
+            DMNType inputType = DMNTypeRegistry.UNKNOWN;
             if ( inputValuesText != null ) {
                 inputValues = textToUnaryTestList( ctx,
                                                    inputValuesText,
@@ -392,6 +394,7 @@ public class DMNEvaluatorCompiler {
             } else if ( ic.getInputExpression().getTypeRef() != null ) {
                 QName inputExpressionTypeRef = ic.getInputExpression().getTypeRef();
                 BaseDMNTypeImpl typeRef = (BaseDMNTypeImpl) model.getTypeRegistry().resolveType(resolveNamespaceForTypeRef(inputExpressionTypeRef, ic.getInputExpression()), inputExpressionTypeRef.getLocalPart());
+                inputType = typeRef;
                 inputValues = typeRef.getAllowedValuesFEEL();
             }
             CompiledExpression compiledInput = feel.compileFeelExpression(
@@ -404,6 +407,7 @@ public class DMNEvaluatorCompiler {
                     dtName,
                     index );
             inputs.add( new DTInputClause(inputExpressionText, inputValuesText, inputValues, compiledInput ) );
+            inputTypes.add(inputType);
         }
         java.util.List<DTOutputClause> outputs = new ArrayList<>();
         index = 0;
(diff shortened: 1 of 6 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -394,8 +394,10 @@ private DMNExpressionEvaluator compileDecisionTable(DMNCompilerContext ctx, DMNM
                 BaseDMNTypeImpl typeRef = (BaseDMNTypeImpl) model.getTypeRegistry().resolveType(resolveNamespaceForTypeRef(inputExpressionTypeRef, ic.getInputExpression()), inputExpressionTypeRef.getLocalPart());
                 inputValues = typeRef.getAllowedValuesFEEL();
             }
+            ctx.setVariable("?", model.getTypeRegistry().resolveType(DMNModelInstrumentedBase.URI_FEEL, "string"));
```
etirelli: this doesn't look correct. The type of ? is the same type as the inputEntry, not hard coded string.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -394,8 +394,10 @@ private DMNExpressionEvaluator compileDecisionTable(DMNCompilerContext ctx, DMNM
...
             CompiledExpression compiledInput = feel.compileFeelExpression(
                     ctx,
+                    Collections.singletonMap("?", inputExpressionText),
```
etirelli: not sure about this...

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
```diff
@@ -94,12 +94,15 @@ public void registerFEELFunctions(Collection<FEELFunction> feelFunctions) {
...
 
-    public CompiledExpression compileFeelExpression(DMNCompilerContext ctx, String expression, DMNModelImpl model, DMNElement element, Msg.Message errorMsg, Object... msgParams) {
+    public CompiledExpression compileFeelExpression(DMNCompilerContext ctx, Map<String, Object> presetValues, String expression, DMNModelImpl model, DMNElement element, Msg.Message errorMsg, Object... msgParams) {
         CompilerContext feelctx = feel.newCompilerContext();
 
         for ( Map.Entry<String, DMNType> entry : ctx.getVariables().entrySet() ) {
             feelctx.addInputVariableType( entry.getKey(), ((BaseDMNTypeImpl) entry.getValue()).getFeelType() );
         }
+        for ( Map.Entry<String, Object> entry : presetValues.entrySet() ) {
+            feelctx.addInputVariable( entry.getKey(), entry.getValue() );
```
etirelli: the value of ? is not known at compile time, only its type.

kurobako: @etirelli do you mean that header itself is an expression and thus we should inject it only after we figured out what its value is?

etirelli: @kurobako ? is a variable, and its value is only known at runtime. Maybe the confusion is that you think "?" is a macro to replace with the "text" of the header, but it is not. It is really a variable... it could have been called "myVariable", but it was called "?" in the 1.2 spec.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -480,16 +486,24 @@ private DMNExpressionEvaluator compileDecisionTable(DMNCompilerContext ctx, DMNM
...
+                                ut.getText(),
+                                node.getIdentifierString(),
+                                index + 1);
+                    } finally {
+                        ctx.exitFrame();
+                    }
                 }
+                final CompiledExpressionImpl compiledInput = (CompiledExpressionImpl) inputs.get(i).getCompiledInput();
                 rule.getInputEntry().add( (c, x) -> tests.stream().anyMatch( t -> {
+                    c.setValue("?", compiledInput.evaluate(c));
```
etirelli: @kurobako although this is technically correct, I think we will face a performance issue because it will force the expression to be re-evaluated for each row of the decision table. In our example, it means that if our input expression is Person.name and the decision table has 1000 rows, this will be evaluated 1000 times. We need to set this in the context outside of the lambda as I mentioned in the e-mail.

kurobako: @etirelli I see. I was afraid evaluation context might be different outside of lambda and we get a different value. I will rewrite this part.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -1762,5 +1762,16 @@ private void assertDROOLS2286(DMNContext context) {
         assertEquals(2, resultObject.size());
     }
 
+    @Test
+    public void testDROOLS2274() {
```
baldimir: Please name the test based on the functionality it tests. I know there are other tests with JIRA numbers, but that's not a good practice, because it removes the "quicklook what the test is about" possibility.

baldimir (changes requested):

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
kurobako: sure, not a problem

etirelli:
Closing as this PR was replaced by #1890

PR #1760, opened 2018-02-09
DROOLS-2307 DMN enhanced for loop 

tarilabs:
/cc @etirelli @baldimir

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/ForExpressionNode.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/ForExpressionNode.java
@@ -16,15 +16,18 @@
 
 package org.kie.dmn.feel.lang.ast;
 
-import org.antlr.v4.runtime.ParserRuleContext;
-import org.kie.dmn.feel.lang.EvaluationContext;
-import org.kie.dmn.feel.lang.Type;
-import org.kie.dmn.feel.lang.types.BuiltInType;
-
+import java.math.BigDecimal;
+import java.math.MathContext;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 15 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/ForExpressionNode.java
```diff
@@ -109,27 +118,56 @@ public Type getResultType() {
...
+            BigDecimal end = (BigDecimal) rangeEnd;
+            fi = new ForIteration(name, start, end);
+        }
         return fi;
     }
 
+    private void valueMustBeANumber(EvaluationContext ctx, Object value) {
+        if (!(value instanceof BigDecimal)) {
+            ctx.notifyEvt(astEvent(Severity.ERROR, Msg.createMessage(Msg.VALUE_X_NOT_A_VALID_ENDPOINT_FOR_RANGE_BECAUSE_NOT_A_NUMBER, value), null));
+            throw new EndpointOfRangeNotOfNumberException();
```
etirelli, line 146: @tarilabs we need to be a bit conservative on the use of exceptions because they are really bad on performance. This is ok, as it only happens on edge cases, but avoiding is always good when possible.

etirelli (approved):

etirelli:
@tarilabs excellent job as usual!

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/ForExpressionNode.java
tarilabs: @etirelli indeed I limited its use here, simply because I didn't have the chance for a full AST node refactoring using Either just yet :stuck_out_tongue_closed_eyes:

PR #1762, opened 2018-02-12
DROOLS-2316 wrong ordering of ItemDefinition for compilation 

tarilabs:

( added @winklerm as per request on all PR I submit onwards )

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/ItemDefinitionDependenciesTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/ItemDefinitionDependenciesTest.java
@@ -87,7 +87,7 @@ public class ItemDefinitionDependenciesTest {
     }
     
     @Test
-    public void test1() {
+    public void testOrdering1() {
         ItemDefinition tCollateralRiskCategory = build("tCollateralRiskCategory");
         ItemDefinition tCreditRiskCategory     = build("tCreditRiskCategory");
         ItemDefinition tAffordabilityCategory  = build("tAffordabilityCategory");
@@ -123,7 +123,7 @@ public class ItemDefinitionDependenciesTest {
     }
     
     @Test
-    public void test2() {
+    public void testOrdering2() {
         ItemDefinition tMortgageType      = build("tMortgageType");
         ItemDefinition tObjective         = build("tObjective");
         ItemDefinition tRequested         = build("tRequested", tMortgageType, tObjective);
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/ItemDefinitionDependenciesTest.java
```diff
@@ -151,4 +152,35 @@ public void test2() {
         assertTrue("Index of tMortgageType < tLoanTypes", orderedList.indexOf(tMortgageType) < orderedList.indexOf(tLoanTypes));
         assertTrue("Index of tConformanceType < tLoanTypes", orderedList.indexOf(tConformanceType) < orderedList.indexOf(tLoanTypes));
     }
+    
+    @Test
+    public void test3() {
```
baldimir: Would it be possible to rename this test method to something like testOrdering() or similar? It would be probably good to rename also other test methods in this class which names are too generic.

baldimir (approved):
Please see my comment (minor thing, not blocker).

baldimir (changes requested):
Please see my previous comment.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/ItemDefinitionDependenciesTest.java
tarilabs: @baldimir I would have expected this comment :) , but what should I name it? testOrdering1, testOrdering2, testOrdering3 ? what would be a difference, is just 3 different data sets over which I test this `Comparator`.
Currently these are just 3 scenarii and potentially more scenarii in the future for tests.

baldimir (approved):

etirelli:
@tarilabs thanks for the quick fix!

@baldimir I think this is good to merge, but I was thinking if it would be good to create a test set for this ordering issue that generates all possible combinations of dependencies, lets say between 4 or 5 types and tests that the ordering algorithm works? We can't do many more than that, because the number of possible permutations with dependencies blows up really fast, but at least would give us some extra confidence given that we already had a couple bugs on this particular method.

etirelli (approved):

baldimir:
@etirelli agree with the idea of the generated test. Could you please create a JIRA for that and assign it to me?

tarilabs:
@baldimir https://issues.jboss.org/browse/DROOLS-2318 thanks :)

btw just for reference, before we were using an approximation for the topological sorting algorithm. Now it I would expect it to be stable, and bug to only highlights real bug in the implementation, not parts which the approximation didn't cover

PR #1769, opened 2018-02-14
DROOLS-2322 Move Kie extended FEEL function into appropriate profile

tarilabs:

please stand-by as I will provide below some commentary.


tarilabs:
Commit e097a77 shows the minimum necessary to move the Kie extended FEEL function into the appropriate profile

Commit afb739f highlight a fix for 2 potential bug which are highlighted thanks to this task.
1. DMNDTExpressionEvaluator was creating its own EvaluationContext to evaluate the DT. But this is wrong, as UnaryTest would rely on the EC to contain all the necessary functions, if they are used in an UT inside a cell. For example one test had a DT with a cell `< today()` but at this point because the evaluator was creating manually the EC, it did not contain the extended function "today", hence when the UT was evaluating, it was not resolving the expected function.
This is highlighted by this change: https://github.com/kiegroup/drools/pull/1769/commits/afb739f7aabad30f9f04488f989cdfdb381ff0d0#diff-6db8b96ef6f3c1bdb9e19639abe56e6cL74
In general, imho we should not create anymore the EvaluationContext manually for the internals but rely on the FEELImpl to provide the appropriate one.

2.  DMNInvocationEvaluator is relying on the `functionLocator` to resolve the function to be invoked, but in case otherwise it defaulted directly to the hard-coded list of built-in function (100% DMN Conformance level 3 only). This is wrong, because in such case the evaluator should actually check if this is the function is present in the list of configured FEEL function by the profile, which would contain the configuration-defined function on top of the DMN-built-in ones.
This is highlighted by this change: https://github.com/kiegroup/drools/pull/1769/commits/afb739f7aabad30f9f04488f989cdfdb381ff0d0#diff-30252cdd1a22702d3adac2e50b148376L92

Commit deb89fa realigns the Signavio profile for the function list, which in this case should re-use the Kie extended FEEL function explicitly. This is more a "proof" that the work in the aboves commit is producing the expected results.

Final commits are cosmetics and final realignment changes.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDTExpressionEvaluator.java
```diff
@@ -53,12 +54,12 @@
 
     private final DMNNode           node;
     private       DTInvokerFunction dt;
-    private       FEELImpl          feel;
+    private       FEELImpl              feel;
```
baldimir: Nitpicking comment about tab character misformatting this variable :)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNInvocationEvaluator.java
```diff
@@ -55,8 +56,13 @@
     private final String     functionName;
     private final List<ActualParameter> parameters = new ArrayList<>();
     private final BiFunction<DMNContext, String, FEELFunction> functionLocator;
+    private final FEEL feel;
 
-    public DMNInvocationEvaluator(String nodeName, DMNElement node, String functionName, Invocation invocation, BiFunction<DMNContext, String, FEELFunction> functionLocator ) {
+    /**
+     * @param functionLocator function to be used to resolve the FEELFunction to be invoked.
+     * @param feel in fase functionLocator is not able to resolve the desired function, it will be used for checking the resolution against the configured/built-in FEEL functions.
```
baldimir: Nitpicking comment no. 2 - typo.. should be "in case..."

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDTExpressionEvaluator.java
tarilabs: Thank you for spotting.
There is no tabs, actually.
This is the "enforced strict formatting" which I'm forced to use which failed, again.
Correcting manually now.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNInvocationEvaluator.java
tarilabs: thx

baldimir (approved):

tarilabs:

given I introduced f119379 81fd79a also to fix for a related issue (enhanced for-loop) I believe this will need a new review of the additional commits, please?

etirelli (approved):

PR #1776, opened 2018-02-19
DROOLS-2339 Reproducer of FileSystemResource bug that only manifests on Windows

rsynek:
[DO NOT MERGE] - reproducer - and beware - reproducible on Windows only.


baldimir:
This can be closed, because it was fixed by this: https://github.com/kiegroup/drools/pull/1927

PR #1777, opened 2018-02-19
DROOLS-2318 DMN: Add generated tests that check ItemDefinition sorting

baldimir:
@tarilabs @etirelli could you please review? 

JIRA: https://issues.jboss.org/browse/DROOLS-2318


baldimir:
After logging off I realized that I have wrong one thing - when I start e.g. with 3rd item I still add it as first. So situations like 1st item with no deps, 2nd item with no deps but 3rd with deps is uncovered. Will fix that tomorrow.

baldimir:
Still needs some tweaking. I need to avoid circular dependencies.

baldimir:
Closing this for now, I will open a new PR today.

PR #1781, opened 2018-02-20
Proper parsing of the Decision Services

SimonRinguette:
This PR contains the required code modification to properly parse DMN decision services.

Be patient with me as this is my first PR to the official code base. I tried to not screw the file layout and limit the changes to a a minimum.

There are a few things to consider here:

1- The DMN 1.1 serialization is broken and do not actually support serializing decision service. It is clearly an oversight in the spec as nobody really implemented it. There are two ways to go about it: change the dmn.xsd in the test cases and keep the roundtrip test, remove the roundtrip test and keep the dmn.xsd without modification (UnmarshalMarshalTest.java).

2- The xstream reading of the decision service inputData is "hacked: because it re-use the same xml localName: inputData which is also used for the actual InputData. This is why the DecisionConverter class has a "hack" to preperly read the inputData as a DMNElementReference.

I've included two test cases (read and round trip) with this modification.


SimonRinguette:
Note that I've added a getDecisionService() method on the definitions model, an alternative could be to fully consider them DRG elements (as they probably should be) but i'm not sure of the potential side effects of that

etirelli:
@SimonRinguette thank you Simon, it looks good to me, but will wait for @tarilabs to take a look as he knows the serialization code better. 

My real only concern is changing the XSD: is it really necessary?

tarilabs:
Thank you @SimonRinguette , and I have the same question.
Additionally, I believe anyway changing the XSD is not an ideal strategy. If a user is using in the xml something like: 
```
<definitions ... xsi:schemaLocation="http://www.omg.org/spec/DMN/20151101/dmn.xsd dmn.xsd " ... >
```
any change we perform on our xsd won't make a difference.

Can you share some brother details why this modification is necessary, and then we evaluate the best course of action as necessary. For instance I'm missing to understand two points sorry: why the xsd as per the spec is "broken" and how a decision service serialized in v1.2 is going to look like instead, please? We can do an online meeting if preferred.
Thanks!

SimonRinguette:
All right, i'm not proposing to patch the XSD lightly. The issue mainly is that in DMN 1.1, the XSD schema as it stands does not allow to serialize a DecisionService element under a Definitions element. DecisionService has a substitution group for namedElement and there is no way to serialize that... basically, the XSD defines the DecisionService element and the tDecisionService type but they did not correctly craft the rest of the XSD to add an anchor point to serialize them.

In 1.2, tDecisionService is drgElement exactly as my change suggest. The 1.2 serialization of decsion services is exactly this very small change.

There are to me 3 course of action possible to support decision services:

1- Delete the round trip test case and don't change the XSD
2- Don't write the decision services (only read them) so that we keep it 1.1 purely compliant while writing (but not reading), we would still need to remove the test case as it would fail
3- Accept that 1.1 has a clear omission and make that small change to the XSD.

I'm proposing 3 considering that, the DMN 1.2 RTF recognize that the 1.1 schema is broken and DecisionService can't be serialized and that they fixed the schema in the exact same way.

I'm really religious about respecting the XSD normally but this is where I have to take exception. Its broken!

tarilabs:
Thank you for this feedback it helps a lot to understand more the context on the DMN spec side. I still have a doubt however what is the final goal for this PR: is it to persist DecisionService somehow on v1.1 until the next version v1.2 is out?

If so:
What about we serialize for 1.1 the DecisionService inside the extension elements, or something on those lines, please?

I never implied the xsd change proposal was done too lightly. I'm just afraid even doing so on our end would not imply end-user doing this on their end (see my example with schemalocation), so I'm really afraid they would not be compatible/on the same page.

SimonRinguette:
@tarilabs Since both you and @etirelli seems to prefer an alternative that would keep the XSD, I can amend the PR to instead serialize in the extension elements. This will make sure that there is no XSD change but will result in a weird looking serialization. The extensionElements can hold any element that is coming from an external namespace... i will need to create a holder element to do that... it would look like that:

```
<dmn:extensionElements>
         <drools:decisionServices>
                          <dmn:decisionService>....</dmn:decisionSerice>
                          <dmn:decisionService>....</dmn:decisionSerice>
                          <dmn:decisionService>....</dmn:decisionSerice>
         </drolls:decisionServices>
</dmn:extensionElements>
```
Would this be an acceptable solution?

tarilabs:
I believe, although not certain, it could be something along the lines of:
```
<dmn:extensionElements>
                          <dmn11fix:decisionService>....</dmn11fix:decisionSerice>
                          <dmn11fix:decisionService>....</dmn11fix:decisionSerice>
                          <dmn11fix:decisionService>....</dmn11fix:decisionSerice>
</dmn:extensionElements>
```
What do you think? naturally the prefix can be different, just to give an idea.

I would wait for @etirelli to comment nevertheless.

SimonRinguette:
Actually, this could work but would redefine the decisionService element in a different namespace. My suggestion at least re-use the decisionService element proposed by XSD

SimonRinguette:
Also, best practices at RH question, would you prefer me to submit an additional commit to this PR and then someone can squash them once accepted or rewrite this commit and loose the history?

tarilabs:
> My suggestion at least re-use the decisionService element proposed by XSD

Oh sorry I missed that; would that "container" shade the problem of fitting the decisionService directly under the extensionElements? Because I tried without the "container" and indeed got a validation error even with multialiasing the dmn namespace url. Will try this also from this side.

With the dmn11fix you could have a reference to an xsd that contained the decisionService as currently drafted for v1.2.
 
I don't have any preference between them, the one you evaluate is the most linear for implementation and end-user point of view.

> best practices at RH question, would you prefer me to submit an additional commit to this PR and then someone can squash them once accepted or rewrite this commit and loose the history?

There is no hard requirement there, you can follow either paths as you prefer, the most convenient. In case of multiple commits we do squash by default (so at the end they would be almost equivalent), unless there is an explicit request to non-squash the commit for any specific reason (I don't believe it is this case).

tarilabs:
[edited previous comment it was mis formatted]

tarilabs:
Okay I see, this indeed validates:
```
<p:definitions exporter="" exporterVersion="" expressionLanguage="http://www.omg.org/spec/FEEL/20140401" id="idvalue0" label="" name="" namespace="http://tempuri.org" typeLanguage="http://www.omg.org/spec/FEEL/20140401" xmlns:p="http://www.omg.org/spec/DMN/20151101/dmn.xsd" xmlns:x="x" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.omg.org/spec/DMN/20151101/dmn.xsd dmn.xsd ">
  <p:extensionElements>
   <x:asd>
	<p:decisionService  label="" name="" >
	  <p:description>p:description</p:description>
	  <p:extensionElements/>
	  <p:outputDecision href="http://tempuri.org"/>
	  <p:encapsulatedDecision href="http://tempuri.org"/>
	  <p:inputDecision href="http://tempuri.org"/>
	  <p:inputData href="http://tempuri.org"/>
	</p:decisionService>
   </x:asd>
  </p:extensionElements>
```
thank you for suggesting this additional option; I was not aware it would have also fixed the issue.

SimonRinguette:
@tarilabs @etirelli I've changed the implementation to revert the XSD changes and serialize in an extension. The second commit should be squashed with the first

etirelli:
ok to test

etirelli (approved):

PR #1790, opened 2018-02-26
[6.5.x] DROOLS-2263: Increase coverage

jomarko:
@Rikkola please have a look

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/util/ListSplitterTest.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/util/ListSplitterTest.java
@@ -14,7 +14,6 @@
 */
 package org.drools.workbench.models.datamodel.util;
 
-import org.drools.workbench.models.datamodel.util.ListSplitter;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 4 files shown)


Rikkola (changes requested):
So happy with the tests. Not 100% sure about the trim, but we talked with @jomarko about this in IRC and he is investigating it.

jomarko:
Ok investigation ended up with this.
- Declare condition column that checks **String field** of a fact, and operator **is contained in comma separated list**
- In the table data append 2 rows
- In the first added cell type **"a,b",c**
- In the second added cell type **c,"a,b"**
- Check the source
  - Without this PR 
    - For the first case, expected source is generated,
    - For the second one **name in("c", ""a", "b"")**
  - With this PR
    - For both first and second case expected content is generated

The same behavior ca be seen if during column definition we declare value lists as:
- **'"a,b"',c**
- **c,'"a,b"'**

jomarko:
@Rikkola could you please have a look on added tests:
- `changeQuoteCharacterTrimming_Eight`
- `changeQuoteCharacterTrimming_Nine`

They fails, however I am not sure if they are valid, please let me know.

Rikkola:
Argh, my eyes. Those are like regepx. I'll try to see if they are valid or not :-)

Rikkola:
@jomarko Investigated and I think I was wrong about the " character. So it looks like it is not used to group values like the ' character. So `"a","b"` gives `"a"` and `"b"` and` "a,b"` gives `"a` and `b"`.

If you do `'"a,b"'` you get `"a,b"`.
If you do `'a,b'` you get `a,b`.

So in the end the " is just another character. Due to this the test _Eight and _Nine should actually fail.

jomarko:
@Rikkola well, so I removed those invalid testcases, please have a look when you have a moment.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/util/ListSplitterTest.java
```diff
@@ -95,4 +94,76 @@ public void changeQuoteCharacterSkipTrimming() throws Exception {
...
+                                                  "\"Boston,'Helsinki',Finland,Prague\"");
+        assertEquals(1, split.length);
+        assertEquals("Boston,'Helsinki',Finland,Prague", split[0]);
+    }
+
+    @Test
+    public void changeQuoteCharacterTrimming_Seven() throws Exception {
+        final String[] split = ListSplitter.split("\"",
+                                                  true,
+                                                  "\"Boston,'Helsinki',Finland,\"Prague\"\"");
```
Rikkola, line 165: @jomarko And as seen here the " is not just a character. Meaning I was wrong again. Quite confusing. Not actually sure if _Eight and _Nine were bugs or not.

Rikkola (approved):

PR #1793, opened 2018-02-27
DROOLS-1679 revert and empty-list if any error in FEEL list filtering

tarilabs:
... when filter returns empty collection when there are errors
evaluating filter.

Continued from commit 6c8d3f9a1b6c3ba420445c99f50a65f066a01ca7
https://github.com/kiegroup/drools/pull/1514

@etirelli can you kindly check for me the additional tests are correct, in the sense they are indeed respecting the DMN v1.1 specification, please? Thx
(I am discovering these edge cases as a side-effect of implementing the compilation)

/cc @baldimir @kurobako @winklerm

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/FilterExpressionNode.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/FilterExpressionNode.java
@@ -23,6 +23,7 @@ import java.util.List;
 import org.antlr.v4.runtime.ParserRuleContext;
 import org.kie.dmn.api.feel.runtime.events.FEELEvent.Severity;
 import org.kie.dmn.feel.lang.EvaluationContext;
+import org.kie.dmn.feel.lang.impl.SilentWrappingEvaluationContextImpl;
 import org.kie.dmn.feel.lang.types.BuiltInType;
 import org.kie.dmn.feel.util.Msg;
 
@@ -66,9 +67,8 @@ public class FilterExpressionNode
         try {
             if( filter.getResultType() != BuiltInType.BOOLEAN ) {
                 // check if index
-                Object f = filter.evaluate( ctx );
-                if (f == null) return null;
-                if (f instanceof Number ) {
+                Object f = filter.evaluate(new SilentWrappingEvaluationContextImpl(ctx)); // I need to try evaluate filter first, ignoring errors; only if evaluation fails, or is not a Number, it delegates to try `evaluateExpressionsInContext`
+                if (f != null && f instanceof Number) {
                     // what to do if Number is not an integer??
                     int i = ((Number) f).intValue();
                     if ( i > 0 && i <= list.size() ) {
@@ -109,7 +109,10 @@ public class FilterExpressionNode
             // using Root object logic to avoid having to eagerly inspect all attributes.
             ctx.setRootObject(v);
 
-            Object r = this.filter.evaluate( ctx );
+            // a filter would always return a list with all the elements for which the filter is true.
+            // In case any element fails in there or the filter expression returns null, it will only exclude the element, but will continue to process the list.
+            // In case all elements fail, the result will be an empty list.
+            Object r = this.filter.evaluate(new SilentWrappingEvaluationContextImpl(ctx)); // evaluate filter, ignoring errors 
             if( r instanceof Boolean && ((Boolean)r) == Boolean.TRUE ) {
                 results.add( v );
             }
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELListsTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELListsTest.java
@@ -69,8 +69,15 @@ public class FEELListsTest extends BaseFEELTest {
                 {"[ {x:1, y:2}, {x:2, y:3} ][x = 0]", Collections.emptyList(), null },
 
                 // Other filtering
-                {"[\"a\", \"b\", \"c\"][a]", null, FEELEvent.Severity.ERROR },
-
+                {"[\"a\", \"b\", \"c\"][a]", Collections.emptyList(), null }, // DROOLS-1679
+                {"{ a list : [ { a : false, b : 2 }, { a : true, b : 3 } ], r : a list[a] }.r", Arrays.asList( new HashMap<String, Object>() {{ put("a", true);  put("b", BigDecimal.valueOf( 3 )); } } ), null },
+                {"{ a list : [ { a : false, b : 2 }, { a : null, b : 3 }, { b : 4 } ], r : a list[a] }.r", Collections.emptyList(), null },
+                {"{ a list : [ \"a\", \"b\", \"c\" ], x : 2, a : a list[x]}.a", "b", null },
+                {"{ a list : [ { x : false, y : 2 }, { x : true, y : 3 } ], x : \"asd\", a : a list[x] }.a", Arrays.asList( new HashMap<String, Object>() {{ put("x", true);  put("y", BigDecimal.valueOf( 3 )); } } ), null },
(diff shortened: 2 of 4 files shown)


baldimir (approved):
Looks good, however I didn't check the spec compliance. I am leaving this to @etirelli.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELListsTest.java
```diff
@@ -70,6 +70,11 @@
 
                 // Other filtering
                 {"[\"a\", \"b\", \"c\"][a]", null, FEELEvent.Severity.ERROR },
+                {"{ a list : [ { a : false, b : 2 }, { a : true, b : 3 } ], r : a list[a] }.r", Arrays.asList( new HashMap<String, Object>() {{ put("a", true);  put("b", BigDecimal.valueOf( 3 )); } } ), null },
+                {"{ a list : [ \"a\", \"b\", \"c\" ], x : 2, a : a list[x]}.a", "b", null },
+                {"{ a list : [ { x : false, y : 2 }, { x : true, y : 3 } ], x : \"asd\", a : a list[x] }.a", Arrays.asList( new HashMap<String, Object>() {{ put("x", true);  put("y", BigDecimal.valueOf( 3 )); } } ), null },
```
tarilabs:
```
{ a list : [ { x : false, y : 2 }, { x : true, y : 3 } ], x : \"asd\", a : a list[x] }
```

`a list` is a list
`x` is NOT a number
spec Table 54: Semantics of lists
page 127 of the printed PDF

> e1 is a list and e2 is an integer

first row is NOT this case
so
it might be row 3

> e1 is a list and type(FEEL(e2 , s')) is boolean

as long as when X in the special scope S' would exist, as a boolean
which it does
because the element of the list contains a definition for an attribute X
in the special scope S' : `{ x : false`

tarilabs (commented):
summary of as-discussed on IRC with @baldimir

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELListsTest.java
```diff
@@ -70,6 +70,11 @@
...
+                {"{ a list : [ { x : false, y : 2 }, { x : true, y : 3 } ], x : false, a : a list[x] }.a", Arrays.asList( new HashMap<String, Object>() {{ put("x", true);  put("y", BigDecimal.valueOf( 3 )); } } ), null },
+                {"{ a list : [ { x : false, y : 2 }, { x : true, y : 3 } ], x : null, a : a list[x] }.a", Arrays.asList( new HashMap<String, Object>() {{ put("x", true);  put("y", BigDecimal.valueOf( 3 )); } } ), null },
```
etirelli: @tarilabs all 5 test cases you are adding here make sense to me. An attribute in the relation will shadow any variable previously defined.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/FilterExpressionNode.java
```diff
@@ -94,27 +94,29 @@ public Object evaluate(EvaluationContext ctx) {
...
+                // element using the name item, unless the list element is a context that contains the key "item".
+                ctx.setValue("item", v);
+
+                // using Root object logic to avoid having to eagerly inspect all attributes.
+                ctx.setRootObject(v);
+
+                Object r = this.filter.evaluate(ctx);
+                if (r == null) {
+                    // I was expecting to handle "filter" as a Predicate, but is not a Predicate.
+                    return null;
```
etirelli: I am not sure about this @tarilabs . 

An error when evaluating the filter for one element should not stop the evaluation. My understanding is that an error in this case will work as a "false" filter, and as such, simply skip the element. In other words, shouldn't this be "continue" instead of "return null"? Or, in case the error is on the predicate itself, not in the evaluation of a given element, then should be "break" instead of "return null".

I am pretty sure there is a ticket that says an error on a filter should return an empty list, not null, and we changed the behaviour in the past to do so.

etirelli: Page 113 of the spec says: "The filter expression is evaluated for each item in list, and a list containing only items where the filter expression is true is returned."

Based on other similar cases in the spec, for a filter that returns any value other than "true", including "null" or errors, the result should be an empty list, not null.

tarilabs: > I am pretty sure there is a ticket that says an error on a filter should return an empty list, not null, and we changed the behaviour in the past to do so.

That is indeed DROOLS-1679, which did not cover correctly for these additional cases, just line 72 in the test (just above the new introduced tests).

tarilabs: > Based on other similar cases in the spec, for a filter that returns any value other than "true", including "null" or errors, the result should be an empty list, not null.

You might be right, the problem is we don't have yet test cases which cover this possibility (as the CI is green). I will write some tests to make sure this coverage of edge-cases is not exposing bugs in other scenarii.

tarilabs: > Based on other similar cases in the spec, for a filter that returns any value other than "true", including "null" or errors, the result should be an empty list, not null.

@etirelli In other words we are saying original DROOLS-1679 was actually not a bug.

I tried to make a test case which is a bit more "edge":
```
{ a list : [ { a : false, b : 2 }, { a : null, b : 3 }, { b : 4 } ], r : a list[a] }.r
```
based on the spec, we are saying for
`a list[a]`
the filer: `a`
> is evaluated for each item in list, and a list containing only items where the filter expression is true is returned

so:
- when evaluated on the first element `{ a : false, b : 2 }` is false boolean-literal, hence not selected, 
- when evaluated on the second element `{ a : null, b : 3 }` is the null-literal, hence not selected, and 
- when evaluated on the third element `{ b : 4 }` it would be an error, hence null, hence not selected

But following this logic, the original test provided with DROOLS-1679:
```
["a", "b", "c"][a]
```
should return an empty list, not a null-literal and an error as per requested DROOLS-1679 itself.

What do you think @etirelli ?

etirelli: @tarilabs I believe your analysis is correct, and so DROOLS-1679 was a mistake. We can bring this up today at the TCK meeting and hear some opinions.

etirelli: Matteo, I discussed this further today with Bruce and he is actually having problems with this case when importing data from XML, as XML not always have all the attributes set. 

Can you go ahead and fix this as per our previous discussion? I.e., a filter would always return a list with all the elements for which the filter is true. In case any element fails in there or the filter expression returns null, it will only exclude the element, but will continue to process the list. In case all elements fail, the result will be an empty list.

Also, for cases like the 3rd bullet in your example, a missing attribute in a map should be considered null, not an error. 

So, in the following example:
```
{
    people : [ { firstName : "bob" }, { firstName : "max" } ]
    result : people[ lastName = null ]
}
```

result will be evaluated to:
```
 [ { firstName : "bob" }, { firstName : "max" } ]
```

tarilabs: Implemented, and this test is also included.

tarilabs:
I rebased the PR branch on the following:

> a filter would always return a list with all the elements for which the filter is true. In case any element fails in there or the filter expression returns null, it will only exclude the element, but will continue to process the list. In case all elements fail, the result will be an empty list. Also a missing attribute in a map [being the map an element in the list] should be considered null, not an error.

/cc @etirelli @baldimir @kurobako @winklerm

etirelli (approved):

etirelli:
Looks good @tarilabs, thanks!

PR #1794, opened 2018-02-27
DMN adjust logging level for ItemDefinitionDependenciesGeneratedTest

tarilabs:
@baldimir this was generating quite many logging lines, but I believe the original intention is for those to be printed out only if now needs to debug any issue with the dependency ordering.

What do you think?


baldimir (approved):
+1, I wasn't sure if I should log this as info or debug.

PR #1796, opened 2018-02-27
[DROOLS-2284] rules with more than three variable references in conse…

btison:
…quence require functions and objects with higher arity


mariofusco:
ok to test

PR #1802, opened 2018-03-06
Fix test failing on Windows

baldimir:
The modified test failed on Windows, because a Person class couldn't be found. 

@mariofusco could you please review?


mariofusco:
ok to test

PR #1804, opened 2018-03-07
Decision service null pointer exception fix

SimonRinguette:
@tarilabs There was a small issue with your modification if you have a model without extensions. Can you have a look at this fix please.


tarilabs:
Ok to test

baldimir:
@SimonRinguette @tarilabs hello, do we have a test for this NPE case? If not, we should create one, so we have this covered for future regression testing.

tarilabs:
@baldimir 

> @SimonRinguette @tarilabs hello, do we have a test for this NPE case? If not, we should create one, so we have this covered for future regression testing.

Proposed here: https://github.com/SimonRinguette/drools/pull/2

PR #1806, opened 2018-03-08
Generate OSGi metadata for drool-pmml

ctron:


drools-pmml/pom.xml
```diff
@@ -26,6 +26,7 @@
   </parent>
 
   <artifactId>drools-pmml</artifactId>
+  <packaging>bundle</packaging>
```
mariofusco, line 29: Afaik drools-ppml is no longer supported. @lanceleverich can you please confirm?

drools-pmml/pom.xml
```diff
@@ -84,6 +91,10 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
```
mariofusco, line 96: Shouldn't you declare here the packages imported and exported by this module?

ctron: Not necessarily. As the plugin will pick that up automatically. Only in the case something goes wrong with that you need to start declaring things.

ghost:

ghost:

mariofusco:
drools-ppml is no longer supported

PR #1817, opened 2018-03-15
The description attribute of decision service is now properly parsed

SimonRinguette:
@tarilabs I realised today that my inexperience with XStream Parser showed a little bit. I was not calling super on a method which prevented the parsing of the description attribute of decision services. I have made the correction and added a test case.


tarilabs:
Ok to test

baldimir (approved):
Looks good, thanks!

tarilabs (approved):
Good catch thanks!

PR #1823, opened 2018-03-20
DROOLS-2403 Close BufferedWriter after use

baldimir:


baldimir:

baldimir:
Fails from downstream build are not related to the PR.

PR #1832, opened 2018-03-22
DROOLS-1663 Kie DMN doesn't support IMPORT decisions between DMN files

tarilabs:
Requires https://github.com/kiegroup/droolsjbpm-knowledge/pull/300

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
@@ -72,13 +72,13 @@ public class DMNAssemblerService implements KieAssemblerService {
     }
 
     @Override
-    public void addResources(Object kbuilder, List<ResourceWithConfiguration> resources, ResourceType type) throws Exception {
+    public void addResources(Object kbuilder, Collection<ResourceWithConfiguration> resources, ResourceType type) throws Exception {
         KnowledgeBuilderImpl kbuilderImpl = (KnowledgeBuilderImpl) kbuilder;
         DMNCompilerImpl dmnCompiler = (DMNCompilerImpl) kbuilderImpl.getCachedOrCreate(DMN_COMPILER_CACHE_KEY, () -> getCompiler(kbuilderImpl));
         DMNMarshaller dmnMarshaller = dmnCompiler.getMarshaller();
         if (resources.size() == 1) {
             // quick path:
-            internalAddResource(kbuilderImpl, dmnCompiler, resources.get(0), Collections.emptyList());
+            internalAddResource(kbuilderImpl, dmnCompiler, resources.iterator().next(), Collections.emptyList());
             return;
         }
         List<DMNResource> dmnResources = new ArrayList<>();
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
@@ -23,7 +23,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 import javax.xml.namespace.QName;
 
@@ -44,9 +43,11 @@ import org.kie.dmn.core.api.DMNFactory;
 import org.kie.dmn.core.compiler.DMNCompilerConfigurationImpl;
 import org.kie.dmn.core.compiler.DMNCompilerImpl;
 import org.kie.dmn.core.compiler.DMNProfile;
+import org.kie.dmn.core.compiler.ImportDMNResolverUtil;
 import org.kie.dmn.core.compiler.profiles.ExtendedDMNProfile;
 import org.kie.dmn.core.impl.DMNKnowledgeBuilderError;
 import org.kie.dmn.core.impl.DMNPackageImpl;
+import org.kie.dmn.feel.util.Either;
 import org.kie.dmn.model.v1_1.Definitions;
 import org.kie.dmn.model.v1_1.Import;
 import org.kie.internal.builder.ResultSeverity;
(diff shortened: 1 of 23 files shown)


tarilabs:
[ as noted on JIRA, please notice these PRs are relevant only for importing ItemDefinition(s) and BKM(s) ]

tarilabs:
Jenkins test again

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -59,12 +71,68 @@ public ResourceType getResourceType() {
...
+        DMNCompilerImpl dmnCompiler = (DMNCompilerImpl) kbuilderImpl.getCachedOrCreate(DMN_COMPILER_CACHE_KEY, () -> getCompiler(kbuilderImpl));
+        DMNMarshaller dmnMarshaller = dmnCompiler.getMarshaller();
+        if (resources.size() == 1) {
+            // quick path:
+            internalAddResource(kbuilderImpl, dmnCompiler, resources.iterator().next(), Collections.emptyList());
+            return;
+        }
+        List<DMNResource> dmnResources = new ArrayList<>();
+        for (ResourceWithConfiguration r : resources) {
+            Definitions definitions = dmnMarshaller.unmarshal(r.getResource().getReader());
```
baldimir, line 86: Nitpicking comment No.1: This could be extracted to a separate method. Something like unmarshallResources(...).

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -59,12 +71,68 @@ public ResourceType getResourceType() {
...
+            QName modelID = new QName(definitions.getNamespace(), definitions.getName());
+            DMNResource dmnResource = new DMNResource(modelID, r, definitions);
+            dmnResources.add(dmnResource);
+        }
+        // enrich with imports
+        for (DMNResource r : dmnResources) {
+            for (Import i : r.getDefinitions().getImport()) {
```
baldimir, line 93: Nitpicking comment No.2: This could be extracted to a separate method. Something like resolveImports(...).

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -59,12 +71,68 @@ public ResourceType getResourceType() {
...
     @Override
-    public void addResource(Object kbuilder, Resource resource, ResourceType type, ResourceConfiguration configuration)
-            throws Exception {
+    public void addResources(Object kbuilder, Collection<ResourceWithConfiguration> resources, ResourceType type) throws Exception {
+        KnowledgeBuilderImpl kbuilderImpl = (KnowledgeBuilderImpl) kbuilder;
+        DMNCompilerImpl dmnCompiler = (DMNCompilerImpl) kbuilderImpl.getCachedOrCreate(DMN_COMPILER_CACHE_KEY, () -> getCompiler(kbuilderImpl));
+        DMNMarshaller dmnMarshaller = dmnCompiler.getMarshaller();
+        if (resources.size() == 1) {
+            // quick path:
+            internalAddResource(kbuilderImpl, dmnCompiler, resources.iterator().next(), Collections.emptyList());
```
baldimir, line 81: Question: Will there be a possibility of an external import? I mean some external resource outside of the kjar or from URL. Asking because for this one file import is not resolved (and that is correct in current case).

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNBaseNode.java
```diff
@@ -53,6 +54,36 @@ public String getName() {
         return source != null ? source.getName() : null;
     }
 
+    @Override
+    public String getModelNamespace() {
+        if (source != null) {
+            Object parent = source.getParent();
+            while (!(parent instanceof Definitions)) {
```
baldimir: Nitpicking comment No.3: This could be extracted to a separate method. E.g. findParent().

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNBaseNode.java
```diff
@@ -53,6 +54,36 @@ public String getName() {
...
+            return ((Definitions) parent).getNamespace();
+        }
+        return null;
+    }
+
+    @Override
+    public String getModelName() {
+        if (source != null) {
+            Object parent = source.getParent();
+            while (!(parent instanceof Definitions)) {
```
baldimir: See my previous comment. The extracted method can be reused here.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: Sorry not following: what should I extract? only this line 85?

baldimir: No, sorry for not being exact. I meant the whole logic block here. Same applies for all nitpicking comments.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: Would do later, as currently import resolution is limited in functionality.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNCompilerTest.java
```diff
@@ -120,4 +127,37 @@ public void testRecursiveFunctions() {
...
+        for (DMNMessage message : dmnModel.getMessages()) {
+            System.out.println(message);
+        }
+
+        DMNContext context = runtime.newContext();
+        context.set("A Person", mapOf(entry("name", "John"), entry("age", 47)));
+
+        DMNResult evaluateAll = runtime.evaluateAll(dmnModel, context);
+        for (DMNMessage message : evaluateAll.getMessages()) {
+            System.out.println(message);
```
baldimir: Probably leftover from debugging. Please use asserts or logger instead if this is still needed here.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNCompilerTest.java
```diff
@@ -120,4 +127,37 @@ public void testRecursiveFunctions() {
...
+        }
+        System.out.println(evaluateAll);
```
baldimir: Same as previous comment.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: but why should I extract it? This part of code is not reused -unless I am missing something.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNBaseNode.java
tarilabs: for `getModel()` and `getModelNamespace` I agree will append commit asap

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNCompilerTest.java
tarilabs: will append commit

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
baldimir: Just for readability. But this is really just a nitpicking, so for me it is good as it is if you like it better as it is.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: not as far as I know. @etirelli ?

etirelli: This might be necessary in the future, but support is not implemented yet. For now, the imported model has to be reachable by the kie-container classloader.
I don't think we need to worry at the moment about "external" imports as that would have additional concerns, like authentication and authorization, that are not covered by the spec at the moment.

etirelli (approved):

etirelli:
@tarilabs this looks good to go as the first version of the implementation (i.e., supports for data types and BKMs).

PR #1840, opened 2018-03-26
RHBA-623_7.5.x

lanceleverich:
* Added method to PMML4Compiler to retrieve field definitions for fields that are the result/output of an earlier sub-model
    * Added methods to AbstractModel to find MiningField and OutputField entries from the model
    * Updated the modelChain template so that it only attempts to create PMMLRequestData objects using the fields that are available in the controlling model
    * Added a test to the MiningmodelTest for modelChain mode models
    * Added the test_mining_model_modelchain.pmml file as a test model


lanceleverich:
Jenkins retest please

PR #1845, opened 2018-03-30
DROOLS-2439 Manage empty relation cell as null

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
@@ -16,15 +16,14 @@
 
 package org.kie.dmn.backend.marshalling.v1_1.xstream;
 
+import org.kie.dmn.model.v1_1.DMNModelInstrumentedBase;
+import org.kie.dmn.model.v1_1.InformationItem;
+import org.kie.dmn.model.v1_1.Relation;
+
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.converters.MarshallingContext;
 import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
-import org.kie.dmn.model.v1_1.DMNModelInstrumentedBase;
-import org.kie.dmn.model.v1_1.Expression;
-import org.kie.dmn.model.v1_1.InformationItem;
-import org.kie.dmn.model.v1_1.LiteralExpression;
-import org.kie.dmn.model.v1_1.Relation;
 
 public class RelationConverter extends ExpressionConverter {
     public static final String EXPRESSION = "expression";
@@ -38,26 +37,9 @@ public class RelationConverter extends ExpressionConverter {
         if (COLUMN.equals(nodeName)) {
             r.getColumn().add((InformationItem) child);
         } else if (ROW.equals(nodeName)) {
-            org.kie.dmn.model.v1_1.List row = (org.kie.dmn.model.v1_1.List) child;
-            normalizeEmptyCells(row);
-            r.getRow().add(row);
+            r.getRow().add((org.kie.dmn.model.v1_1.List) child);
         } else {
             super.assignChildElement(parent, nodeName, child);
-        }
-    }
-
-    /**
-     * Normalize empty cell in DMN Relation as null expression.
-     */
-    private void normalizeEmptyCells(org.kie.dmn.model.v1_1.List row) {
-        // DROOLS-2439
-        for (Expression e : row.getExpression()) {
-            if (e instanceof LiteralExpression) {
-                LiteralExpression literalExpression = (LiteralExpression) e;
-                if (literalExpression.getText().isEmpty()) {
-                    literalExpression.setText("null");
-                }
-            }
         }
     }
 
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
@@ -56,7 +56,7 @@ public class RelationConverter extends ExpressionConverter {
                 LiteralExpression literalExpression = (LiteralExpression) e;
                 if (literalExpression.getText().isEmpty()) {
                     literalExpression.setText("null");
-                    }
+                }
             }
         }
     }
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
@@ -16,14 +16,16 @@
 
 package org.kie.dmn.backend.marshalling.v1_1.xstream;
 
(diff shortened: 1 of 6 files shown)


baldimir (approved):

tarilabs:
Rebased due to recent conflict with `ValidatorTest.java`

manstis:
@tarilabs So when a DMN file containing empty Relation cells is loaded by Drools XML->Model unmarshaller empty cells become "null". This is then passed to the UI unmarshaller that will take "null" literally and create cells containing the text "null". When saved from the UI the Relation will have cells containing "null" and no longer empty. Is this a problem? If "null" FEEL is required by the runtime can the same not be infered only at runtime without affecting the Models?

tarilabs:
> If "null" FEEL is required by the runtime can the same not be infered only at runtime without affecting the Models?

Sorry @manstis but _this_ model (kie-dmn-model) has the scope for the engine itself, and an Expression must be present. How would I know otherwise if the LiteralExpression is the one nested inside a Relation? Doing this "reflectively" at runtime would be a performance hit for accommodating a simplification which is for the purpose of UI only. 

(the standard AFAIK requires an expression, and if "empty" then requires a LiteralExpression for `null`)

IFF, and I repeat IFF, we want to similarly adopt that an empty Relation's cell stands for `null` in the authoring, than this be managed on the converters and on the models for the UI only.

I think you raise a good point @manstis but I believe this is the wrong repo =P
I will raise a JIRA affecting the marshaller/converter for the wb.

@etirelli your feedback please?

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/RelationConverter.java
```diff
@@ -37,12 +38,29 @@ protected void assignChildElement(Object parent, String nodeName, Object child)
...
+    /**
+     * Normalize empty cell in DMN Relation as null expression.
+     */
+    private void normalizeEmptyCells(org.kie.dmn.model.v1_1.List row) {
+        // DROOLS-2439
+        for (Expression e : row.getExpression()) {
+            if (e instanceof LiteralExpression) {
+                LiteralExpression literalExpression = (LiteralExpression) e;
+                if (literalExpression.getText().isEmpty()) {
+                    literalExpression.setText("null");
```
etirelli: Matteo, I am not totally sure about this. Ideally we should not change the model, but just support the semantics that if there is an empty LiteralExpression, then it should be treated as null. As @manstis said, this will end up affecting the UI as well. 
Can we do this without changing the model?

tarilabs:
@etirelli @manstis 
Concerning this comment: https://github.com/kiegroup/drools/pull/1845#pullrequestreview-109491931

My understanding is that anyway we are going against the DMN spec v1.1
From the spec:
image
An empty string is not a valid FEEL expression. This is because if I feed an empty string to the parser for "10.3.1.2 Grammar rules" of the spec, it would return a parse error. Which would in turn result in a null during evaluation, but still would be an error (which we actually don't want).

### Current approach
The current approach was to implement an explicit normalization.

That is, an empty text in LiteralExpression during unmarshalling is normalized explicitly as a `null` LiteralExpression.

For what concering the UI, we could have managed in the UI model. We have separate models already, and this could be well fitted in the role of the converters.

I believe this is similar approaches other vendors are doing --e.g.: the normalization of the namespaces, and assimilated.

This would have also benefited from a Validation point of view: we have indeed a rule to check a LiterlExpression to be a valid one (an empty string is not a valid FEEL expression).

Once marshalled back into xml from the kie-dmn-model, this would have then resulted in a correct model.

### Alternative approach
I will work now for looking an alternative approach which doesn't explicitly normalize on the kie-dmn-model.

Personally, I preferred the explicit approach, but if the constraint is that "_Ideally we should not change the model_", then I comply.

manstis:
@tarilabs Great research. I also note the Grammar states an Expression can be:

>33. literal = simple literal | "null" ;
>34. simple literal = numeric literal | string literal | Boolean literal | date time literal ;
>35. string literal = '"' , { character – ('"' | vertical space) }, '"' ;

Using string literal "" would solve the issue entirely.

WDYT?

tarilabs:
> Using string literal "" would solve the issue entirely.
> WDYT?

@manstis sorry but I don't concur with your conclusion.
A FEEL empty String, is a String.
A FEEL `null` is a null.
Not the same type and the semantic we want there is that the empty cell in the DMN Relation is managed as a `null` value, not a FEEL empty String.

btw I believe I have already an approach for the compilation/execution, I just need to confirm and to update Validation rule to take into account the exception.

Will keep posted ;)

manstis:
@tarilabs Fair enough.. but you're setting the ```LiteralExpression``` text to ```"null"``` which is not ```null``` either. I was merely proposing that setting the ```LiteralExpression``` text to ```""``` would avoid the need to handle ```"null"``` in the UI model (un)marshalling.. I'll wait for your posting :-)

tarilabs:
> but you're setting the LiteralExpression text to "null" which is not null either. 
> I was merely proposing that setting the LiteralExpression text to "" would avoid the need [...]

I think @manstis we are not speaking of the same thing.

I set LiteralExpression text, with Java, as `"null"`; because text is a java.lang.String field of the pojo LiteralExpression.
Once at FEEL parsing, the Antlr parser will receive the character sequence of `null` (without the quotes).
This is a valid FEEL expression and conforming to the FEEL grammar.

Therefore I don't agree that I was _setting LiteralExpression text to "null"_, as you state. 
What I was setting is actually a Java String literal, which is passed to the Antlr parser as the character sequence of n-u-l-l which for the FEEL grammar would become the FEEL `null` value.

I'm not sure what you mean _by setting the LiteralExpression text to ""_ 

If I set this with Java `""` then is the problem, because an empty character sequence to the Antlr parser for the FEEL grammar would indeed throw the parsing error, as we expect.

If I set this with Java `"\"\""` then is a semantic problem, this is a valid FEEL expression which would result in a FEEL String, a FEEL empty String, but this is not what we want.

Am I missing something?

manstis:
@tarilabs I doubt you're missing anything.. more likely I am (or was).

I understand what you're saying now.. about the text being passed to ANTLR for parsing :-)

tarilabs:
btw I take back when I was saying:
> we have indeed a rule to check a LiterlExpression to be a valid one (an empty string is not a valid FEEL expression).

We are actually NOT doing this.

What we actually have, a rule to check that a DMN DecisionTable cannot contain an empty-text cell. 

Sorry.

tarilabs:
Proposed alternative approach as requested with commit 5a0065a .

jomarko:
@manstis @tarilabs thanks for this conversation, however it brings my concern, what values should we be able to put into UI cell and how they should be interpreted/stored in xml and vice versa.

tarilabs:
@jomarko while technically Relation could contain in each row's cell a general `Expression`, they are most commonly intended for being an analogous to a table in a relational database, hence _by default_ they should be/contain a `LiteralExpression`.
@etirelli can correct me ;)
/cc @manstis 
If we need to discuss this specific point further however I propose creating a separate and more appropriate thread, please?

manstis:
@tarilabs @jomarko Yes, @etirelli agreed the simplification that ```Relation``` will only contain ```LiteralExpression``` (although both the DMN model and UI support any ```Expression```). @jomarko One day our "plain text" boxes will be FEEL expression builders to ensure the form and syntax of (Literal) expressions. The storage of the values are handled by the DMN XML<->DMN model marshallers.

manstis:
@tarilabs Preferring your new solution :-)

etirelli:
Correct, relations contain literal expression. I can't see today relations having any other type of expression on their cells, unless something changes in the spec.

baldimir:
Rereviewed, +1 :)

etirelli (approved):

PR #1847, opened 2018-03-30
RHBA-626 Tree model does not work on models expecting numeric output

lanceleverich:
* Updated the process that generates the Java classes to now include a TreeNode so that the node's outcome is of the type needed
* Updated the AbstractModel so that retrieving the mining fields does not exclude any fields
* Added a method to the Treemodel to return the source for a TreeNode java class
* Added context to the name of the generated TreeNode objects, in the templates that use TreeNode
* Created an mvel template to be used in generating the TreeNode java class


mariofusco:
ok to test

PR #1856, opened 2018-04-04
DROOLS-2419 schema validation for definition xml element with prefix

tarilabs:


tarilabs:
Although this does not fix any bug (in fact DROOLS-2419 not a bug) and is just a test, it is helpful to demonstrate xsi:schemaLocation works also for prefixed DMN xml elements.
Additionally it could be helpful also as a non-regression for DROOLS-2448 eventually.

etirelli (approved):

PR #1857, opened 2018-04-05
DROOLS-2447 DMNValidator for schema to use an embed version of DMN XSD

tarilabs:
With this change a DMNValidator called for schema validation, would always use the embedded version of the DMN XSD, rather than relying on the `xsi:schemaLocation` inside each single dmn model XML file (which the user could point actually at any xsd).

/cc @etirelli @baldimir @kurobako @winklerm

### DIFF
--- a/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/DMNValidatorImpl.java
+++ b/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/DMNValidatorImpl.java
@@ -73,7 +73,7 @@ public class DMNValidatorImpl implements DMNValidator {
             schema = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI)
                                   .newSchema(new StreamSource(DMNValidatorImpl.class.getResourceAsStream("org/omg/spec/DMN/20151101/dmn.xsd")));
         } catch (SAXException e) {
-            e.printStackTrace();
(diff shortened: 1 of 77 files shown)


etirelli (approved):

baldimir (approved):

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/DMNValidatorImpl.java
```diff
@@ -70,7 +70,8 @@
     static Schema schema;
     static {
         try {
-            schema = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI).newSchema();
+            schema = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI)
+                                  .newSchema(new StreamSource(DMNValidatorImpl.class.getResourceAsStream("org/omg/spec/DMN/20151101/dmn.xsd")));
         } catch (SAXException e) {
             e.printStackTrace();
```
winklerm: I am wondering if we could avoid using e.printStackTrace() here:
- If I understand correctly, the schema should always be available, so we might be able to just throw a (runtime) exception here to avoid a NPE later. 
- If it is expected that the schema creation may fail (and we do not want to fail the static initializer by throwing an exception), the exception ``e`` could be just logged.

What do you think?

winklerm (approved):
Thanks!

etirelli (approved):

PR #1862, opened 2018-04-11
Remove org.eclipse.sisu.plexus since it seems not relevant in sourcecode

ryanzhang:
This PR is trying to remove the potential  dependencies not referenced on org eclipse sisu plexus.  
I can't find the reference in source code and the eclipse sisu plexus is not able to rebuild from source.

### DIFF
--- a/kie-ci/pom.xml
+++ b/kie-ci/pom.xml
@@ -139,16 +139,6 @@
       <artifactId>aether-transport-http</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.sisu</groupId>
-      <artifactId>org.eclipse.sisu.plexus</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>javax.el</groupId>
-          <artifactId>javax.el-api</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
       <groupId>org.apache.ant</groupId>
       <artifactId>ant</artifactId>
     </dependency>


ryanzhang:

ryanzhang:
@manstis Hi, Could l ask you to review the fail test please?
is it sisu still relevant or import in our user case?

manstis:
@ryanzhang IMO sisu/plexus has been incredibly important to our use of Maven API (and has caused its fair deal of headaches over the years). Since the test fails at a line loading resources from ```META-INF/plexus/``` I think the failure is even more of a worry as it looks like we deliberately added this test to cover some issue in the past.

@mariofusco was the author of this test; and probably has a greater understanding of the significance of its failure; so I'd defer further comment to him. @mariofusco opinions please.. sisu/plexus has been a pain in the past and I get scared thinking about changing it!

ryanzhang:
@manstis  Thanks for your checking. Honestly, I am scared thinking of change it as well.
The pain in productized it is that the upstream sisu plexus depends on some resource external and not even existed. So it's simply can't be rebuilt.
I don't see sisu is explicitly referenced by the code and don't know how important it to us. 
Wondering if there is any alternative applies here.

manstis:
@ryanzhang If memory serves me well; sisu is used by the Maven libraries to provide it's own Dependency Injection mechanism at runtime. Perhaps @mariofusco may have better recollection.

mariofusco:
ok to test

mariofusco:
That dependency is necessary.

PR #1863, opened 2018-04-11
(6.4.z) [RHBRMS-3089] clear defined types cache in ProjectClassLoader when a declared type is changed during an incremental compilation

elguardian:
(cherry picked from commit a44f3ec2fd34f2c7fa2b3b8787be2f5339b1a963)

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/IncrementalCompilationTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/IncrementalCompilationTest.java
@@ -4213,4 +4213,73 @@ public class IncrementalCompilationTest extends CommonTestMethodBase {
         ksession.insert("test");
         ksession.fireAllRules();
     }
+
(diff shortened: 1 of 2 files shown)


mariofusco:
ok to test

baldimir (approved):
Looks ok, backport looks the same as original commit.

mariofusco:
I didn't backport the fix because it didn't work on 6.5.x in a specific case (see IncrementalCompilationTest.testIncrementalCompilationWithRedeclares). The fix that works for 7.x doesn't work on 7. This is because in drools 7.x I modified the building process a lot first to implement the parallel engine, then for the parallel compilation and finally to accommodate the executable model. I tried to surgically take the changes related to this issue that I made on 7 and backport to 6.5.x but I just ended up with a bit mess and had to rollback everything. At the moment my advice is to avoid trying to backport this fix to 6.5.x.

mareknovotny:
@mariofusco ok, thanks for the explanation, I was confused by JIRA, where you acked. So we need to decline everything.

PR #1869, opened 2018-04-16
JBPM-7164: Add support to TypedWorkItem(Handler)s

evacchi:
Proposal for extending the WorkItem(Handler) models with support for typed POJOs for parameters.

in particular
- typed handlers are injected (and wrapped into an untyped handler) here https://github.com/kiegroup/drools/pull/1869/files#diff-2538833330c9feb51801037bd74c6d81R39
- conversion between Map and beans (domain objects) is here
https://github.com/kiegroup/drools/pull/1869/files#diff-7f6a52814c93ebe889c01041b2b77832R41

Depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/303


evacchi:
let's close this for now

PR #1870, opened 2018-04-16
DROOLS-1480 Support timezones in date and time variables

kurobako:
https://issues.jboss.org/browse/DROOLS-1480

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
@@ -575,7 +575,7 @@ public class DMNRuntimeTest {
         assertThat( ctx.get("cHour"), is( BigDecimal.valueOf( 0 ) ) );
         assertThat( ctx.get("cMinute"), is( BigDecimal.valueOf( 0 ) ) );
         assertThat( ctx.get("cSecond"), is( BigDecimal.valueOf( 1 ) ) );
-        assertThat( ctx.get("cTimezone"), is( Duration.parse( "PT-1H" ) ) );
+        assertThat( ctx.get("cTimezone"), is( "GMT-01:00" ) );
         assertThat( ctx.get("years"), is( BigDecimal.valueOf( 1 ) ) );
         assertThat( ctx.get("seconds"), is( BigDecimal.valueOf( 14 ) ) );
 
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
@@ -28,6 +28,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.TemporalAccessor;
 import java.time.temporal.TemporalQueries;
+import java.util.TimeZone;
 
 import org.kie.dmn.api.feel.runtime.events.FEELEvent.Severity;
 import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
@@ -63,7 +64,7 @@ public class DateAndTimeFunction
         if (!DateFunction.BEGIN_YEAR.matcher(val).find()) { // please notice the regex strictly requires the beginning, so we can use find.
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "from", "year not compliant with XML Schema Part 2 Datatypes"));
         }
-        
+
         try {
             if( val.contains( "T" ) ) {
                 return FEELFnResult.ofResult(FEEL_DATE_TIME.parseBest(val, ZonedDateTime::from, OffsetDateTime::from, LocalDateTime::from));
@@ -95,7 +96,7 @@ public class DateAndTimeFunction
         if (!(time instanceof LocalTime || (time.query(TemporalQueries.localTime()) != null && time.query(TemporalQueries.zone()) != null))) {
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "time", "must be an instance of LocalTime or (it must contain localTime AND zone)"));
         }
-        
+
         try {
             if( date instanceof LocalDate && time instanceof LocalTime ) {
                 return FEELFnResult.ofResult( LocalDateTime.of( (LocalDate) date, (LocalTime) time ) );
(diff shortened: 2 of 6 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
```diff
@@ -280,8 +286,10 @@ public static PropertyValueResult getDefinedValue(final Object current, final St
             }
         }
 
-        // before returning, coerce "result" into number.
-        result = coerceNumber(result);
+        if (result instanceof Number) {
```
tarilabs: not needed, this is the logic of coercion.

kurobako: Thanks for noticing, right, it remains from an earlier version of this PR where coercion would actually nullify the value, doesn't matter why now. I will update this PR.

kurobako:
@tarilabs I updated my PR according to @etirelli 's recommendations, could you please review?

kurobako:
14:57:30 ERROR: Error cloning remote repo 'origin'

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -575,7 +575,7 @@ public void testDateAndTime() {
         assertThat( ctx.get("cHour"), is( BigDecimal.valueOf( 0 ) ) );
         assertThat( ctx.get("cMinute"), is( BigDecimal.valueOf( 0 ) ) );
         assertThat( ctx.get("cSecond"), is( BigDecimal.valueOf( 1 ) ) );
-        assertThat( ctx.get("cTimezone"), is( Duration.parse( "PT-1H" ) ) );
+        assertThat( ctx.get("cTimezone"), is( "GMT-01:00" ) );
```
tarilabs, line 578: @kurobako the original time string which should impact to originate this result, is "`00:00:01-01:00`".
Hence I am not sure if `GMT-01:00` is the expected result, or _any_ valid identifier is allowed here.
/cc @etirelli

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateAndTimeFunction.java
```diff
@@ -28,8 +28,9 @@
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.TemporalAccessor;
 import java.time.temporal.TemporalQueries;
+import java.util.TimeZone;
 
-import org.kie.dmn.api.feel.runtime.events.FEELEvent.Severity;
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
```
tarilabs: minor, but fyi maybe the editor swapped the import hence you have a few modification in the file itself accordingly (this is not a request for modification, just fyi).

tarilabs (commented):
just uncertain about one specific result, besides of that I understand the modification is now been applied to the "core" function looks good to me thx!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
etirelli: This is on the edge of what the spec says, so there is no clear answer on what it should return. My personal understanding is that it would return null as the timezone is not defined on the input string, only the offset. We could make an argument that returning "GMT-01:00" is a better result, but I don't think this is what other vendors are doing (on interoperability). I don't mind leaving it as it is, if you guys think it is better, but my preference would be to return null I think.

kurobako: @etirelli @tarilabs the issue here is that we are using java time classes internally and thus to represent time with an offset we have to use ZonedDateTime which happens to have ZoneId.
Thus, after DateAndTimeFunction returned its result we can't distinguish whether this time value was created using an offset or timezone.
When neither timezone nor offset is provided, DateTime type we use internally has no ZoneId encoded and we correctly return empty result when user tries to call .timezone on such type.

Bottom line is without reworking how we internally represent date and time types I can't see a way to be super strict on this.

kurobako:
@tarilabs thanks for the notes. It was probably not very critical but I fixed some shady imports and tabs idea inserted. I think we can merge this now (see my last comment), wdyt?

tarilabs:
@kurobako I concur the only way to report null on a non-provided-timezoned FEEL Date[x] would be to change internal representation. Given however this is touching a very sensitive behavior, have you re-run the TCK with this change applied? What is the result, please?

tarilabs:
> shady imports and tabs idea inserted

is the code style correctly setup in your IDEA? My understanding is that the code style (which I undersood we are required to apply _stricly_) should not have any issue.

kurobako:
It was freshly installed idea at that moment with absolutely no settings applied, so that's why. Not to worry.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
etirelli: Ok, lets keep as it is for now.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
```diff
@@ -256,9 +260,15 @@ public static PropertyValueResult getDefinedValue(final Object current, final St
...
                 case "time offset":
-                case "timezone":
                     result = Duration.ofSeconds(((Temporal) current).get(ChronoField.OFFSET_SECONDS));
                     break;
+                case "timezone":
+                    try {
+                        result = TimeZone.getTimeZone(ZoneId.from((TemporalAccessor) current)).getDisplayName();
+                        break;
+                    } catch (DateTimeException e) {
+                        return PropertyValueResult.notDefined();
```
etirelli: what will happen with this return value? 
Usually it is better to avoid the DateTimeException by checking if the "current" variable supports ZoneId and if it is set. Usually creating exceptions is very expensive and hits performance. If not possible, then catching for the exception is the lesser evil.

kurobako: @etirelli you are right, exception here could be avoided with ease. fixed.

kurobako:
jenkins retest this

kurobako:
@etirelli @tarilabs 
I think I found a way to return empty result for timezone query for date and time variable created with an offset but no timezone.
First, we need to actually return OffsetDateTime instanced from date and time function if user provided us with an offset. I initially assumed there was a deeper meaning we were returning ZonedDateTime instead but apparently it was unintentional.
Second, we need to use stricter API when requesting time: 
- temporal.query(TemporalQueries.zone()) and APIs that use this internally automatically convert offset info encoded into DateTime object into zone id and thus are useless
- temporal.query(TemporalQueries.zoneId()) is stricter and gives us zone id if and only if DateTime was created using timezone

I implemented these changes and it seems to work. Locally all tests pass perfectly fine. I also checked TCK; it's good.

tarilabs:
Awesome @kurobako thanks a lot!!

etirelli:
@kurobako excellent!

etirelli:
@tarilabs can you please coordinate with @mariofusco to get this PR merged?

PR #1871, opened 2018-04-20
DROOLS-1663 fix parameter name dependency resolution for DT in Context

tarilabs:
node dependencies which are not coming from the same model, but from an
imported model, needs to be prefixed in the dependency-name with the
"alias." prefix with the DMN Import name used during import.

@etirelli : this should fix the problem reported by Simon.

/cc @baldimir @kurobako @winklerm

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
@@ -585,6 +585,7 @@ public class DMNEvaluatorCompiler {
                                               Msg.IMPORT_NOT_FOUND_FOR_NODE_MISSING_ALIAS,
                                               new QName(depEntry.getValue().getModelNamespace(), depEntry.getValue().getModelName()),
                                               node);
+                        return null;
                     }
                     parameterNames.add(importAlias.get() + "." + depEntry.getKey()); // this dependency is from an imported model, need to add parameter with "alias." DMN Import name prefix.
                 }
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
@@ -3,6 +3,7 @@ package org.kie.dmn.core.compiler;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -11,6 +12,7 @@ import javax.xml.namespace.QName;
 import org.kie.dmn.api.core.DMNMessage;
 import org.kie.dmn.api.core.DMNType;
 import org.kie.dmn.api.core.ast.BusinessKnowledgeModelNode;
+import org.kie.dmn.api.core.ast.DMNNode;
 import org.kie.dmn.api.core.ast.DecisionNode;
 import org.kie.dmn.core.api.DMNExpressionEvaluator;
 import org.kie.dmn.core.api.EvaluatorResult;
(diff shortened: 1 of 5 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
```diff
@@ -0,0 +1,54 @@
+package org.kie.dmn.core.imports;
```
winklerm, line 17: Please include the header with copyright, see for instance \[1].

\[1] https://github.com/tarilabs/drools/blob/1bef5a3077a812058af7f39bc6a5186cffb638c8/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/InputDataCompiler.java

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/util/Msg.java
```diff
@@ -6,6 +6,7 @@
     // consolidated
     public static final Message2 UNSUPPORTED_ELEMENT                                 = new Message2( DMNMessageType.UNSUPPORTED_ELEMENT, "Element %s with type='%s' is not supported." );
     public static final Message2 IMPORT_NOT_FOUND_FOR_NODE                           = new Message2( DMNMessageType.IMPORT_NOT_FOUND, "Required import not found: %s for node '%s' " );
+    public static final Message2 IMPORT_NOT_FOUND_FOR_NODE_MISSING_ALIAS             = new Message2(DMNMessageType.IMPORT_NOT_FOUND, "Required import not found: %s for node '%s'; missing DMN Import name alias.");
```
winklerm: Just a bit of nitpicking ;) please keep the indentation consistent.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -568,7 +570,25 @@ private DMNExpressionEvaluator compileDecisionTable(DMNCompilerContext ctx, DMNM
...
+                                              DMNMessage.Severity.ERROR,
+                                              dt.getParent(),
+                                              model,
+                                              null,
+                                              null,
+                                              Msg.IMPORT_NOT_FOUND_FOR_NODE_MISSING_ALIAS,
+                                              new QName(depEntry.getValue().getModelNamespace(), depEntry.getValue().getModelName()),
+                                              node);
+                    }
+                    parameterNames.add(importAlias.get() + "." + depEntry.getKey()); // this dependency is from an imported model, need to add parameter with "alias." DMN Import name prefix.
```
winklerm: When ``importAlias.isPresent()`` is ``false``, the ``importAlias.get()`` would throw a ``NoSuchElementException``, is it expected?

winklerm (changes requested):
Please see the comments/questions inline, otherwise looks good to me.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
tarilabs: You are right, but as well the end-user would have obtained an error message, as per the lines above this statement. If the import alias is missing, it would be impossible to establish a correct parameter names, hence the compilation should not terminate correctly.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/util/Msg.java
tarilabs: True, sorry this is induced by one of the problems with the _strict_ code style.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
tarilabs: done, thanks for spotting.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
winklerm: Thanks for explanation - in that case, I would suggest either not calling the parameterNames.add() statement at all, or throwing a more descriptive exception (which might be redundant in this case as you are already returning a well-descriptive error message, but might be needed if you want the compilation to fail - in that case you might not get all error messages, though).

My concern is that if compilation fails with generic ``NoSuchElementException``, this exception is probably the first thing that the client using this code sees and it is not very helpful telling them what went wrong. It will also break the for loop which would result in not reporting error messages for all missing imports, if I understand correctly.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
winklerm: Thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
tarilabs: fair enough, aligned with the style originally implemented in this compiler, to return a `null` node just after notifying an error message.

tarilabs: (also to be note, v1.1 of DMN does not oblige to define a DMN Import `name` but it is de-facto expected. This is enforced in v1.2. So the probability of this Optional to be empty is very low)

winklerm: Thanks, sounds good!

winklerm (approved):
Thanks!

baldimir (approved):

etirelli (approved):
Fix looks good, but I am wondering if there are other places with similar issue? We might need a focused testing effort around imports. Anyway, this is good to go.

tarilabs:
>  wondering if there are other places with similar issue?

I had the same doubt, but I checked for code call sites using the getDependencies() method, and this was the only one using the _keys_ (the name String). Hence while it doesn't rule out any other potential bug, this fix is quite comprehensive on its own imho based on this.

PR #1877, opened 2018-04-27
Remove $context field from enums

ghost:
When compiling a kjar which includes a kbase from another kjar containing an enum, the compilation fails with message like:

    New declaration of com.fsecure.droolsenumrepro.rules.common.FooEnum does not include field $context

See https://github.com/fs-aikito/drools-enum-repro for demonstration of this issue (run `mvn clean install` at root level)


mariofusco:

PR #1882, opened 2018-04-30
DROOLS-2317 FEEL Syntax error on function(bkm) containing `for` keyword

kurobako:
https://issues.jboss.org/browse/DROOLS-2317

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
@@ -1913,6 +1913,15 @@ public class DMNRuntimeTest {
     }
 
     @Test
+    public void testFor() {
+        // DROOLS-2317
+        DMNRuntime runtime = DMNRuntimeUtil.createRuntime("Dynamic composition.dmn", this.getClass());
+        DMNModel dmnModel = runtime.getModel("http://www.trisotech.com/definitions/_c2d86765-c3c7-4e1d-b1fa-b830fa5bc529", "Dynamic composition");
+        assertThat(dmnModel, notNullValue());
+        assertThat(DMNRuntimeUtil.formatMessages(dmnModel.getMessages()), dmnModel.hasErrors(), is(false));
+    }
+
+    @Test
     public void testProductFunction() {
         DMNRuntime runtime = DMNRuntimeUtil.createRuntime( "product.dmn", this.getClass() );
         DMNModel model = runtime.getModel( "http://www.trisotech.com/dmn/definitions/_40fdbc2c-a631-4ba4-8435-17571b5d1942", "Drawing 1" );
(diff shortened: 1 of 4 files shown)


etirelli:
@kurobako this change looks good to me. We are having jenkins problems, did you try your change locally? if all tests pass, we can go ahead with the merge.

etirelli (approved):

kurobako:
@etirelli in fact I was unable to verify these changes locally before sending a PR because latest master at the moment was failing tests locally. Do we have some tests running as part of the build process that depend on external infrastructure by chance? 
Well, anyway, today master tests worked fine for me but this branch did not. Surprisingly enough, FlightRebookingTest fails branch because somehow relaxing conditions for verifying name parts prevents flight numbers in the example to be evaluated properly. I guess it might have something to do with us selectively swallowing validation exceptions. I am investigating that and will update this PR soon.

kurobako:
@etirelli @tarilabs @baldimir this will be complicated.
I investigated the reason why FlightRebookingTest fails: dmn files contain 'and' as part of literal expression, like this:
Flights[ From = Original Flight.From and To = Original Flight.To and Departure > Original Flight.Departure and Status = "scheduled" and has capacity( item, Rebooked ) ][1]

checkVariableNamePart, the function which behaviour I changed is used by ParserHelper.followUp method returning boolean; FlightRebookingTest works correctly only if "and" as name part is considered invalid and thus followUp returns false. We need followUp to return false in this situation otherwise FlightNumbers won't be evaluated.

Natually, in other cases like the one I was working on followUp returning false causes no viable alt error in antlr because we need antlr to follow up after encountering "for" and possibly other keywords as part of the variable name, like in:

Federal Tax for Item
Otherwise, antrl will think we have a malformed for expression here.
I might come up with the way to sew this together but I suggest we talk about a proper way to resolve this. Should followUp really be wired to checkVariableNamePart in the first place, maybe we can start thinking from there?

tarilabs:
> FlightRebookingTest works correctly only if "and" as name part is considered invalid and thus followUp returns false

Unfortunately we cannot do that, as "and" is totally valid as part of a name.

At the moment the only general direction I could think of, would be to try having the follow-up to check if the current followup is a valid name AND the _current **total**_ followup is the beginning of any symbol name in the current scope. If you want, you can give it a try in this direction, but don't overspend efforts please in this direction as I believe is important for @etirelli to confirm is a valid one.

kurobako:
@etirelli updated this PR. Very questionable workaround, but that's the best I was able to come up with without refactoring whole parser helper followup machinery.

etirelli (changes requested):
@kurobako sorry, but I am not sure this change is correct now. This latest solution seems to be fixing the symptom, not the root cause. You shouldn't need a second list of keywords to bypass the check.

kurobako:
jenkins retest this

kurobako:
@etirelli fair enough. I updated this PR.

kurobako:
@etirelli this PR was updated, please take another look

etirelli (approved):

etirelli:
@kurobako looks better now! thanks!

@tarilabs can you please double check and merge if everything is ok?

PR #1893, opened 2018-05-04
DROOLS-2521 Implement FEEL functions in preparation for DMN 1.2

kurobako:
https://issues.jboss.org/browse/DROOLS-2521

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/EvenFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/EvenFunction.java
@@ -1,0 +1,30 @@
+package org.kie.dmn.feel.runtime.functions.extended;
+
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
+import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
+import org.kie.dmn.feel.runtime.functions.BaseFEELFunction;
+import org.kie.dmn.feel.runtime.functions.FEELFnResult;
+import org.kie.dmn.feel.runtime.functions.ParameterName;
+
+import java.math.BigDecimal;
+
+public class EvenFunction 
+        extends BaseFEELFunction {
+    public static final EvenFunction INSTANCE = new EvenFunction();
+
+    private static final BigDecimal TWO = BigDecimal.valueOf(2);
+
+    EvenFunction() {
+        super("even");
+    }
+
+    public FEELFnResult<Boolean> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        if ( number.remainder( BigDecimal.ONE ).signum() != 0 ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot have non-zero fractional part" ) );
+        }
+        return FEELFnResult.ofResult( BigDecimal.ZERO.compareTo( number.remainder(TWO) ) == 0);
+    }
+}
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ExpFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ExpFunction.java
@@ -1,0 +1,26 @@
+package org.kie.dmn.feel.runtime.functions.extended;
+
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
+import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
+import org.kie.dmn.feel.runtime.functions.BaseFEELFunction;
+import org.kie.dmn.feel.runtime.functions.FEELFnResult;
+import org.kie.dmn.feel.runtime.functions.ParameterName;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+
+public class ExpFunction
+        extends BaseFEELFunction {
+    public static final ExpFunction INSTANCE = new ExpFunction();
+
+    ExpFunction() {
+        super("exp");
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        return FEELFnResult.ofResult( new BigDecimal( Math.exp( number.doubleValue() ), MathContext.DECIMAL128 ) );
+    }
+}
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/KieExtendedDMNFunctions.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/KieExtendedDMNFunctions.java
@@ -34,11 +34,20 @@ public class KieExtendedDMNFunctions {
                                                                          new NowFunction(),
                                                                          new TodayFunction(),
                                                                          new AbsFunction(),
-                                                                         new ModuloFunction(),
+                                                                         ModuloFunction.INSTANCE,
                                                                          new ProductFunction(),
                                                                          new CodeFunction(),
                                                                          new InvokeFunction(),
-                                                                         new SplitFunction(),
+                                                                         SplitFunction.INSTANCE,
+                                                                         StddevFunction.INSTANCE,
+                                                                         ModeFunction.INSTANCE,
+                                                                         AbsFunction.INSTANCE,
+                                                                         SqrtFunction.INSTANCE,
+                                                                         LogFunction.INSTANCE,
+                                                                         ExpFunction.INSTANCE,
+                                                                         EvenFunction.INSTANCE,
+                                                                         OddFunction.INSTANCE,
+                                                                         MedianFunction.INSTANCE
     };
 
     public static FEELFunction[] getFunctions() {
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/LogFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/LogFunction.java
@@ -1,0 +1,26 @@
+package org.kie.dmn.feel.runtime.functions.extended;
+
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
+import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
+import org.kie.dmn.feel.runtime.functions.BaseFEELFunction;
+import org.kie.dmn.feel.runtime.functions.FEELFnResult;
+import org.kie.dmn.feel.runtime.functions.ParameterName;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+
+public class LogFunction
+        extends BaseFEELFunction {
+    public static final LogFunction INSTANCE = new LogFunction();
+
+    LogFunction() {
+        super("log");
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        return FEELFnResult.ofResult( new BigDecimal( Math.log( number.doubleValue() ), MathContext.DECIMAL128 ) );
+    }
+}
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,18 +14,23 @@
  * limitations under the License.
  */
 
-package org.kie.dmn.signavio.feel.runtime.functions;
+package org.kie.dmn.feel.runtime.functions.extended;
 
+import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.kie.dmn.api.feel.runtime.events.FEELEvent;
 import org.kie.dmn.api.feel.runtime.events.FEELEvent.Severity;
 import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
 import org.kie.dmn.feel.runtime.functions.BaseFEELFunction;
 import org.kie.dmn.feel.runtime.functions.FEELFnResult;
 import org.kie.dmn.feel.runtime.functions.ParameterName;
+import org.kie.dmn.feel.util.EvalHelper;
 
 public class ModeFunction
         extends BaseFEELFunction {
@@ -40,22 +45,27 @@ public class ModeFunction
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "list", "cannot be null"));
         }
         if (list.isEmpty()) {
-            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "list", "cannot be empty"));
+            return FEELFnResult.ofResult( Collections.emptyList() );
         }
 
-        Map<?, Long> collect = list.stream().collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+        Map<BigDecimal, Long> collect = list.stream().map(EvalHelper::getBigDecimalOrNull).collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
 
         long maxFreq = collect.values().stream().mapToLong(Long::longValue).max().orElse(-1);
 
-        List<?> mostFrequents = collect.entrySet().stream()
-                .filter(kv -> kv.getValue() == maxFreq)
-                .map(Map.Entry::getKey)
-                .collect(Collectors.toList());
+        List<BigDecimal> mostFrequents = collect.entrySet().stream()
+                                       .filter(kv -> kv.getValue() == maxFreq)
+                                       .map(Map.Entry::getKey)
+                                       .collect(Collectors.toList());
 
-        if (mostFrequents.size() == 1) {
-            return FEELFnResult.ofResult(mostFrequents.get(0));
-        } else {
-            return FEELFnResult.ofResult(mostFrequents.stream().sorted().collect(Collectors.toList()));
+        return FEELFnResult.ofResult(mostFrequents.stream().sorted().collect(Collectors.toList()));
+    }
+
+    public FEELFnResult<Object> invoke(@ParameterName("n") Object[] list) {
+        if ( list == null ) {
+            // Arrays.asList does not accept null as parameter
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "n", "the single value list cannot be null" ) );
         }
+
+        return invoke( Arrays.asList( list ) );
     }
 }
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/OddFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/OddFunction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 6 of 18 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
```diff
@@ -0,0 +1,66 @@
...
+// based on the examples of calculations, stddev is supposed to return sample standard deviation, not population standard deviation
+public class StddevFunction
+        extends BaseFEELFunction {
+    public static final StddevFunction INSTANCE = new StddevFunction();
+
+    StddevFunction() {
+        super("stddev");
+    }
+
+    // written in verbose manner in case BigDecimal math will be needed internally to improve precision
```
baldimir: I think we should go for the BigDecimals. The list is homogenous filled with BigDecimals, so we have them from the start. Am I correct @etirelli?

kurobako: @baldimir I think this will be performance hit. If needed we can allow some flags for precision maybe, same for sqrt. I can implement BigDecimal square root no problem, but do we actually have users that need it?

kurobako: @tarilabs maybe you could share your thoughts about this as well?

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/feel/runtime/FEELExtendedFunctionsTest.java
```diff
@@ -204,7 +204,7 @@
                 {"median([2, 5, 10, 12, 34, 35])", new BigDecimal("11"), null},
                 {"median([2, 5, 11, 12, 34, 35])", new BigDecimal("11.5"), null},
                 {"median([5, 10, 12, 34, 35])", new BigDecimal("12"), null},
-                {"mode([1, 2, 4, 4, 5, 6])", new BigDecimal("4"), null},
+                {"mode([1, 2, 4, 4, 5, 6])", Arrays.asList(new BigDecimal("4")), null}, // DMN 1.2 examples expect single element lists in such cases
```
tarilabs: No, Sorry I disagree. The example in DMN v1.2 draft shows `[ 6 ]` because in DMN a singleton element list is equivalent to the singleton element itself.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
tarilabs: It is actually the DMN spec that **_requires_** us to use BigDecimal. We cannot have internal calculation based on double as this could potentially loose precision. We have no choice but to use BigDecimal everywhere (like the rest of the code).

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/SplitFunction.java
```diff
@@ -48,7 +49,7 @@ public SplitFunction() {
         try {
             int f = processFlags( flags );
             Pattern p = Pattern.compile( delimiter, f );
-            String[] split = p.split( input );
+            String[] split = p.split( input , -1);
```
tarilabs, line 52: I am not sure if this is really required.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
```diff
@@ -39,22 +45,27 @@ public ModeFunction() {
...
+        List<BigDecimal> mostFrequents = collect.entrySet().stream()
                                        .filter(kv -> kv.getValue() == maxFreq)
                                        .map(Map.Entry::getKey)
                                        .collect(Collectors.toList());
 
-        if (mostFrequents.size() == 1) {
-            return FEELFnResult.ofResult(mostFrequents.get(0));
-        } else {
-            return FEELFnResult.ofResult(mostFrequents.stream().sorted().collect(Collectors.toList()));
+        return FEELFnResult.ofResult(mostFrequents.stream().sorted().collect(Collectors.toList()));
```
tarilabs: I am not sure about this change in behavior, for reasons explained in the test, ref below.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
baldimir: @kurobako we discussed this on some previous Monday meeting. I think we agreed also with Edson we can stick with doubles for now. But not 100% sure. Anyway, I'm ok with this as it is.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
```diff
@@ -39,22 +45,27 @@ public ModeFunction() {
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "list", "cannot be null"));
         }
         if (list.isEmpty()) {
-            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "list", "cannot be empty"));
+            return FEELFnResult.ofResult( Collections.emptyList() );
```
tarilabs: This is a change in behaviour if compared to Signavio profile which should be checked, before we can have it "shared" between 1.2 and Signavio extension?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ExpFunction.java
```diff
@@ -0,0 +1,25 @@
...
+
+    ExpFunction() {
+        super("exp");
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        return FEELFnResult.ofResult( new BigDecimal( Math.exp( number.doubleValue() ) ) );
```
tarilabs: We cannot convert to Double for the same reasons motivated below.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/EvenFunction.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public FEELFnResult<Boolean> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        if ( number.remainder( BigDecimal.ONE ).signum() != 0 ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot have non-zero fractional part" ) );
+        }
+        return FEELFnResult.ofResult( BigInteger.ZERO.equals( number.toBigInteger().remainder(TWO) ) );
```
tarilabs: I am just not 100% sure why the conversion to BigInteger is needed here.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/LogFunction.java
```diff
@@ -0,0 +1,25 @@
...
+
+    LogFunction() {
+        super("log");
+    }
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        return FEELFnResult.ofResult( new BigDecimal( Math.log( number.doubleValue() ) ) );
```
tarilabs: cannot convert to double value, same reason

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/OddFunction.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public FEELFnResult<Boolean> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        if ( number.remainder( BigDecimal.ONE ).signum() != 0 ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot have non-zero fractional part" ) );
+        }
+        return FEELFnResult.ofResult( !BigInteger.ZERO.equals( number.toBigInteger().remainder(TWO) ) );
```
tarilabs: I am just not 100% sure why the conversion to BigInteger is needed here.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/SqrtFunction.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public FEELFnResult<BigDecimal> invoke(@ParameterName( "number" ) BigDecimal number) {
+        if ( number == null ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "cannot be null" ) );
+        }
+        if ( number.signum() < 0 ) {
+            return FEELFnResult.ofError( new InvalidParametersEvent( FEELEvent.Severity.ERROR, "number", "is negative" ) );
+        }
+        // high precision sqrt for big decimals is available since java 9 if needed
+        return FEELFnResult.ofResult( new BigDecimal( Math.sqrt( number.doubleValue() ) ) );
```
tarilabs: cannot convert to double value, same reason

tarilabs (commented):
Just a few comments, I'm available to discuss on IRC/BlueJeans if preferred.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
tarilabs: @baldimir "previous" meeting like 2018-05-07? Then I was not aware of this decision sorry @kurobako you will ignore my comments re Double values in my current review.

baldimir: @tarilabs we should leave this to @etirelli. I am also not 100% sure if I remember this correctly.

tarilabs: IIRC between _exp()_ and _sqrt()_, one can be done already directly on BigDecimal, while another need manual implementation, I think I seen some libraries which implement the one which is missing from JDK.

kurobako: @tarilabs BigDecimal square root is one loop, we don't need libraries to do this if the need arises

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/feel/runtime/FEELExtendedFunctionsTest.java
kurobako: I got confused by the examples. I guess. This is good, I will modify my PR.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/SplitFunction.java
kurobako: It is. Examples expect "a,b,," to be split as "a", "b", "", ""

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
kurobako: @tarilabs then the question is what are our backward compatibility rules towards Signavio? this will be in DMN 1.2, if Signavio wants to stick to 1.1 we do need to split for sure.

kurobako: I assumed Signavio is to follow the standard when it comes mode function and such.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/SplitFunction.java
tarilabs: I see now, thanks!

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
tarilabs: For Signavio unfortunately the documentation is limited, so normally I used to manually test the corner cases.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/OddFunction.java
kurobako: You mean o just divide big decimals after checking for presence of fractional part?
I guess we can do that, remainder should be exact in this case iirc. Let me check the implementation. Thanks.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
tarilabs: That's what I'm saying, and then we need to find/import the code for the other function.

kurobako:
@tarilabs @baldimir regarding big decimal vs double math, I get a bit of a mixed message here.
 I don't see much in the spec besides: " Implementations are allowed to limit precision to 34 decimal digits and to round toward the nearest neighbor with ties favoring the even neighbor." and if I understand this correctly BigDecimal.doubleValue() should do just fine. Am I wrong?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
kurobako: Let's waif for Edson on this one maybe? I wouldn't split our implementations even more unless we are really forced to.

tarilabs:
From the spec:

> [FEEL Numbers] are equivalent to Java BigDecimal with MathContext DECIMAL128

This is how the code base also is currently structured, in other words I'm not against changes but I'm for promoting consistency.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/EvenFunction.java
etirelli: +1 looking at the API, BigDecimal has a remainder() method.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ExpFunction.java
etirelli: I told @kurobako to use a simple implementation on this first iteration converting to double if no alternative for BigDecimal exists. In particular, Euler's number is irrational, so a precise representation of it at the Math_Context_128 will be expensive to compute. Caching would be an alternative, but at this point, using the double representation given by Math.E is probably enough.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/LogFunction.java
etirelli: Same logic as previous comment, lets start with a double implementation and improve later. Having said that, the BigDecimal should still be created using our helper method to ensure MATH_CONTEXT_128.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ExpFunction.java
etirelli: BigDecimal should be instantiated using the helper method to ensure MATH_CONTEXT_128.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
etirelli: We should stick with the DMN 1.2 definition where a conflict exists.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
etirelli: We should stick with the DMN 1.2 definition.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/SqrtFunction.java
etirelli: Same as before, lets use the simple implementation first.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StddevFunction.java
etirelli: wherever a BigDecimal API exists, we should stick to it as per the DMN spec.

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/feel/runtime/FEELExtendedFunctionsTest.java
etirelli: just to be clear, I believe the spec requires returning a list for the mode() function.

tarilabs: Maybe I'm getting a bit confused myself, but to me _the mode_ is normally a single value, and only in particular case of discrete distribution would be a set of values.
image
Here the spec is giving an example with `[ 6 ]` but my interpretation was again the "user friendliness" that in FEEL a single element list is equivalent to the singleton element itself.
I'm fine with any direction, but just a suggesion to ensure consistency later across vendors and TCK imho the spec could further specify it always need to be a list in the case of single value. Too late for DMN v1.2 but potentially next minor iteration and agreement on the TCK? What do you think @etirelli ?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
tarilabs: Hence we need to verify this edge case on the Signavio manually, and act accordingly.

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/feel/runtime/FEELExtendedFunctionsTest.java
etirelli: @tarilabs this was discussed in the RTF and the equivalence of singleton lists was changed. In DMN 1.2, the coercion only happens from a singleton list '[a]' to a single element 'a', not vice-versa, and only happens when necessary, not automatically. So the mode() function will always return a list, but if that list is used in an expression or assigned to a non-list variable, then and only then, the list is unwrapped.

tarilabs: Okay thanks for this info @etirelli.

tarilabs: @kurobako this line change is OK. Please leave it as-is (eg: return a list of 1 element)

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/ModeFunction.java
tarilabs: This change is OK for DMN v1.2 definition.
I am not 100% sure how Signavio behaves on this case, tho.

kurobako:
@etirelli @tarilabs 
I updated this PR again, adding more precision to tests, precise sqrt and optimized stddev calculation. It shouldn't be very expensive as precision is limited by MathContext.DECIMAL_128. If needed, we can have a flag to short-circuit the calculation if we only care about first 16 or 32 digits. Otherwise, we are good to go I guess.

etirelli (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/KieExtendedDMNFunctions.java
```diff
@@ -34,11 +34,19 @@
...
-                                                                         new SplitFunction(),
+                                                                         SplitFunction.INSTANCE,
+                                                                         StddevFunction.INSTANCE,
+                                                                         ModeFunction.INSTANCE,
+                                                                         AbsFunction.INSTANCE,
+                                                                         SqrtFunction.INSTANCE,
+                                                                         LogFunction.INSTANCE,
+                                                                         ExpFunction.INSTANCE,
+                                                                         EvenFunction.INSTANCE,
+                                                                         OddFunction.INSTANCE
```
SimonRinguette: This is missing MedianFunction.INSTANCE in the list

SimonRinguette (changes requested):
The KieExtendedDMNFunctions is missing the median function, everything else look good

kurobako:
@SimonRinguette @etirelli thanks! updated my PR

tarilabs:
Thanks for the update @kurobako as this avoid the previous proposal where Signavio functions was going to be moved into Kie-extended-profile when those might be specific to the vendor implementation, especially for DMN v1.1. 

We have a doubt however now with the PR updated, some of the Kie-extended-profile function have been duplicated into the `kie-dmn-signavio` module. For instance the `NowFunction`. Is there any specific reason for this duplication? That was unnecessary before. Can you clarify that please?

We are going to move ahead with merging the PR anyway, as this provides preview for v1.2 _and_ avoid any move from the Signavio into the Kie-extended-profile. We can always clean-up the duplication in the opposite direction eventually later, once clarified.

kurobako:
@tarilabs I decided to duplicate it because if we are going to maintain backwards compatibility for signavio it would make sense to have functions in different directories to make a clear distinction. The need to implement the same change in both extended and signavio functions is improbable because of the same reason while having two separate directories prevents possible unwanted change in signavio behaviour in commits to come.

PR #1896, opened 2018-05-10
DROOLS-2537 Add reproducer for kcontext issue

baldimir:


mariofusco:
Fixed by https://github.com/kiegroup/drools/commit/419fd35752694a6dd2a79195252b7d687647622b

PR #1899, opened 2018-05-15
DROOLS-2530 Wrong evaluation of parameter named "product"

kurobako:
https://issues.jboss.org/browse/DROOLS-2530

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/EvaluationContextImpl.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/EvaluationContextImpl.java
@@ -20,6 +20,7 @@ import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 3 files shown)


kurobako:
@tarilabs I think we might want to additionally log warnings in case some name from the model shadows built-in function like 'product' in this case. What do you think?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/impl/EvaluationContextImpl.java
```diff
@@ -140,8 +141,9 @@ public boolean isDefined(String[] name) {
     @Override
     public Map<String, Object> getAllValues() {
         Map<String, Object> values = new HashMap<>(  );
-        for (ExecutionFrame frame : stack) {
-            values.putAll( frame.getAllValues() );
+        Iterator<ExecutionFrame> it = stack.descendingIterator();
```
tarilabs, line 144: implementing reversing order as expected. thanks.

tarilabs (approved):
Unfortunately I know @etirelli view on this kind of features is that such additional check would hit performance quite heavily (thinking about how many times this method is called).
So, I would refrain from introducing this check for now.

However, you are suggesting a feature which might be possible to some extent in the validation. I suggest you raise a new JIRA as "enhancement" (ie: not bug) and we discuss this potential check later in more details.

What do you think?

Besides this correction looks consistent, thanks!

tarilabs:
@baldimir didn't realize yet you were not tagged for review, for me is fine, unless there are further comments I'll ask Mario to merge.

baldimir (approved):
Looks ok, thanks.

PR #1901, opened 2018-05-15
DROOLS-2390 Run tests from drools-compiler "integrationtests.operators" package with executable model

baldimir:
- Tests migrated to test-compiler-integration module
- I moved many rules directly to the tests from DRL files, so one can see the test rules right in the test method. 
- To run a test class with the executable model, switch test class parameter generation from

`TestParametersUtil.getKieBaseCloudConfigurations(false);`

to 

`TestParametersUtil.getKieBaseCloudConfigurations(true);`

@mariofusco @lucamolteni could you please review? @mariofusco when merging, please don't squash commits, so the git move commit is in the history (so we don't lose git history on the files).


baldimir:

baldimir:
Downstream fail not related to the PR.

PR #1909, opened 2018-05-18
DROOLS-2543 "?" doesn't work with multi-conditions in decision table

tarilabs:


tarilabs:
This is better to double-check and review with Edson.

tarilabs:
Discussed with @etirelli this is good enough fix for the specific problem identified by this JIRA.
/cc @kurobako @baldimir @winklerm

baldimir (approved):

PR #1912, opened 2018-05-21
DROOLS-2575 Run DRL integration tests with executable model

baldimir:
@mariofusco @lucamolteni could you please review? @mariofusco when merging, please don't squash the commits, so the forst git mv commit is there for history preservation.


baldimir:

baldimir:
Looking at the failures.

PR #1920, opened 2018-05-23
fix DecisionTableProviderImpl ignoring CSV type for drt templates

karim789:
So the issue is that the ruleTemplate declared in kmodule.xml is aware that it's csv, but the information is not used by the ExternalSpreadsheetCompiler and default to XLS and it fails.

### DIFF
--- a/drools-decisiontables/src/main/java/org/drools/decisiontable/DecisionTableProviderImpl.java
+++ b/drools-decisiontables/src/main/java/org/drools/decisiontable/DecisionTableProviderImpl.java
@@ -53,7 +53,20 @@ public class DecisionTableProviderImpl
         ExternalSpreadsheetCompiler converter = new ExternalSpreadsheetCompiler();
         for ( RuleTemplateConfiguration template : configuration.getRuleTemplateConfigurations() ) {
             try {
-                drls.add(converter.compile(resource.getInputStream(), template.getTemplate().getInputStream(), template.getRow(), template.getCol()));
+                InputType inputType;
+                switch ( configuration.getInputType() ) {
+                case XLS :
+                case XLSX :
+                	inputType= InputType.XLS;
+                	break;
+                case CSV : 
+                	inputType= InputType.CSV;
+                	break;
+                default:
+                	throw new IllegalArgumentException("unsuported input type : "+configuration.getInputType());
+                }
+                
+				drls.add(converter.compile(resource.getInputStream(), template.getTemplate().getInputStream(), inputType,template.getRow(), template.getCol()));
             } catch (IOException e) {
                 logger.error( "Cannot open " + template.getTemplate(), e );
             }


karim789:
I don't know what is the release process, but a patch on 7.7 would be nice, because that's a blocking issue for us.
We don't want to version xls file in git because it's binary, and prefer csv because we can see the differences because it's plain text.

baldimir:
Hi, thanks for contribution! Could you please provide a JUnit test case in this PR reproducing the issue? If this fix gets approved, it would help to prevent similar problem in the future. Please, also create a tracking ticket in our JIRA [1] for this problem in DROOLS project [2]. Thanks!

[1] https://issues.jboss.org
[2] https://issues.jboss.org/browse/DROOLS

mbiarnes:
ok to test

ghost:

ghost:

ghost:

ghost:

baldimir:
Hi @karim789, could you please provide a test case for this that shows the problem?

karim789:
@baldimir Hello, sorry, I will not have time to do that.
Also that issue arrose when trying to use the kmodule, so the test should probably be done there.
I gave up using kmodule and did programatically, this gave me more flexibility.

jomarko:
@mariofusco Hi, @baldimir asked me to investigate this PR. 

I think we should close this PR as the description as insufficient and without more specific description of the use case we can not confirm the fix.

The only entry point where the touched code is invoked is [here](https://github.com/kiegroup/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java#L379). It means that use would have to use `KnowledgeBuilder` as:

```java
dtableConf.addRuleTemplateConfiguration(templateResource, rowIndex, colIndex);
kBuilder.add(dtableResource, ResourceType.DTABLE, dtableConf)
```

However from the PR description is not clear if @karim789 intended to fix such scenario.

We have already tests of `ExternalSpreadsheetCompiler` where we prove we can have `rule template` in `drl file` and `template data` in `csv file`. See [testLoadCsv()](https://github.com/kiegroup/drools/blob/master/drools-decisiontables/src/test/java/org/drools/decisiontable/ExternalSpreadsheetCompilerTest.java).

karim789:
@jomarko
The description is the first post is sufficient if you know what your library does.

So we fix your buggy library, then you reject the fixes because you don't even understand your own code base.
And you ask to put in whole knew integration tests on use cases that aren't covered at all, hence the bug that I faced.

You really know how to shoot yourself in the foot. xD

mariofusco:
@karim789 maybe we don't understand our own code base, but for sure you don't understand how to collaborate on an open source project.

You cannot throw some random "fix" and expect that we will merge it without any doubts and questions. The fact that we asked more than once for a reproducer it means that your description is not crystal clear as you think and we haven't been able to reproduce the problem ourselves. And I'm not willing to merge a "fix" for a bug that we cannot see.

To cut it short your description is not enough. I firmly believe in the motto: **"talk is cheap, show me the code". Even and especially for test cases.**

jomarko:
@karim789 `kiegroup` organization has simple 2 rules for contribution.

- file ticket in jira so we can track change, see https://github.com/kiegroup/drools/pull/1920#issuecomment-391973800
- provide unit test for your changes

You haven't accomplish any of this prerequisites so we can not merge your changes. 

I am totally sure you understand touched code better than me. The same for many other users of our code. My focus is not the engine part, I just helped my colleague who had other more important tasks at the moment.  

We can not merge your changes without unit test and risk other users will be affected in negative way. I am happy to review your PR once you file jira with description of the problem and steps to reproduce.

karim789:
@jomarko
The need was at work, in a big corporation where GitHub write acces is blocked, and of course where you can't bring code out of the corporation without beeing fired.
I decided to not use kmodule.xml to configure my rules because of this bug and also because I prefered configure in Spring.

But since I found the bug, and found the fix was easy, I still made a pull request with the fix, but I did that from my home not from the work place.

That's why I won't submit more than that, because I don't want to work half a day at home for free when my employer should pay me for it.
He should even donate to you and others for maintaining and developing drools, but that won't happen. They are too greedy for that.

I think your rules are good, but I can't follow them.
If my employer give me write access to GitHub I would do it on my work time, but I guess if they blocked write access, it's because they don't want us to do it.

So if someone needs this someday, at least there is a fix, and he can complete the PR with tests.

baldimir:
Hi @karim789, there was a reproducer available for your problem so we investigated this and found out that your fix is correct. The reproducer and the fix will be merged as part of https://github.com/kiegroup/drools/pull/2354, if all pull request checks are green. I expect the fix will be available in 7.23.0.Final release. Thanks for contribution!

PR #1921, opened 2018-05-23
DROOLS-2513 Add support for embedded test execution

kurobako:
https://issues.jboss.org/browse/DROOLS-2513

### DIFF
--- a/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/extensions/ComponentConverter.java
+++ b/kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/extensions/ComponentConverter.java
@@ -1,0 +1,52 @@
+package org.kie.dmn.backend.marshalling.v1_1.xstream.extensions;
+
+import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.io.HierarchicalStreamReader;
+import org.kie.dmn.backend.marshalling.CustomStaxReader;
+import org.kie.dmn.backend.marshalling.v1_1.xstream.DMNModelInstrumentedBaseConverter;
+import org.kie.dmn.model.v1_1.DMNModelInstrumentedBase;
+import org.kie.dmn.model.v1_1.extensions.Component;
+import org.kie.dmn.model.v1_1.extensions.Value;
+
+import java.util.Map;
+
+public class ComponentConverter extends DMNModelInstrumentedBaseConverter {
+    public ComponentConverter(XStream xstream) {
+        super(xstream);
+    }
+
+    @Override
+    protected void assignChildElement(Object parent, String nodeName, Object child) {
+        switch ( nodeName ) {
+            case "value": {
+                ( (Component) parent ).setValue( (Value) child );
+                break;
+            }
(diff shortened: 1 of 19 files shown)


kurobako:
@baldimir @tarilabs @etirelli FYI

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -484,6 +491,47 @@ private void reportFailure(DMNDecisionResultImpl dr, DMNMessage message, DMNDeci
...
+            for ( ResultNode resultNode : testCase.getResultNode() ) {
+                DecisionNode decisionNode = model.getDecisionByName( resultNode.getName() );
+                evaluateDecision( ctx, result, decisionNode, typeCheck );
+            }
+        }
+        return result == null ? new DMNResultImpl( model) : result;
+    }
+
+    private Object eval(String s) {
+        Object result = FEELCodeMarshaller.INSTANCE.unmarshall(null, s);
```
tarilabs: While there are reason I like this :), unfortunately I don't believe will be 100% correct. This is because for instance FEEL doesn't have a date literal, while Test Cases if I recall correctly use the attribute "xsd:date" and then enter the date as a literal from XML Schema value. You may want to check how TCK Test do it, so to try follow same strategy here.

tarilabs (commented):
Thank you @kurobako !
This is already a great work here to lay down the infrastructure to execute the embedded tests!
Just a couple comments here or in the code below.

I am not 100% sure we need these elements to extend in the `DMNModelInstrumentedBase` hierarchy, both object model (and converters analogously), given they are not part of the DMN standard specification itself.

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_1/xstream/extensions/ComponentConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class ComponentConverter extends DMNModelInstrumentedBaseConverter {
+    public ComponentConverter(XStream xstream) {
+        super(xstream);
+    }
+
+    @Override
+    protected void assignChildElement(Object parent, String nodeName, Object child) {
+        switch ( nodeName ) {
+            case "value": {
```
baldimir: Nitpicking: The Strings from the switch could be converted to constants, so they could be reused.

baldimir (approved):

kurobako:
updated this PR

kurobako:
jenkins retest this

kurobako:
ScorecardReasonCodeTest.testReasonCodesCombinations:221->assertReasonCode:247 expected:<OCC01> but was:<null>

this test runs successfully when alone

kurobako:
There are test failures on current master.

baldimir:
@kurobako, @mariofusco will take a look.

baldimir:
@kurobako could you please rebase latest master changes? This should be fixed now.

kurobako:
jenkins retest this

kurobako:
good :)

kurobako:
jenkins retest this

PR #1925, opened 2018-05-28
Drools 2591

MeyerNils:
See https://issues.jboss.org/browse/DROOLS-2591 for details


mariofusco:
@MeyerNils Both your test and fix looks ok, thanks a lot. The only problem is that I cannot merge pull request against an old version branch. Would you mind to resend this PR against master?

PR #1926, opened 2018-05-28
[DROOLS-2591] fix reloading of updated SNAPSHOT KJars

MeyerNils:
As requested


mariofusco:

PR #1927, opened 2018-05-28
[DROOLS-2339] fix handling of FileSystemResource on Windows

mariofusco:


baldimir (approved):

baldimir:
Tested in Windows environment and the test passes, so approved.

PR #1928, opened 2018-05-29
[DROOLS-2595] average accumulate function should return null instead …

mariofusco:
…of zero when no facts are matched


baldimir:
Looks good. When the conflict is fixed, approving.

mariofusco:
Merged by https://github.com/kiegroup/drools/commit/664e4f849fc1639efea2f6c7a626c75c5da692d5

PR #1931, opened 2018-05-31
[RHBRMS-3099] NullPointerException in BaseClassFieldReader.writeExternal() when a global method is used in LHS and serialize package

elguardian:
both commits are required in order to backport properly RHBRMS-3099

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/AccessorKey.java
+++ b/drools-core/src/main/java/org/drools/core/base/AccessorKey.java
@@ -44,7 +44,7 @@ public class AccessorKey
         final int PRIME = 31;
         int result = 1;
         result = PRIME * result + className.hashCode();
-        result = PRIME * result + ( (fieldName == null) ? 0 : fieldName.hashCode() );
+        result = PRIME * result + ( (this.fieldName == null) ? 0 : this.fieldName.hashCode() );
         result = PRIME * result + type.name().hashCode();
         this.hashCode = result;
     }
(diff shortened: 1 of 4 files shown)


baldimir:
ok to test

drools-pmml/.gitignore
```diff
@@ -1,13 +1 @@
-/target
```
baldimir: This file shouldn't be changed.

baldimir (changes requested):
Please see my comment.

drools-core/src/main/java/org/drools/core/base/AccessorKey.java
```diff
@@ -44,7 +44,7 @@ public AccessorKey(String className,
         final int PRIME = 31;
         int result = 1;
         result = PRIME * result + className.hashCode();
-        result = PRIME * result + ( (fieldName == null) ? 0 : fieldName.hashCode() );
+        result = PRIME * result + ( (this.fieldName == null) ? 0 : this.fieldName.hashCode() );
```
baldimir, line 47: This was not part of the original fix from [1]. @mariofusco is it ok to have this change here? I would say so, but I want to doublecheck this. 

[1] https://issues.jboss.org/browse/DROOLS-2517

elguardian: The test case fails without this fix ; Otherwise it fails after serialization invoking "equals" method in AccessorKey.

PR #1938, opened 2018-06-05
DROOLS-2612 Change default artifact version to 1.0.0

baldimir:
@mariofusco could you please review?


baldimir:

PR #1943, opened 2018-06-08
DROOLS-1663 Kie DMN support for IMPORT decisions between DMN files

tarilabs:
@etirelli just for some visibility on the code so far, and this should solve the problems discussed. I am waiting however for the clarification and alignment on the authoring side, in order to be sure this is correct.

Once authoring side is aligned, I can make a proper testing for 3 layers of import, which the current code should cover based on the understanding of "nested imports", not flattened import.

If the strategy is flattened imports, I need to undo some code and align, but again I expect I will have proper testing files to leverage.

Also before merging there is an opportunity for optimisation, when the FEEL layer calls again at the DMN level, currently for Function Invocation is taking the whole EvaluationContext, but I believe I could just clone everything except: the RootExecutionFrame, the ExecutionFrame introduced by the profiles.


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFEELCtxWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public boolean isDefined(String name) {
+        return wrapped.isDefined(name);
+    }
+
+    @Override
+    public DMNContext clone() {
+        return new DMNContextImpl(wrapped.getAllValues());
```
tarilabs, line 65: Q1: this is a DMNContext created to perform a FunctionInvocation from the FEEL layer back to the DMN layer (thinking BKM invocations, for instance). Here to clone the DMNContext, which is done by all Evaluators as a strategy, all symbols from FEEL EvaluationContet are collected --including from RootExecutionFrame and custom ExecutionFrame containing profile functions. Is this a problem?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
```diff
@@ -66,6 +66,82 @@ public void testImportDependenciesForDTInAContext() {
...
+        DMNModel dmnModelL3 = runtime.getModel("http://www.trisotech.com/dmn/definitions/_820c548c-377d-463e-a62b-bb95ddc4758c",
+                                               "L3 Do say hello");
+        assertThat(dmnModelL3, notNullValue());
+        for (DMNMessage message : dmnModelL3.getMessages()) {
+            LOG.debug("{}", message);
+        }
+
+        DMNContext context = runtime.newContext();
+        context.set("Another Name", "Bob");
+        context.set("L2import", mapOf(entry("Person name", "John")));
```
tarilabs: Q2: because `L2import` is the name of the DMN Import at the 3rd layer, here the InputData beloning to the second layer "Person name" need to be nested with the Map. Should I make so that `DMNContext.set()` checks the key, first parameter, and if it contains a dot (`.`) will take care of this map-nesting internally?

tarilabs (commented):
@etirelli final commits if you can kindly check of the 2 questions below, please? From my side is otherwise good to merge --and eventually deal with the 2 questions below later? Kindly let me know

/cc @kurobako @baldimir @winklerm

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNModelImpl.java
```diff
@@ -156,7 +164,7 @@ public DecisionNode getDecisionByName(String name) {
             return null;
         }
         for ( DecisionNode dn : this.decisions.values() ) {
-            if (dn.getName() != null && name.equals(dn.getName()) && getNamespace().equals(dn.getModelNamespace())) {
+            if (dn.getName() != null && name.equals(dn.getName())) {
```
jomarko: `dn.getName() != null` is not necessary if we will use `Objects.equals(name, dn.getName())`.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNModelImpl.java
```diff
@@ -203,7 +211,7 @@ public BusinessKnowledgeModelNode getBusinessKnowledgeModelByName(String name) {
             return null;
         }
         for ( BusinessKnowledgeModelNode bkm : this.bkms.values() ) {
-            if ( bkm.getName() != null && name.equals( bkm.getName() ) ) {
+            if (bkm.getName() != null && name.equals(bkm.getName())) {
```
jomarko: same comment about `Objects.equals()`

jomarko (commented):
I am not going to approve or decline, just small suggestion for code simplification.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
```diff
@@ -66,6 +66,82 @@ public void testImportDependenciesForDTInAContext() {
...
+    @Test
+    public void testImport2BKMs() {
+        DMNRuntime runtime = DMNRuntimeUtil.createRuntimeWithAdditionalResources("Do_say_hello_with_2_bkms.dmn",
+                                                                                 this.getClass(),
+                                                                                 "Saying_hello_2_bkms.dmn");
+
+        DMNModel importedModel = runtime.getModel("http://www.trisotech.com/dmn/definitions/_16a48e7a-0687-4c2d-b402-42925084fa1a",
+                                                  "Saying hello 2 bkms");
+        assertThat(importedModel, notNullValue());
+        for (DMNMessage message : importedModel.getMessages()) {
```
baldimir: This is used at more places, so it could be extracted to a separate method, something like logMessages(Messages).

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
baldimir: I personally like better the approach with dotted key:

`context.set("L2import.Person name", "John");`

It looks more user friendly to me.

tarilabs: Thinking about it @etirelli @baldimir , this would also simplify when a API user need to pass-in a complex type, like "Person", as instead of building the map manually:

```
context.set("Person", mapOf(entry("name", "John"), entry("age", 32)));
```

 the user could directly do something like:

```
context.set("Person.Name", "John");
context.set("Person.Age", 32);
```

What do you think?

tarilabs: (and I reiterate from my side we could merge this PR as-is without a definitive answer on this point, and deal with this Q2 later)

tarilabs: [ why would GitHub hide this comment now? Is actually still relevant ]

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFEELCtxWrapper.java
etirelli: I can't think of a problem from the top of my head. I guess we will need to try and if a problem is found later, we fix it.

etirelli (approved):

etirelli:
@tarilabs lets go ahead with the merge and lets release for Trisotech to try it out as well.

PR #1952, opened 2018-06-20
[DO NOT MERGE] DROOLS-2643: Do not generate extraneous parentheses

jomarko:
Do not generate extraneous parentheses around accumulate source pattern if accumulate function is used.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -771,22 +771,27 @@ public class RuleModelDRLPersistenceImpl
 
                 buf.append(" from accumulate ( ");
                 if (pattern.getSourcePattern() != null) {
+
+                    // https://issues.jboss.org/browse/DROOLS-2643
+                    final boolean isFunctionUsed = pattern.getFunction() != null && !pattern.getFunction().isEmpty();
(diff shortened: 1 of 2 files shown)


jomarko:
Still adding some tests, opened in advance to be able build:

jomarko:

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -771,22 +771,27 @@ public void visitFromAccumulateCompositeFactPattern(final FromAccumulateComposit
 
                 buf.append(" from accumulate ( ");
                 if (pattern.getSourcePattern() != null) {
+
+                    // https://issues.jboss.org/browse/DROOLS-2643
```
manstis, line 775: Hi @jomarko IMO this is a hack and does not cover all cases.

I had a quick look and it might be better to remove the ```isSubPattern``` parameter (from **all** of the ```visitXYZ...``` methods) and use the ```LHSGeneratorContext.getDepth()``` to check if the pattern is a sub-pattern. When debugging the ```FromCompositeFactPattern``` is indicated as not being a sub-pattern (although it is) which leads to the extra brackets being added. The methods invoked by the "reflective visitor" are hard-coding the "sub pattern" flag to ```false``` (in recent years we added the "context" which can now be better used to determine the "depth" of a pattern and hence whether it is a sub-pattern).

manstis: @jomarko See https://github.com/kiegroup/drools/pull/1953

PR #1971, opened 2018-06-29
[FINDBUGS] First bunch of fixes for problems reported by FindBugs in drools-core

baldimir:
@mariofusco could you please review? I made one separate commit for each reported thing I fixed. So if you want to check the original problem that ended with a code change, you could find it.


baldimir:

baldimir:

PR #1973, opened 2018-07-02
[DROOLS-2695] make beta node indexes to support type coercion

mariofusco:


mariofusco:
ok to test

mariofusco:
ok to test

PR #1977, opened 2018-07-11
Removed all references to PMML from drools-compiler module

lanceleverich:
Created PMMLAssemblerService to compile packages that contain PMML resources.

DO NOT MERGE: There is still a intermittent error that occurs in the a set of unit tests for the drools-scorecards module. I wanted to start getting this reviewed while I hunt down the reason for the error.


baldimir:
@jiripetrlik see this.

drools-pmml/src/main/java/org/drools/pmml/pmml_4_2/PMML4Compiler.java
```diff
@@ -70,7 +70,7 @@
     public static final String PMML = PMML_NAMESPACE + ".descr";
     public static final String SCHEMA_PATH = "xsd/org/dmg/pmml/pmml_4_2/pmml-4-2.xsd";
     public static final String BASE_PACK = PMML_DROOLS.replace('.','/');
-
```
jiripetrlik: It seems that there are only unrelated formatting changes in this file. Please remove these changes from this PR.

drools-workbench-models/drools-workbench-models-guided-scorecard/src/main/java/org/drools/workbench/models/guided/scorecard/backend/GuidedScoreCardProviderImpl.java
```diff
@@ -24,8 +24,8 @@
 import org.dmg.pmml.pmml_4_2.descr.Scorecard;
 import org.dmg.pmml.pmml_4_2.descr.True;
 import org.drools.compiler.compiler.GuidedScoreCardProvider;
-import org.drools.compiler.compiler.PMMLCompiler;
```
jiripetrlik, line 27: Please remove imports completely instead of commenting them out. It is possible to return back to the older version using GIT history.

kie-pmml/src/main/java/org/kie/pmml/assembler/PMMLAssemblerService.java
```diff
@@ -0,0 +1,288 @@
...
+import org.kie.api.io.ResourceType;
+import org.kie.api.io.ResourceWithConfiguration;
+import org.kie.internal.builder.KnowledgeBuilderResult;
+import org.kie.internal.io.ResourceFactory;
+import org.kie.pmml.pmml_4_2.PMML4Compiler;
+import org.kie.pmml.pmml_4_2.PMMLResource;
+
+public class PMMLAssemblerService implements KieAssemblerService {
+
+	private List<KnowledgeBuilderResult> results;
```
jiripetrlik: It seems that this variable is used to save some values, but these values are never used.

jiripetrlik (commented):
Thank you @lanceleverich for PR. Just a few minor comments. Can you be more specific about scorecards error. I've compiled Drools from your branch and there are no test errors in kie-pmml and drools-scorecards module. Maybe I missed something?

lanceleverich:
The requested changes have been made.
As for the intermittent error, I have not been able to get it to happen today.  See https://issues.jboss.org/browse/DROOLS-2592 for details about the issue.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/KnowledgeBuilderTest.java
```diff
@@ -459,7 +456,7 @@ public void testAddPackageArray() throws Exception {
     public void testResourceMapping() throws Exception {
         ServiceRegistryImpl serviceRegistry = (ServiceRegistryImpl) ServiceRegistry.getInstance();
         serviceRegistry.reset();
-
+/*
```
baldimir, line 459: Commented out code. Please remove it.

baldimir:
Just one minor comment about commented out code. Please see inline.

jiripetrlik:
@lanceleverich Thank you for changes! It is possible to get intermittent error by running "testPointsBelow" test multiple times. It usually fails in 1 of 20 runs. You can use "repeat until failure" function in Intelij IDEA to simply reproduce this error. However, it seems that the error is unrelated to this PR. Does it make sense to block this PR by it?

mariofusco:
@lanceleverich please review that test following @jiripetrlik suggestion to try to reproduce the failure.

lanceleverich:
I agree with @jiripetrlik in that the error is not related to this PR. I am continuing to work on finding what the underlying cause is, but I think we could move forward with merging the PR.

PR #1984, opened 2018-07-17
DROOLS-2746 DMN Invocation parameters resolution with imported function

tarilabs:
/cc @etirelli @baldimir @winklerm


baldimir (approved):

etirelli (approved):
Please check about the function with a dot in the name as a separate ticket.

PR #1985, opened 2018-07-18
[DROOLS-2740] Move ExecutableCommand to public API

danielezonca:
@mariofusco 
Move ```ExecutableCommand``` interface to public API (without ```canRunInTransaction``` method that remains internal).
Create internal ```TransactionalCommand``` to keep ```canRunInTransaction``` method.

[Changed ```InternalExecutable```](https://github.com/danielezonca/drools/blob/31150b7d27d812b3490454a909f9819dba7b0d7c/drools-core/src/main/java/org/drools/core/fluent/impl/InternalExecutable.java#L31) to consider by default all ```Command``` that are not instance of ```TransactionalCommand``` as able to run in transaction (similar to previous behaviour where default value of ```canRunInTransaction``` for all ```ExecutableCommand``` was ```true```)

To be merged together with
https://github.com/kiegroup/droolsjbpm-knowledge/pull/314
https://github.com/kiegroup/drools/pull/1985
https://github.com/kiegroup/jbpm/pull/1281
https://github.com/kiegroup/droolsjbpm-integration/pull/1521


MarianMacik:
Just one note - since org.drools.core.command.impl.ExecutableCommandImpl is no longer available and since we moved it now under org.kie.api - is there any chance that somebody was using it as "public" before although it was internal? In that case, I think it is a backward-incompatible change. Just to be aware of that.
@danielezonca your opinion?

danielezonca:
@MarianMacik 
I don't know when org.drools.core.command.impl.ExecutableCommandImpl was removed but yes this change could break some custom code that was using internal ExecutableCommand interface.
I can easily add a deprecated internal ExecutableCommand interface to warn the user to move to the public API or just add a release note.
@mariofusco wdyt?

PR #1992, opened 2018-07-24
[RHDM-645] "java.security.AccessControlException: WFSM000001: Permission check failed" when enabled security-manager

elguardian:
avoiding common pool in parallel streams so it checks properly with the security manager

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
@@ -43,6 +43,7 @@ import java.util.Stack;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ForkJoinPool;
 import java.util.function.Supplier;
 
 import org.drools.compiler.builder.impl.errors.SrcError;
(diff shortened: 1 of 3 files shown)


drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1397,19 +1398,26 @@ private void compileRulesLevel(PackageDescr packageDescr, PackageRegistry pkgReg
...
-                                this.results.addAll(results);
+                                    this.results.addAll(results);
+                                }
                             }
-                        }
-                    });
+                        })
+                ).get();
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("An error has occurred during rules compilation", e);
```
baldimir: Not sure, if we want the error just logged. What do you think @mariofusco?

mariofusco: @elguardian @baldimir No, in this way the error will be swallowed. It needs to be rethrown (wrapped in a RuntimeException if necessary).

elguardian:
@baldimir any progress with the review ? a pushed a new commit the changes asked

elguardian:
ok to test

baldimir (approved):
Sorry, I missed the new commit. Approving.

PR #1995, opened 2018-07-26
DROOLS-2791 Test DMN Import with 3 layers of antipattern of importing ..

tarilabs:
Decision and InputData.

Test DMN Import transitive 3 level diamond

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
@@ -205,6 +205,52 @@ public class ImportsTest {
         assertThat(evaluateAll.getDecisionResultByName("Say the Greeting to Person").getResult(), is("Hello, John"));
     }
 
+    @Test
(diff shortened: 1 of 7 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
```diff
@@ -205,6 +205,52 @@ public void testImportHardcodedDecisions() {
...
     }
 
+    @Test
+    public void testImportTransitive() {
+        DMNRuntime runtime = DMNRuntimeUtil.createRuntimeWithAdditionalResources("Sayhello1ID1D.dmn",
+                                                                                 this.getClass(),
+                                                                                 "ModelB.dmn",
+                                                                                 "ModelB2.dmn",
+                                                                                 "ModelC.dmn");
+        {
```
baldimir: I think this would be better split into 3 separate tests. Something like this [1].

[1] https://gist.github.com/baldimir/5e13ffd8aa7a557e47ceb9d986a4a18c

baldimir (approved):

PR #1997, opened 2018-07-27
[DROOLS-2923] Improve DMN execution performance

ignacio-ifm:
The DMN feel evaluation runs the same Java reflection calls for previously seen Objects, which is not performant. A cache of known accessors is added to avoid useless computation. The cache is a simple HashMap and thus does not have an eviction policy. This has an impact on memory but it is assumed that the amount of different Classes mapped to DMN rules is neglible for a single application and is naturally bounded.

Additionally, a cache is added to the normalization of variable names that was also detected as a performance bottleneck, same logic applies there.

On a simple perfomance test against a similar boolean rule using DRL the results are:
Before: DMN is x5 times slower than DRL
After: DMN is x2 times slower than DRL

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
@@ -52,6 +52,7 @@ import org.kie.dmn.core.impl.DMNPackageImpl;
 import org.kie.dmn.feel.util.Either;
 import org.kie.dmn.model.api.Definitions;
 import org.kie.dmn.model.api.Import;
+import org.kie.dmn.feel.util.EvalHelper;
 import org.kie.internal.builder.ResultSeverity;
 import org.kie.internal.utils.ChainedProperties;
 import org.slf4j.Logger;
@@ -73,6 +74,7 @@ public class DMNAssemblerService implements KieAssemblerService {
 
     @Override
     public void addResources(Object kbuilder, Collection<ResourceWithConfiguration> resources, ResourceType type) throws Exception {
+        EvalHelper.clearGenericAccessorCache();
         KnowledgeBuilderImpl kbuilderImpl = (KnowledgeBuilderImpl) kbuilder;
         DMNCompilerImpl dmnCompiler = (DMNCompilerImpl) kbuilderImpl.getCachedOrCreate(DMN_COMPILER_CACHE_KEY, () -> getCompiler(kbuilderImpl));
         DMNMarshaller dmnMarshaller = dmnCompiler.getMarshaller();
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
@@ -28,11 +28,11 @@ import java.time.temporal.ChronoField;
 import java.time.temporal.Temporal;
 import java.time.temporal.TemporalQueries;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TimeZone;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiPredicate;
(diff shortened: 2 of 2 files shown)


ghost:

ghost:

triceo:
ok to test

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
```diff
@@ -48,8 +49,18 @@
     public static final Logger LOG = LoggerFactory.getLogger( EvalHelper.class );
     private static final Pattern SPACES_PATTERN = Pattern.compile( "[\\s\u00A0]+" );
 
+    private static final Map<String, Method> accessorCache = new HashMap<>();
+    private static final Map<String, String> normalizationCache = new HashMap<>();
+
     public static String normalizeVariableName(String name) {
-        return SPACES_PATTERN.matcher( name.trim() ).replaceAll( " " );
+		if (normalizationCache.containsKey(name)){
```
mariofusco: Are we caching the result of the application of a regular expression to a String here? Unless there are very good reason to do this, and we can demonstrate it with a benchmark, this seems really a bad idea.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
```diff
@@ -313,13 +324,25 @@ public static Object getValue(final Object current, final String property) {
      */
     public static Method getGenericAccessor(Class<?> clazz, String field) {
         LOG.trace( "getGenericAccessor({}, {})", clazz, field );
-        return Stream.of( clazz.getMethods() )
+
+        String accessorQualifiedName = new StringBuilder(clazz.getCanonicalName())
```
mariofusco: Caching an accessor method could make a bit more sense. Maybe we can just make it nicer using the computeIfAbsent() method.

mariofusco (changes requested):
Should this stuff be thread safe? If not it's ok, but I'd keeè only the cache of the accessor method and not of the regular expression.

baldimir (changes requested):
Hi, sorry this won't work without cache eviction, because of incremental compilation. Imagine a new version of kjar is loaded at runtime with different definitions. This cache will contain the first versions forever.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
ignacio-ifm: Although in my tests it showed about 10% performance improvement I'll focus the PR in the reflection side, which has the bigger impact. So I removed this improvement.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
ignacio-ifm: Yes, computeIfAbsent() made it nicer!

I didn´t quite get the "Caching an accessor method could make a bit more sense" part. I built the string as key because otherwise you would need a map within a map (clazz > field > Method) and would make the look up slower. You suggest something else as key?

ignacio-ifm:
@mariofusco I changed it to a ConcurrentHashMap for Thread safety.

@baldimir what eviction strategy would you recommend that covers the dynamic loading case and still improves the performance? Maybe a direct eviction when the reload is detected? Can you point me where in the code would that happen?

tarilabs:
Hi @ignacio-ifm
thank you for this PR. Unfortunately without a mechanism for eviction the cache would actually create troubles during a KieContainer upgrade use-case. The KieContainer upgrade was/is a foreseen use-case, but we didn't have an explicit JUnit test for this in the codebase; I added it as part of PR #2000 .

tarilabs:
Jenkins test again

tarilabs (changes requested):
Demonstrated by #2000 problem with cache without eviction policy. @mariofusco do you have a suggestion for a callback to be placed in `EvalHelper`, please?

ghost:

ghost:

tarilabs:
(I am working on a proposal for a cache eviction based on the original content of this PR, preserving commit authoring --will keep posted)

tarilabs:
@ignacio-ifm unfortunately the master branch diverged quite a bit in the meantime, so this requires a rebase. Let me know if that is feasible from your side to rebase, otherwise I will create on my side a new PR (preserving commit authoring as possible) with all changes.

ignacio-ifm:
@tarilabs I did the rebase. There was a minor conflict on some imports with a strange org.kie.dmn.model.api.(v1_1).Definitions I kept the ones without the version.

Please check.

tarilabs:
> There was a minor conflict on some imports 

Correct, this package changed in between when this PR originally raised, and currently on master, hence the git diff unmatched because it was in the nearby lines of the diff.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
```diff
@@ -73,6 +74,7 @@ public ResourceType getResourceType() {
 
     @Override
     public void addResources(Object kbuilder, Collection<ResourceWithConfiguration> resources, ResourceType type) throws Exception {
+        EvalHelper.clearGenericAccessorCache();
```
tarilabs, line 77: @mariofusco cache eviction has been added. Can you kindly merge with squash and title "`DROOLS-2923 Improve DMN execution performance`", please?

tarilabs (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/assembler/DMNAssemblerService.java
tarilabs: (better to rename the PR title as well with the same, but I don't have authorization to do it)

tarilabs:
@ignacio-ifm thanks again and this was merged with the cache eviction discussed.

I seize the chance to mention that we are currently working on a next generation of optimisations which will include benefit also for FEEL evaluations too. We will update the Release Notes and communicate when this will be available, I suggest for you to eventually try that out too, and feedback will be appreciated.

ignacio-ifm:
@tarilabs That sounds great. I'll do. Thanks!

PR #1999, opened 2018-07-30
[DROOLS-2766] allow symbols in output of compiled DMN decision tables

mariofusco:


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/DTableModel.java
```diff
@@ -73,10 +73,43 @@ public DTableModel( DMNFEELHelper feel, String namespace, String dtName, Decisio
...
         this.dtable = new DecisionTableImpl( dtName, outputs );
-        initOutputs(feel);
+    }
+
+    public DTableModel compileAll( DMNFEELHelper feel, DMNCompilerContext ctx ) {
+        CompilerContext feelctx = feel.newCompilerContext();
+        feelctx.setDoCompile( true );
+        ctx.getVariables().forEach( (k, v) -> feelctx.addInputVariableType( k, ((BaseDMNTypeImpl ) v).getFeelType() ) );
+
+        initInputs(feel, feelctx);
```
tarilabs: can you please rename these as InputClauses and OutputClauses where necessary to align with DMN definitions, please?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/DTableModel.java
```diff
@@ -73,10 +73,43 @@ public DTableModel( DMNFEELHelper feel, String namespace, String dtName, Decisio
...
+        for (DRowModel row : rows) {
+            row.compiledOutputs = row.outputs.stream().map( expr -> (CompiledFEELExpression) feel.compile( expr, feelctx ) ).collect( toList() );
+        }
+    }
+
+    private void initOutputs( DMNFEELHelper feel, CompilerContext feelctx ) {
+        for (DOutputModel output : outputs) {
+            String outputValuesText = getOutputValuesText( output.outputClause );
+            if (outputValuesText != null) {
+                output.outputValues = feel.evaluateUnaryTests( outputValuesText, variableTypes );
```
tarilabs: I am not sure about this: output clauses are meant to be FEEL expression, not unarytest. I believe here should be calling down to FEEL to compile for an expression, not the unary test.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/Null.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.dmn.core.compiler.execmodelbased;
+
+public class Null {
```
tarilabs: This is in the wrong module. It needs to be moved into the kie-dmn-feel in order to be reachable/used from FEEL layer.

tarilabs (commented):
Some little changes required, otherwise good and as always thank you!!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/DTableModel.java
tarilabs: please ignore previous comment, misunderstanding. This is not a (row-)output but the OutputClause Unarytest of the header for the table

PR #2006, opened 2018-08-01
[DROOLS-2766] fix error management + other minor improvements in comp…

mariofusco:
…iled DMN decision tables


tarilabs (approved):

mariofusco:
The failing test is in scorecard module, so for sure it is a false negative unrelated with this pull request.

PR #2007, opened 2018-08-01
DROOLS-2822 FEEL augment not() heuristic for function invocation

tarilabs:
/cc @etirelli @baldimir @jiripetrlik @evacchi please notice I tested locally to bypass slow-Jenkins and it works.

Until the proper not() heuristic solution is in place, this should do as well.


etirelli (approved):

etirelli:
looks good @tarilabs

PR #2008, opened 2018-08-02
DROOLS-2824: adding support for context entry evaluator listeners

etirelli:


kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/DMNRuntimeEventListener.java
```diff
@@ -26,6 +26,10 @@
 
     void afterEvaluateBKM(AfterEvaluateBKMEvent event);
 
+    default void beforeEvaluateContextEntry(BeforeEvaluateContextEntryEvent event) {}
```
etirelli: made the method "default" for interface backward compatibility. I wonder if we should make all methods in that interface default?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerUtils.java
```diff
@@ -41,43 +32,73 @@
     private static final Logger logger = LoggerFactory.getLogger( DMNRuntimeEventManagerUtils.class );
 
     public static void fireBeforeEvaluateDecision( DMNRuntimeEventManager eventManager, DecisionNode decision, DMNResult result) {
-        BeforeEvaluateDecisionEvent event = new BeforeEvaluateDecisionEventImpl( decision, result );
-        notifyListeners( eventManager, l -> l.beforeEvaluateDecision( event ) );
+        if( eventManager.hasListeners() ) {
```
etirelli, line 35: optimization to avoid creating the event instance when there are no registered listeners

tarilabs: the proper solution would have been instead of guarding with an `if` to move the `new` instantiation inside the lambda, will approve anyway and change later

tarilabs (approved):

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/DMNRuntimeEventListener.java
baldimir: I would say so, because in that case when someone uses this interface, they just implement the event handling they need.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/DMNRuntimeEventManager.java
```diff
@@ -46,6 +46,13 @@
      */
     Set<DMNRuntimeEventListener> getListeners();
 
+    /**
+     * Returns true if there are registered listeners, false otherwise
+     *
+     * @return
+     */
+    boolean hasListeners();
```
baldimir, line 54: This could have a default - `getListeners() != null && !getListeners.isEmpty()`

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNContextEvaluator.java
```diff
@@ -119,10 +125,14 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult d
                                                Msg.ERR_EVAL_CTX_ENTRY_ON_CTX,
                                                ed.getName(),
                                                name );
+                        DMNRuntimeEventManagerUtils.fireAfterEvaluateContextEntry( eventManager, name, ed.getName(), entryVarId, entryExprId, er.getResult(), result );
                         return new EvaluatorResultImpl( results, ResultType.FAILURE );
                     }
                 } catch ( Exception e ) {
                     logger.error( "Error invoking expression for node '" + name + "'.", e );
+                    String entryVarId = ed.getContextEntry().getVariable() != null ? ed.getContextEntry().getVariable().getId() : null;
```
baldimir: The variable null check is repeated. There could be an overloaded version of `DMNRuntimeEventManagerUtils.fireAfterEvaluateContextEntry` which would take ed.getContextEntry() as an argument and handle the nullchecks internally.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerImpl.java
```diff
@@ -54,6 +54,11 @@ public void removeListener(DMNRuntimeEventListener listener) {
         return listeners;
     }
 
+    @Override
+    public boolean hasListeners() {
```
baldimir, line 58: As mentioned before, this could be made a default in the interface.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/DMNRuntimeEventListener.java
etirelli: Ok, I will do so.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerUtils.java
etirelli: instantiating the event inside the lambda would do so for every listener, while the way it is, it instantiates only once and sends the same event to all listeners. Happy to change, but I think this is more efficient.

tarilabs: > I think this is more efficient.

It depends who you ask, if I understand correctly there is another school of thought which would say is more correct to have another instance =P
image

tarilabs: but anyway imho it can stay as-is. The priority for this PR is another

etirelli:
I addressed some of the concerns raised in the PR. @mariofusco @tarilabs can I ask you please to merge this once the build succeeds? Thanks.

tarilabs (approved):

baldimir (approved):

PR #2010, opened 2018-08-02
[DROOLS-2766] support for question mark variable in evaluation contex…

mariofusco:
…t of compiled DMN decision table


tarilabs (approved):
thanks!

PR #2013, opened 2018-08-03
[DROOLS-2766] support for null values in compiled DMN dtables + minor…

mariofusco:
… refactor


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/FeelValue.java
```diff
@@ -14,11 +14,16 @@
  * limitations under the License.
  */
 
-package org.kie.dmn.feel.util;
+package org.kie.dmn.core.compiler.execmodelbased;
 
-public class Null {
+public class FeelValue {
```
tarilabs, line 19: ok

tarilabs (approved):
Thanks to both you and @evacchi !

PR #2014, opened 2018-08-06
DROOLS-2766 Do not wrap in `eq()` when `?` is used in a test

evacchi:
quick & dirty, but straight to the point

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
@@ -255,13 +255,7 @@ public class DMNFEELHelper {
         }
         variableTypes.put( "?", columntype );
 
-        FEEL_1_1Parser parser = FEELParser.parse(
-                null,
-                input,
-                variableTypes,
-                Collections.emptyMap(),
-                feel.getCustomFunctions(),
-                Collections.emptyList());
+        FEEL_1_1Parser parser = FEELParser.parse(null, input, variableTypes, Collections.emptyMap(), feel.getCustomFunctions(), Collections.emptyList());
 
         FEEL_1_1Parser.ExpressionListContext tree = parser.expressionList();
         DirectCompilerVisitor v = new DirectCompilerVisitor(variableTypes);
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
@@ -22,6 +22,7 @@ import org.kie.dmn.core.util.Msg;
 import org.kie.dmn.core.util.MsgUtil;
 import org.kie.dmn.feel.FEEL;
 import org.kie.dmn.feel.codegen.feel11.CompilerBytecodeLoader;
+import org.kie.dmn.feel.codegen.feel11.DirectCompilerResult;
 import org.kie.dmn.feel.codegen.feel11.DirectCompilerVisitor;
 import org.kie.dmn.feel.lang.CompiledExpression;
 import org.kie.dmn.feel.lang.CompilerContext;
@@ -30,6 +31,7 @@ import org.kie.dmn.feel.lang.Type;
 import org.kie.dmn.feel.lang.impl.EvaluationContextImpl;
 import org.kie.dmn.feel.lang.impl.FEELEventListenersManager;
 import org.kie.dmn.feel.lang.impl.FEELImpl;
+import org.kie.dmn.feel.lang.types.BuiltInType;
 import org.kie.dmn.feel.parser.feel11.FEELParser;
 import org.kie.dmn.feel.parser.feel11.FEEL_1_1Parser;
 import org.kie.dmn.feel.runtime.FEELFunction;
@@ -46,7 +48,7 @@ public class DMNFEELHelper {
     private static final Logger logger = LoggerFactory.getLogger( DMNFEELHelper.class );
 
     private final ClassLoader classLoader;
-    private final FEEL                   feel;
+    private final FEELImpl               feel;
     private final FEELEventsListenerImpl listener;
     private final List<FEELProfile> feelProfiles = new ArrayList<>();
 
@@ -58,7 +60,7 @@ public class DMNFEELHelper {
         this.classLoader = classLoader;
         this.feelProfiles.addAll(feelProfiles);
         this.listener = new FEELEventsListenerImpl();
-        this.feel = createFEELInstance();
+        this.feel = (FEELImpl) createFEELInstance();
     }
 
     private FEEL createFEELInstance() {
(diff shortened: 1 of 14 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerVisitor.java
```diff
@@ -479,6 +479,9 @@ public DirectCompilerResult visitExpressionList(FEEL_1_1Parser.ExpressionListCon
                         // being a range, need the `in` operator.
                         DirectCompilerResult replaced = createUnaryTestExpression(childCtx, child, UnaryOperator.IN);
                         exprs.add(replaced);
+                    } else if (child.resultType == BuiltInType.BOOLEAN) {
```
tarilabs: Sorry, not entirely convided this will work in all cases --can you please check `child.expression` contains a JavaParser.NameExpr for "?" as well? Sometimes is not so easy to determine the type of the result, hence here it would go into BuiltinType.ANY and won't match.

tarilabs (changes requested):
The strategy looks okay, just an actual if check i would do differently.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerVisitor.java
tarilabs: > contains a JavaParser.NameExpr for "?" as well?

actually contains a JavaParser.NameExpr for "?" in place of the current "resultType".
This would also better align with the spec (although we would leave the result type check for boolean at runtime)

mariofusco:
@evacchi @tarilabs Did you check that this PR fixes the question mark tests that were failing when enabling the dtables compilation?

evacchi:
Hi @tarilabs 

Here's the situation so far:

-  broken because fields are not sorted the right way (a_b forward-references a, b) 
`testDecisionTablesQuestionMarkInWithUnaryTests[true]`

- these fail for the same reason: expected to FAIL, but SUCCEED -- possibly an error in tests; i.e. [null, true] should evaluate to true and not raise an ERROR (at most a WARNING) 
  * `testDecisionTablesQuestionMarkInNonBooleanFunction[true]`
  * `testDecisionTablesQuestionMarkOnly[true]`

- most probably failing because of `not()`
  `testDTUsingEqualsUnaryTestWithVariable2[true]`

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
```diff
@@ -249,18 +251,26 @@ public void onEvent(FEELEvent event) {
...
-        ParseTree tree = parser.expressionList();
-        DirectCompilerVisitor v = new DirectCompilerVisitor(variableTypes, true);
-        return new CompilerBytecodeLoader().getSourceForUnaryTest(packageName, className, input, v.visit(tree));
+        FEEL_1_1Parser parser = FEELParser.parse(
+                null,
+                input,
+                variableTypes,
+                Collections.emptyMap(),
+                feel.getCustomFunctions(),
+                Collections.emptyList());
```
tarilabs: :thinking:

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSemanticMappings.java
```diff
@@ -211,7 +211,20 @@ private static Comparable asComparable(Object s) {
         }
         return (T) actual;
     }
-    
+
+    public static Boolean coerceToBoolean(EvaluationContext ctx, Object value) {
+        if (value instanceof Boolean) return (Boolean) value;
+
+        ctx.notifyEvt( () -> new ASTEventBase(
+                FEELEvent.Severity.ERROR,
```
tarilabs: shouldn't be WARN also in this case?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
evacchi: I can't read STUFF captain
can't read stuff

I can revert that anyway

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSemanticMappings.java
evacchi: right. I have reverted the EVAL part, though, wanted @etirelli's POV before. Will ping him later

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerVisitor.java
```diff
@@ -468,30 +489,57 @@ public DirectCompilerResult visitExpressionList(FEEL_1_1Parser.ExpressionListCon
...
+//                    } else if (isExtendedUnaryTest(childCtx)) {
+//                        DirectCompilerResult replaced = createUnaryTestExpression(childCtx, child, UnaryOperator.TEST);
+//                        exprs.add(replaced);
+//                    } else {
+//                        // implied a unarytest for the `=` equal operator.
+//                        DirectCompilerResult replaced = createUnaryTestExpression(childCtx, child, UnaryOperator.EQ);
+//                        exprs.add(replaced);
+////                        exprs.add(child);
+//                    }
+//                }
```
evacchi: will remove this commented code

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/DTableModel.java
```diff
@@ -138,7 +140,9 @@ private void initRows( CompilerContext feelctx, Map<String, CompiledFEELExpressi
         int index = 1;
         for (DRowModel row : rows) {
             int rowIndex = index;
-            row.compiledOutputs = row.outputs.stream().map( expr -> compileFeelExpression( dt, feel, feelctx, Msg.ERR_COMPILING_FEEL_EXPR_ON_DT_RULE_IDX, compilationCache, expr, rowIndex ) ).collect( toList() );
+            row.compiledOutputs = row.outputs.stream().map( expr ->
+                                                                    compileFeelExpression( dt, feel, feelctx, Msg.ERR_COMPILING_FEEL_EXPR_ON_DT_RULE_IDX, compilationCache, expr, rowIndex ) )
+                    .collect( toList() );
```
evacchi: this change is useless, will revert

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNFEELHelper.java
tarilabs: > I can revert that anyway

For me it can stay, but I would like to highlight FYI this is not as per codestyle standard, it should align on the null where the null is side to the open paren

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSemanticMappings.java
etirelli: When is this error generated?

evacchi:
Please notice that previous situation for Unaries was:

compilation failed:

> DMNDecisionTableRuntimeTest.testDTUsingEqualsUnaryTestWithVariable1
> DMNDecisionTableRuntimeTest.testDTUsingEqualsUnaryTestWithVariable2
> DMNRuntimeTest.testListContainmentDT
> DMNRuntimeTest.testNotListInDT

now compilation DOES not fail, test results are... controversial

evacchi:
Let's close this. We have to rework this from a grammar perspective so that we are more confident about the ambiguities

PR #2015, opened 2018-08-07
[RHPAM-2] NullPointerException during transaction rollback when using CMT

elguardian:
if it is not the transaction owner we try to rely on the transaction registry
and set the current tx associated as a rollback

### DIFF
--- a/drools-persistence/drools-persistence-api/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
+++ b/drools-persistence/drools-persistence-api/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java
@@ -262,8 +262,18 @@ public class JtaTransactionManager
             		this.ut.rollback();
         		}
         	} else {
-        		getUt().setRollbackOnly();
-        	}
+                // use transaction sync registry if available. this works if there is tx associated
+                if (this.tsr != null) {
+                    try {
+                        ((TransactionSynchronizationRegistry) this.tsr).setRollbackOnly();
+                    } catch ( IllegalStateException e) { 
+                        getUt().setRollbackOnly();
+                    }
+                } else {
+                    // if no transaction sync registry available fallback to user transaction
+                    getUt().setRollbackOnly();
+                }
+            }
         } catch ( Exception e ) {
             logger.warn( "Unable to rollback transaction", e);
             throw new RuntimeException( "Unable to rollback transaction",


mariofusco:

elguardian:
any feedback @mariofusco

mswiderski (approved):
change is ok, was wondering if there is anyway to make a test for it? @MarianMacik maybe jbpm container tests would help here...

MarianMacik:
Just to be clear what tx tests we have in jbpm-in-container:
We have transactional tests which call EJB here:
https://github.com/kiegroup/jbpm/blob/master/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/test/ejbservices/tx/ETransactionTest.java

So here we test our EJBs which are by default set to CMT. Before we call these beans we start tx ourselves so even though our EJBs are CMT they will only join transaction, right? If that's true, engine is not tx owner, then we can create a new test in this class or in the class: https://github.com/kiegroup/jbpm/blob/master/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/test/ejbservices/process/EStartProcessTest.java - in this one we don't use user managed transactions at all (container will create tx for us). So we can include exactly the reproducer from the JIRA (e.g. throw exception in Script Task) in any of these two classes - maybe both use cases are worth implementing?

elguardian:
thx @MarianMacik for the detailed explanation.
Note: I don't think EJB join transaction, by default the behaviour is required so if the tx does not exist it will create a new one. Keep in mind that if you are joining a tx not controlled by container, that defeats the idea of CMT (somebody from outside would control the tx. I am presuming this maybe you manage to do that somehow)

elguardian:
@MarianMacik  Not really sure if that is what you have in mind
https://github.com/kiegroup/jbpm/pull/1309

MarianMacik:
@elguardian My point was that by default it is required - so if there isn't any transaction, it will create a new one, but if there is already one (in this situation created by us), it will just join the already running tx. I think it makes sense. So yes, you are right, this partly defeats the idea of CMT, but it still can be used for situations where you call EJB-component from non-ejb component like remote client for example. Hope this clarifies it a little bit.

mariofusco:

mswiderski:
Jenkins retest this

mswiderski (approved):

PR #2016, opened 2018-08-10
DROOLS-2875 Problem compiling PMML when multiple output fields are present

lanceleverich:
- scorecardOutputGeneration.drlt - Updated rule names to add the target field for situations where there are multiple output fields defined.
- regParams.drlt - Changed the insertion of RegTable objects to be insert instead of insertLogical
- Updated test files to include multiple fields tagged as output
- Fixed Miningmodel.java so that the getMiningFields method would also return any fields that were used by sub-models. This removes a limitation that previously existed where the all fields that were in a sub-model's mining schema had to be included in the mining models mining schema.


kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/PMML4ExecutionHelper.java
```diff
@@ -264,10 +267,18 @@ public synchronized PMML4Result submitRequest(PMMLRequestData request)
...
+
+        KieRuntimeLogger logger = loggerFileName != null ?
+                ((InternalRuleUnitExecutor)executor).addFileLogger(loggerFileName) : null;
+        try {
+            requestData.insert(request);
+            baseResultHolder = new PMML4Result(request.getCorrelationId());
+            resultData.insert(baseResultHolder);
+            executor.run(ruleUnitClass);
+        } finally {
+            if (logger != null)
```
mariofusco: nitpicking: missing { here

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/Miningmodel.java
```diff
@@ -122,6 +122,13 @@ public PMMLOutputField findOutputField( String fieldName) {
                 fields.remove(field);
             }
         }
+        if (this.segmentation != null) {
+            segmentation.getMiningSegments().forEach(ms -> {
+                PMML4Model model = ms.getModel();
+                List<PMMLMiningField> innerFields = model.getMiningFields();
+                innerFields.stream().filter(fld -> !fields.contains(fld)).forEach(fld -> {fields.add(fld);});
```
mariofusco: I guess that the whole code  in this if block could be rewritten as

`segmentation.getMiningSegments().stream().flatMap( ms -> ms.getModel().getMiningFields().stream() ).filter(fld -> !fields.contains(fld)).forEach( fields::add )`

mariofusco (approved):

PR #2017, opened 2018-08-14
DROOLS-2607 Add rounding to the BigDecimal average accumulator

baldimir:
@tarilabs @etirelli could you please review while @mariofusco is away? The division uses the same scale as before, because `count` is an int. (Without rounding it uses `this.scale() - divisor.scale()`, with rounding it uses just `this.scale()`). 

JIRA: https://issues.jboss.org/browse/DROOLS-2607

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/accumulators/BigDecimalAverageAccumulateFunction.java
+++ b/drools-core/src/main/java/org/drools/core/base/accumulators/BigDecimalAverageAccumulateFunction.java
@@ -20,8 +20,8 @@ import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
-import java.io.Serializable;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 
 /**
  * An implementation of an accumulator capable of calculating average values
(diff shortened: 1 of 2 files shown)


triceo:
@baldimir Missing a regression test.

baldimir:
@triceo What an irony, right? :) Sorry, will add it.

baldimir:
@triceo reproducer added.

drools-core/src/main/java/org/drools/core/base/accumulators/BigDecimalAverageAccumulateFunction.java
```diff
@@ -95,7 +95,7 @@ public void reverse(AverageData data,
      * @see org.kie.base.accumulators.AccumulateFunction#getResult(java.lang.Object)
      */
     public Object getResult(AverageData data) {
-        return data.count == 0 ? BigDecimal.ZERO : data.total.divide( BigDecimal.valueOf( data.count ) );
+        return data.count == 0 ? BigDecimal.ZERO : data.total.divide(BigDecimal.valueOf( data.count ), RoundingMode.HALF_EVEN );
```
tarilabs, line 98: sorry but I am missing something: isn't the average of `{0, 0, 1}` actually 0.3333.. depending on the context? Shouldn't we just pass in `MathContext.DECIMAL_128` instead of an explicit rounding?

baldimir: Yes, you are right, but I wanted to preserve the original behaviour. If you average Integers, it scales to 0.

tarilabs: ok thanks for the clarification

tarilabs (approved):

baldimir:
Test fail not related - it is a PMML test already reported.

baldimir:
jenkins retest this

PR #2019, opened 2018-08-17
DROOLS-1393 - Investigate why JpaOptLockPersistentStatefulSessionTest…

mswiderski:
… is failing with Hibernate 5

PR #2020, opened 2018-08-17
RHPAM-214 - Add testOptimisticLockInterceptorMaxRetry test

MarianMacik:
@mswiderski The PR with a new test. I have also changed the current test so it doesn't use Thread.sleep() to wait for threads, but now it does it the proper way using join().


mswiderski:
included in https://github.com/kiegroup/drools/pull/2019

PR #2022, opened 2018-08-22
RHDM-717: add reproducer

winklerm:
Added a reproducer for NPE when opening a live query with null parameters (RHDM-717).


mariofusco:
Test case merged and fixed by https://github.com/kiegroup/drools/commit/534735c99db0432c3bc9a1722b49b55ef5649046

PR #2027, opened 2018-08-28
DROOLS-2923 PR #1997 test case for benchmark

tarilabs:
/cc @etirelli @baldimir @jiripetrlik forgot to include the benchmark test reported/described in the jira sorry


mariofusco (commented):
+1

PR #2039, opened 2018-08-31
DO NOT MERGE: DROOLS-2924  Factor out Unit behavior to allow pluggable backends

evacchi:
This is still partially incomplete, but it should pass all unit-related tests -- a change is required in `kie-api` to enable it though, otherwise it uses the current impl: I will open a separate PR. 

Currently I just want feedback / code review for the code. The meat is in

- UnitExecutor
- UnitSession (and subtypes -- currently RuleUnitSession and GuardedRuleUnitSession) 
- UnitScheduler

**UnitExecutor** exposes the public API. Forwards units to the scheduler, and, for `run()`, it pulls from the scheduler one UnitSession at a time, and invokes its `fireAllRules()`, until no more UnitSessions are available. Notice that the *InternalRuleUnitExecutor* interface is implemented as a wrapper around UnitExecutor (`InternalUnitExecutorDelegate`) because they do not need to share state. I think that eventually we might even get rid of this interface altogether.

**UnitSession** wraps a (stateful) session and transparently manages binding and unbinding of the Unit. To the Executor, the session is invisible. There are two types, RuleUnitSession and GuardedRuleUnitSession; the latter only differs from the former, because it holds a set of `Activation` records. Each `RuleUnitSession` also contains a collection `GuarderRuleUnitSessions` that it refers to.  

**UnitScheduler** inserts rule units and guarded rule units to an internal stack-like structure. For each RuleUnitSession, it also checks if it contains any GuardedRuleUnitSessions (notice this is recursive, because a GuardedRuleUnitSession is a subclass of RuleUnitSession); if it does, and these are `active` (they have >0 Activation records) then it schedules the Guarded session(s) as well.

I would like to review a bit how the guarding mechanism is implemented, because I think there could be a way to streamline it a little bit.

(once complete: please do squash before merge)

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
+++ b/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
@@ -55,6 +55,7 @@ public class AgendaGroupQueueImpl
 
     private InternalWorkingMemory workingMemory;
     private boolean               autoDeactivate = true;
+    private boolean               keepWhenEmpty = false;
     private Map<Long, String>     nodeInstances  = new ConcurrentHashMap<Long, String>();
 
     private volatile              boolean hasRuleFlowLister;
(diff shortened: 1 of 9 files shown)


evacchi:
Jenkins please retest this

evacchi:
Jenkins please retest this

evacchi:
this work will probably end up in the next version. let's close this PR for now

PR #2048, opened 2018-09-05
[RHDM-719] reproducer showing that guided rule templates do not work …

mariofusco:
…when a variable is used only in LHS


mariofusco:
We decided to close this issue as "won't fix".

PR #2052, opened 2018-09-06
RHPAM-1500 PMML compilation problem with weightedAverage mining operator

lanceleverich:
* Removed @Ignore from MiningModelWeightedAverageTest
* Added a method to get the Double value from the MiningSegmentWeight
* Fixed errors in the weightedAvg.mvel including
  - making sure that the correct PMML4Result is used for setting a MiningSegmentWeight
  - making sure that the correct PMML4Result gets the weighted average for its target field
  - making sure that the all possible SegmentExecutions are inserted prior to starting any SegmentExecution
  - updating the "Check Segment Can Fire" rules to update the proper SegmentExecution, instead of inserting a new one
* Fixed tests in MiningmodelTest that were broken by the above fixes


baldimir (approved):

jiripetrlik:
@lanceleverich @mariofusco 
Looks good to me. Can we cherry-pick it also to 7.11.x?

PR #2059, opened 2018-09-06
DROOLS-2823  FEEL Parser: refactor `not` unary negation and `not` function call; fix support to wildcard `?`

evacchi:
All relevant tests should be passing.

Please DO squash during merge.

 Also, [**POKER: 5** 🃏](https://www.youtube.com/watch?v=fPmruHc4S9Q)


kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerTest.java
```diff
@@ -170,7 +171,7 @@ public void test_exponentiationExpression() {
         assertThat(parseCompileEvaluate("3 ** null"), nullValue());
     }
 
-    @Test
+    @Test @Ignore("this is not valid syntax")
```
baldimir: We should either make this into a negative test, or remove it completely. Ignores get lost in the void of time usually.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
```diff
@@ -54,7 +55,7 @@
     @Test
     public void test_Dash() {
         assertThat(parseCompileEvaluate("-", 1), is(Arrays.asList(true)));
-        assertThat(parseCompileEvaluate("-, -", 1), is(Arrays.asList(true, true)));
+        // assertThat(parseCompileEvaluate("-, -", 1), is(Arrays.asList(true, true)));
```
baldimir: Commented out assert - please either delete it if not needed or change to a proper one.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
```diff
@@ -79,9 +80,9 @@ public void test_positiveUnaryTestIneq_forEQ() {
         assertThat(parseCompileEvaluate("<47, 47", 1), is(Arrays.asList(true, false)));
     }
 
-    @Test
+    @Test @Ignore("all function not() evaluate to null when arg is not boolean")
```
baldimir: Same as my previous `@Ignore` comment

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
```diff
@@ -58,21 +58,21 @@
...
-                {"10 in ( not( 5, (5+5), (20+10) ) )", Boolean.FALSE, null},
-                {"10 in ( not( 5, (20+10) ) )", Boolean.TRUE, null},
-                {"10 in ( not( >5*20 ) )", Boolean.TRUE , null},
-                {"10 in ( not( 10 ), not( 20 ) )", Boolean.TRUE , null},
-                {"10 in ( not( null, 10 ) )", Boolean.FALSE , null},
-                {"10 in ( not( 5, 10 ) )", Boolean.FALSE , null},
-                {"null in ( not( 10, null ) )", Boolean.FALSE , null},
-                {"\"Boston\" in ( not( \"Toronto\", \"Montreal\" ) )", Boolean.TRUE , null},
-                {"\"Boston\" in ( not( \"Toronto\", \"Boston\" ) )", Boolean.FALSE , null},
+                // {"10 in ( not( <5, >=20, =15, !=10 ) )", Boolean.TRUE, null},
```
baldimir: Are these not relevant anymore?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/examples/SimpleDecisionTablesTest.java
```diff
@@ -29,6 +29,7 @@
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
+@Ignore("Deprecated for DMN 1.2")
```
baldimir: I think these should not be ignored, because they are still relevant for 1.1, which we still support.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELValuesConstantsTest.java
```diff
@@ -32,7 +32,7 @@
                 { "false", Boolean.FALSE , null},
                 // dash is an unary test that always matches, so for now, returning true.
                 // have to double check to know if this is not the case
-                { "-", UnaryTest.class , null},
+                // { "-", UnaryTest.class , null},
```
baldimir: Commented out case, is it not relevant anymore?

baldimir (changes requested):
Please see my comments.

tarilabs:
@baldimir most of the comments are here temporarily to highlight for non-BC / was-wrong tests. I think we need to all-agree (also with @etirelli ) those were wrong first. I am going later today to comment on them in order to try facilitate this agreement and moving forward with this PR. Simply deleting them I don't think is the answer, maybe we should add at the beginning of the `//` the reason why are wrong --if we agree

tarilabs:
> Also, POKER: 5 :black_joker: 

yeah sure if we don't count the ace in the hole of the work done on previous sprints? =P

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/WrongConstraintsInItemDefinition.dmn
```diff
@@ -27,7 +27,7 @@
           <semantic:text>-</semantic:text>
         </semantic:inputEntry>
         <semantic:outputEntry id="_aef76a51-de39-488c-bf54-fdb3e193613b">
-          <semantic:text>-</semantic:text>
+          <semantic:text>"DEFAULT"</semantic:text>
```
tarilabs, line 30: In a Decision Table, output entry (output column), is NOT expected to contain `-`, so we changed this DMN model in the decision table output.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerVisitor.java
```diff
@@ -1145,6 +1366,9 @@ private DirectCompilerResult visitQuantExpr(Expression quantOp, IterationContext
     public DirectCompilerResult visitNameRef(FEEL_1_1Parser.NameRefContext ctx) {
         String nameRefText = ParserHelper.getOriginalText(ctx);
         Type type = scopeHelper.resolveType(nameRefText).orElse(BuiltInType.UNKNOWN);
+        if (nameRefText.equals("?")) {
```
tarilabs, line 1335: @evacchi can you please raise a JIRA to record for the potential edge-case where `myref.?`, please?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/ASTBuilderVisitor.java
```diff
@@ -95,6 +96,60 @@ public ASTBuilderVisitor(Map<String, Type> inputTypes) {
...
+    }
+
+    @Override
+    public BaseNode visitUnaryTests_negated(FEEL_1_1Parser.UnaryTests_negatedContext ctx) {
+        BaseNode name = ASTBuilderFactory.newNameRefNode( ctx.not_key(), BuiltInType.BOOLEAN ); // negating a unary tests: BOOLEAN-type anyway
+        ListNode value = (ListNode) visit( ctx.positiveUnaryTests() );
+        return ASTBuilderFactory.newListNode(ctx, Collections.singletonList(buildNotCall(ctx, name, value)))    ;
+    }
+
+    private BaseNode buildNotCall(ParserRuleContext ctx, BaseNode name, ListNode params) {
```
tarilabs: @evacchi is this a DIFF or the method has simply moved?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerTest.java
tarilabs: The syntax `not ...` without a parens `(` is not valid syntax.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
tarilabs: A unary test is either a full "don't care" operator, `-`, but cannot be a list of don't cares.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
tarilabs: A unarytest is either a negated unary test, meaning it starts with a `not(` and ends with `)`, or either by the grammar in this case mean the actual function `not` invokation here, for which would evaluate to `null, null, null`

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/examples/SimpleDecisionTablesTest.java
tarilabs: The syntax for "decision table" function in FEEL is inconsistent with the grammar.
From a FEEL expression, there is NO way to include the don't care operator `-`, hence this should be enough to demonstrate it was a mistake/bug also on DMN v1.1

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/parser/feel11/FEELParserTest.java
```diff
@@ -231,7 +231,7 @@ public void testLogicalNegation() {
         BaseNode neg = parse( inputExpression );
 
         assertThat( neg, is( instanceOf( FunctionInvocationNode.class ) ) );
-        assertThat( neg.getResultType(), is( BuiltInType.BOOLEAN ) );
+        assertThat( neg.getResultType(), is( BuiltInType.UNKNOWN ) );
```
tarilabs, line 234: at present we are not resolving type for the (result of) function invocatin

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
tarilabs: FEEL expression, and it does not exists a not function call invocation for arity=4, would result in error.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
```diff
@@ -58,21 +58,21 @@
...
+                // {"10 in ( not( <5, >=20, =10 ) )", Boolean.FALSE, null},
```
tarilabs: FEEL expression, and it does not exists a not function call invocation for arity=3, would result in error.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
```diff
@@ -58,21 +58,21 @@
...
+                // {"10 in ( not( <5 ) )", Boolean.TRUE, null},
```
tarilabs: FEEL expression, and function `not` call expects a boolean as the param, would result in error.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
```diff
@@ -58,21 +58,21 @@
...
+                // {"10 in ( not( (10..20] ) )", Boolean.TRUE, null},
+                // {"10 in ( not( 10 ) )", Boolean.FALSE, null},
+                // {"10 in ( not( 5 ) )", Boolean.TRUE, null},
```
tarilabs: all 3 aboves, FEEL expression, and function `not` call expects a boolean as the param, would result in error.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
```diff
@@ -58,21 +58,21 @@
...
+                // {"10 in ( not( 5, (5+5), (20+10) ) )", Boolean.FALSE, null},
+                // {"10 in ( not( 5, (20+10) ) )", Boolean.TRUE, null},
+                // {"10 in ( not( >5*20 ) )", Boolean.TRUE , null},
+                // {"10 in ( not( 10 ), not( 20 ) )", Boolean.TRUE , null},
+                // {"10 in ( not( null, 10 ) )", Boolean.FALSE , null},
+                // {"10 in ( not( 5, 10 ) )", Boolean.FALSE , null},
+                // {"null in ( not( 10, null ) )", Boolean.FALSE , null},
+                // {"\"Boston\" in ( not( \"Toronto\", \"Montreal\" ) )", Boolean.TRUE , null},
+                // {"\"Boston\" in ( not( \"Toronto\", \"Boston\" ) )", Boolean.FALSE , null},
```
tarilabs: the rest of the commented-out expressions, apply the same of the 2 comments above, either does not exist for n-arity the function call of `not`, or either for 1-arity the function invocation of `not` expexts a boolean as a parameter.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELOperatorsTest.java
```diff
@@ -31,7 +31,7 @@
                 { "not( false )", Boolean.TRUE , null},
                 { "not( 10 = 3 )", Boolean.TRUE , null},
                 { "not(list contains([1,2,3,4,5,6], 3))", Boolean.FALSE , null},
-                { "not( \"foo\" )", UnaryTest.class , null},
+                { "not( \"foo\" )", null, FEELEvent.Severity.ERROR},
```
tarilabs, line 34: FEEL expression, and function not call expects a boolean as the param, would result in error.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELOperatorsTest.java
```diff
@@ -54,7 +54,7 @@
                 { "10 in ( > 50, < 5 )", Boolean.FALSE , null},
                 { "10 in ( > 5, < -40 )", Boolean.TRUE , null},
                 { "null in ( > 20, null )", Boolean.TRUE , null},
-                { "null in -", Boolean.TRUE , null},
+                { "null in -", null, FEELEvent.Severity.ERROR},
```
tarilabs, line 57: Invalid syntax. From a FEEL expression (grammar rule1) there is no way to reach the "don't care" operator `-`

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELValuesConstantsTest.java
tarilabs: Invalid syntax. From a FEEL expression (grammar rule1) there is no way to reach the "don't care" operator `-`

tarilabs (approved):
@etirelli @baldimir with this changes, the inconsistency of the `not` call is resolved on the existing FEEL grammar. This would imply some of the old tests are no longer valid if we agree on the rationale I tried to summarize at my best below.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerTest.java
evacchi: we can do either @baldimir I have left these commented out / @Ignore'd tests for the sole purpose of reviewing what we are removing with this patch. Will *actually* remove once every party agrees with the removals

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/ASTBuilderVisitor.java
evacchi: moved. I had previously deleted that, then restored it

tarilabs: @evacchi do you think possible to restore in original position as many methods as possible here, in order to keep DIFF only for behavior change, please? Not mandatory, but would be really nice from history preserving. thx.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerTest.java
baldimir: What will happen if a user writes it like this? I mean we could leave it as a negative test, to test that we behave well in this kind of situation (test we raise a meaningful error or return null or something...).

kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
```diff
@@ -436,6 +462,10 @@ between_key
     ;
 
 not_key
+    : Not_Key
+    ;
+
+Not_Key
```
etirelli: why do we need a rule for "not_key" and another rule for "Not_Key"?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSemanticMappings.java
```diff
@@ -150,7 +150,7 @@ public static Boolean notExists(
 
         for (Object test : tests) {
             Boolean r = applyUnaryTest(ctx, test, target);
-            if (r == null || r) return false;
+            if (r != null && r) return false;
```
etirelli: just double checking that this is really the intent? change looks odd..

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
etirelli: we should just remove the test completely instead of commenting.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/codegen/feel11/DirectCompilerUnaryTestsTest.java
etirelli: either remove the test or make it into a negative test.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/examples/SimpleDecisionTablesTest.java
etirelli: Yeah, this was a "hack" I implemented to fix a bug in the DMN 1.1 spec, so we can just straight out remove.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELExpressionsTest.java
etirelli: Just remove all the commented out tests

kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
evacchi: so, unless there is real motivation for that (it might, I don't know) lexer rules should be capitalized, as to signify they are done at the lexing phase. Otherwise, they are resolved at the parsing level, which might not do what we expect. at this line I needed to refer to the lexing token Not
https://github.com/kiegroup/drools/pull/2059/files/c82693e41ce4713ca466fb6947e4a046e697979f#diff-29688d5ce72a1273e981bcaf48314708R366
However I had to re-introduce the non-lexing rule `not_key` so that the existing "visitors" would not break

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSemanticMappings.java
evacchi: I think it had to do with getting right the logic for short-circuiting on error. But we solved that somewhere else, so we can revert.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/ASTBuilderVisitor.java
evacchi: doing it right now, just a minor notice: the method, as restored, does not really make sense to be there that much (it's now basically unrelated to the surrounding code)

evacchi:
I think all comments have been addressed :)

baldimir (approved):
Thanks!

kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
etirelli: Ok, it would be good then to move the rule to the end of the file with the other lexer rules to avoid confusion. But change is fine. Thanks for the explanation.

etirelli (approved):

tarilabs (approved):

kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
evacchi: done !

PR #2060, opened 2018-09-10
DROOLS-2972 Don't use Collections.emptyList() as a class provided to PMML template

baldimir:
- This fixes the failing Scorecard tests

Collections.emptyList() returns an EmptyList instance, which is then checked with reflection during RuleUnit generation process - mvel tries to call size() on the EmptyList through reflection, which fails, because EmptyList is an internal private class of Collections class (so it is not accessible with reflection). 

@jiripetrlik @lanceleverich please be aware that the randomness of the failures indicates that there might be another bug in the RuleUnit generation process, because it implies that EmptyList class was not generated all the time for the RuleUnit. Mentioning this because it could be a good area to extend test coverage, so we find out what is going on in case of the randomness.

### DIFF
--- a/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/PMML4Compiler.java
+++ b/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/PMML4Compiler.java
@@ -546,20 +546,6 @@ public class PMML4Compiler {
         return javaClasses;
     }
 
-    public List<PMMLResource> precompile(String fileName, ClassLoader classLoader, KieBaseModel rootKieBaseModel) {
-        InputStream is = getInputStreamByFileName(fileName);
-        List<PMMLResource> resources = null;
-        if (is != null) {
-            try {
-                resources = precompile(is, classLoader, rootKieBaseModel);
-            } catch (Exception e) {
-                PMMLError err = new PMMLError("Unable to retrieve pre-compiled resources for PMML: " + e.getMessage());
-                this.results.add(err);
-            }
-        }
-        return (resources != null) ? resources : new ArrayList<>();
-    }
-
     public List<PMMLResource> precompile(InputStream stream, ClassLoader classLoader, KieBaseModel rootKieBaseModel) {
         List<PMMLResource> resources = new ArrayList<>();
         KieServices services = KieServices.Factory.get();
--- a/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/AbstractModel.java
+++ b/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/AbstractModel.java
@@ -251,7 +251,11 @@ public abstract class AbstractModel<T> implements PMML4Model {
                     .distinct()
                     .collect(Collectors.toList());
         }
-        return list != null ? list : new ArrayList<>();
+        // DO NOT USE Collections.emptyList() here!!! emptyList() returns an EmptyList instance
+        // which is an internal class from Collections class, therefore not accessible
+        // by reflection - when mvel parses a RuleUnit template and there is
+        // a size call, it tries to invoke the method using reflection
+        return list != null ? list : new ArrayList<>(0);
     }
 
     public List<ExternalBeanRef> getExternalMiningFields() {
--- a/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/PMMLDataField.java
+++ b/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/PMMLDataField.java
@@ -15,48 +15,44 @@
  */
 package org.kie.pmml.pmml_4_2.model;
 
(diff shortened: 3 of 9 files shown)


baldimir:
The plan is to run 10 Jenkins runs without any Scorecard test fail, 5 to go. (I will keep updating this comment)

baldimir:
Fail not related.

baldimir:
jenkins retest this

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/PMML4Compiler.java
```diff
@@ -557,7 +557,7 @@ private InputStream getInputStreamByFileName(String fileName) {
                 this.results.add(err);
             }
         }
-        return (resources != null) ? resources : Collections.emptyList();
+        return (resources != null) ? resources : new ArrayList<>();
```
triceo: If it's so important to not have `Collections.emptyList()` here, perhaps there should be a comment saying so? Otherwise I can just see some smart code reviewer "fixing this" in the future.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/PMMLDataField.java
```diff
@@ -84,7 +85,7 @@ public String getCompactUpperCaseName() {
     	if (dataDictionaryField != null) {
     		return dataDictionaryField.getIntervals();
     	}
-    	return Collections.emptyList();
+    	return new ArrayList<>();
```
triceo: And while we're doing that, let's size the list to `0` so that we don't waste memory needlessly.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/AbstractModel.java
```diff
@@ -251,7 +251,7 @@ public String getModelPackageName() {
                     .distinct()
                     .collect(Collectors.toList());
         }
-        return list != null ? list : Collections.emptyList();
+        return list != null ? list : new ArrayList<>();
```
triceo: In fact, thinking about it, perhaps there should be a utility method that does that, with the proper comment?

triceo (changes requested):
This is a weird quirk. IMHO Should be centralized and documented.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/AbstractModel.java
lanceleverich: The fix proposed here may address some issue, but it does not address the issue that I have seen with drools-scorecards; mainly that there are some tests that periodically fail. I have applied the fixes from this PR to my local machine and I still get the intermittent errors in the ScorecardReasonCodeTest. Please see https://issues.jboss.org/browse/DROOLS-2592 to see my latest notes and results from my testing.

baldimir: Yes @lanceleverich, this fix addresses just 25 scorecards tests that were failing at the same time. I can take a look on that other issue in another PR if needed.

baldimir: @triceo I don't think that a utility method for a oneliner is necessary.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/PMML4Compiler.java
baldimir: Removed the whole method, because I found out it was not used anywhere.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/PMMLDataField.java
baldimir: Removed the whole method, because I found out it was not used anywhere.

kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/model/AbstractModel.java
baldimir: @triceo comment and initial size added.

baldimir:
jenkins retest this

triceo (approved):

baldimir:
@lanceleverich just a note for future work - we need to cleanup the PMML code, because there are e.g. many unused methods. Would you mind if I do it when having a bit of time?

jiripetrlik:
@baldimir 
+1 for cleanup of unused methods. It would be really useful.

jiripetrlik (approved):

baldimir:
jenkins retest this

baldimir:
jenkins retest this

baldimir:
jenkins retest this

PR #2061, opened 2018-09-10
[RHPAM-2] NullPointerException during transaction rollback when using CMT (#2015)

elguardian:
if it is not the transaction owner we try to rely on the transaction registry
and set the current tx associated as a rollback


baldimir (approved):

mswiderski:
this was not approved as far as I know for 7.1 so closing

PR #2067, opened 2018-09-12
DROOLS-2592 PMML - Fix race conditions caused by corner cases of used "after" and "before" operators

baldimir:
- Running ScorecardReasonCodeTest, it happened sometimes, that the old and new fact instances handling ReasonCode result value had the same timestamp, therefore simple "after" used in the PMML rules wasn't enough
- I found out that there is also "before" operator used at three places, so I applied similar fix also there. 

@lanceleverich @jiripetrlik @mariofusco could you please check, if the rules are correct? These fixes should stabilize all remaining unstable PMML tests.


baldimir:
jenkins retest this

baldimir:
Retesting just to be sure about race conditions.

baldimir:
jenkins retest this

PR #2069, opened 2018-09-13
RHPAM-1509 [7.11.x] Mining model modelChain mode does not compile

lanceleverich:
* Fixed an error, in the PMML4Compiler, that was caused when instead of creating a copy of a DataField object the object was being updated
* Fixed problems in the test data...
  - Models need to have a modelName attribute
  - A target element, in a MiningSchema, must have a corresponding DataField defined


jiripetrlik (approved):
Looks good to me.

PR #2072, opened 2018-09-17
[RHPAM-1184] URLDecoder: Illegal hex characters in escape (%) patternwhen a rule has a name with "%" 

elguardian:
moving decoding the file name to the EJC compilation unit adapter so memory file system
is not inconsistent anymore regarding the names


elguardian:
ok to test

baldimir:
@mariofusco All failures are related to the kie-ci failures with the offline feature, including that DMN test (I tested that locally with the offline "fix" and it passes).

baldimir (approved):

PR #2078, opened 2018-09-21
[findbugs] Fix array compare in executable model and other fixes

baldimir:
@mariofusco please review. 

The globals.remove() thing is a type problem. globals is a collection of FactData, so the types don't match in the cases I changed.


baldimir:
@tarilabs please also review.

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/DMNValidatorImpl.java
```diff
@@ -235,7 +237,7 @@ public ValidatorBuilderImpl(DMNValidatorImpl dmnValidatorImpl, Validation[] opti
                                 results.addAll(model.getMessages());
                                 otherModel_DMNModels.add(model);
                             } else {
-                                Objects.requireNonNull(model);
+                                throw new IllegalStateException("Compiled model is null!");
```
tarilabs, line 240: Wouldn't it be better then to do something like: https://github.com/kiegroup/drools/blob/e5d8ec4416a34679954208fc8c15dde4746f1777/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java#L97

What do you think @baldimir

PR #2081, opened 2018-09-21
DROOLS-3012 DMN API to evaluate direct-dependency imported Decisions

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
@@ -16,6 +16,8 @@
 
 package org.kie.dmn.core.imports;
 
+import java.util.function.Function;
+
 import org.junit.Test;
 import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNModel;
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/imports/ImportsTest.java
```diff
@@ -277,5 +279,40 @@ public void testImportingID() {
...
+        testAllowDMNAPItoEvaluateDirectDependencyImportedDecisions_evaluateResultsAndCheck(runtime,
+                                                                                           dmnModel,
+                                                                                           context -> runtime.evaluateById(dmnModel,
+                                                                                                                           context,
+                                                                                                                           "http://www.trisotech.com/dmn/definitions/_ae5b3c17-1ac3-4e1d-b4f9-2cf861aec6d9#_f7fdaec4-d669-4797-b3b4-12b860de2eb5",
+                                                                                                                           "_96df766e-23e1-4aa6-9d5d-545fbe2f1e23"));
+    }
+
+    private void testAllowDMNAPItoEvaluateDirectDependencyImportedDecisions_evaluateResultsAndCheck(DMNRuntime runtime, final DMNModel dmnModel, Function<DMNContext, DMNResult> fn) {
```
triceo, line 305: A quick tip: when you have to create a separate method that will be reused multiple times within a single test, just with different arguments each time, it's a sign that you should consider splitting the test in two. And if the method were to be reused even more often, then perhaps even a parameterized test is in order.

The biggest benefit being that failures in one invocation will not prevent you from seeing failures in the other.

tarilabs: Thank you @triceo for your feedback, but I am getting a confused here because if I mix your comment also with code re-use generally from @baldimir about assertions groups, it would end up in 4 methods here for a total of pragmatically 2 test?

triceo: I'm not sure I understand, let me try. 

I don't have a problem with having the 3 assertions below in one method - they are all related to the same functionality, and can not be further split by parameterization. So, in my opinion, it's 2 tests which will each call this method once.

That said, we're probably over-thinking this already. :-)

baldimir (approved):

PR #2085, opened 2018-09-24
DROOLS-3026 DMN model get[Decision|Input]by[Name|Id] consistency

tarilabs:


baldimir (approved):
Approving, thanks for this, nice simplification.

PR #2096, opened 2018-10-04
DROOLS-3070 Ensure DMN tests run parameterized with executable model

baldimir:
@tarilabs @etirelli @evacchi please review. 

Sorry for larger PR, but I did also a cleanup of the testing code. 

- Adds parameters to run tests with executable model to tests that do evaluation. 
- Fixes a problem that left side of "add" expression like "a+2" was not cast to BigDecimal

I separated the cleanup to separate commits, so when reviewing, you can check the changes by taking a look at individual commits.


kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/KieExtendedBaseFEELTest.java
```diff
@@ -1,81 +0,0 @@
...
-import static org.hamcrest.Matchers.nullValue;
-import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-
-@RunWith(Parameterized.class)
-public abstract class KieExtendedBaseFEELTest {
```
tarilabs, line 41: Sorry @baldimir but can we discuss this removal, please?
This had the advantage of make it super easy to distinguish test which belonged to one of the extended profile.

baldimir: I can create a child of BaseFEELTest that will just override the addAdditionalParams method. Would that be ok?

tarilabs: I am troubled that we ended up adding more parameters, when I thought we agreed to have it refactored to minimise the number of parameters :sweat_smile: 

I suggest the following way forward; given this change (which I believe was actually targeting code-reuse just for the method of test) is a bit difficult to extract from current PR, we go forward as-is. But during refactor of the tests, I would suggest we limit parameters only to different kind of execution configuration eg: only exec model, or interpreted. Data not as parameter. Test which are based on an extension does not need a parameter. What do you think?

baldimir: +Infinity :)

tarilabs (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompilerBytecodeLoader.java
```diff
@@ -60,12 +60,15 @@ public TemplateLoader(ClassLoader parent) {
...
         public Class<?> load(MemoryFileSystem pStore, String string) {
-            Map<String, Class<?>> loaded = new HashMap<>();
+            Class<?> loadedClass = null;
             for (Entry<String, byte[]> kv : pStore.getMap().entrySet() ) {
-                String className = kv.getKey().substring(0, kv.getKey().lastIndexOf(".class")).replaceAll("/", ".");
-                loaded.put(className, defineClass(className, kv.getValue(), 0, kv.getValue().length));
+                final String className = kv.getKey().substring(0, kv.getKey().lastIndexOf(".class")).replaceAll("/", ".");
+                final Class<?> definedClass = defineClass(className, kv.getValue(), 0, kv.getValue().length);
+                if (string.equals(className)) {
+                    loadedClass = definedClass;
```
mariofusco, line 68: once the class is found it could immediately return it there instead of storing it in a variable and continuing the for-loop

PR #2098, opened 2018-10-05
[DROOLS-3080] SpreadsheetCompiler generates wrong LHS order

tkobayas:
- Test case


drools-decisiontables/src/test/java/org/drools/decisiontable/SpreadsheetCompilerUnitTest.java
```diff
@@ -742,4 +742,20 @@ public void checkLhsBuilderFixValue() {
...
+
+        System.out.println(drl);
+        assertNotNull( drl );
+        int firstLine = drl.indexOf( "accumulate(Person(name == \"John\", $a : age); $max:max($a))" );
+        int secondLine = drl.indexOf( "$p:Person(name == \"John\", age == $max)" );
+
+        assertTrue( firstLine > -1 );
+        assertTrue( secondLine > -1 );
+        assertTrue( "Lhs order is wrong. firstLineIndex = " + firstLine + ", secondLineIndex = " + secondLine , firstLine < secondLine );
+    }
```
jomarko, line 756: We could shorten the test slightly if we use `assertj` library see the code below, would you mind to incorporate?
```java
@Test
    public void testLhsOrder() {
        // DROOLS-3080
        final SpreadsheetCompiler converter = new SpreadsheetCompiler();
        String drl = converter.compile("/data/LhsOrder.xls",
                                       InputType.XLS);

        Assertions.assertThat(drl.split("\n"))
                .as("Lhs order is wrong")
                .containsSequence("accumulate(Person(name == \"John\", $a : age); $max:max($a))",
                                  "$p:Person(name == \"John\", age == $max)");
    }
```

jomarko (commented):
@tkobayas thanks for the reproducer, one comment to code from my site.

drools-decisiontables/src/test/java/org/drools/decisiontable/SpreadsheetCompilerUnitTest.java
tkobayas: @jomarko Thanks! Updated.

jomarko (approved):
Approving, but do not merge until we have fix or add ignore annotation and mention in jira we need to unignore once fixed.

mariofusco:
Replaced and fixed by https://github.com/kiegroup/drools/pull/2101

PR #2099, opened 2018-10-05
[DROOLS-3076] KieSession pool

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerSessionsPoolImpl.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.drools.core.impl.AbstratcKieSessionsPool;
+import org.drools.core.impl.StatefulSessionPool;
+import org.drools.core.impl.StatelessKnowledgeSessionImpl;
+import org.kie.api.runtime.KieContainerSessionsPool;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.KieSessionConfiguration;
+import org.kie.api.runtime.StatelessKieSession;
+
+public class KieContainerSessionsPoolImpl extends AbstratcKieSessionsPool implements KieContainerSessionsPool {
```
baldimir: Typo here - Abstratc...

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+public class SessionsPoolTest {
+
+    @Test
+    public void testKieSessionsPool() {
+        KieContainerSessionsPool pool = getKieContainer().newKieSessionsPool( 1 );
+
+        KieSession ksession = pool.newKieSession();
+        checkKieSession( ksession );
+        ksession.dispose();
```
baldimir: I know this is simple use, but please add the ksession.dispose() and pool.shutdown() into a finally block.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
```diff
@@ -0,0 +1,162 @@
...
+        assertSame( ksession, ksession2 );
+        assertNull( ksession2.getGlobal( "list" ) );
+        checkKieSession( ksession2 );
+
+        pool.shutdown();
+
+        try {
+            ksession.insert( "test3" );
+            fail("after pool shutdown all sessions created from it should be disposed");
+        } catch (Exception e) { }
```
baldimir: These exception catch blocks should catch specific Exception type, because this one could catch some unexpected error that we would like to take a look.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
```diff
@@ -0,0 +1,162 @@
...
+            fail("after pool shutdown it shouldn't be possible to get sessions from it");
+        } catch (Exception e) { }
+    }
+
+    @Test
+    public void testKieSessionsPoolInMultithreadEnv() throws InterruptedException, ExecutionException {
+        KieContainerSessionsPool pool = getKieContainer().newKieSessionsPool( 4 );
+
+        final int THREAD_NR = 10;
+        final ExecutorService executor = Executors.newFixedThreadPool(THREAD_NR, r -> {
```
baldimir: Please shutdown the executor in a finally block.

drools-core/src/main/java/org/drools/core/impl/AbstratcKieSessionsPool.java
```diff
@@ -0,0 +1,65 @@
...
+package org.drools.core.impl;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.kie.api.runtime.Environment;
+import org.kie.api.runtime.KieSessionConfiguration;
+import org.kie.api.runtime.KieSessionsPool;
+
+public abstract class AbstratcKieSessionsPool implements KieSessionsPool {
```
baldimir: Typo...Abstratc

drools-core/src/main/java/org/drools/core/impl/StatelessKnowledgeSessionImpl.java
```diff
@@ -59,56 +59,75 @@
...
+        this.pool = pool;
+        wmCreated = new AtomicLong(1);
     }
 
     public InternalKnowledgeBase getKnowledgeBase() {
         return this.kBase;
     }
 
+    private StatefulKnowledgeSession newWorkingMemory() {
+        StatefulKnowledgeSessionImpl ksession = pool != null ? pool.get() : createWorkingMemory();
```
baldimir, line 105: Just thinking if we shouldn't have a general pool holder somewhere (I mean one place for holding the pool(s) reference), so we don't need to hold the reference to the pool in the session impl. But not 100% sure.

drools-core/src/main/java/org/drools/core/util/ScalablePool.java
```diff
@@ -0,0 +1,94 @@
...
+    private final T[] fixedSizePool;
+    private volatile int cursor;
+
+    private java.util.Queue<T> dynamicPool;
+
+    private final Supplier<? extends T> supplier;
+    private final Consumer<? super T> resetter;
+    private final Consumer<? super T> disposer;
+
+    public ScalablePool( int initialSize, Supplier<? extends T> supplier, Consumer<? super T> resetter, Consumer<? super T> disposer ) {
```
baldimir, line 38: This could be unit-tested so no-one accidentally changes this in the future.

baldimir (changes requested):
Please see comments inline.

baldimir:
jenkins retest this

PR #2100, opened 2018-10-08
Test Cleanup - Call dispose after kieSession is not needed

jomarko:
@baldimir @mariofusco please have a look, this is some cleanup of tests I went trough during #1920 investigation so no jira filed for this.


baldimir (approved):

baldimir (changes requested):
Please put the dispose into try-finally block.

jomarko:
@baldimir thanks for review, I added try-finally block as separate commit.

baldimir (approved):
Thanks!

PR #2101, opened 2018-10-08
[DROOLS-3080] output lhs patterns generated from decision tables in r…

mariofusco:
…ight order


baldimir (approved):

baldimir:
Good for me if checks are green.

PR #2103, opened 2018-10-08
[BAPL-794] allow to set Calendars on a KieSession in a declartive way

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -549,6 +551,17 @@ private void registerLoggers(KieSessionModelImpl kSessionModel, KieRuntimeEventM
         }
     }
 
+    private void registerCalendars(KieSessionModelImpl kSessionModel, KieSession kSession) {
+        for (Map.Entry<String, String> entry : kSessionModel.getCalendars().entrySet()) {
+            try {
+                Calendar calendar = (Calendar) getClassLoader().loadClass( entry.getValue() ).newInstance();
+                kSession.getCalendars().set( entry.getKey(), calendar );
+            } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+                log.error( "Cannot instance calendar " + entry.getKey(), e );
```
baldimir, line 560: Just a nitpicking comment for a doublecheck - the exception is just logged here. Is this intended?

baldimir (approved):
Approving, however with 1 question.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
mariofusco: @baldimir I believe that just logging here is fine: if the calendar is not in use I don't want to make the whole ksession creation to fail and if it is it will fail later when it will try to use it.

PR #2107, opened 2018-10-11
[DROOLS-3088] FEEL Compiler: refactor to AST-based tree walk

evacchi:
Current compiler backend is harder to maintain because it closely follows the parse tree that ANTLR generates. We could improve the tree that ANTLR generates, but luckily a lot of work already has been done to build an AST, and we can build a lot on top of that. Reusing the separate AST representation allows us

1) to better represent compilation stages 
2) to manage less cases, because the number of AST node is much smaller the number of parse tree node (the visitor is much smaller!)
3) with less cases to manage, it is also easier to separate compilation concerns and to further improve by adding more processing phases: e.g. a name resolution phase and a type-checking phase (and possibly different compiler backends)

In this PR:

- a minor refactoring of the grammar, where we only use named tokens in the lexer, and remove inline token definitions (`'my-token'`) in the grammar, so that lexing and parsing are better distinguished
- we add a new AST node (used also in the interpreter) to deal with the root of the UnaryTests grammar (used in Decision Tables). This node supports being flagged as Negated, to better distinguish the `not` case in Decision Tables (different from the `not()` function)  
- we better support the question mark `?` special name (here called a `Wildcard`) -- now it cannot be mistaken as a wildcard if it occurs within a qualified expression (`foo.?.bar` is not a wildcard, `?.foo.bar` is) 
- we add and implement a Visitor for the AST nodes
- we support compiling the AST visitor down to a `DirectCompilerResult`, i.e. the same type returned by current compiler -- (`ASTCompilerVisitor`)
- we support AST rewriting (`ASTUnaryTestTransform`) to reduce an AST for unary tests, to a "canonical" form that can be processed by the simple AST compiler (e.g.: [1, 2, 3] is rewritten to [=1, =2, =3])

All tests are currently passing including TCK's.

Further work will be done in a follow-up PR, but it's unrelated to the compiler backend itself, rather to the way components interact at higher-level ( compiler vs. interpreter vs. public APIs vs. internal APIs. see: https://issues.jboss.org/browse/DROOLS-3112 )

@tarilabs for review, cc @etirelli 
@baldimir can you take a look and run the benchmarks on this PR so that we are sure we are not regressing

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
+++ b/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
@@ -377,9 +377,9 @@ nameRef
        )  nameRefOtherToken*
     ;
 
-nameRefOtherToken // added some special cases here: we should rework a bit the lexing part, though --ev
-    : { helper.followUp( _input.LT(1), _localctx==null ) }? //~('('|')'|'['|']'|'{'|'}'|'>'|'<'|'='|'!'|'/'|'*'|'in'|',')
-        ~(LPAREN|RPAREN|LBRACK|RBRACK|LBRACE|RBRACE|LT|GT|EQUAL|DIV|MUL|IN|COMMA)
+nameRefOtherToken
+    : { helper.followUp( _input.LT(1), _localctx==null ) }?
+        ~(LPAREN|RPAREN|LBRACK|RBRACK|LBRACE|RBRACE|LT|GT|EQUAL|BANG|DIV|MUL|IN|COMMA)
     ;
 
 /********************************
@@ -710,10 +710,14 @@ ADD : '+';
 SUB : '-';
 MUL : '*';
 DIV : '/';
+BANG
+    : '!'
+    ;
 
 NOT
     : 'not'
     ;
+
 
 Identifier
 	:	JavaLetter JavaLetterOrDigit*
--- a/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
+++ b/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
@@ -56,23 +56,23 @@ textualExpression
 
 // #41
 parameters
-    : '(' ')'                       #parametersEmpty
-    | '(' namedParameters ')'       #parametersNamed
-    | '(' positionalParameters ')'  #parametersPositional
+    : LPAREN RPAREN                       #parametersEmpty
+    | LPAREN namedParameters RPAREN       #parametersNamed
+    | LPAREN positionalParameters RPAREN  #parametersPositional
     ;
 
 // #42 #43
 namedParameters
-    : namedParameter (',' namedParameter)*
+    : namedParameter (COMMA namedParameter)*
     ;
 
 namedParameter
-    : name=nameDefinition ':' value=expression
+    : name=nameDefinition COLON value=expression
     ;
 
 // #44
 positionalParameters
-    : expression ( ',' expression )*
+    : expression ( COMMA expression )*
     ;
 
 // #46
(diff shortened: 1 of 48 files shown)


kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
```diff
@@ -372,115 +378,117 @@ nameRef
     ;
 
 nameRefOtherToken // added some special cases here: we should rework a bit the lexing part, though --ev
-    : { helper.followUp( _input.LT(1), _localctx==null ) }? ~('('|')'|'['|']'|'{'|'}'|'>'|'<'|'='|'!'|'/'|'*'|'in'|',')
+    : { helper.followUp( _input.LT(1), _localctx==null ) }? //~('('|')'|'['|']'|'{'|'}'|'>'|'<'|'='|'!'|'/'|'*'|'in'|',')
+        ~(LPAREN|RPAREN|LBRACK|RBRACK|LBRACE|RBRACE|LT|GT|EQUAL|DIV|MUL|IN|COMMA)
```
tarilabs: missing `'!'` case?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/ASTUnaryTestTransform.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Override
+    public UnaryTestSubexpr visit(UnaryTestListNode n) {
+        List<BaseNode> collect =
+                new ArrayList<>();
+        for (BaseNode e : n.getElements()) {
+            UnaryTestSubexpr accept = e.accept(this);
+            if (accept.isWildcard()) {
+                collect.add(asWildcard(accept.node));
```
tarilabs: didn't you want to in-line this `asWildcard()` call ?

tarilabs (approved):

kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
evacchi: good catch

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/Functions.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (java != null) {
+
+            String className = java.get("class");
+            String methodSignature = java.get("method signature");
+            if (className == null || methodSignature == null) {
+                throw new FEELCompilationError(Msg.createMessage(Msg.UNABLE_TO_FIND_EXTERNAL_FUNCTION_AS_DEFINED_BY, methodSignature));
+            }
+            Expression methodCallExpr = FunctionDefs.asMethodCall(className, methodSignature, paramNames);
+//            DirectCompilerResult parameters = visit(ctx.formalParameters());
```
baldimir: Nitpicking: Commented out code, please delete or resolve.

baldimir (approved):
Just one nitpicking comment. Other than that, the new classes could be covered with unit tests. But we have existing coverage, so approving. Take that as an idea for future work (that we can share between us:)).

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/Functions.java
evacchi: fix pushed

evacchi:
> Other than that, the new classes could be covered with unit tests. But we have existing coverage, so approving. Take that as an idea for future work

Good point, thanks a lot for feedback

manstis:
It'd be good to add licenses to the new files... (a code review is beyond my ability, but was being nosy).

evacchi:
Good point @manstis I can't seem to be able to configure my editor properly 😅will do

evacchi:
will wait for the release branch to be cut before we merge

evacchi:
jenkins please retest this

evacchi:
jenkins please retest this

evacchi:
jenkins please retest this

evacchi:
come on Jenkins...

Jenkins please retest this

evacchi:
*I should change naming convention for branches* :angel: 

Jenkins please retest this

evacchi:
@tarilabs shall we merge this?

PR #2118, opened 2018-10-24
[DROOLS-3181] fix VirtualFileSystem discovery after changes in eap 6.…

mariofusco:
…4.21


baldimir:
I tested the fix with Wildfly 14.0.1 and it works.

baldimir (approved):

PR #2119, opened 2018-10-24
DROOLS-3180 Split RuleUnitRegistry to two separate classes

baldimir:
- Splits RuleUnitRegistry to RuleUnitDescriptionRegistry and RuleUnitDescriptionLoader
- I had to change methods in InternalKnowledgePackage and InternalKnowledgeBase interfaces, therefore I did also a naming cleanup - there were methods named as if they handled rule units, however they handled rule unit descriptions.

@mariofusco could you please review? I will run a downstream build with this because of the interface changes.


baldimir:

baldimir:
The downstream fails look unrelated.

PR #2124, opened 2018-10-31
Enable commands to return value using outIdentifier

jiripetrlik:


jiripetrlik:
@mariofusco Hello, can you please merge?

PR #2125, opened 2018-10-31
DROOLS-3217 DMN compilation via kie-maven-plugin wrong lookup at runtime

tarilabs:
/cc @baldimir @etirelli


baldimir (approved):
Works with my KJar. Approving. Thanks for the fix!

PR #2126, opened 2018-11-01
RHDM-766 PMML running on kie-server with REST API

lanceleverich:
Not quite ready for merge, but please start reviewing


lanceleverich:
Updated the approach to dealing with output. Now uses a template that generates individual java classes for each Output element and/or MiningField that is of usage PREDICTED or TARGET. This has caused several unit tests to break, so the PR should not be merged.

lanceleverich:
No longer the preferred PR

PR #2135, opened 2018-11-03
DROOLS-3228: Rules don't produce the correct result anymore after a session reset

liebharc:
Hello,

I did some further investigations on the issue reported with https://issues.jboss.org/browse/DROOLS-3228 . For the main issue it seems that just the order _memory.getSegmentMemory()_ and  _memory.reset()_ must be maintained (as it was with Drools 7.12.0.Final). 

I've also found another issue with session reset in that a sequential session can't be used anymore after a reset as it refuses to add new elements to the agend group queue. 

Executing https://github.com/liebharc/JavaRules/blob/master/src/test/java/com/github/liebharc/JavaRules/UnitTestBase.java should show all the described issues as test failures.

There doesn't seem to be specific test cases for AgendaGroupQueueImpl and ConcurrentNodeMemories - or at least I couldn't find them. 

I'm looking forward to hear your thoughts.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
+++ b/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
@@ -167,7 +167,7 @@ public class AgendaGroupQueueImpl
     }
 
     public void add(final Activation activation) {
-        if ( lastRemoved != null ) {
+        if ( lastRemoved != null && !sequential) {
             // this will only be set if sequential. Do not add Match's that are higher in salience + load order than the lastRemoved (fired)
             if ( lastRemoved == activation || lastRemoved  == visited || PhreakConflictResolver.doCompare( lastRemoved, activation ) < 0 ) {
                 return;
--- a/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
+++ b/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
@@ -1122,6 +1122,8 @@ public class StatefulKnowledgeSessionImpl extends AbstractRuntime
         this.processRuntime = null;
 
         this.initialFactHandle = initInitialFact(kBase, null);
+
+        this.alive = true;
     }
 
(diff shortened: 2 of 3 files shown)


ghost:

ghost:

ghost:

ghost:

mariofusco:
ok to test

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -1122,6 +1122,8 @@ public void reset() {
         this.processRuntime = null;
 
         this.initialFactHandle = initInitialFact(kBase, null);
+
+        this.alive = true;
```
mariofusco, line 1126: I cannot do this. If I'd do this it would be possible to keep using a KieSession created from a pool even after it has been disposed.

liebharc: Ah that makes sense. In that case we have to update our code to use the session pools. I will update the PR and remove this part of the change. 

I see that there are test failures but can't access the links (I double checked to use the public link). Is this the reason for the failure or is there a second issue?

drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
```diff
@@ -167,7 +167,7 @@ public int size() {
     }
 
     public void add(final Activation activation) {
-        if ( lastRemoved != null ) {
+        if ( lastRemoved != null && !sequential) {
```
mariofusco, line 170: This is correcting the symptoms not cause. The actual fix is setting lastRemoved to null in the reset() method.

mariofusco:
I made another pull request fixing both the sequential mode and the memories reset (this second fix is the one you suggested thanks a lot for it) and having an unit test for each of these problems to avoid future regressions.

Replaced by https://github.com/kiegroup/drools/pull/2136

liebharc:
Thanks!

PR #2136, opened 2018-11-05
[DROOLS-3228] fix node memories and agenda queue reset for sessions pool

mariofusco:


baldimir (approved):
Approving when green.

PR #2139, opened 2018-11-08
Fix passing parameters to various DB drivers

rsynek:
@MarianMacik please review, will test with DBs meanwhile


rsynek:
The compile downstream job seems to be failing on enforcer due to maven version running the build being too low. In the job config, the version is not specified, which is probably the root cause.

The changes were tested on DB matrix.

rsynek:

MarianMacik (approved):
Thanks a lot!

baldimir (approved):

PR #2140, opened 2018-11-08
DROOLS-3279 DMN DRGElement typeRef to allow FEEL Any

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
@@ -587,6 +587,15 @@ public class DMNCompilerImpl implements DMNCompiler {
                     }
                 }
             }
+            if (!(localElement instanceof ItemDefinition)) {
+                // If this is an InformationItem, e.g. is a variable of a DRGElement, it allows all types in the language.
+                // if instead it was an ItemDefinition, we only limited to DMNTypeRegistryV12.ITEMDEF_TYPEREF_FEEL_BUILTIN as per specification chapter, and we are not entering in this branch.
+                for (String biName : BuiltInType.UNKNOWN.getNames()) { // additional check types: Any.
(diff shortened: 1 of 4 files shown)


jomarko (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -587,6 +587,15 @@ public static QName getNamespaceAndName(DMNModelInstrumentedBase localElement, M
                     }
                 }
             }
+            if (!(localElement instanceof ItemDefinition)) {
+                // If this is an InformationItem, e.g. is a variable of a DRGElement, it allows all types in the language.
+                // if instead it was an ItemDefinition, we only limited to DMNTypeRegistryV12.ITEMDEF_TYPEREF_FEEL_BUILTIN as per specification chapter, and we are not entering in this branch.
```
etirelli: Matteo, just to make sure we are on the same page, it is allowed to have an itemDefinition attribute of type "Any", correct?

tarilabs: @etirelli no, and I did it indeed on purpose to align with the spec:

image

...and `date`.

Do you want me to extend to ItemDefinition too?

etirelli: Unfortunately I think this is an oversight on the spec. Specially for integration with external services/layer, I think it is useful to be able to define an attribute on item definition as "Any". It is akin to, on a java class, set an attribute as type Object. In other words, it allows it to hold a reference to any value.

tarilabs: understood. Will align code accordingly

baldimir (approved):

PR #2143, opened 2018-11-08
DROOLS-3282: Should not move to next AgendaGroup even if current is empty

evacchi:
This is used for Units' scheduler-based impl. Figured we better track this feature/bug-fix in its own Jira. 

@mariofusco Not sure how to test

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
+++ b/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java
@@ -55,6 +55,7 @@ public class AgendaGroupQueueImpl
 
     private InternalWorkingMemory workingMemory;
     private boolean               autoDeactivate = true;
+    private boolean               keepWhenEmpty  = false;
     private Map<Long, String>     nodeInstances  = new ConcurrentHashMap<Long, String>();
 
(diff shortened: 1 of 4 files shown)


mariofusco:
@evacchi I'm afraid this change is breaking the sequential agenda, see failures in https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/drools-pullrequests/255/

evacchi:
good to know, will look into that

evacchi:
Jenkins please retest this

evacchi:
@mariofusco I kept a note to self here "check if it's the same agenda group" but I don't remember what I was supposed to actually write in code

mariofusco:
@evacchi This PR is intended to provide a way to suspend the execution of a rule by switching to a different empty agenda-group. However I'm afraid that calling getNextFocus() at each iteration will be performance expensive and moreover I'd prefer to have a cleaner primitive to suspend the execution. Let's discuss this in person.

evacchi:
discussed with @mariofusco. The AgendaGroup stack should *not* suspend execution, it's a stack that must progress execution till the end. We may add a boolean `suspend` == true when execution must halt (reset to false from the body of this method when execution gets suspended)

PR #2151, opened 2018-11-15
DROOLS-3324 Update properties in persistence.xml

baldimir:
@jiripetrlik @mariofusco could you please review? The properties needs to be updated because of Nayarana changes. Without them the tests fail on DB2 database randomly.

### DIFF
--- a/drools-persistence/drools-persistence-jpa/src/test/filtered-resources/META-INF/persistence.xml
+++ b/drools-persistence/drools-persistence-jpa/src/test/filtered-resources/META-INF/persistence.xml
@@ -21,8 +21,12 @@
         <property name="hibernate.dialect" value="${maven.hibernate.dialect}"/>
         <property name="hibernate.default_schema" value="${maven.jdbc.schema}"/>
          
+        <!-- BZ 841786: AS7/EAP 6/Hib 4 uses new (sequence) generators which seem to cause problems -->
+        <property name="hibernate.id.new_generator_mappings" value="false" />
+
         <!-- The following line is what's used in Hibernate 4 instead of a TransactionManagerLookup class -->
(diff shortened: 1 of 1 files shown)


rsynek (approved):
Looks good, the same property has been used for jBPM

drools-persistence/drools-persistence-jpa/src/test/filtered-resources/META-INF/persistence.xml
```diff
@@ -21,8 +21,12 @@
         <property name="hibernate.dialect" value="${maven.hibernate.dialect}"/>
         <property name="hibernate.default_schema" value="${maven.jdbc.schema}"/>
          
+        <!-- BZ 841786: AS7/EAP 6/Hib 4 uses new (sequence) generators which seem to cause problems -->
```
MarianMacik, line 24: Maybe add `+` after EAP 6 and Hib 4 so somebody else doesn't think that it is no longer needed.

MarianMacik (approved):
I would just improve the comment, otherwise looks good.

drools-persistence/drools-persistence-jpa/src/test/filtered-resources/META-INF/persistence.xml
baldimir: It is the same comment as in jBPM.

jiripetrlik (approved):
Looks good to me.

drools-persistence/drools-persistence-jpa/src/test/filtered-resources/META-INF/persistence.xml
MarianMacik: Then fine by me.

PR #2157, opened 2018-11-19
DROOLS-3332 Add reproducer for handling large DRL file

baldimir:
@mariofusco this is a reproducer for https://issues.jboss.org/browse/DROOLS-3332.


mariofusco:
Fixed by #2161

PR #2160, opened 2018-11-19
[DROOLS-1905] URLDecoder: Illegal hex characters in escape (%) pattern when a rule has a name with "%" 

elguardian:
fixing the problem for incremental build

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderSetImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderSetImpl.java
@@ -42,7 +42,6 @@ import org.kie.internal.builder.KnowledgeBuilder;
 import org.kie.internal.builder.KnowledgeBuilderError;
 
 import static java.util.Arrays.asList;
-
 import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.filterFileInKBase;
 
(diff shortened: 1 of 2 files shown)


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderSetImpl.java
```diff
@@ -248,10 +246,10 @@ public Reader getReader() throws IOException {
             throw new UnsupportedOperationException();
         }
 
-        private String decode( final String resourceName ) {
+        private String decode(final String resourceName) {
             try {
-                return URLDecoder.decode( resourceName, "UTF-8" );
-            } catch ( UnsupportedEncodingException e ) {
+                return URLDecoder.decode(resourceName, "UTF-8");
+            } catch (Exception e) {
```
baldimir: Hi, just doublechecking if this is really what we want to do here. Usually catching general Exception is a bad practice, because it could catch exceptions we don't want to catch. @mariofusco what do you think?

elguardian:
hi @baldimir just changed to the problem of getting an argument with no proper encoding.

baldimir:
Thanks, I will wait for checks to complete and approve afterwards.

baldimir (approved):

elguardian:
@baldimir any progress here ?

mariofusco:
ok to test

PR #2169, opened 2018-11-28
DROOLS-3351 [DMN Designer] Font size information is not being persisted

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-backend/src/test/java/org/kie/dmn/backend/marshalling/v1_2/UnmarshalMarshalTest.java
+++ b/kie-dmn/kie-dmn-backend/src/test/java/org/kie/dmn/backend/marshalling/v1_2/UnmarshalMarshalTest.java
@@ -33,6 +33,8 @@ import javax.xml.transform.stream.StreamSource;
 import org.junit.Test;
 import org.kie.dmn.api.marshalling.DMNMarshaller;
 import org.kie.dmn.model.api.Definitions;
+import org.kie.dmn.model.api.dmndi.DMNShape;
+import org.kie.dmn.model.api.dmndi.DMNStyle;
 import org.kie.dmn.model.v1_2.KieDMNModelInstrumentedBase;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


etirelli (approved):

kie-dmn/kie-dmn-backend/src/test/java/org/kie/dmn/backend/marshalling/v1_2/UnmarshalMarshalTest.java
```diff
@@ -48,13 +50,15 @@
...
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 
 public class UnmarshalMarshalTest {
 
     private static final StreamSource DMN12_SCHEMA_SOURCE = new StreamSource(UnmarshalMarshalTest.class.getResource("/DMN12.xsd").getFile());
+    private static final DMNMarshaller MARSHALLER = new org.kie.dmn.backend.marshalling.v1x.XStreamMarshaller();
```
baldimir, line 61: This will share the marshaller instance between tests. Is that ok? It could be created as part of a `@Before` method. In that case a fresh clean marshaller instance will be used in each of the tests.

baldimir (changes requested):
One comment.

kie-dmn/kie-dmn-backend/src/test/java/org/kie/dmn/backend/marshalling/v1_2/UnmarshalMarshalTest.java
tarilabs: Indeed that's okay, the Mashaller is meant to be shared

baldimir (approved):

kie-dmn/kie-dmn-backend/src/test/java/org/kie/dmn/backend/marshalling/v1_2/UnmarshalMarshalTest.java
tarilabs: on second though I realized we are not really sharing it :D but I still believe it is fine with the current code in this PR, as the Marshaller at least is not meant to encapsulate any state. Keep as in current PR code would later show if any problem. Agree @baldimir ?

tarilabs:
Jenkins test again

PR #2170, opened 2018-11-28
DROOLS-3379: RuntimeService API should not require a KieRuntime to create an instance

evacchi:
Reflecting changes in https://github.com/kiegroup/droolsjbpm-knowledge/pull/347, we use the new API in the (known) runtimes: 

1. cleaned up Bayes, updated DMN.
2. updated StatefulKnowledgeSessionImpl so that it forwards to the new KieRuntimeFactory public API in `kie-api`

@mariofusco @tarilabs


tarilabs (approved):

evacchi:
ping @etirelli for review as per @tarilabs suggestion

PR #2171, opened 2018-11-30
Drools 2851 Verifier: Add support for ranges

Rikkola:
https://issues.jboss.org/browse/DROOLS-2851

There is quite a few code formatting changes here.
The actual change is adding the `FieldRange.java` in and then using it in `SingleRangeCheck`.
I also added tests for this.

@jomarko @manstis if you two could please take a look. This PR is not urgent.


Rikkola:
I'll fix the conflicts :)

drools-verifier/drools-verifier-api/pom.xml
```diff
@@ -3,20 +3,20 @@
...
-    <description>Drools Workbench - Verifier Api</description>
+  <name>Drools Workbench - Verifier Api</name>
+  <description>Drools Workbench - Verifier Api</description>
 
-    <properties>
-        <java.module.name>org.drools.verifier.api</java.module.name>
-    </properties>
+  <properties>
+    <java.module.name>org.drools.verifier.api</java.module.name>
+  </properties>
```
Rikkola, line 20: Just removed spaces from the start here and in the other pom.xml files. No changes in these.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/SingleRangeCheck.java
```diff
@@ -93,9 +93,9 @@ private void checkRanges(Collection<ObjectField> rangeFields, Map<PartitionKey,
 
             for (ObjectField field : rangeFields) {
                 if ("Integer".equals(field.getFieldType())) {
-                    checkMonodimensionalRange(partition, dimensions, field, conditionIndex, IntegerRange::new, Integer.MIN_VALUE, Integer.MAX_VALUE);
+                    checkMonodimensionalRange(partition, dimensions, field, conditionIndex, IntegerRange::new, FieldRange.getIntegerMinValue(field), FieldRange.getIntegerMaxValue(field));
                 } else {
-                    checkMonodimensionalRange(partition, dimensions, field, conditionIndex, NumericRange::new, Double.MIN_VALUE, Double.MAX_VALUE);
+                    checkMonodimensionalRange(partition, dimensions, field, conditionIndex, NumericRange::new, FieldRange.getDoubleMinValue(field), FieldRange.getDoubleMaxValue(field));
```
Rikkola: The 4 lines above have the changes in this class. Everything else is my code formatter acting up.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/index/model/ObjectField.java
```diff
@@ -16,18 +16,39 @@
...
                        final AnalyzerConfiguration configuration) {
         super(factType,
               fieldType,
               name,
               configuration);
+        this.range = Optional.of(fieldRange);
+    }
+
+    public Optional<FieldRange> getRange() {
+        return range;
```
Rikkola, line 52: Added the range property for this field. At the moment it is actually not used by anything else than the unit test.

Rikkola:
@jomarko @manstis added licenses and resolved conflicts.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/index/model/FieldRange.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.verifier.core.index.model;
+
+public class FieldRange<T> {
```
jomarko, line 18: I juts would like to open discussion about this class. I see methods for `max` are similar and the same holds true for `min` methods. I am wondering if we can reduce the repetition of code by these changes:
- Change constructor to take `final ObjectField field` as parameter
- Don't make `FieldRange` generic by `<T>`
- Have no static methods in class
- Have just two public methods
  - `public Number getMaxValue()`
  - `public Number getMinValue()`

Not saying it is doable, but wanted to ask.

Rikkola: I gave this some thought.

First. The Integer and Double/Number needs to be separate. This is because with Integer there is no gap between 0 and 1, but with Double there is.

Second is removing the static methods. I'm not a fan of them, but since not every field has a range removing them would mean there is a need to check if the optional exists and if not, then use the default min or max value. Now the method is static, but it does everything in one location.

For these reasons I would keep it as it is for now. I imagine there will be more added to this in the future. Since there are more ranges than just numeric ones. For example enums or the illegal range might be between 0-10 for example. Currently there is no support for this. Then it might be worth rethinking the ObjectField structure.

manstis (commented):
From what I understand (it's Monday and close to Christmas, that's my excuse) it all looks fine to me.

However I'd appreciate some comments in `RangeCheckTest` to explain the tests a bit more (I found it difficult to understand the assertions in the tests... perhaps I being a bit thick!)

Rikkola:
No the code is indeed full of strange variables tSomething and random letters. Also it is for a feature that is not used anywhere so I'll make it easier to read.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        ageConditionColumn = new Column(1, analyzerConfiguration);
+        animalTypeConditionColumn = new Column(2, analyzerConfiguration);
+        actionColumn = new Column(3, analyzerConfiguration);
+
+        objectType = new ObjectType("mock.Type",
+                                    analyzerConfiguration);
+
+        kittenObjectField = new ObjectField("mock.Type", "String", "tKitten", analyzerConfiguration);
+        riskObjectField = new ObjectField("mock.Type", "String", "risk", analyzerConfiguration);
```
jomarko: Shouldn't we use `tRisk` instead of `risk` to be consistent with `tKitten` and `tAge`?

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private void addRow(final int rowNumber,
+                        final int ageGreaterThanMoreThanColumnValue,
+                        final String animalTypeColumn,
+                        final String actionColumnValue) {
+
+        final Pattern typePattern = new Pattern("mock.Type",
+                                                objectType,
+                                                analyzerConfiguration);
+        final Field ageField = new Field(ageObjectField, "mock.Type", "Integer", "tAge", analyzerConfiguration);
```
jomarko: I think we should extract `mock.Type` and field names as class constants.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,212 @@
...
+        objectType.getFields().add(kittenObjectField);
+        objectType.getFields().add(riskObjectField);
+    }
+
+    @Test
+    public void runTheCheckWithoutRange() {
+        ageObjectField = new ObjectField("mock.Type", "Integer", "tAge", analyzerConfiguration);
+        objectType.getFields().add(ageObjectField);
+
+        final List<Issue> issues = runAnalysis();
```
jomarko, line 38: I started the test locally and the `issues` contains two totally same objects, is it fine/expected, is it due to mocked environment?

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        assertEquals("Uncovered range starting from value -2147483648 in partition [Cat]", issues.get(0).getDebugMessage());
+    }
+
+    @Test
+    public void runTheCheckWithRange() {
+        ageObjectField = new ObjectField("mock.Type", "Integer", "tAge", new FieldRange(0, Integer.MAX_VALUE), analyzerConfiguration);
+        objectType.getFields().add(ageObjectField);
+
+        final List<Issue> issues = runAnalysis();
```
jomarko, line 62: Same as above, I started the test locally and the `issues` contains two totally same objects, is it fine/expected, is it due to mocked environment?

jomarko (commented):
Few more comments/questions.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
Rikkola: Sure

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
Rikkola: Ok

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
Rikkola: Best I investigate this, could be a bug.

Rikkola: The test is bad. I'm calling both start() and analyse(), but going to still check if I do the same in kie-wb-common/drools-wb.

jomarko: Let me know, when I should have a look again.

Rikkola:
I updated some of these locally, but giving priority to 7.2.1 issues I have atm. Just to make sure they get in.

jomarko:
@Rikkola sounds reasonable to me. thanks for update.

manstis:
@Rikkola Please let me know when this is ready for re-review. Thanks.

Rikkola:
@jomarko @manstis Ok. Updated, fixed and cleaned up. Let me know if more is needed. Trying to improve the API, tests and documentation for V&V.

Rikkola:
Error is in drools-wb-verifier-guided-decision-table-adapter so need to check that.

Rikkola:
jenkins execute downstream build

Rikkola:
Fixed now. Had to keep the analyse() method public for now.

jomarko:
@Rikkola Hi, I tried locally, and spotted the behavior below. So would like to ask you if this is something caused by this PR or was present also before.

- Open sample mortgages
- Create new guided decision table
- Add one row
- Add one column 
  - Simple condition column
  - $a : Applicant(age > 10)
- Analysis panel **has not** message about missing range
- Save and reopen
- Analysis panel **has** message about missing range

Could you please confirm behavior was same also before your changes?

No comments to code, only this behavior check question.

Rikkola:
@jomarko same issue exists with and without these changes.

jomarko (approved):
@Rikkola thank you.

Rikkola:
@manstis ready for a review.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,213 @@
...
+                index,
+                analyzerConfiguration);
+
+        analyzer.start();
+
+        return issues;
+    }
+
+    private void addRow(final int rowNumber,
+                        final int ageGreaterThanMoreThanColumnValue,
```
manstis: `ageGreaterThanOrEqualToColumnValue` might be a better name (if I was being picky)

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,213 @@
...
+                                  analyzerConfiguration);
+        row.getConditions().add(ageGreaterThanOrEqualCondition);
+        row.getConditions().add(ageLessThanCondition);
+        row.getPatterns().add(typePattern);
+        row.getActions().add(riskAction);
+        index.getRules().add(row);
+    }
+
+    private void addRow(final int rowNumber,
+                        final int ageGreaterThanMoreThanColumnValue,
```
manstis: `ageGreaterThanOrEqualToColumnValue` might be a better name.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
manstis: Sorry @Rikkola but I am confused :-)

I ran the test and it now returns 1 issue with the message in this PR. The issue has "row numbers" 3 and 4 which are for animal "Cat" and "Dog". IMO the reported row numbers are correct but the message should include "Dog" too? Or should there be two issues one for "Cat" and the other for "Dog" each with a single row number?

manstis: I guess I don't understand why the row for "Dog" is not also reporting an issue?

manstis: I _think_ I'm expecting:

1. An uncovered range for "Cat" (as 0>=age<12 appears not to be covered).
2. An uncovered range for "Dog" (as 0>=age<12 appears not to be covered).
3. An uncovered range for "Dog" (as 15>=age<18 appears not to be covered).

manstis (commented):
@Rikkola Well... everything looks hunky dory, if only I understood your test assertions.

If you can impart your expert understanding to my "porridge for brains" I'd be happier!

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
tarilabs: image
this was my original comment, for reference.

Rikkola: @manstis Yes so @tarilabs brought the same issue up. I think the reason why it works like this is to save resources when running the Verifier in webapp. Another reason is that not every table aims to be complete, especially in the Drools side so if each issue is reported you can get as many errors as you have rows.

The plan is to add a configuration for both modes. One for first issue reporting and one for full run. Since I think V&V uses the same approach in other locations too.

Rikkola:
To make the addition easier to understand I think I'll add the improved messaging into this PR. There is nobody waiting for this PR to go in, it affects nothing, so why not.

Rikkola:
@manstis @tarilabs gaps reported with a better reporting object and covering the mentioned gaps. Unit tests changed to reflect the change.

@jomarko you might want to take a new look too.

drools-verifier/drools-verifier-api/src/main/java/org/drools/verifier/api/reporting/gaps/MissingRange.java
```diff
@@ -0,0 +1,39 @@
+package org.drools.verifier.api.reporting.gaps;
```
manstis, line 16: License please.

drools-verifier/drools-verifier-api/src/main/java/org/drools/verifier/api/reporting/gaps/PartitionCondition.java
```diff
@@ -0,0 +1,32 @@
+package org.drools.verifier.api.reporting.gaps;
```
manstis, line 16: License please.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/cache/RuleInspectorCache.java
```diff
@@ -98,19 +97,21 @@ public void newColumn(final Column column) {
...
-                        .filter(Check::hasIssues)
-                        .map(Check::getIssue)
-                        .collect(toSet())
-        );
-        issues.addAll(generalChecks.stream()
-                              .filter(Check::hasIssues)
-                              .map(Check::getIssue)
-                              .collect(toSet())
-        );
+final         Set<Issue> issues = new HashSet<>();
```
manstis: Formatting.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/cache/RuleInspectorCache.java
```diff
@@ -98,19 +97,21 @@ public void newColumn(final Column column) {
...
+        for (final RuleInspector inspector : all()) {
+            for (Check check : inspector.getChecks()) {
+                if (check.hasIssues()) {
+                issues.addAll(check.getIssues());
```
manstis: Formatting.. (any reason to move from streams to nested loops!?)

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/NumericRange.java
```diff
@@ -0,0 +1,60 @@
...
+    public NumericRange(final List<ConditionInspector> conditionInspectors) {
+        super(conditionInspectors);
+    }
+
+    @Override
+    protected Consumer<ConditionInspector> getConditionParser() {
+        return c -> {
+            final FieldCondition cond = (FieldCondition) c.getCondition();
+            final Operator op = resolve(cond.getOperator());
+            switch (op) {
```
manstis, line 38: Why do you not cover `EQUALS` (like the `IntegerRange`?)

manstis (commented):
The test gives the results I'd have expected. LGTM (other than a few silly errors).

drools-verifier/drools-verifier-api/src/main/java/org/drools/verifier/api/reporting/gaps/MissingRangeIssue.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private Collection<MissingRange> uncoveredRanges;
+    private List<PartitionCondition> partition;
+
+    public MissingRangeIssue() {
+    }
+
+    public MissingRangeIssue(final Severity severity,
+                             final CheckType checkType,
+                             final HashSet<Integer> es,
```
jomarko: What does `es` stand for please?

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/cache/RuleInspectorCache.java
jomarko: +1 for Michael's comment

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/DetectConflictingRowsCheck.java
```diff
@@ -64,6 +66,6 @@ protected Issue makeIssue(final Severity severity,
 
         issue.setDebugMessage(new RuleInspectorDumper(ruleInspector).dump() + " ## " + new RuleInspectorDumper(other).dump());
 
-        return issue;
+        return Arrays.asList(issue);
```
jomarko: I wonder if we should use `Collections.singletonList` as we have one item list. I read [this](https://stackoverflow.com/questions/26027396/arrays-aslist-vs-collections-singletonlist/26039540#26039540) and now I think it could be fine, what do you think?

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/DetectDeficientRowsCheck.java
```diff
@@ -61,11 +62,13 @@ protected Severity getDefaultSeverity() {
...
 
     @Override
-    protected Issue makeIssue(final Severity severity,
-                              final CheckType checkType) {
-        return new Issue(severity,
-                         checkType,
-                         new HashSet<>(Arrays.asList(ruleInspector.getRowIndex() + 1))
+    protected List<Issue> makeIssues(final Severity severity,
+                                     final CheckType checkType) {
+        return Arrays.asList(
```
jomarko: Please see comment above for all `makeIssues` implementations.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/base/Check.java
```diff
@@ -16,14 +16,16 @@
...
+
 import org.drools.verifier.api.reporting.Issue;
 import org.drools.verifier.core.configuration.CheckConfiguration;
 
 public interface Check {
 
     boolean check();
 
-    Issue getIssue();
+    List<Issue> getIssues();
```
jomarko, line 28: Just question, looks like important API change, Doesn't cause this change troubles somewhere in `drools-wb` for example?

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/PartitionKey.java
```diff
@@ -0,0 +1,66 @@
...
+    public String toString() {
+        return Arrays.toString(keys);
+    }
+
+    boolean hasNulls() {
+        return Stream.of(keys).anyMatch(Objects::isNull);
+    }
+
+    public boolean subsumes(final PartitionKey other) {
+        return IntStream.range(0, keys.length).allMatch(i -> keys[i] == null || keys[i].equals(other.keys[i]));
```
jomarko: `Objects.equals(keys[i], other.keys[i])` won't throw a NPE if `other.keys[i]` is null, do we want to use it?

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/RangeError.java
```diff
@@ -0,0 +1,57 @@
...
+        return new MissingRangeIssue(severity,
+                                     checkType,
+                                     new HashSet<>(ruleInspectors.stream().map(r -> r.getRowIndex() + 1).collect(toSet())),
+                                     partitionKey.getConditions(),
+                                     uncoveredRanges
+        ).setDebugMessage(getMessage());
+    }
+
+    private String getMessage() {
+        return "Uncovered range found";
```
jomarko: I see this is used just in kie-wb-common as log message, however maybe could contain also fact/field, or some more details.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/checks/RangeCheckTest.java
```diff
@@ -0,0 +1,264 @@
...
+        index = new IndexImpl();
+
+        ageConditionColumn = new Column(1, analyzerConfiguration);
+        animalTypeConditionColumn = new Column(2, analyzerConfiguration);
+        actionColumn = new Column(3, analyzerConfiguration);
+
+        objectType = new ObjectType(OBJECT_TYPE_NAME,
+                                    analyzerConfiguration);
+
+        kittenObjectField = new ObjectField(OBJECT_TYPE_NAME, "String", "kitten", analyzerConfiguration);
```
jomarko: I think we should introduce constants for these fields `kitten` ... as we use it on multiple places. Constants will prevent us from updating just subset of occurrences in the class.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/index/model/FieldRangeTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+@RunWith(MockitoJUnitRunner.class)
+public class FieldRangeTest {
+
+    @Mock
+    ObjectField objectField;
```
jomarko: Never used field, please remove it.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/index/model/FieldRangeTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    public void getIntegerMinValueException() {
+        assertEquals(Integer.MIN_VALUE, FieldRange.getIntegerMinValue(new ObjectField("factType",
+                                                                                      "fieldType",
+                                                                                      "name",
+                                                                                      new FieldRange(null,
+                                                                                                     null),
+                                                                                      mock(AnalyzerConfiguration.class))));
```
jomarko: Could we please merge tests which use the same `ObjectField` constructor and then do more assertions per test.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/util/DateTimeFormatProviderMock.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.drools.verifier.core.util;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.drools.verifier.core.configuration.DateTimeFormatProvider;
+
+public class DateTimeFormatProviderMock
```
jomarko, line 24: We have such class `drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/test/java/org/drools/workbench/services/verifier/plugin/client/testutil/DateTimeFormatProviderMock.java` is there a chance to get rid of one of them?

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/util/UUIDKeyProviderMock.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.verifier.core.util;
+
+import org.drools.verifier.core.index.keys.UUIDKeyProvider;
+
+public class UUIDKeyProviderMock
```
jomarko, line 21: Similar question `drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/test/java/org/drools/workbench/services/verifier/plugin/client/testutil/UUIDKeyProviderMock.java`

jomarko (commented):
### Code review notes
Few code review notes, points to discuss better to say.

jomarko:
### Manual review
I haven't found difference in behavior without and with this PR.

drools-verifier/drools-verifier-api/src/main/java/org/drools/verifier/api/reporting/gaps/MissingRangeIssue.java
Rikkola: A really good question.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/DetectConflictingRowsCheck.java
Rikkola: Makes sense.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/base/Check.java
Rikkola: The change only affects the verifier module. Internally it now uses a list from each check, but the issues from checks are collected inside the Verifier module, relisted and then passed to drools-wb and other locations in the same format as before.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/PartitionKey.java
Rikkola: Sure it is better.

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/index/model/FieldRangeTest.java
Rikkola: Sure

drools-verifier/drools-verifier-core/src/test/java/org/drools/verifier/core/util/UUIDKeyProviderMock.java
Rikkola: This would require a test dependency from the tests of another module. I think a better practice would be to create a separate module for the classes like we have in appformer, but I would wait for until there is more than two classes that need to be moved to that module. The need for this might come once I go forward.

drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/NumericRange.java
Rikkola: I think I should.

Rikkola:
jenkins please execute downstream build

jomarko:
@Rikkola do I understand you added just formatting changes after review? Am I right that no re-review is needed?

manstis (commented):
Getting there :-)

A couple of points:

1) The `row numbers` seem wrong (out by 1). Could this be caused by [this](https://github.com/kiegroup/drools/blob/47b3fc1e4defffd593bfac6edd41fe1fa8e2a3e0/drools-verifier/drools-verifier-core/src/main/java/org/drools/verifier/core/checks/gaps/RangeError.java#L49)?

2) In `RangeCheckTest#L183` I added an additional rule `addRow(5, Integer.MIN_VALUE, 12, "ok");` and was expecting one of the Dog gaps to disappear but it appeared to remain (`Integer.MIN_VALUE < age < 11` was still reported). Did I do something stupid?

Rikkola:
1. Maybe best if I disable row numbers for this asset type, since the row number can be misleading. The gap "edge" can be formed from multiple lines.
2. I'll make more tests to see if the issue is there.

Rikkola:
@manstis @jomarko updated the PR with another commit that fixes min/max value issues.

jomarko:
@Rikkola @manstis Hi, here my position is not nice. We work on feature which complexity grows with the size of validated table. The added tests looks valid, however they use just small tables. For me is very hard to understand the implementation. So I decided to do manual check. It seems to me we have troubles with validation feature, regardless of changes in this PR, please see my [comment](https://issues.jboss.org/browse/DROOLS-2851?focusedCommentId=13712025&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13712025)

Rikkola:
@jomarko the problem for any Verifier is that if you can see what is missing you don't need it and if you need it the table is too hard to understand. That is why I hope that at some point we either improve the visual reporting or we generate unit tests that show what is wrong with the table.

I'll see if can find time to take a look at the issue you reported. My fear is that the problem is in web worker <-> main app synchronization and not on my codes, but need to figure some way to validate my suspection. This is because all the unit tests I create for editing the table work fine.

jomarko:
@Rikkola ok, thank you, however this PR is opened too long time, @manstis could you please have a look? If you are fine I think this should be merged and then we can address the issue I mentioned above in a separate ticket.

manstis:
@jomarko I think @Rikkola is right; the problem of the Issues Panel not always displaying has been a recurrent issue with how the V&V is executed and results displayed from the Guided Decision Table Editor and does not relate to this PR. IMO this PR can be merged (although @Rikkola I'd like to understand why removal of the row numbers is the best solution.. they only seemed to be +1 out of sync' with what I'd have expected them to be).

Rikkola:
@manstis I'm have doubts about if the "edge" of a gap is always a row. I fear it might sometimes a combination of rows, but need to verify this with some tests.

jomarko:
@Rikkola @manstis what is the status of this PR please?

Rikkola:
I think I'll add the row info back. I can only think of few situations where having the row number might be confusing, when the row has "or" in it, but this is a really rare situation and even then the information is correct.

Rikkola:
jenkins please retest this

Rikkola:
@manstis @tarilabs I could use another look. The row numbers are back in.

tarilabs:
Not sure how much I can contribute _to this specific_ PR, as I see it does not make use of the DMN Validation data structure. On other hand, if this helps going into that direction I of course trust you @Rikkola ;)

Rikkola:
@tarilabs In this case. Your marked as a reviewer so as long as your not against this then we are good, just thought I'll check with you to be safe.

tarilabs:
> Your marked as a reviewer so as long as your not against this then we are good, just thought I'll check with you to be safe.

That's the thing, I'm not sure why I am listed as a reviewer in this module :smile:
Anyhow, I never have reasons to veto work in other modules, I trust you this is a needed change for steps in the proper direction.
As discussed please bear in mind the end goal is for the your verifier to populate the required data structure to the kie-dmn-validation module. When you have that available, I can provide some actual pragmatic review there ;)

manstis (approved):

Rikkola:
@kiegroup/gatekeepers can you please revert. This breaks some tests in drools-wb that I need to fix.

manstis:
@Rikkola Done. See https://github.com/kiegroup/drools/pull/2316

Rikkola:
@manstis thank you.

PR #2172, opened 2018-11-30
[DROOLS-3392] workaround lack of xjc executable in JDK11

mariofusco:


kie-pmml/pom.xml
```diff
@@ -157,6 +157,64 @@
...
+              <groupId>com.sun.istack</groupId>
+              <artifactId>istack-commons-runtime</artifactId>
+              <version>3.0.7</version>
+              <type>jar</type>
+              <overWrite>false</overWrite>
+              <outputDirectory>${project.basedir}/target/lib</outputDirectory>
+            </artifactItem>
+          </artifactItems>
+        </configuration>
+      </plugin>
```
mareknovotny, line 217: @mariofusco this is much better than adding jar files into git. It would be good to use version properties for the versions and declare them in kie-parent, if not possible declare them in drools parent.

baldimir (approved):

PR #2176, opened 2018-12-03
[DROOLS-3385] use a concurrent map to store classes declaration in kn…

mariofusco:
…owledge base


baldimir (approved):

rsynek (approved):
Looks good, preliminary tests passed several times, so let's merge and see in the next nightly build.

PR #2179, opened 2018-12-06
[DROOLS-3367] Created DMNRuntimeFluent and related APIs

danielezonca:
Implement DMN fluent API plus small changes of existing API

It supports at the moment only a single run definition per time because it registers `DMNModel`, `DMNContext` etc without any concept like "DMN session id". I think it is not really needed because DMN is stateless so we can assume user are working only on a single model per time but it can be easily updated in the future.

@mariofusco @tarilabs @kkufova  

PRs list
https://github.com/kiegroup/droolsjbpm-knowledge/pull/350
https://github.com/kiegroup/drools/pull/2179

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/fluent/SetDMNInputCommand.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/fluent/SetDMNInputCommand.java
@@ -19,6 +19,7 @@ package org.kie.dmn.core.fluent;
 import org.kie.api.command.ExecutableCommand;
 import org.kie.api.runtime.Context;
 import org.kie.dmn.api.core.DMNContext;
+import org.kie.dmn.api.core.DMNRuntime;
 import org.kie.dmn.core.api.DMNFactory;
 import org.kie.internal.command.RegistryContext;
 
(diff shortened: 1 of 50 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/fluent/SetDMNInputCommand.java
```diff
@@ -0,0 +1,44 @@
...
+        this.name = name;
+        this.value = value;
+    }
+
+    @Override
+    public Void execute(Context context) {
+        RegistryContext registryContext = (RegistryContext) context;
+        DMNContext dmnContext = registryContext
+                .computeIfAbsent(DMNContext.class,
+                                 clazz -> DMNFactory.newContext());
```
tarilabs: here you may want to create a DMNContext actually from DMNRuntime instead

tarilabs (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/fluent/SetDMNInputCommand.java
danielezonca: Done :)

kkufova (approved):
Looks good to me, except for the `return null;` statement. :-)

danielezonca:

PR #2178, opened 2018-12-05
BXMSDOC-3377: Update HelloWorldExample

ron-murhammer:
Update HelloWorldExample to align with doc updates in https://issues.jboss.org/browse/BXMSDOC-3377.

Enable audit logging for the HelloWorldExample to the target directory so that it can be tested in the "Audit" view in JBDS: https://docs.jboss.org/drools/release/7.15.0.Final/drools-docs/html_single/Examples/HelloWorldExample/helloworld_auditview1.png

PR #2185, opened 2018-12-07
BXMSDOC-3377: Update HelloWorldExample - fix compilation

MarianMacik:
Hi, @mariofusco, there is a compilation failure as a result of #2178. I hope this can fix that.


MarianMacik:
Downstream compiled failed, but on an unrelated change. Drools compilation passed.

mariofusco:
Fixed by https://github.com/kiegroup/drools/commit/13aff7acd17d574a182ca926b5019c66ecb21c42

PR #2188, opened 2018-12-10
DROOLS-3419: Bump heap size for findbugs in drools-compiler Maven build

evacchi:
- add config param to bump to 1GB [(default is 512MB)](https://gleclaire.github.io/findbugs-maven-plugin/findbugs-mojo.html#maxHeap)
- reference: https://stackoverflow.com/questions/13876912/can-not-execute-findbugs-java-lang-outofmemoryerror-java-heap-space-while-run/13879398#13879398


evacchi:
@mariofusco the downstream build is broken but because of the workbench (I think they are already aware of the issue)

PR #2189, opened 2018-12-11
DROOLS-2373 : BRL 'is contained in the (comma separated) list' does not produce multiselect dropdown

Rikkola:
https://issues.jboss.org/browse/DROOLS-2373

https://github.com/kiegroup/drools/pull/2189
https://github.com/kiegroup/kie-wb-common/pull/2359
https://github.com/kiegroup/drools-wb/pull/1025

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/InterpolationVariable.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/InterpolationVariable.java
@@ -15,61 +15,67 @@
  */
 package org.drools.workbench.models.datamodel.rule;
 
+import java.util.Objects;
+
 public class InterpolationVariable {
 
     private String varName;
     private String dataType;
     private String factType;
     private String factField;
+    private String operator;
 
     public InterpolationVariable() {
         //For Errai marshalling...
     }
 
-    public InterpolationVariable( String varName,
-                                  String dataType ) {
+    public InterpolationVariable(String varName,
+                                 String dataType) {
         this.varName = varName;
         this.dataType = dataType;
     }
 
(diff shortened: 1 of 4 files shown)


Rikkola:
@jomarko here is the BRL fix.

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/InterpolationVariable.java
```diff
@@ -21,55 +21,73 @@
...
-                                                                                                                           that.factField );
+        return equalOrNull(this.varName,
+                           that.varName)
+                && equalOrNull(this.dataType,
+                               that.dataType)
+                && equalOrNull(this.factType,
+                               that.factType)
+                && equalOrNull(this.factField,
+                               that.factField)
+                && equalOrNull(this.operator,
```
jomarko: Thinking if we need this `equalOrNull`, it seems to me `Objects.equals` has same behavior.
```java
@Override
    public boolean equals(Object obj) {
        if (obj == null) {
            return false;
        }
        if (!(obj instanceof InterpolationVariable)) {
            return false;
        }
        InterpolationVariable that = (InterpolationVariable) obj;
        return Objects.equals(this.varName,
                              that.varName)
                && Objects.equals(this.dataType,
                                  that.dataType)
                && Objects.equals(this.factType,
                                  that.factType)
                && Objects.equals(this.factField,
                                  that.factField)
                && Objects.equals(this.operator,
                                  that.operator);
    }
```

jomarko (approved):
One comment but approving in advance.

manstis (approved):

jomarko:
@Rikkola @manstis hi, this seems to me as fine to be merged. Is there anything that I am missing and block us from merging? Asking because we have to create 7.14.x cherry pick until Friday.

PR #2192, opened 2018-12-12
[DROOLS-3563] minimize calls to feel variable names normalization

mariofusco:


mariofusco:
@etirelli @tarilabs I know the change looks trivial but benchmarks result are the following for me

BEFORE

Benchmark                                                          (dmnFileName)  Mode  Cnt    Score    Error  Units
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision  example_short_names.xml    ss   45  187.426 ± 10.207  ms/op
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision   example_long_names.xml    ss   45  328.034 ±  6.303  ms/op

AFTER

Benchmark                                                          (dmnFileName)  Mode  Cnt   Score   Error  Units
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision  example_short_names.xml    ss   45  77.886 ± 7.317  ms/op
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision   example_long_names.xml    ss   45  79.747 ± 4.470  ms/op

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextImpl.java
```diff
@@ -39,12 +40,12 @@ public DMNContextImpl(Map<String, Object> entries) {
...
     @Override
     public Object set(String name, Object value) {
-        return getCurrentEntries().put(name, value);
+        return getCurrentEntries().put( EvalHelper.normalizeVariableName( name ), value);
     }
 
     @Override
     public Object get(String name) {
-        return getCurrentEntries().get(name);
+        return getCurrentEntries().get( EvalHelper.normalizeVariableName( name ) );
```
etirelli: If the variable name is normalized on the set() method, is it necessary to normalize on the get() as well?

mariofusco: @etirelli I'm not entirely sure tbh, I thought that somebody could ask for a value passing a non-normalized name, but I'll check what happens if I remove it.

mariofusco: @etirelli You were right, normalization on get is also not necessary. New benchmark results are

Benchmark                                                          (dmnFileName)  Mode  Cnt   Score   Error  Units
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision  example_short_names.xml    ss   45  72.983 ± 3.376  ms/op
DMNEvaluateDecisionNameLengthBenchmark.evaluateDecision   example_long_names.xml    ss   45  70.427 ± 3.009  ms/op

Now the long names benchamark is around 4.5 times faster than before :)

tarilabs:
This is promising.
However, I am afraid this is breaking some of the prerequisites which originally @etirelli wrote in the README.md about FEEL being insensitive to number of spaces. If that is fine, this is good but won't be enough as the model nodes names will still need be normalized --just wild guessing I need to build a test case to actually confirm the problem.

mariofusco:
@tarilabs if you see a potential problem with this, I'd need a reproducer for it. We can delay this discussion to tomorrow if you're busy today, but that normalize method is clearly the elephant in the room and the more we avoid calling it the better it is.

mariofusco:
ok to test

tarilabs:
@mariofusco please rename this PR title to link to DROOLS-3563 instead. Thanks.

tarilabs (changes requested):
As discussed there are safer and less-risky idea which can be implemented, and in turn that make this idea of minimizing surface of normalization less relevant.
You can please close this PR now.

PR #2199, opened 2018-12-20
DROOLS-3404 

lanceleverich:
Finish removing 'declare' statements from generated rules
@Ignored tests in guided scorecards that don't work due to empty field names


lanceleverich:
Closed as the changes have been incorporated into a new PR for RHDM-766

PR #2200, opened 2018-12-21
JBPM-8107 - Kjar added as dependency of spring boot application is not properly found when creating class path kie container

mswiderski:
…- e.g. spring boot

still thinking how to write test for this ... I tested that with spring boot app and works as expected though no idea how to make a test case ... without bringing in entire spring boot or similar to the picture...


mswiderski:
Jenkins retest this

mariofusco (approved):

PR #2205, opened 2019-01-03
DROOLS-3469 DMN xml NS definition on non-root xml elements

tarilabs:
/cc @etirelli @baldimir @jiripetrlik @danielezonca

PR 1/2


etirelli (approved):

danielezonca (approved):
I tested this PR in the WB and the error is no more present :+1: :+1:

jomarko (approved):

kkufova (approved):

PR #2198, opened 2018-12-17
Serialization with SecurityManager (DROOLS-3429)

aboukhal:
Serialization of rule packages always failed when using a SecurityManager; extended the special handling of CompiledInvokers to also consider SafePredicateExpression and SafeConsequence wrappers

PR #2207, opened 2019-01-08
DROOLS-3429 Fix serialization of compiled invokers when SecurityManager is enabled

baldimir:
- Extension of fix proposed here: https://github.com/kiegroup/drools/pull/2198
- Added test case

There were more "Safe*" classes that wrap CompiledInvoker instances, therefore I applied similar fix everywhere where needed. 

@mariofusco could you please review? @aboukhal please also take a look. I made changes to your proposed code.


baldimir:
jenkins retest this

baldimir:
Fail not related to the changes

PR #2212, opened 2019-01-10
[DROOLS-3505] Parse error with executable-model, java dialect, modify…

tkobayas:
… block and comma

- Unit test


mariofusco:
Fixed

PR #2213, opened 2019-01-11
[DROOLS-3508] Build error with executable-model, multiple DRLs of the…

tkobayas:
… same package, mvel dialect

- Unit test


mariofusco:
Fixed

PR #2221, opened 2019-01-18
DROOLS-3534 DMN built-in FEEL sample standard deviation for one sample

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/twovaluelogic/StddevFunctionTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/twovaluelogic/StddevFunctionTest.java
@@ -16,16 +16,16 @@
 
 package org.kie.dmn.feel.runtime.functions.twovaluelogic;
 
-import org.junit.Before;
-import org.junit.Test;
-import org.kie.dmn.feel.runtime.events.InvalidParametersEvent;
-import org.kie.dmn.feel.runtime.functions.FunctionTestUtil;
-
 import java.math.BigDecimal;
 import java.math.MathContext;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


etirelli (approved):

etirelli:
Thank you @tarilabs , good catch.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/twovaluelogic/StddevFunctionTest.java
```diff
@@ -42,28 +42,12 @@ public void invokeNumberNull() {
...
-
-    @Test
-    public void invokeNumberDoubleWithDecimalPart() {
-        FunctionTestUtil.assertResult(stddevFunction.invoke(10.1d), BigDecimal.ZERO);
-    }
-
-    @Test
-    public void invokeNumberFloat() {
-        FunctionTestUtil.assertResult(stddevFunction.invoke(10.1f), BigDecimal.ZERO);
+    public void invokeSingleNumber() {
```
baldimir, line 45: These are meant as unit tests so, the tests were separated on purpose - when something is wrong, it shows exactly what is wrong while testing also other cases. Please leave them separated.

baldimir (changes requested):
Just one comment.

baldimir (approved):
Approving after discussing on Slack

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/twovaluelogic/StddevFunctionTest.java
tarilabs: [as discussed on Slack, test have changed semantically from returning a number in different cases, to always return an error in case of a single parameter number.]

PR #2231, opened 2019-01-28
[DROOLS-3558] "NoSuchElementException: No value present" with executa…

tkobayas:
…ble-model, Map and binding

- Unit test


mariofusco:
Test case merged and fixed

PR #2232, opened 2019-01-28
[DROOLS-3559] executable-model misses to raise an error for a non-exi…

tkobayas:
…siting method with Mvel dialect

- Unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/795fe4e6fe0e86499b5e8a30f4a191e3ceb0cfba

PR #2237, opened 2019-02-03
DROOLS-3594: adding support for interval-based algebra. 

etirelli:


etirelli:
jenkins test this

etirelli:
@tarilabs @baldimir this is ready for your review.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/AfterFunction.java
```diff
@@ -0,0 +1,107 @@
...
+    public FEELFnResult<Boolean> invoke(@ParameterName( "value" ) Comparable value, @ParameterName( "range" ) Range range) {
+        if ( value == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "value", "cannot be null"));
+        }
+        if ( range == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range", "cannot be null"));
+        }
+        try {
+            boolean result = ( range.getHighBoundary() == Range.RangeBoundary.CLOSED && value.compareTo( range.getHighEndPoint() ) > 0 ) ||
+                    ( range.getHighBoundary() == Range.RangeBoundary.OPEN && value.compareTo( range.getHighEndPoint() ) >= 0 );
```
baldimir, line 64: You don't need to check that the boundary is OPEN, after checking it is CLOSED. Similarly as in the last invoke method comparing 2 ranges. This comment applies for all invokes with ranges (please check also other functions).

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/FinishedByFunction.java
```diff
@@ -0,0 +1,89 @@
...
+    public FEELFnResult<Boolean> invoke(@ParameterName( "range1" ) Range range1, @ParameterName( "range2" ) Range range2) {
+        if ( range1 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range1", "cannot be null"));
+        }
+        if ( range2 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range2", "cannot be null"));
+        }
+        try {
+            boolean result =
+                    range1.getHighBoundary() == Range.RangeBoundary.CLOSED && range2.getHighBoundary() == Range.RangeBoundary.CLOSED &&
```
baldimir, line 60: Is this boundary check necessary? This way we omit the OPEN boundaries.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/FinishesFunction.java
```diff
@@ -0,0 +1,89 @@
...
+    public FEELFnResult<Boolean> invoke(@ParameterName( "range1" ) Range range1, @ParameterName( "range2" ) Range range2) {
+        if ( range1 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range1", "cannot be null"));
+        }
+        if ( range2 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range2", "cannot be null"));
+        }
+        try {
+            boolean result =
+                    range1.getHighBoundary() == Range.RangeBoundary.CLOSED && range2.getHighBoundary() == Range.RangeBoundary.CLOSED &&
```
baldimir, line 60: Same comment as in finishedBy about boundaries.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/StartedByFunction.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public FEELFnResult<Boolean> invoke(@ParameterName( "range" ) Range range, @ParameterName( "value" ) Comparable value) {
+        if ( value == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "value", "cannot be null"));
+        }
+        if ( range == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range", "cannot be null"));
+        }
+        try {
+            boolean result = ( range.getLowBoundary() == Range.RangeBoundary.CLOSED && value.compareTo( range.getLowEndPoint() ) == 0 );
```
baldimir, line 43: Shouldn't this work also for open boundaries? I mean comparing with range.getLowBoundary() + 1?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/StartedByFunction.java
```diff
@@ -0,0 +1,89 @@
...
+    public FEELFnResult<Boolean> invoke(@ParameterName( "range1" ) Range range1, @ParameterName( "range2" ) Range range2) {
+        if ( range1 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range1", "cannot be null"));
+        }
+        if ( range2 == null ) {
+            return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, "range2", "cannot be null"));
+        }
+        try {
+            boolean result =
+                    range1.getLowBoundary() == Range.RangeBoundary.CLOSED && range2.getLowBoundary() == Range.RangeBoundary.CLOSED &&
```
baldimir, line 60: Same comment about boundaries as before. Shouldn't this work also with OPEN boundaries? Please think about that also regarding other functions. I won't comment that again, because I would have to put it are more places.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/interval/StartsFunctionTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    public void invokeParamSingleAndRange() {
+        FunctionTestUtil.assertResult( startsFunction.invoke( "a",
+                new RangeImpl( Range.RangeBoundary.CLOSED, "a", "f", Range.RangeBoundary.CLOSED )),
+                Boolean.TRUE );
+        FunctionTestUtil.assertResult( startsFunction.invoke( "f",
+                new RangeImpl( Range.RangeBoundary.CLOSED, "a", "f", Range.RangeBoundary.CLOSED )),
+                Boolean.FALSE );
+        FunctionTestUtil.assertResult( startsFunction.invoke( "a",
```
baldimir, line 58: Just an example of a case with OPEN boundaries as I commented before and should work: 

`        FunctionTestUtil.assertResult( startsFunction.invoke( "b",
                new RangeImpl( Range.RangeBoundary.OPEN, "a", "f", Range.RangeBoundary.CLOSED )),
                Boolean.FALSE );`

baldimir (changes requested):
See comments please.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/AfterFunction.java
etirelli: I believe I do need to check for OPEN, otherwise it would allow a CLOSED boundary to have a compareTo() result == 0. I could try to rewrite the complete expression as nested if's, but I think the alternative would be more complicated to understand and heavier from a performance perspective than just checking for OPEN and CLOSED. Please let me know if I misunderstood your suggestion.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/FinishedByFunction.java
etirelli: Yes, the algebra is only defined for CLOSED intervals. In some cases, like the after() function, I relaxed the rules to allow for an OPEN interval. 
I guess, in this particular case, one can argue that as long as both boundaries are the same, either OPEN or CLOSED, this function would work, but I am not 100% sure of the side-effects of this assumption. 
If you think I *should* relax the restriction here, I can go ahead and change, but I would prefer to be on the safe side for this first implementation.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/FinishesFunction.java
etirelli: See previous answer.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/StartedByFunction.java
etirelli: No, you can't do +1 because we are talking about continuous intervals. We are not dealing with integer numbers, we are dealing with real numbers, and strings, etc.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/StartedByFunction.java
etirelli: see previous answer.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/FinishedByFunction.java
baldimir: Thanks for explanation. OK for of then as it is.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/interval/StartsFunctionTest.java
etirelli: No, that is not correct. The immediate successor of "a" is NOT "b". 
Lets use a simple example, we know that "a" < "b", but the following is also true: "a" < "aa" < "b". Also "a" < "aa" < "aaa" < "b". Etc.
In fact, since we are dealing with unicode characters here, the immediate successor of "a" is "a\u0000". But this will make practical use cases not viable, so I prefer to restrict the cases to closed intervals.

etirelli:
@baldimir thanks for the review, and please let me know if my answers make sense. In case I misunderstood any of your suggestions, please let me know and I will fix it.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/StartedByFunction.java
baldimir: Yes agree, forgot about continous ranges here where there isn't a discrete step.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/AfterFunction.java
baldimir: Yes you are right. How about the other way? Accepting all greater than 0 or OPEN equal 0? Either way this is just nitpicking. So I agree to have it also as it is.

baldimir:
Thanks Edson, yes they make sense. Will approve this.

baldimir (approved):

PR #2243, opened 2019-02-06
[DROOLS-3606] From with modify fires unexpected rule

tkobayas:
- Unit test

This unit test rule doesn't look very meaningful (retrieve Person *from* Person) but the same is reproducible with a rule retrieving Address *from* Person.


mariofusco:
Test case merged and fixed

PR #2255, opened 2019-02-14
[DROOLS-3653] Unexpected node sharing by function with equals

tkobayas:
- Unit test


mariofusco:
Fixed by https://github.com/kiegroup/drools/commit/9e35e1b27226d586fc0974491e0369bb3fc3e818

PR #2258, opened 2019-02-15
RHPAM-1875 Fixes for JSON marshalling/unmarshalling of PMML

lanceleverich:


drools-core/src/main/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommand.java
```diff
@@ -176,7 +178,22 @@ public PMML4Result execute(Context context) {
...
+                    if (!clazz.isAssignableFrom(pi.getValue().getClass())) {
+                        try {
+                            Object o = clazz.getDeclaredConstructor(pi.getValue().getClass()).newInstance(pi.getValue());
+                            pi.setValue(o);
+                        } catch (Throwable t) {
+                            resultHolder.setResultCode("ERROR-3");
+                            return resultHolder;
+                        }
+                    }
+                }
```
mariofusco, line 196: I'm not sure of which kind of issue you're trying to fix here. Can you explain the problem with a practical example? I can be wrong, but this seems more a workaround that an actual fix.

lanceleverich: The problem is that for some reason the incoming `PMMLRequestData` isn't getting deserialized using the registered `JSONMarshallerPMMLRequest `extension. Whatever mechanism it is using, deserializes the `ParameterInfo` objects and creates them with the value set to a String object. This code attempts to set the value to an object of the correct type.
If this code is commented out, then when executing the ApplyScorecardModelIntegrationTest or the ApplyDecisionTreeModelIntegrationTest (both found in droolsjbpm-integration/kie-server-parent/kie-server-tests/kie-server-integ-tests-drools), the tests will fail when using JSON as the MarshallingFormat. The failure will occur when the generated rules attempt to create a new object, based upon the class in ParameterInfo.type, but with a String value.

PR #2259, opened 2019-02-18
DROOLS-3661 Add reproducer

baldimir:


mariofusco:
Reproducer included by https://github.com/kiegroup/drools/commit/426130b76d7d460dcc61f5bd8dabc37f254be52d

PR #2261, opened 2019-02-18
DROOLS-3670: Test now() between two date times

jomarko:
@tarilabs could you please have a look? I created the "is office open.dmn"  using our DMN designer I and would expect the decision "is open" will return false but it is returning "null". Am I missing something?


jomarko:
@tarilabs Should I just add `@Ignore` annotation to the added test so we can merge the PR?

tarilabs:
I would prefer to cherry-pick this commit when I will work for the fix, thx

tarilabs:
I included your commit in https://github.com/kiegroup/drools/pull/2267
I believe you can close this PR as superseeded by the above

jomarko:
@tarilabs thank you.

PR #2264, opened 2019-02-20
[JBPM-8097] jBPMHelper no longer sets default values for H2

elguardian:
you can specify the datasource name now
remove asserts as some code can be use as examples (jbpm-examples)

### DIFF
--- a/kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
+++ b/kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
@@ -27,11 +27,8 @@ import javax.persistence.Persistence;
 
 import org.h2.tools.DeleteDbFiles;
 import org.h2.tools.Server;
-import org.junit.Assert;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.junit.Assert.assertNotNull;
 
 public class PersistenceUtil {
 
@@ -122,12 +119,21 @@ public class PersistenceUtil {
     }
 
     /**
+     * This method creates default pooling datasource
+     * @return a PoolingDataSource
+     */
+    public static PoolingDataSourceWrapper setupPoolingDataSource() {
+        return setupPoolingDataSource(getDatasourceProperties());
+    }
+
+    /**
      * This method uses the "jdbc/testDS1" datasource, which is the default.
      * @param dsProps The properties used to setup the data source.
      * @return a PoolingDataSource
      */
     public static PoolingDataSourceWrapper setupPoolingDataSource(Properties dsProps) {
-        return setupPoolingDataSource(dsProps, "jdbc/testDS1");
+        String datasourceName = dsProps.getProperty("datasourceName", "jdbc/testDS1");
+        return setupPoolingDataSource(dsProps, datasourceName);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
```diff
@@ -121,13 +118,23 @@ public static void cleanUp(Map<String, Object> context) {
 
     }
 
+    /**
+     * This method uses the "jdbc/testDS1" datasource, which is the default.
```
MarianMacik: Well, this is only true if datasourceName property is missing.

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
```diff
@@ -121,13 +118,23 @@ public static void cleanUp(Map<String, Object> context) {
...
+     * @param dsProps The properties used to setup the data source.
+     * @return a PoolingDataSource
+     */
+    public static PoolingDataSourceWrapper setupPoolingDataSource() {
+        return setupPoolingDataSource(getDatasourceProperties());
+    }
+
     /**
      * This method uses the "jdbc/testDS1" datasource, which is the default.
```
MarianMacik, line 130: Same here.

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
```diff
@@ -164,7 +171,7 @@ private static Properties getDefaultProperties() {
                     "className", "maxPoolSize", "allowLocalTransactions" };
             String[] defaultPropArr = { "", "", "", "jdbc:h2:tcp://localhost/JPADroolsFlow", "sa", "", "org.h2.Driver",
                     "org.h2.jdbcx.JdbcDataSource", "16", "true" };
-            Assert.assertTrue("Unequal number of keys for default properties", keyArr.length == defaultPropArr.length);
+            assert keyArr.length == defaultPropArr.length : "Unequal number of keys for default properties";
```
baldimir: Is there a reason for using java asserts here instead of Assert class?

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
```diff
@@ -188,7 +195,7 @@ public static Properties getDatasourceProperties() {
         System.setProperty("h2.lobInDatabase", "true");
 
         InputStream propsInputStream = PersistenceUtil.class.getResourceAsStream(DATASOURCE_PROPERTIES);
-        assertNotNull(propertiesNotFoundMessage, propsInputStream);
+        assert propsInputStream != null : propertiesNotFoundMessage;
```
baldimir: Same question as in previous comment.

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
elguardian: cannot be uses by jbpm-examples as those are not in test scope

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
elguardian: cannot be uses by jbpm-examples as those are not in test scope

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
baldimir: Please change it to something like:
```
if (keyArr.length != defaultPropArr.length) {
    throw new IllegalStateException("Unequal number of keys for default properties!");
}
```
or remove it completely please. The problem is that java asserts need to be enabled by a command line argument, so they are not active by default.

kie-test-util/src/main/java/org/kie/test/util/db/PersistenceUtil.java
baldimir: Please change it to something like:
```
if (propsInputStream == null) {
    throw new IllegalStateException(propertiesNotFoundMessage);
}
```
or remove it completely please. The problem is that java asserts need to be enabled by a command line argument, so they are not active by default.

elguardian:
done
@MarianMacik

baldimir (approved):
Approving if green.

MarianMacik (approved):

PR #2267, opened 2019-02-25
DROOLS-3670 DMN `between` FEEL operator alignments

tarilabs:

/cc @jomarko

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
@@ -1854,6 +1854,7 @@ public class DMNRuntimeTest extends BaseInterpretedVsCompiledTest {
 
     @Test
     public void testNowBetweenTwoDates() {
+        // DROOLS-3670 DMN `between` FEEL operator alignments
         final DMNRuntime runtime = DMNRuntimeUtil.createRuntime("is office open.dmn", this.getClass());
         final DMNModel dmnModel = runtime.getModel("https://github.com/kiegroup/drools/kie-dmn/_19170B18-B561-4EB2-9D38-714E2442710E",
                                                    "is office open");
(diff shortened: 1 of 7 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -1852,6 +1852,25 @@ public void testVerifyExtendedKieFEELFunction_now() {
...
+
+        final DMNContext context = DMNFactory.newContext();
+        context.set("an office", prototype(entry("opened from", LocalDateTime.of(2018, 12, 31, 8, 0)),
+                                           entry("opened till", LocalDateTime.of(2018, 12, 31, 16, 0))));
+
+        final DMNResult dmnResult = runtime.evaluateAll(dmnModel, context);
+        assertThat(DMNRuntimeUtil.formatMessages(dmnResult.getMessages()), dmnResult.hasErrors(), is(false));
+        LOG.debug("{}", dmnResult);
+        assertThat(dmnResult.getDecisionResultByName("is open").getEvaluationStatus(), is(DecisionEvaluationStatus.SUCCEEDED));
+        assertThat(dmnResult.getDecisionResultByName("is open").getResult(), notNullValue());
```
jomarko, line 1871: Shouldn't we check also the value? To me the `notNullValue()` doesn't seem to be sufficient.

jomarko (commented):
Just one question.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
tarilabs: behind `is open` decision there is the `now()` function invocation which is not deterministic, because it depends on the machine system clock. 
This change transform the test into a deterministic version, by ensuring the result is evaluated without error and is not-`null`, which by 3vl of the FEEL spec would be a valid value in other cases.

jomarko (approved):
@tarilabs ok, thank you.

baldimir (approved):

PR #2272, opened 2019-03-04
[DROOLS-3716] Use jboss annotations api instead of javax

sutaakar:
To align with WildFly/EAP


sutaakar:
@baldimir Please take a look

baldimir (approved):

PR #2274, opened 2019-03-06
[PLANNER-1433] Add or clause test with constraint match

Fonifan:
Test reproducing issue described in [PLANNER-1433](https://issues.jboss.org/browse/PLANNER-1433).

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -163,12 +163,6 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.optaplanner</groupId>
-      <artifactId>optaplanner-core</artifactId>
-      <version>7.18.0-SNAPSHOT</version>
-      <scope>test</scope>
-    </dependency>
 
   </dependencies>
 
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -163,7 +163,13 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
-    
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-core</artifactId>
+      <version>7.18.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


drools-compiler/pom.xml
```diff
@@ -163,7 +163,13 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
-    
+    <dependency>
```
rsynek: Drools should never depend on OptaPlanner; it's the opposite - OptaPlanner depends on Drools. This way you would introduce a circular dependency.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
```diff
@@ -9134,4 +9135,34 @@ public void testMvelJitDivision() {
 
         assertEquals(1, fired);
     }
+
+    @Test
+    public void testOrClause() {
+        //PLANNER-1433
+        String drl = "package org.drools.compiler.integrationtests;\n"
+                + "import " + Person.class.getName() + ";\n"
+                + "import " + HardSoftScoreHolder.class.getName() + ";\n"
```
rsynek: Could you please replace the HardSoftScoreHolder reference by a very simple mock to avoid declaring a dependency on OptaPlanner?

rsynek (changes requested):

drools-compiler/src/test/java/org/drools/compiler/integrationtests/Misc2Test.java
Fonifan: Surely.

rsynek:
@Fonifan a rule of thumb when you create a PR with a reproducer of some issue:
- put the issue ID in the commit message as follows: "[PLANNER-1433] Add or clause test with constraint match"
- provide a reasonable description of the PR, e.g. informing it's a reproducer for (a link to the issue)

This way people reviewing the PR can quickly get the context, and, as a result, you will receive the feedback sooner.

PR #2276, opened 2019-03-07
[PLANNER-1433] Add or clause bug reproducer

Fonifan:
This test reproduces issue described in [PLANNER-1433](https://issues.jboss.org/browse/PLANNER-1433).

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/OrClauseTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/OrClauseTest.java
@@ -1,0 +1,36 @@
+package org.drools.compiler.integrationtests;
+
+import org.drools.compiler.Person;
+import org.junit.Test;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieSession;
+import org.kie.internal.utils.KieHelper;
+
+import static org.junit.Assert.assertEquals;
+
+public class OrClauseTest {
+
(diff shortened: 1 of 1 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/OrClauseTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieSession;
+import org.kie.internal.utils.KieHelper;
+
+import static org.junit.Assert.assertEquals;
+
+public class OrClauseTest {
+
+    @Test
+    public void testOrClause() {
```
rsynek, line 14: @Fonifan would you please rename the test method to describe precisely what you're testing?

e.g: "twoConditionsInOrClause_shouldFireRuleOnce"

rsynek (changes requested):

mariofusco:
The rule is firing twice and this is not a bug but just how Drools is intended to work. In Drools a rule with a 'or' is totally equivalent to having 2 separated rules. I agree that this is questionable, but we cannot change this, first because it would be very hard to fire the rule only once, and second because if we do so we will break backward compatibility.

rsynek:
Closing this PR as the behaviour we are experiencing turned out to be expected and the fix must be done in OptaPlanner.

PR #2282, opened 2019-03-14
DROOLS-3715 Fix kjar path handling on Windows

baldimir:
It is a bit harder to create a reproducer for this, so I tested it manually with SpringBoot reproducer from https://issues.jboss.org/browse/JBPM-8107. I will try to create a regression test for this next. 

@mariofusco @mswiderski could you please review?

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ZipKieModule.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ZipKieModule.java
@@ -139,7 +139,7 @@ public class ZipKieModule extends AbstractKieModule implements InternalKieModule
                 }); 
             } 
         } else if (urlPath.indexOf( '!' ) > 0) {
-            urlPath = urlPath.substring( urlPath.lastIndexOf( '!' ) + 1 );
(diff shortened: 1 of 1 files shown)


baldimir:

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ZipKieModule.java
```diff
@@ -139,7 +139,7 @@ private void indexZipFile(java.io.File jarFile) {
                 }); 
             } 
         } else if (urlPath.indexOf( '!' ) > 0) {
-            urlPath = urlPath.substring( urlPath.lastIndexOf( '!' ) + 1 );
+            urlPath = urlPath.substring( urlPath.lastIndexOf( '!' ) + 1 ).replace("\\", "/");
```
lukaszs7, line 142: Shouldn't be replaceAll ?

baldimir: No, the only difference between replace() and replaceAll() is that replaceAll() uses regular expression as a parameter. Please see JavaDoc for more information [1].

[1] https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replace-java.lang.CharSequence-java.lang.CharSequence-

PR #2287, opened 2019-03-26
DROOLS-3795 : Added new row with action in between of guided decision table is replicating in source for other below rows

Rikkola:
https://issues.jboss.org/browse/DROOLS-3795

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2010 Red Hat, Inc. and/or its affiliates.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
  * the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
@@ -23,7 +23,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.StringTokenizer;
 
 import org.drools.workbench.models.datamodel.rule.ActionExecuteWorkItem;
 import org.drools.workbench.models.datamodel.rule.ActionFieldList;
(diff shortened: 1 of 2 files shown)


Rikkola:
Issues with running tests. I suspect this is due to setup in the tests, since workbench works fine. 

```
[ERROR]   BRLRuleModelTest.testUpdateModifyMultipleFieldsWithMultipleSkipped4:1110->assertEqualsIgnoreWhitespace:1783
Expecting:
  <"//from row number: 1
//desc
rule "Row 1 null"
dialect "mvel"
when
  x : Context( )
then
  modify( x ) {
    setF1( "v1" ),
    setF3( "v3" )
}
end
">
to be equal to:
  <"//from row number: 1
//desc
rule "Row 1 null"
        dialect "mvel"
        when
                x : Context( )
        then
                x.setF1( "v1" );
                x.setF3( "v3" );
end

">
```

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java
```diff
@@ -980,4 +983,11 @@ private FieldConstraint makeSingleFieldConstraint(ConditionCol52 c,
                                            allColumns,
                                            data);
     }
+
+    private class LabelledAction {
+
+        String boundName;
+        boolean isUpdate;
+        IAction action;
+    }
```
Rikkola: I'll move this back to it's old location.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java
```diff
@@ -324,11 +323,22 @@ private void addAction(IAction action,
...
         } else if (action instanceof ActionSetField) {
-            final ActionSetField af = (ActionSetField) action;
-            binding = af.getVariable();
+            final ActionSetField original = (ActionSetField) action;
+
+            final ActionSetField asf = new ActionSetField(original.getVariable());
+            asf.setFieldValues(original.getFieldValues());
+
+            action = asf;
+            binding = original.getVariable();
```
Rikkola, line 348: Need to make new copies or else the changes are copied to each row below the current one.

manstis: Hi @Rikkola can you explain why changing how _columns_ are created affects _row_ data?

My suspicion would have been where we generate a single rule from a row data.. something is cached and re-used. I know this PR is _draft_ so thought it best to ask now rather than later.

Rikkola: @manstis the action here comes from the column header. So later when we add field values to that same action they are also added into the BRL header. This then copies the action for each row below.

Still giving this more attention, but made an early PR so @jomarko can take a look.

jomarko (commented):
@Rikkola Hi, the fix is step in a good way, now just the appropriate rows in generated rules are affected by row addition into the grid, however we lost the `modify()` in the generated rules. I mean I have column declared as `Modify value of ...` however the generated source contains just `$myObject.setValue...` there is no `modify($myObject)`.

Rikkola:
@jomarko yes the tests were picking this up. Updated the PR with changes and all tests should now pass.

Rikkola:
@manstis ok now I could use a review please.

The problem is that unless the action are cloned the original is used and edited later on. This is not an issue when the action is defined on the row, but it is if the action was defined in the header. BRL columns for example define the actions in the header. This means the table content is now changed and the ActionSetField actions for the same object are change to "modify" even if the column header says only set should be done. It also copies the ActionSetField for every row after the problem row.

All in all a tricky issue that only happens when both BRL action and set/update column exist for the same object.

jomarko (approved):
@Rikkola thank you!

MarianMacik:
jenkins retest this

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTDRLPersistence.java
manstis: Thanks for the explanation @Rikkola. It makes perfect sense.

manstis (approved):

PR #2290, opened 2019-03-28
RHPAM-1815 Renamed the file ScoringStrategiesTest.java...

lanceleverich:
... to ScorecardScoringStrategiesTest.java. This makes it the last set of tests to be executed, which fixes the referenced issue, that only occurs when building in a Windows environment. It is unclear what the tests do that causes the issue, as the issue only shows up when the tests contained in the renamed file are executed before other tests in the same module (drools-scorecards).

### DIFF
--- a/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/compiler/FieldMarker.java
+++ b/kie-pmml/src/main/java/org/kie/pmml/pmml_4_2/compiler/FieldMarker.java
@@ -1,0 +1,65 @@
+package org.kie.pmml.pmml_4_2.compiler;
+
+import org.dmg.pmml.pmml_4_2.descr.DATATYPE;
+import org.kie.api.definition.type.Key;
+
+public class FieldMarker {
+
+    @Key
+    private String name;
+    private DATATYPE dataType;
+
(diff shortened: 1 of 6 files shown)


baldimir:
This doesn't solve the issue. It just reorders the test execution. Is that correct?

jiripetrlik:
@lanceleverich 
Would it be possible to investigate further this issue to find a root cause? In this PR we hide the problem not solve it.

lanceleverich:
From what I have been able to figure out, the problem is somehow related to how JUnit and the class loader are behaving in the Windows environment. The individual tests all execute without issues if the command 'mvn clean test  -Dtest=_nameOfTestClass_' is used. I can spend more time digging into the issue, but it does not appear that this is an issue related to the lack of case-support for file names, as I originally thought it was.

lanceleverich:
New update puts the old file name back on ScoringStrategiesTest, and replaces some **declare** statements that were in the rules that are used to generate the rules from PMML. This means there is no longer the reliance on order of execution for the tests.

baldimir (approved):
Please remove the commented out declares. Otherwise looks good. Thanks. Approving if green.

mareknovotny:
we need this to be also on 7.18.x branch as today is the cut off

PR #2297, opened 2019-04-02
DROOLS-3836 Handle possible null date value returned from trigger

baldimir:
JIRA: https://issues.jboss.org/browse/DROOLS-3836

@mariofusco could you please review? Especially the serialization parts in ProtobufOutputMarshaller and ObjectTypeNode.ExpireJobContextTimerOutputMarshaller. 

- Fixes the serialization - the trigger can return null from hasNextFireTime() method. 
- I also fixed other occurrences where the possible null was not handled properly. 
- Added test case reproducing the NPE. However it was a race condition between the expiration timer and the main thread, so to reproduce the issue it was necessary to run the reproducer in a loop.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufOutputMarshaller.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufOutputMarshaller.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufOutputMarshaller.java
```diff
@@ -834,12 +834,15 @@ public int compare(TimerJobInstance o1,
...
+                    ProtobufMessages.Trigger.PointInTimeTrigger.newBuilder();
+            if (pitTrigger.hasNextFireTime() != null) {
+                pitBuilder.setNextFireTime(pitTrigger.hasNextFireTime().getTime());
+            }
             return ProtobufMessages.Trigger.newBuilder()
                     .setType( ProtobufMessages.Trigger.TriggerType.POINT_IN_TIME )
-                    .setPit( ProtobufMessages.Trigger.PointInTimeTrigger.newBuilder()
-                            .setNextFireTime( pinTrigger.hasNextFireTime().getTime() )
-                            .build() )
+                    .setPit( pitBuilder.build() )
```
mariofusco: I'm not sure of the impact of this, but if the trigger doesn't have a next fire time I think it is pointless to unmarshall this at all. In other words in this case the method should just return null (and probably the null should be managed upstream) instead of returning a trigger without any fire time.

drools-core/src/main/java/org/drools/core/reteoo/ObjectTypeNode.java
```diff
@@ -663,8 +663,9 @@ public void write(JobContext jobCtx,
 
             DefaultJobHandle jobHandle = (DefaultJobHandle) ejobCtx.getJobHandle();
             PointInTimeTrigger trigger = (PointInTimeTrigger) jobHandle.getTimerJobInstance().getTrigger();
-            outputCtx.writeLong( trigger.hasNextFireTime().getTime() );
-
+            if (trigger.hasNextFireTime() != null) {
```
mariofusco: I believe that if we change the unmarshall as I suggested also this check could be avoided.

drools-core/src/main/java/org/drools/core/reteoo/ObjectTypeNode.java
```diff
@@ -675,13 +676,16 @@ public void write(JobContext jobCtx,
...
             return ProtobufMessages.Timers.Timer.newBuilder()
-                                                .setType( ProtobufMessages.Timers.TimerType.EXPIRE )
-                                                .setExpire( ProtobufMessages.Timers.ExpireTimer.newBuilder()
-                                                                                               .setHandleId( expireAction.getFactHandle().getId() )
-                                                                                               .setNextFireTimestamp( trigger.hasNextFireTime().getTime() )
-                                                                                               .build() )
-                                                .build();
+                    .setType(ProtobufMessages.Timers.TimerType.EXPIRE)
+                    .setExpire(expireTimerBuilder.build())
+                    .build();
```
mariofusco: This is the dual: a trigger without a next fire time shouldn't be marshalled at all.

drools-core/src/main/java/org/drools/core/time/impl/PseudoClockScheduler.java
```diff
@@ -218,7 +218,7 @@ private synchronized long runCallBacksAndIncreaseTimer( long increase ) {
 
     public synchronized long getTimeToNextJob() {
         TimerJobInstance item = (TimerJobInstance) queue.peek();
-        return (item != null) ? item.getTrigger().hasNextFireTime().getTime() - this.timer.get() : -1;
+        return (item != null && item.getTrigger().hasNextFireTime() != null) ? item.getTrigger().hasNextFireTime().getTime() - this.timer.get() : -1;
```
mariofusco: This should also never happen.

mariofusco:
@baldimir I added a few comments to your PR, but the TLDR of this is that is pointless to keep in the system a trigger without a fire time.

baldimir:
@mariofusco I updated the PR according to our conversation on IRC. I still handle some nulls because when serializing, the timers are running (we should think about if this is a problem). I also did that a trigger is not created when its timestamp doesn't fall into any calendar.

PR #2310, opened 2019-04-16
DROOLS-3841: [DMN Designer] Included Models - An error is raised when a DMN file with imports is saved

manstis:
See https://issues.jboss.org/browse/DROOLS-3841

@tarilabs @mariofusco This PR shows the error reported in the referenced JIRA.

It comes about by the Workbench adding the second DMN file (the one _importing_ the other) is added incrementally to `KieBuilder`. This is how the workbench builds projects incrementally...  A full build is OK.


manstis:
IMO it appears `DMNAssemblerService` is at fault.

When handling a single additional resource it fails to try and resolve the single resources possible import dependency. Simply commenting out the "quick path" reveals that the issue is slightly more complex as presumably `ImportDMNResolverUtil` does not consider DMN files already assembled!?

tarilabs:
IMO `KieAssemblerService#addResources` was added to cover build order, but it seems to me by the API usage you have in the test, that the KieAssembler is not covering this use case, when there is a dependency from a previously added resources which is either: not passed to the collection of resources (again) nor I have a notification of removed assets (as they been compiled but not removed)

manstis:
@tarilabs PR updated to include your PR on my fork.

Could you please approve?

tarilabs (approved):

jomarko (approved):
Thank you, checked with workbench where included models were not turned off and it fixed the issue.

PR #2314, opened 2019-04-23
DROOLS-3927 DMN compile error in codegen when maven plugin and expr in

tarilabs:
input clause


baldimir (approved):
Approving if green. You could eventually add the test DMN files here https://github.com/kiegroup/droolsjbpm-integration/tree/master/kie-plugins-testing/src/test/projects/kjar-7-with-dmn. I recommend adding the test DMN there if the codegen refactor is not going to happen in near future.

PR #2319, opened 2019-04-30
DROOLS-3779 DMN DT Analysis Subsumption&Contraction for DMN Decision Table

tarilabs:


tarilabs:
```
[Message [id=1, level=ERROR, path=pom.xml, line=0, column=0
   text=maven pom.xml found, but unable to read
org.apache.maven.project.ProjectBuildingException: 1 problem was encountered while building the effective model
[FATAL] Non-readable POM : input contained no data @ 
```

Jenkins test again

tarilabs:
>  just out of interest what's the reason for changing from List to Collection (and then a concrete HashSet)?

Program against interfaces the most generic interface is preferable when possible, for the former question.
Faster lookups for the latter question.

> Did the list contain duplicate entries in some circumstances?

I don't expect this scenario, dt rule are scanned once.

> But then again IIRC you don't implement hashCode() and equals(..) on your Subsumption or Contraction classes?!

Correct, I don't need to define equivalence for those classes, at least for now.

baldimir (approved):

PR #2332, opened 2019-05-07
DROOLS-3432 : While using the same template key with difference objects/attributes causes an exception when switching to the Data tab in a Guided Rule Template

Rikkola:
https://issues.jboss.org/browse/DROOLS-3432
If the same variable is used for two fields the UI defaults to String cell editor.

PRs:
https://github.com/kiegroup/drools/pull/2332
https://github.com/kiegroup/drools-wb/pull/1134

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/util/InterpolationVariableCollector.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/util/InterpolationVariableCollector.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.models.datamodel.rule.util;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.TreeMap;
+
+import org.drools.workbench.models.datamodel.rule.InterpolationVariable;
+
+import static org.drools.workbench.models.datamodel.rule.RuleModel.DEFAULT_TYPE;
+
+public class InterpolationVariableCollector {
+
+    private final String defaultType;
+    private final Map<Integer, InterpolationVariable> byColumn = new TreeMap<Integer, InterpolationVariable>();
+
+    public InterpolationVariableCollector(final Map<InterpolationVariable, Integer> map,
+                                          final String defaultType) {
+        this.defaultType = defaultType;
+        removeDuplicates(map);
+    }
+
+    public InterpolationVariableCollector(final Map<InterpolationVariable, Integer> map) {
+        this(map,
+             DEFAULT_TYPE);
+    }
+
+    private void removeDuplicates(Map<InterpolationVariable, Integer> map) {
+        for (Map.Entry<InterpolationVariable, Integer> entry : map.entrySet()) {
+
+            final InterpolationVariable newVar = entry.getKey();
+            final Optional<InterpolationVariable> oldColumnData = findByVariable(byColumn,
+                                                                                 newVar);
+
+            if (oldColumnData.isPresent()) {
+                if (isThereANeedToSetDefaultType(newVar, oldColumnData.get())) {
+                    oldColumnData.get().setDataType(defaultType);
+                }
+            } else {
+                InterpolationVariable variable = new InterpolationVariable(newVar.getVarName(),
+                                                                           newVar.getDataType(),
+                                                                           newVar.getFactType(),
+                                                                           newVar.getFactField());
+                variable.setOperator(newVar.getOperator());
+                byColumn.put(entry.getValue(),
+                             variable);
+            }
+        }
+    }
+
+    public boolean isThereANeedToSetDefaultType(final InterpolationVariable newVar,
+                                                final InterpolationVariable oldVar) {
+        if (Objects.equals(newVar.getDataType(), oldVar.getDataType())
+                && Objects.equals(newVar.getFactType(), oldVar.getFactType())
+                && Objects.equals(newVar.getFactField(), oldVar.getFactField())) {
+            return false;
+        } else {
+            return true;
+        }
+    }
+
+    private Optional<InterpolationVariable> findByVariable(final Map<Integer, InterpolationVariable> result,
+                                                           final InterpolationVariable newVar) {
+        for (final InterpolationVariable columnData : result.values()) {
+            if (Objects.equals(columnData.getVarName(), newVar.getVarName())) {
+
+                return Optional.of(columnData);
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    public Map<InterpolationVariable, Integer> getMap() {
+        final HashMap<InterpolationVariable, Integer> result = new HashMap<InterpolationVariable, Integer>();
+
+        int index = 0;
+        for (Integer integer : byColumn.keySet()) {
+            result.put(byColumn.get(integer),
+                       index);
+            index++;
+        }
+
+        return result;
+    }
+}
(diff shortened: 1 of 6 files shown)


Rikkola:
@jomarko @manstis yet again something to review.

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -144,6 +148,17 @@ public int getColsCount() {
         return result;
     }
 
+    private Optional<InterpolationVariable> findByVariable(final Map<Integer, InterpolationVariable> result,
+                                                           final InterpolationVariable newVar) {
+        for (final InterpolationVariable columnData : result.values()) {
```
manstis: You could use streams/filter, but this works as just well.

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -144,6 +148,17 @@ public int getColsCount() {
         return result;
     }
 
+    private Optional<InterpolationVariable> findByVariable(final Map<Integer, InterpolationVariable> result,
```
manstis: This method should probably live in the `Collector` class?

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -256,4 +271,43 @@ public void setValue(String varName,
...
+
+        private void removeDuplicates(Map<InterpolationVariable, Integer> map) {
+            for (Map.Entry<InterpolationVariable, Integer> entry : map.entrySet()) {
+
+                final InterpolationVariable newVar = entry.getKey();
+                final Optional<InterpolationVariable> oldColumnData = findByVariable(byColumn,
+                                                                                     newVar);
+
+                if (oldColumnData.isPresent()) {
+                    oldColumnData.get().setDataType(DEFAULT_TYPE);
```
manstis: During testing of `drools-wb-webapp` I found if I added two Template Keys for the same data-type it was being overridden to `DEFAULT_TYPE` whereas, if the data-types are the same we can keep it "as is". For example, I added `Applicant.age < $default` and `Applicant.age >= $default` (OK a stupid rule!) and the data-grid default to a `TextBox` when a widget for `Integer` was perfectly valid.

manstis: And possibly also check if it's the same Fact/Field too in case someone does something like this:
```
$a1 : Applicant( creditRating = $default ) 
Applicant( this != $a1, creditRating = $default, age > 100 )
```
and it's an enumeration!? (just a thought, not tried whether an enum selector would still be presented?!)

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
Rikkola: It should

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
Rikkola: The lower example actually works. The above does not since the operator is not the same. I'll fix that.
Enumerations can be used, but I worry about how dependent enums work. I will check that.

Rikkola:
@manstis Updated the PR with requested changes.

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -256,4 +261,71 @@ public void setValue(String varName,
...
+                                 variable);
+                }
+            }
+        }
+
+        public boolean isThereANeedToSetDefaultType(final InterpolationVariable newVar,
+                                                    final InterpolationVariable oldVar) {
+            if (newVar.getDataType().equals(oldVar.getDataType())
+                    && Objects.equal(newVar.getFactType(), oldVar.getFactType())
+                    && Objects.equal(newVar.getFactField(), oldVar.getFactField())) {
```
jomarko: Please use `java.util.Objects` and please use `equals` in the same way for all three lines.

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -256,4 +261,71 @@ public void setValue(String varName,
...
+                return false;
+            } else {
+                return true;
+            }
+        }
+
+        private Optional<InterpolationVariable> findByVariable(final Map<Integer, InterpolationVariable> result,
+                                                               final InterpolationVariable newVar) {
+            for (final InterpolationVariable columnData : result.values()) {
+                if (columnData.getVarName().equals(newVar.getVarName())) {
```
jomarko: Could we use `Objects.equals` also here please?

drools-workbench-models/drools-workbench-models-guided-template/src/test/java/org/drools/workbench/models/guided/template/shared/TemplateModelTest.java
```diff
@@ -49,4 +50,89 @@ public void testAddRowInvalidColumnCount() throws Exception {
...
+
+        assertEquals(2, m.getInterpolationVariablesList().length);
+        assertEquals("Integer", m.getInterpolationVariablesList()[0].getDataType());
+        assertEquals("String", m.getInterpolationVariablesList()[1].getDataType());
+    }
+
+    private SingleFieldConstraint makeConstraint(final String factType,
+                                                 final String fieldName,
+                                                 final String fieldType,
+                                                 final String variableName, String operator) {
```
jomarko: Wouldn't be the `operator` nicer on separate line, and `final`?

drools-workbench-models/drools-workbench-models-guided-template/src/main/java/org/drools/workbench/models/guided/template/shared/TemplateModel.java
```diff
@@ -256,4 +261,71 @@ public void setValue(String varName,
...
+                                 variable);
+                }
+            }
+        }
+
+        public boolean isThereANeedToSetDefaultType(final InterpolationVariable newVar,
+                                                    final InterpolationVariable oldVar) {
+            if (newVar.getDataType().equals(oldVar.getDataType())
+                    && Objects.equals(newVar.getFactType(), oldVar.getFactType())
+                    && Objects.equals(newVar.getFactField(), oldVar.getFactField())) {
```
jomarko: Is there a reason why just 2/3 lines use `Objects.equals` ?

jomarko (commented):
Code review comments in. Manual check pending.

drools-workbench-models/drools-workbench-models-guided-template/src/test/java/org/drools/workbench/models/guided/template/shared/TemplateModelTest.java
Rikkola: It would, but I have a good reason. Too much beer too soon.

jomarko (approved):
Approving and increasing coverage https://github.com/kiegroup/drools/pull/2364

jomarko:
@manstis hi, is your review still pending?

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/util/InterpolationVariableCollector.java
```diff
@@ -0,0 +1,102 @@
...
+                variable.setOperator(newVar.getOperator());
+                byColumn.put(entry.getValue(),
+                             variable);
+            }
+        }
+    }
+
+    public boolean isThereANeedToSetDefaultType(final InterpolationVariable newVar,
+                                                final InterpolationVariable oldVar) {
+        if (newVar.getDataType().equals(oldVar.getDataType())
```
manstis: I'm surprised @jomarko didn't report this! 

It'd be more consistent to use `Objects.equals(...)` on all three checks.

manstis (commented):
@Rikkola Just a tiny comment.. otherwise good for me. Thank-you.

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/util/InterpolationVariableCollector.java
Rikkola: @manstis update done.

manstis (approved):

PR #2346, opened 2019-05-14
DROOLS-4021 Fix executable model file path on Windows

baldimir:
@mariofusco @lucamolteni please review. 

The original form was causing all executable model tests to fail when run on Windows. The reason is that our memory filesystem doesn't support Windows file separator.


lucamolteni:
@baldimir I still Paths.get is a better API than concatenating String.

Isn't it possible to get the unix String from a windows Path created using Paths.get?

baldimir:
I don't know about such option in Paths.

lucamolteni:
@baldimir then let's merge it like this thanks

PR #2349, opened 2019-05-15
Drools 4035 Remove kie-soup-project-datamodel-api dependency from drools-scenario-simulation-api

gitgabrio:
@danielezonca @yesamer  @jomarko 

See https://issues.jboss.org/browse/DROOLS-4035

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-api/pom.xml
+++ b/drools-scenario-simulation/drools-scenario-simulation-api/pom.xml
@@ -19,11 +19,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.kie.soup</groupId>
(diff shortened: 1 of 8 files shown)


jomarko (changes requested):
This is incompatible change, that breaks the class `org.drools.workbench.screens.scenariosimulation.client.editor.ScenarioSimulationEditorPresenter` and maybe others from `drools-wb` repository.

yesamer:
> This is incompatible change, that breaks the class `org.drools.workbench.screens.scenariosimulation.client.editor.ScenarioSimulationEditorPresenter` and maybe others from `drools-wb` repository.

Yes, I think a PR on drools-wb project is required to make that project compatible with these changes.
@gitgabrio is working on it.

PR #2350, opened 2019-05-15
[DROOLS-4028] Implemented reverse function from object to expression in Test Scenario runner

danielezonca:
This PR implement a reverse function for the expression to be able to obtain the expression starting from the actual value. It also contains a small enhancement for the message in the popover

https://issues.jboss.org/browse/DROOLS-4028

@gitgabrio @yesamer @jomarko

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
@@ -23,4 +23,6 @@ public interface ExpressionEvaluator {
     boolean evaluateUnaryExpression(Object rawExpression, Object resultValue, Class<?> resultClass);
 
     Object evaluateLiteralExpression(String className, List<String> genericClasses, Object raw);
+
+    String reverseFunction(Object value);
 }
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
@@ -24,5 +24,5 @@ public interface ExpressionEvaluator {
 
     Object evaluateLiteralExpression(String className, List<String> genericClasses, Object raw);
 
-    String reverseFunction(Object value);
+    String fromObjectToExpression(Object value);
 }
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
@@ -108,7 +108,7 @@ public abstract class AbstractRunnerHelper {
                 return ResultWrapper.createResult(entry.getValue());
             }
         }
-        return ResultWrapper.createErrorResult(null, null);
+        return ResultWrapper.createErrorResultWithErrorMessage("No direct mapping available");
     }
 
     public List<ScenarioExpect> extractExpectedValues(List<FactMappingValue> factMappingValues) {
@@ -200,12 +200,16 @@ public abstract class AbstractRunnerHelper {
         }
     }
 
-    protected ScenarioResult fillResult(FactMappingValue expectedResult, FactIdentifier factIdentifier, Supplier<ResultWrapper<?>> resultSupplier) {
+    protected ScenarioResult fillResult(FactMappingValue expectedResult,
+                                        FactIdentifier factIdentifier,
+                                        Supplier<ResultWrapper<?>> resultSupplier,
+                                        ExpressionEvaluator expressionEvaluator) {
         ResultWrapper<?> resultValue = resultSupplier.get();
 
-        if (!resultValue.isSatisfied()) {
-            expectedResult.setErrorValue(resultValue.getResult());
-
+        if (!resultValue.isSatisfied() && resultValue.getErrorMessage().isPresent()) {
+            expectedResult.setExceptionMessage(resultValue.getErrorMessage().get());
+        } else if (!resultValue.isSatisfied()) {
+            expectedResult.setErrorValue(expressionEvaluator.reverseFunction(resultValue.getResult()));
         } else {
             expectedResult.resetStatus();
         }
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
@@ -209,7 +209,7 @@ public abstract class AbstractRunnerHelper {
         if (!resultValue.isSatisfied() && resultValue.getErrorMessage().isPresent()) {
             expectedResult.setExceptionMessage(resultValue.getErrorMessage().get());
         } else if (!resultValue.isSatisfied()) {
-            expectedResult.setErrorValue(expressionEvaluator.reverseFunction(resultValue.getResult()));
+            expectedResult.setErrorValue(expressionEvaluator.fromObjectToExpression(resultValue.getResult()));
         } else {
             expectedResult.resetStatus();
         }
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
@@ -44,6 +44,7 @@ import org.kie.dmn.api.core.DMNResult;
 import org.kie.dmn.api.core.ast.DecisionNode;
 
 import static org.drools.scenariosimulation.backend.runner.model.ResultWrapper.createErrorResult;
+import static org.drools.scenariosimulation.backend.runner.model.ResultWrapper.createErrorResultWithErrorMessage;
 import static org.drools.scenariosimulation.backend.runner.model.ResultWrapper.createResult;
 import static org.kie.dmn.api.core.DMNDecisionResult.DecisionEvaluationStatus.SUCCEEDED;
 
@@ -107,7 +108,10 @@ public class DMNScenarioRunnerHelper extends AbstractRunnerHelper {
                 FactMapping factMapping = simulationDescriptor.getFactMapping(factIdentifier, expressionIdentifier)
                         .orElseThrow(() -> new IllegalStateException("Wrong expression, this should not happen"));
 
-                ScenarioResult scenarioResult = fillResult(expectedResult, factIdentifier, () -> getSingleFactValueResult(factMapping, expectedResult, decisionResult, expressionEvaluator));
+                ScenarioResult scenarioResult = fillResult(expectedResult,
+                                                           factIdentifier,
+                                                           () -> getSingleFactValueResult(factMapping, expectedResult, decisionResult, expressionEvaluator),
+                                                           expressionEvaluator);
 
                 scenarioRunnerData.addResult(scenarioResult);
             }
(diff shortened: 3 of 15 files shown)


jomarko:
@danielezonca wrong link in jira description i think

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
```diff
@@ -108,7 +108,7 @@ public void run(KieContainer kieContainer, SimulationDescriptor simulationDescri
                 return ResultWrapper.createResult(entry.getValue());
             }
         }
-        return ResultWrapper.createErrorResult(null, null);
+        return ResultWrapper.createErrorResultWithErrorMessage("No direct mapping available");
```
jomarko, line 111: I miss some test that would check this case

jomarko (commented):
@danielezonca thank you for PR, have two comments about coverage and still think the linking to jira is not appropriate. I am not sure what to verify during manual check.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
```diff
@@ -122,7 +126,10 @@ protected ResultWrapper getSingleFactValueResult(FactMapping factMapping,
         Object resultRaw = decisionResult.getResult();
         final DMNDecisionResult.DecisionEvaluationStatus evaluationStatus = decisionResult.getEvaluationStatus();
         if (!SUCCEEDED.equals(evaluationStatus)) {
-            return createErrorResult(SUCCEEDED, evaluationStatus);
+            return createErrorResultWithErrorMessage("The decision " +
```
jomarko, line 129: We miss coverage for this case

danielezonca:
@jomarko 
The Jira link is correct, I fixed the title of the PR :)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
danielezonca: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
danielezonca: Done

jomarko (approved):
Thank you

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
```diff
@@ -200,12 +200,16 @@ public void validateAssertion(List<ScenarioResult> scenarioResults, Scenario sce
...
+    protected ScenarioResult fillResult(FactMappingValue expectedResult,
+                                        FactIdentifier factIdentifier,
+                                        Supplier<ResultWrapper<?>> resultSupplier,
+                                        ExpressionEvaluator expressionEvaluator) {
         ResultWrapper<?> resultValue = resultSupplier.get();
 
-        if (!resultValue.isSatisfied()) {
-            expectedResult.setErrorValue(resultValue.getResult());
-
+        if (!resultValue.isSatisfied() && resultValue.getErrorMessage().isPresent()) {
```
gitgabrio, line 209: @danielezonca 
Could you pls semplify this
if (!!resultValue.isSatisfied()) {
     resultValue.getErrorMessage().ifPresent(_lambda)
expectedResult.setErrorValue(expressionEvaluator.reverseFunction(resultValue.getResult()));
} else ...

or something similar?

yesamer (approved):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
```diff
@@ -28,25 +29,25 @@
...
-    private ResultWrapper(T result, T expected) {
-        this.satisfied = false;
-        this.result = result;
         this.expected = expected;
+        this.errorMessage = errorMessage;
     }
 
     public static <T> ResultWrapper<T> createResult(T result) {
-        return new ResultWrapper<>(result, true);
+        return new ResultWrapper<>(result, null, true, null);
```
gitgabrio, line 42: @danielezonca 
Surely there is a reason, but why did you removed the overloaded constructors and then call the one left with two "null"s ?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
```diff
@@ -28,25 +29,25 @@
...
     }
 
     public static <T> ResultWrapper<T> createErrorResult(T result, T expected) {
-        return new ResultWrapper<>(result, expected);
+        return new ResultWrapper<>(result, expected, false, null);
+    }
+
+    public static <T> ResultWrapper<T> createErrorResultWithErrorMessage(String errorMessage) {
+        return new ResultWrapper<>(null, null, false, errorMessage);
```
gitgabrio, line 50: @danielezonca 
About same comment as above

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
```diff
@@ -169,6 +169,36 @@ public static Object convertValue(String className, Object cleanValue, ClassLoad
...
     }
 
+    public static String revertValue(Object cleanValue) {
+        if (cleanValue == null) {
+            return "null";
+        }
+
+        Class<?> clazz = cleanValue.getClass();
+
+        if (clazz.isAssignableFrom(String.class)) {
```
gitgabrio, line 179: @danielezonca 
could you please create a map/list/enum or something to 
1) avoid all this "else if"
2) whenever a new class (if ever) has to be managed that way, we have to simply add to that collection/enum ? Maybe a Map with  Function<T,R>s as values?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
```diff
@@ -23,4 +23,6 @@
     boolean evaluateUnaryExpression(Object rawExpression, Object resultValue, Class<?> resultClass);
 
     Object evaluateLiteralExpression(String className, List<String> genericClasses, Object raw);
+
+    String reverseFunction(Object value);
```
gitgabrio: @danielezonca 
Maybe a different name? "reverseFunction" makes me think to a functional interface or something similar, not to a method that retrieve a String out of a given object

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
```diff
@@ -89,4 +90,13 @@ public void expressionTest() {
         assertEquals(1, parsedMap.size());
         assertEquals(BigDecimal.valueOf(1), ((Map<String, Object>) parsedMap.get("first").get("phones")).get("number"));
     }
+
+    @Test
+    public void reverseFunction() {
+        assertEquals("\"Test\"", expressionEvaluator.reverseFunction("Test"));
+        assertEquals("false", expressionEvaluator.reverseFunction(false));
+        assertEquals("1", expressionEvaluator.reverseFunction(BigDecimal.valueOf(1)));
+        assertEquals("date( \"0001-01-01\" )", expressionEvaluator.reverseFunction(LocalDate.of(1, 1, 1)));
```
gitgabrio: @danielezonca 
does this mean that the user has to insert "date( \"0001-01-01\" )" in a LocalDate cell, with yyyy-?-? format?
Beside that, using all "1" does not makes clear what should be the month and what the day

gitgabrio (changes requested):
@danielezonca 
Mostly style comments and a doubt about LocalDate

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
danielezonca: I updated the test with `date("2019-05-13")` to be not ambiguous but keep in mind that this is FEEL `date` function so the format is defined in the standard and is `yyyy-mm-dd`

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
danielezonca: I changed it in `fromObjectToExpression`. Wdyt?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
danielezonca: Yes I tried to do something similar but I don't like the result, it requires much more glue code just to have a "nicer invocation"
https://gist.github.com/danielezonca/593473e44a441d31406990240b4e28e8
Considering that I don't plan to have a pluggable mechanism but this list will stay hardcoded I chose to have this list of `if-else`.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
danielezonca: Same answer as above :)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
danielezonca: Considering that `ResultWrapper` can be initialized only via factory methods I decided to simplify the internal constructor otherwise we had 3 different factory methods and 3 different constructors.
I don't like `null` as explicit parameter but considering I think it is reasonable when hidden inside the class

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
danielezonca: I know you don't like `isPresent` method **:-P** but I cannot simplify the condition because I need to do three different actions:
- result failed and there is an error message -> propagate error message
- result failed and no error message -> propagate actual data
- result success -> clean error state

gitgabrio: @danielezonca 
ok

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
gitgabrio: @danielezonca 
ok

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ResultWrapper.java
gitgabrio: @danielezonca 
As above...

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
gitgabrio: @danielezonca 
ok, gotcha

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluator.java
gitgabrio: @danielezonca 
Thanks!

gitgabrio (approved):
@danielezonca 
fine for me

danielezonca:
@manstis 

Can you please squash and merge this?

PR #2355, opened 2019-05-21
[DROOLS-3806] Improve error reporting messages + minor bug

danielezonca:
This PR cover what is described in the corresponding ticket ( https://issues.jboss.org/browse/DROOLS-3806 ) plus some other improvements in error reporting area:
- (DMN) now if FEEL has a parsing error in Given section the error is properly reported and the evaluation is stopped
- (RULE) when an unsupported operator is used in Given section (i.e. `> 3`) a proper message is provideda and the cell is highlighted
- (General) alert error messages improvement: i.e. if the specified KieSession doesn't exist the error is what it come from the KieContainer (` Impossible to find a KieSession with name test`)
- (General) migrate scenario activator Junit class for project created before backend/api refactoring (move some of them into `drools` repo). You just need to open and save one scesim file
- (DMN) minor performance improvement on model evaluation: moved to new `KieRuntimeFactory` API to avoid creation of an unnecessary `KieSession` just to obtain `DMNRuntime` 

@jomarko @gitgabrio @Rikkola 

NOTE: this PR should fix also https://issues.jboss.org/browse/DROOLS-4059

PS: PR description is probably longer than code changes :)

PRs list:
- https://github.com/kiegroup/drools-wb/pull/1146
- https://github.com/kiegroup/drools/pull/2355

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
@@ -169,7 +169,7 @@ public enum BaseExpressionOperator {
     protected abstract boolean eval(Object rawValue, Object resultValue, Class<?> resultClass, ClassLoader classLoader);
 
     protected Object evaluateLiteralExpression(String className, String value, ClassLoader classLoader) {
-        throw new IllegalStateException("This operator cannot be used into a Given clause");
+        throw new IllegalStateException(name() + " (" + String.join(",", symbols) + ") operator cannot be used into a Given clause");
     }
 
     protected Optional<String> match(String value) {
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
@@ -194,7 +194,7 @@ public enum BaseExpressionOperator {
     protected abstract boolean eval(Object rawValue, Object resultValue, Class<?> resultClass, ClassLoader classLoader);
 
     protected Object evaluateLiteralExpression(String className, String value, ClassLoader classLoader) {
-        throw new IllegalStateException(toString() + " operator cannot be used into a Given clause");
+        throw new IllegalStateException(toString() + " operator cannot be used in a GIVEN clause");
     }
 
     protected Optional<String> match(String value) {
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
@@ -46,6 +46,11 @@ public enum BaseExpressionOperator {
                     .collect(Collectors.toList());
             return results.size() != 0 && results.stream().allMatch(a -> a);
         }
+
+        @Override
+        public String toString() {
+            return "AND ( ; )";
+        }
     },
     LIST_OF_VALUES(1, "[") {
         @Override
@@ -66,6 +71,11 @@ public enum BaseExpressionOperator {
             return Stream.of(rawValue.substring(1, ((String) raw).length() - 1).split(","))
                     .map(String::trim)
                     .collect(Collectors.toList());
+        }
+
+        @Override
+        public String toString() {
+            return "OR ( [ ] )";
         }
     },
     EQUALS(2, "=") {
@@ -94,6 +104,11 @@ public enum BaseExpressionOperator {
             }
             return Objects.equals(resultValue, parsedResults);
         }
+
+        @Override
+        public String toString() {
+            return "Equal ( = )";
+        }
     },
     NOT_EQUALS(3, "!", "!=", "<>") {
         @SuppressWarnings("unchecked")
@@ -109,6 +124,11 @@ public enum BaseExpressionOperator {
             }
 
             return !operator.eval(valueToTest, resultValue, resultClass, classLoader);
+        }
+
+        @Override
+        public String toString() {
+            return "Not Equal ( !, !=, <> )";
         }
     },
     RANGE(4, "<", ">", "<=", ">=") {
@@ -140,6 +160,11 @@ public enum BaseExpressionOperator {
                     throw new IllegalStateException(new StringBuilder().append("This should not happen ").append(operator).toString());
             }
         }
+
+        @Override
+        public String toString() {
+            return "Range ( <, >, <=, >= )";
+        }
     };
 
     final List<String> symbols;
@@ -169,7 +194,7 @@ public enum BaseExpressionOperator {
     protected abstract boolean eval(Object rawValue, Object resultValue, Class<?> resultClass, ClassLoader classLoader);
 
     protected Object evaluateLiteralExpression(String className, String value, ClassLoader classLoader) {
-        throw new IllegalStateException(name() + " (" + String.join(",", symbols) + ") operator cannot be used into a Given clause");
+        throw new IllegalStateException(toString() + " operator cannot be used into a Given clause");
     }
 
     protected Optional<String> match(String value) {
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 
 import org.kie.dmn.api.feel.runtime.events.FEELEvent;
 import org.kie.dmn.feel.FEEL;
@@ -91,11 +92,34 @@ public class DMNFeelExpressionEvaluator extends AbstractExpressionEvaluator {
 
     @Override
     protected Object internalLiteralEvaluation(String raw, String className) {
+        EvaluationContext evaluationContext = newEvaluationContext();
+        return executeAndVerifyErrors(feel -> feel.evaluate(raw, evaluationContext));
+    }
+
+    @Override
+    protected boolean internalUnaryEvaluation(String rawExpression, Object resultValue, Class<?> resultClass, boolean skipEmptyString) {
+        if (rawExpression != null && skipEmptyString && rawExpression.isEmpty()) {
+            return true;
+        }
+        EvaluationContext evaluationContext = newEvaluationContext();
+        List<UnaryTest> unaryTests = executeAndVerifyErrors(feel -> feel.evaluateUnaryTests(rawExpression));
+        return unaryTests.stream()
+                .allMatch(unaryTest -> Optional
+                        .ofNullable(unaryTest.apply(evaluationContext, resultValue))
+                        .orElse(false));
+    }
+
+    /**
+     * Common internal method that execute the command and manage error
+     * @param command
+     * @param <T>
+     * @return
+     */
+    protected <T> T executeAndVerifyErrors(Function<FEEL, T> command) {
         AtomicReference<FEELEvent> errorHolder = new AtomicReference<>();
         FEEL feel = newFeelEvaluator(errorHolder);
 
-        EvaluationContext evaluationContext = newEvaluationContext();
-        Object expressionResult = feel.evaluate(raw, evaluationContext);
+        T result = command.apply(feel);
         FEELEvent errorEvent = errorHolder.get();
         if (errorEvent != null) {
             if (errorEvent instanceof SyntaxErrorEvent) {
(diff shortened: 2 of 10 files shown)


drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
```diff
@@ -66,25 +70,57 @@ public String fromObjectToExpression(Object value) {
...
         }
+        AtomicReference<FEELEvent> errorHolder = new AtomicReference<>();
+        FEEL feel = newFeelEvaluator(errorHolder);
+
         EvaluationContext evaluationContext = newEvaluationContext();
         List<UnaryTest> unaryTests = feel.evaluateUnaryTests(rawExpression);
-        if (unaryTests.size() < 1) {
-            throw new IllegalArgumentException("Impossible to parse the expression '" + rawExpression + "'");
+        FEELEvent errorEvent = errorHolder.get();
+        if (errorEvent != null) {
```
jomarko, line 124: Do not we need to check `SyntaxErrorEvent` here? Asking because of `internalLiteralEvaluation` implementation.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperatorTest.java
```diff
@@ -33,34 +40,37 @@
...
-        Assert.assertEquals("", BaseExpressionOperator.EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "= ", classLoader));
-        Assert.assertEquals(null, BaseExpressionOperator.EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "null", classLoader));
-        Assert.assertEquals(null, BaseExpressionOperator.EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "= null", classLoader));
-        Assert.assertEquals(null, BaseExpressionOperator.EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), null, classLoader));
+        assertEquals("Test", EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "= Test", classLoader));
+        assertEquals("", EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "= ", classLoader));
+        assertNull(EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "null", classLoader));
+        assertNull(EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), "= null", classLoader));
+        assertNull(EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), null, classLoader));
```
jomarko, line 56: In scenarios, we can use `;` for null, if I am not wrong, shouldn't we add such test here?

jomarko: Next we added some catch of exceptions into `evaluateLiteralExpression`, I miss tests like:
```java
assertThatThrownBy(() -> EQUALS.evaluateLiteralExpression(String.class.getCanonicalName(), SOME_VALUE, classLoader))
                .hasMessageContaining("Syntax error");
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
```diff
@@ -37,20 +43,36 @@
...
+            expressionEvaluator.evaluateUnaryExpression(new Object(), null, Object.class);
+        } catch (IllegalArgumentException e) {
+            assertEquals("Raw expression should be a string", e.getMessage());
+        }
+
+        try {
+            expressionEvaluator.evaluateUnaryExpression("! true", null, null);
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().startsWith("Impossible to parse the expression"));
+        }
```
jomarko: Assertj has really nice fluent aby to avoid try catch blocks:
```java
Assertions.assertThatThrownBy(() -> expressionEvaluator.evaluateUnaryExpression(new Object(), null, Object.class))
                .hasMessage("Raw expression should be a string");

        Assertions.assertThatThrownBy(() -> expressionEvaluator.evaluateUnaryExpression("! true", null, null))
                .hasMessageContaining("Impossible to parse the expression");
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
```diff
@@ -37,20 +43,36 @@
...
+            expressionEvaluator.evaluateLiteralExpression(String.class.getCanonicalName(), null, "SPEED");
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().startsWith("Error during evaluation:"));
+        }
+
+        try {
+            expressionEvaluator.evaluateLiteralExpression(String.class.getCanonicalName(), null, "\"SPEED");
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().startsWith("Syntax error:"));
+        }
```
jomarko: Please assume assertj comment also here

jomarko (commented):
### Code review
Some comments to code, manual check is pending.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
danielezonca: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
danielezonca: Thanks for the suggest, I'm not that expert with `Assertj` :)
Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperatorTest.java
danielezonca: In general this is not intended to cover all the possible values but `;` is empty string and not `null`

There is not a real parsing in this base expression language so no `Syntax error` messages. Each operator is covered by their tests and only `EQUALS` supports literal expression semantic

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
danielezonca: Correct, I refactored the code and now both method has the same error handling

Rikkola (approved):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
```diff
@@ -169,7 +194,7 @@ public static BaseExpressionOperator findOperator(String rawValue) {
     protected abstract boolean eval(Object rawValue, Object resultValue, Class<?> resultClass, ClassLoader classLoader);
 
     protected Object evaluateLiteralExpression(String className, String value, ClassLoader classLoader) {
-        throw new IllegalStateException("This operator cannot be used into a Given clause");
+        throw new IllegalStateException(toString() + " operator cannot be used into a Given clause");
```
kkufova: Please change to:

> ... operator cannot be used in a GIVEN clause.

kkufova (changes requested):
A couple of notes:

1. In DMN-based test scenarios, if the input is `1hh` instead of `1`, the error message is the following:

> Syntax error: extraneous input 'hh' expecting `<EOF>`

In rule-based test scenarios, the error message is:

> For input string: "1nn"

I think the messages should be united. And I'm not sure about `<EOF>`; can this be changed to something else?

2. If you create a KIE base and a **stateless** KIE session and open a DMN-based test scenario, the Business Central freezes and the following error is thrown:

> Unable to complete your request. The following exception occurred: Cannot find a default KieSession.

I think the error should be at least rephrased, as it does not appear when the KIE session is stateful.

3. Often, the message in the Error reason window starts with "_Exception with detailed message:_". I suggest removing this string completely, it is redundant.

4. Also, if there's an error in a date (such as `1993-05-13hh` or `1993-0513`), the cell is not highlighted.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionOperator.java
danielezonca: Done

danielezonca:
> A couple of notes:
> 
> 1. In DMN-based test scenarios, if the input is `1hh` instead of `1`, the error message is the following:
> 
> > Syntax error: extraneous input 'hh' expecting `<EOF>`
> 
> In rule-based test scenarios, the error message is:
> 
> > For input string: "1nn"
> 
> I think the messages should be united. And I'm not sure about `<EOF>`; can this be changed to something else?

I am not generating these messages but the error comes directly from the expression evaluator. So they cannot be unified, I can only partially change them in Rule based scenario but only for specific situations.

> 
> 1. If you create a KIE base and a **stateless** KIE session and open a DMN-based test scenario, the Business Central freezes and the following error is thrown:
> 
> > Unable to complete your request. The following exception occurred: Cannot find a default KieSession.
> 
> I think the error should be at least rephrased, as it does not appear when the KIE session is stateful.
> 

Fixed

> 1. Often, the message in the Error reason window starts with "_Exception with detailed message:_". I suggest removing this string completely, it is redundant.

Changed to `Error message:`

> 2. Also, if there's an error in a date (such as `1993-05-13hh` or `1993-0513`), the cell is not highlighted.

Fixed

danielezonca:
jenkins retest this please

jomarko (approved):

danielezonca:
This PR should fix also https://issues.jboss.org/browse/DROOLS-4059

kkufova (approved):
I haven't discovered any other bugs. Approving.

:+1:

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
```diff
@@ -132,12 +132,17 @@ protected ResultWrapper getSingleFactValueResult(FactMapping factMapping,
                                                              evaluationStatus);
         }
 
-        for (ExpressionElement expressionElement : factMapping.getExpressionElementsWithoutClass()) {
-            if (!(resultRaw instanceof Map)) {
-                throw new ScenarioException("Wrong resultRaw structure because it is not a complex type as expected");
+        List<ExpressionElement> elementsWithoutClass = factMapping.getExpressionElementsWithoutClass();
+
+        // DMN engine doesn't generate the whole object when no entry of the decision table match
+        if(resultRaw != null) {
```
gitgabrio, line 138: @danielezonca 
Just a question: if we get there, evaluation status is `SUCCEEDED`: does this mean that it is possible to have a successful elaboration with no result?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/IndexedScenarioException.java
```diff
@@ -47,7 +47,8 @@ public void setFileName(String fileName) {
 
     @Override
     public String getMessage() {
-        StringBuilder message = new StringBuilder().append("#").append(index).append(": ").append(super.getMessage());
+        String errorMessage = getCause() != null ? getCause().getMessage() : super.getMessage();
```
gitgabrio, line 50: @danielezonca 
Here `String errorMessage = getCause() != null ? getCause().getMessage() : super.getMessage(); `, _in theory_,  `getCause().getMessage()` and `super.getMessage() `may be **null** too

gitgabrio (commented):
@danielezonca 

Just a couple of questions

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/DMNScenarioRunnerHelper.java
danielezonca: Yes it could be possible because if a decision table has no entry that match with the inputs it can generate a null result with `SUCCESS` as status (it is related with the hit policy if I have understood correctly). And in this case if the output object is a complex one it will be `null` while I was expecting the object with `null` as fields values

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/IndexedScenarioException.java
danielezonca: Yes but I just want to be sure that there is a cause and then show the message. I expect it _should_ not be `null`

gitgabrio (approved):
@danielezonca 
Fine for me

PR #2358, opened 2019-05-22
[DROOLS-1540] also use the ClassLoader containing drools core classes…

mariofusco:
… only when running in a osgi container

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
+++ b/drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
 
 import org.drools.core.util.ClassUtils;
 import org.kie.internal.utils.KieTypeResolver;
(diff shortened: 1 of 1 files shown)


mariofusco:

drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java
```diff
@@ -345,14 +346,18 @@ public URL nextElement() {
...
+        if (getParent() != droolsClassLoader && isOsgiClassLoader(droolsClassLoader)) {
             this.droolsClassLoader = droolsClassLoader;
             if (CACHE_NON_EXISTING_CLASSES) {
                 nonExistingClasses.clear();
             }
         }
     }
 
+    private boolean isOsgiClassLoader(ClassLoader cl) {
+        return Stream.of( cl.getClass().getInterfaces() ).map( i -> i.getSimpleName() ).anyMatch( name -> name.equals( "BundleReference" ) );
```
yurloc, line 358: I think it would be more correct to check the full class name. There are projects that have their own BundleReferences: https://github.com/search?l=&q=BundleReference+extension%3Ajava+filename%3ABundleReference.java+language%3AJava&type=Code

PR #2363, opened 2019-05-24
[RHBRMS-3133] [DROOLS-3050] allow to add via annotation a serialVersi…

mareknovotny:
…onUID on classes generated from declared types


baldimir (approved):
Approved. The changes are the same as in original backport.

PR #2375, opened 2019-06-04
Allows to keep the row numbers in DTAnalysis messages

SimonRinguette:
I need to keep track of the row numbers of the Decision Table to be able correlate them later in the log and properly display them to the user in the UI.

### DIFF
--- a/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/dtanalysis/DMNDTAnalysisMessage.java
+++ b/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/dtanalysis/DMNDTAnalysisMessage.java
@@ -16,6 +16,8 @@
 
 package org.kie.dmn.validation.dtanalysis;
 
+import java.util.Collection;
+
 import org.kie.dmn.api.core.DMNMessageType;
 import org.kie.dmn.core.impl.DMNMessageImpl;
 import org.kie.dmn.validation.dtanalysis.model.DTAnalysis;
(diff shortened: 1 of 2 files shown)


SimonRinguette:
@tarilabs @etirelli are you guys ok with this small modification? This gives me the ability to highlight rows in the UI for better error reporting as we used to have in the XSLT version

ghost:

etirelli:
The change looks fine, I am just not seeing where the new field is populated? Can you provide a simple test case for it?

SimonRinguette:
@etirelli the new field is populated in DTAnalysis.java

tarilabs:
I see @SimonRinguette you need the ruleIDs in to be associated with each of the message.
This looks fine if you need just their reference IDs, but some checks have an higher semantic: for instance "rule X is masked by Y".
I am wondering if you need to reflect this in the "payload" then a simple Collection is not enough, and the datamodel could be associated with the message --instead of the Collection as you are proposing here.
Up to you, I just wanted to highlight this.
If you just need their reference IDs thanks for submitting directly a PR :)

SimonRinguette:
@tarilabs I need the primary rule ids for the message. In the modeler this will highlight the rows id in the message. In the XSLT implementaiton for the rule X masks Y we highlighted X only since it was the cause of the "issue". For overlaps, I'm highlighting both rows since they are equally responsible of the error. It's just a visual queue for the user when dealing with huge tables.

etirelli:
ok to test

tarilabs:
Sorry I thought this was already solved in a different way with the actual analysis message payload.

Jenkins test again

tarilabs:
Jenkins test again

PR #2383, opened 2019-06-11
[DROOLS-4150] make FactHandles serializable

mariofusco:


mariofusco:

drools-core/src/main/java/org/drools/core/common/QueryElementFactHandle.java
```diff
@@ -97,7 +98,13 @@ public void setObject(Object object) {
         this.object = object;
     }    
 
-    public WorkingMemoryEntryPoint getEntryPoint() {
+    public EntryPointId getEntryPointId() {
+        return null;
+        //throw new UnsupportedOperationException( "DisonnectedFactHandle does not support this method" );
```
baldimir: Nitpicking: Please delete this comment line. It is not needed here. Same for the getEntryPoint( InternalWorkingMemory wm ) method.

baldimir (commented):
Would it be possible to write a simple test for FactHandle serialization? Or is it already serialized as part of some test?

baldimir (approved):

drools-core/src/main/java/org/drools/core/common/EventFactHandle.java
```diff
@@ -252,36 +268,39 @@ public void setOtnCount( int otnCount ) {
...
 
     private EventFactHandle cloneWithoutTuples() {
         EventFactHandle clone = new EventFactHandle( getId(),
-                                                     getObject(),
-                                                     getRecency(),
-                                                     getStartTimestamp(),
-                                                     getDuration(),
-                                                     getEntryPoint(),
-                                                     isTraitOrTraitable() );
+                getIdentityHashCode(),
```
mdproctor, line 290: All code should be align on column. not align on indent.

PR #2387, opened 2019-06-13
Update ExcelParser.java

alban5650:
Replaced deprecated POI getCellType with the getCellTypeEnum equivalent.
This allows the ExcelParser to work with POI 3.17 and 4.0.1 versions.


ghost:

mariofusco:
ok to test

deepsnitk:
getCellTypeEnum is now deprecated and not available in Apache POI 5.x. So the old getCellType needs to be used.

From POI 4.1.2 documentation:

getCellTypeEnum()
Deprecated. 
will be removed in 4.2 Will be renamed to getCellType() when we make the CellType enum transition in POI 4.0. See bug 59791.

PR #2389, opened 2019-06-14
DROOLS-4156 DMN noname PMML model and DROOLS-4154

tarilabs:

/cc @manstis @jomarko


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNEvaluatorCompiler.java
```diff
@@ -482,12 +485,31 @@ private DMNExpressionEvaluator compileFunctionDefinitionPMML(DMNCompilerContext
                 logger.trace("theImport: {}", theImport);
                 URL pmmlURL = DMNCompilerImpl.pmmlImportURL(getRootClassLoader(), model, theImport, funcDef);
                 logger.trace("pmmlURL: {}", pmmlURL);
+                DMNImportPMMLInfo pmmlInfo = model.getPmmlImportInfo().get(pmmlDocument);
+                logger.trace("pmmlInfo: {}", pmmlInfo);
+                if (pmmlModel == null || pmmlModel.isEmpty()) {
+                    List<String> pmmlModelNames = pmmlInfo.getModels()
+                                                          .stream()
+                                                          .map(PMMLModelInfo::getName)
+                                                          .filter(x -> x != null)
```
jomarko, line 494: I think would be better to use variable name saying what is content of the stream (name, modelName ...) or the oppostie way, omit the variable name `.filter(Objects::nonNull)`

jomarko (commented):
Just one comment as I like discussion about code.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/util/Msg.java
```diff
@@ -118,6 +118,7 @@
     public static final Message3 FUNC_DEF_COMPILATION_ERR                            = new Message3( DMNMessageType.ERR_COMPILING_FEEL, "Error compiling Java function '%s' on node '%s': %s" );
     public static final Message2 FUNC_DEF_MISSING_ENTRY                              = new Message2( DMNMessageType.INVALID_SYNTAX, "A Java function definition requires both the 'class' and the 'method signature' attributes. Invalid definition for function '%s' on node '%s'" );
     public static final Message2 FUNC_DEF_PMML_MISSING_ENTRY                         = new Message2( DMNMessageType.INVALID_SYNTAX, "A PMML function definition requires both the 'document' and the 'model' attributes. Invalid definition for function '%s' on node '%s'" );
+    public static final Message1 FUNC_DEF_PMML_MISSING_MODEL_NAME                    = new Message1( DMNMessageType.INVALID_SYNTAX, "The PMML function definition did not provide 'model' attribute, while the PMML resource defines the following model names: %s" );
```
jomarko, line 121: Message looks good to me.

manstis:
Testing I get (adding a PMML document name but no model name):
```
[INFO] java.lang.NullPointerException: null
[INFO] 	at org.kie.dmn.core.compiler.DMNEvaluatorCompiler.compileFunctionDefinitionPMML(DMNEvaluatorCompiler.java:491)
[INFO] 	at org.kie.dmn.core.compiler.DMNEvaluatorCompiler.compileFunctionDefinition(DMNEvaluatorCompiler.java:335)
[INFO] 	at org.kie.dmn.core.compiler.DMNEvaluatorCompiler.compileExpression(DMNEvaluatorCompiler.java:144)
[INFO] 	at org.kie.dmn.core.compiler.DecisionCompiler.compileEvaluator(DecisionCompiler.java:125)
[INFO] 	at org.kie.dmn.core.compiler.DMNCompilerImpl.processDrgElements(DMNCompilerImpl.java:405)
[INFO] 	at org.kie.dmn.core.compiler.DMNCompilerImpl.compile(DMNCompilerImpl.java:224)
[INFO] 	at org.kie.dmn.core.compiler.DMNCompilerImpl.compile(DMNCompilerImpl.java:155)
[INFO] 	at org.kie.dmn.core.compiler.DMNCompilerImpl.compile(DMNCompilerImpl.java:139)
```
I'll investigate more after lunch.

manstis:
@tarilabs OK, the NPE has magically disappeared after lunch. Sweet.

However I find the WARNING reported for a missing PMML model name is not propagated through to the `IncrementalResults` returned by `org.drools.compiler.kie.builder.impl.KieBuilderSetImpl.buildChanges(..)`. This appears to only report ERROR messages in the incremental build mechanism. 

This can be demonstrated (using our infamous incremental build test) by first adding a `FunctionDefinition` for PMML with no PMML document or model name and building the `KieModule`. An ERROR is returned stating they need to be provided. Then perform an incremental build passing an updated DMN file that has the PMML document value complete. `DMNEvaluatorCompiler.compileFunctionDefinitionPMML(..)` will now report a WARNING due to the missing model value; however this message does not appear in the incremental build results.

The impact is that in the WB if a User adds a `Function` with PMML and saves it, the ERROR will correctly appear. If they then set the PMML document value and save the ERROR will disappear but the WARNING does not appear. The WARNING only appears if the User performs a full compilation of the project.

So.. we'll either need to make the missing PMML model name an ERROR (easy fix) or change `KieBuilderSetImpl` to also support add/removal of any message type (ERROR, WARNING and INFO). This second option is probably the more complete however may be more effort.

manstis (changes requested):
Please see my [comment](https://github.com/kiegroup/drools/pull/2389#issuecomment-502098526).

tarilabs:
> we'll either need to make the missing PMML model name an ERROR (easy fix) or change KieBuilderSetImpl to also support add/removal of any message type (ERROR, WARNING and INFO). This second option is probably the more complete however may be more effort.

I don't concur the missing PMML model name is an ERROR, as other PMML runtimes are supporting running without the name just fine. That reduces our options I guess :)

manstis:
@tarilabs I agree, preserving the WARNING is far better than making it an ERROR... I was trying to be helpful by proposing alternative solutions (to the issue that WARNINGs are not handled by the incremental build mechanism)...

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/pmml/DMNRuntimeKiePMMLTest.java
```diff
@@ -44,6 +58,41 @@ public void testNoPMMLmodelName() {
...
+        KieFileSystem kfs = ks.newKieFileSystem();
+
+        kfs.write("src/main/resources/org/acme/test_scorecard.pmml", ks.getResources().newClassPathResource("test_scorecard.pmml", DMNRuntimeKiePMMLTest.class));
+        KieBuilder kieBuilder = ks.newKieBuilder(kfs).buildAll();
+        assertEquals(0, kieBuilder.getResults().getMessages(org.kie.api.builder.Message.Level.ERROR).size());
+
+        kfs.write("src/main/resources/org/acme/KiePMMLScoreCard_NOPMMLmodelName.dmn", ks.getResources().newClassPathResource("KiePMMLScoreCard_NOPMMLmodelName.dmn", DMNRuntimeKiePMMLTest.class));
+        IncrementalResults addResults = ((InternalKieBuilder) kieBuilder).createFileSet("src/main/resources/org/acme/KiePMMLScoreCard_NOPMMLmodelName.dmn").build();
+        LOG.debug("getAddedMessages: {}", addResults.getAddedMessages());
+        assertTrue(addResults.getAddedMessages().size() > 0);
```
tarilabs, line 83: @manstis can you kindly confirm this line make an assert for where you would expect to find the WARN messages, please?
On confirmation, I'll ask support for indeed the addedMessages from KieBuilder to report all messages besides errors.

manstis: @tarilabs Yes, that looks to be perfect (I stepped through this test and `KieBuilderSetImpl.buildChanges(..)` L#154 retrieves the "errors" from `KnowledgeBuilderImpl` however skips DMN's warnings (you can see they are present in `KnowledgeBuilderImpl.getErrorList()` L#1981 `results` property).

It'd be great if you could also add a subsequent step in the test to then update ` KiePMMLScoreCard_NOPMMLmodelName.dmn` (perhaps with `KiePMMLScoreCard.dmn`) to then observe there is a (warning) message removed too.

BTW, there is already a "stepped" test in `kie-dmn-core`, `org.kie.dmn.core.incrementalcompilation.WBCompilationTest`. I'll leave the decision with you as to whether the new test should better live there; or with the PMML tests.

tarilabs: > BTW, there is already a "stepped" test in kie-dmn-core, org.kie.dmn.core.incrementalcompilation.WBCompilationTest.

yes, but that is a working test an inherited down the line by jpmml-integrations.
This test instead is only a problem for this module locally.

mariofusco:
@manstis @tarilabs I don't think that I could just add also WARN and INFO among the Messages contained in the IncrementalResults returned by the KieBuilderSet, because this will very likely break many tests that don't expect anything else than ERRORs inside those lists.

At the moment we're creating the KieBuilderSet through which the incremental compilation is performed using the method of the InternalKieBuilder:

`KieBuilderSet createFileSet(String... files)`

My suggestion is to create an overload of this method like

`KieBuilderSet createFileSet(Level minimalLevel, String... files)`

where the additional Level argument is the minimal level of messages that you want in the IncrementalResults.

tarilabs:
Superseeded by https://github.com/kiegroup/drools/pull/2394#pullrequestreview-250926787

PR #2393, opened 2019-06-17
DROOLS-4186 : Guided Rules does not output parenthesis in DRL

Rikkola:
https://issues.jboss.org/browse/DROOLS-4186

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -1313,10 +1313,8 @@ public class RuleModelDRLPersistenceImpl
                                          final String fieldType,
                                          final String value) {
             String workingValue = value.trim();
-            if (workingValue.startsWith("(")) {
+            if (workingValue.startsWith("(") && workingValue.endsWith(")")) {
                 workingValue = workingValue.substring(1);
-            }
-            if (workingValue.endsWith(")")) {
                 workingValue = workingValue.substring(0,
                                                       workingValue.length() - 1);
             }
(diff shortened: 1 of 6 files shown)


Rikkola:
@jomarko @manstis if you could review please. I think the safe way to fix this is removing the surrounding brackets. At least the existing unit tests only check the brackets are removed in these cases.

Rikkola:
@jomarko @manstis if you could review please. I think the safe way to fix this is removing the surrounding brackets. At least the existing unit tests only check the brackets are removed in these cases.

manstis (approved):
Thanks @Rikkola your test is exactly what I was about to suggest on this PR before I scrolled down :-)

jomarko:
No comments to code, however question, the fix is for guided rule templates, as the attached project uses guided rule templates, however I checked simple guided rules, and there is the issue still present, my question is if we should solve also guided rules as part of [DROOLS-4186](https://issues.jboss.org/browse/DROOLS-4186). If not, then I can approve.

Rikkola:
@jomarko I know where I need to fix this, but didn't manage to reproduce with Guided Rules. If you can give me simple ( or less simple ) reproducer I'll do the fix for the other locations.

Rikkola:
@manstis @jomarko I found these 3 locations where we have the same issue ;) I hope I got them all now.

Rikkola:
SonarCloud finds a code smell from an area that I did not edit. The fix it proposes is ok, but not the worst issue in that "code chapter". The class itself could use refactoring, but with the test coverage I know we have for this area I would rather not touch it.

jomarko (approved):
Works fine in all editors, so approving, however want to ask if the video show an issue or not, if a user put quoted string value list, the generated rule has "double quoted" the string values. is it expected? Really not sure.

https://drive.google.com/open?id=166ek_qoOL4EEyjsP8y7XeP_qcaVVJ7M7

Rikkola:
It probably should not add quotes then, but since it has always done so I'm not sure if we should change it. It could be someone is depending on it. These things tend to be tricky to fix due to backwards compatibility.

jomarko:
@Rikkola thank you for response. feel free to merge.

Rikkola:
@kiegroup/gatekeepers Could this be merged please.

PR #2420, opened 2019-07-01
Drools 4268 - FIX XStream configuration

gitgabrio:
@kkufova @danielezonca @yesamer 

See https://issues.jboss.org/browse/DROOLS-4268


drools-scenario-simulation/drools-scenario-simulation-api/src/main/resources/org/drools/scenariosimulation/scesim.xsd
```diff
@@ -0,0 +1,126 @@
+<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
```
danielezonca: XSD introduction is related to client side marshalling that is not part of this PR so I think should not be added here. Wdyt?
If you decide to keep it please remove commented nodes

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
```diff
@@ -37,35 +40,60 @@
...
 
+    public static String getCurrentVersion() {
+        return currentVersion;
+    }
+
+    public static String cleanUpUnusedNodes(String input) {
+        return removeNode(input, "<simulationDescriptor reference\\b[^>]*>(.*?)");
+    }
+
+    private static  String removeNode(String input, String toRemove) {
```
danielezonca: This is not a "remove node" but just a string replace, I think the name is a bit misleading. Wdyt?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/InMemoryMigrationStrategy.java
```diff
@@ -113,4 +127,30 @@
...
+            // Unmarshall the 1.5 format with "older" xstream configuration, that read "reference" attributes
+            Object o = getLocalXStream().fromXML(input);
+            ScenarioSimulationModel model = (ScenarioSimulationModel) o;
+
+            // Marshall the model with the "new" xstream configuration, that does not write "reference" attributes
+            return updateVersion(xmlPersistence.marshal(model), "1.5", "1.6");
+        };
+    }
+
+    private XStream getLocalXStream() {
```
danielezonca, line 145: This method should be used only with files version <= 1.5. Add please a java doc or rename the method

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
```diff
@@ -37,35 +40,60 @@
...
-        xt.alias("SimulationDescriptor", SimulationDescriptor.class);
-
-        xt.alias("Import", Import.class);
+    /**
+     * Method to configure the commonly-used mappings defined in {@see XSTREAM_MAPPINGS}
+     * @param toConfigure
+     */
+    public static void configureXStreamMappings(XStream toConfigure) {
+        Map<String, Class> XSTREAM_MAPPINGS = new HashMap<String, Class>() {
+            {
```
danielezonca: Using this syntax you are creating a subclass of `HashMap` everytime this method is invoked. Can you please refactor to avoid the creation of this map?

danielezonca (changes requested):

baldimir:

drools-scenario-simulation/drools-scenario-simulation-api/src/main/resources/org/drools/scenariosimulation/scesim.xsd
gitgabrio: @danielezonca 
Yes - it currently is not part of this PR. Removed

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
gitgabrio: @danielezonca 
Removed

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
gitgabrio: @danielezonca 
Fixed

gitgabrio:
jenkins please retest this

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/model/Scenario.java
```diff
@@ -37,10 +35,6 @@
     private final List<FactMappingValue> factMappingValues = new ArrayList<>();
 
 
-    public Scenario() {
```
danielezonca, line 43: Empty constructor is needed for CDI so just removing it could be a bit error prone in the future (we need to remember that as soon as we will need another constructor we will need to add back also the empty one).
What about a javadoc note?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
```diff
@@ -134,6 +123,9 @@ public String migrateIfNecessary(String rawXml) {
             case "1.5":
                 migrator = migrator.andThen(getMigrationStrategy().from1_5to1_6());
                 supported = true;
+                break;
+            default:
+                supported = currentVersion.equals(fileVersion);
```
danielezonca, line 128: Sonar complains with `case` without `break` but I don't like to add it only for 1.5 (Sonar is analyzing only new code). I would like to refactor the whole switch (to `if/else` probably) or keep last case without and we will do in another ticket. Wdyt?

danielezonca (changes requested):
Additional two comments

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
gitgabrio: @danielezonca 
I am not really sure about the reason behind that implementation - but changing the logic may require more tests and is out of scope of this PR. What is clearly not-so-nice is the mix of two different approaches: switch/case is meant exactly to provide a "default" value if specific "cases" are not executed. But in the original implementation the "default" value was provided before the switch, and the switch' default was ignored. IMHO for now this fix it is just a readability improvement; later eventually we can dig more on the execution logic behind.

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/model/Scenario.java
gitgabrio: @danielezonca 
I may be wrong, but empty constructor is needed only when there is another (not empty) one: it seems also Simulation and SImulationDescriptor does not explicitly declare it

danielezonca: Yes, I meant that as soon as we will need another constructor we need to remember to add back also the empty constructor because of CDI. So I was suggesting a note in javadoc but just a comment not really a request for change

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationXMLPersistence.java
danielezonca: Fine for me

danielezonca (approved):

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/model/Scenario.java
```diff
@@ -118,7 +106,7 @@ public void resetErrors() {
     }
 
     Scenario cloneScenario() {
-        Scenario cloned = new Scenario(simulationDescriptor);
+        Scenario cloned = new Scenario();
```
yesamer, line 109: In my opinion this method is now misleading: before the change, method 'cloneScenario()' creates a new Scenario instance with the same scenarioDescriptor in scope. Now, it simpy creates a totally different instance of Scenario. Therefore, if the method is still used, it should be renamed to 'createScenario()' or it could be removed, and create a new scenario instance directly in the caller.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/MigrationStrategy.java
```diff
@@ -61,6 +61,12 @@
      */
     Function<String, String> from1_4to1_5();
 
+    /**
+     * Method to obtain the migration function from 1.4 to 1.5
```
yesamer: Probably you mean from 1.5 to 1.6 in this JavaDoc

yesamer (changes requested):

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/model/Scenario.java
gitgabrio: @yesamer 
I don't think so. "SimulationDescriptor" was a leftover unused anywhere. The main point is the cloning of all fact mappings - that's why it is different from a simple "createScenario" - for which, a simple new Scenario() is enough. Beside that, the meant semantic is: create an exact copy of the original: whatever property will be add/removed from Scenario, it is clear it will have to be managed here.

yesamer (approved):

kkufova (approved):
Looks good to me.

:+1: 

Approving.

gitgabrio:
jenkins please retest this

danielezonca:

PR #2421, opened 2019-07-01
[DROOLS-4277] add min and max accumulate functions specific for Integ…

mariofusco:
…ers and Longs


baldimir:

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/CompilationFailuresTest.java
```diff
@@ -98,4 +99,20 @@ public void testBadQueryArg() {
         assertFalse(results.getMessages( Message.Level.ERROR).isEmpty());
     }
 
+    @Test
+    public void testMaxIntegerResultOnDoublePatternShouldntCompile() {
```
baldimir, line 103: Could you please add tests also for the other functions? I mean minI, minL and maxL?

mariofusco: Done

PR #2425, opened 2019-07-03
fix eclipse plugin class not found error

wenjunxiao:
eclipse open *.bpmn with `BPMN2 Process Editor` throw error WorkDefinitions.conf class not found import org.drools.core.process.core.datatype.impl.type.StringDataType because that class has moved to org.jbpm.process.core.datatype.impl.type.StringDataType


ghost:

mareknovotny:
@wenjunxiao  can you file the jira too? and then comply with out contribution rules?
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/README.md

PR #2426, opened 2019-07-04
[DROOLS-4289] avoid FactHandle id overflow during session deserializa…

mariofusco:
…tion

### DIFF
--- a/drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufInputMarshaller.java
+++ b/drools-core/src/main/java/org/drools/core/marshalling/impl/ProtobufInputMarshaller.java
@@ -185,8 +185,11 @@ public class ProtobufInputMarshaller {
                                                                     Environment environment,
                                                                     SessionConfiguration config,
                                                                     ProtobufMessages.KnowledgeSession _session) throws IOException {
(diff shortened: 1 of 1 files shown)


etirelli:
@mariofusco I am not sure this is enough, but it's been a while since I looked into this code. Potential problems that could be caused by a simple id reset are:

* id clashes: if facts ids survive for a long time, is there a check when creating id's to avoid clashes with existing facts?

* related to the above, I believe activations are persisted as id tuples. Do they need to be handled differently if that is the case?

* ids are used to correlate facts in clusters. Do we need to take any measure to avoid getting nodes out of sync?

mariofusco:
@etirelli I confirm this proposed fix is wrong and I'm closing this pull request. FactHandle ids of actual facts aren't indeed regenerated (I thought they were) and so if a fact remained in the session for long time, it's possible that my fix could generate an id clash. What is generated from scratch are ids of synthetic facts (from, accumulate, ...) but they are not flagged in a special way so I don't know if I could reuse them during a deserialization. Moreover I wouldn't implement a mechanism to recycle no longer used ids (provided it would be feasible, and I don't think it is) because it would be quite expensive both memory and performance wise.

The only way I see to structurally solve this problem is turning that id into a long. I believe this could be done relatively easily (I checked that the id is not used as an array index) and possibly without breaking backward compatibility of already serialized sessions. Please let me know if I should attempt to do this.

PR #2427, opened 2019-07-04
Changed DMNAssemblerService#getDMNProfiles method visibility

SimonRinguette:
Simple PR to change the visibility of a method from private to protected. I want to be able to extend this method to provide at runtime additional DMNProfiles instead of relying solely on the system preferences.

The system preferences is problematic for me because of class loader dynamic instantiation is very limited.


ghost:

etirelli (approved):

PR #2430, opened 2019-07-04
DROOLS-4278 Applying PMML model using kie-server fails

lanceleverich:
Changed the ApplyPmmlModelCommand execute method so that it uses a
RegistryContext, instead of a RequestContextImpl.


drools-core/src/main/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommand.java
```diff
@@ -260,11 +276,13 @@ public PMML4Result execute(Context context) {
...
-        execRes.setResult("results", resultHolder);
-        outputs.forEach(out -> {
-            execRes.setResult(out.getName(), out);
-            resultHolder.updateResultVariable(out.getName(), out);
-        });
+        if (execRes != null) {
+            execRes.setResult("results", resultHolder);
+            outputs.forEach(out -> {
+                execRes.setResult(out.getName(), out);
+                resultHolder.updateResultVariable(out.getName(), out);
```
mariofusco: I believe that the returned `resultHolder` should be populated even when `execRes` is `null`.

danielezonca: +1
If you wrap `execRes` with an `Optional` you can remove completely the if and just protect execRes statements with `.ifPresent()`

drools-core/src/test/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommandTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+
+public class ApplyPmmlModelCommandTest {
+
+    @Test(expected = IllegalStateException.class)
+    public void testMissingRequestData() {
+        ApplyPmmlModelCommand cmd = new ApplyPmmlModelCommand();
+        RegistryContext ctx = new ContextImpl().register(KieBase.class, null);
```
danielezonca: I think `.register()` is not necessary here

drools-core/src/main/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommand.java
```diff
@@ -196,17 +196,33 @@ public void setOutIdentifier(String outIdentifier) {
...
         }
+        PMML4Result resultHolder = new PMML4Result(requestData.getCorrelationId());
+        RegistryContext ctx = (RegistryContext) context;
         if (packageName == null) {
             packageName = (String)ctx.get("packageName");
         }
-        KieBase kbase = ((RegistryContext)context).lookup(KieBase.class);
-        PMML4Result resultHolder = new PMML4Result(requestData.getCorrelationId());
+
+        KieBase kbase = lookupKieBase(ctx);
```
danielezonca, line 226: Can you please add a test for this?

danielezonca (changes requested):

drools-core/src/main/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommand.java
lanceleverich: Updated to use suggested ifPresent method so that the resultHolder gets populated, regardless of whether there is an execRes

drools-core/src/test/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommandTest.java
lanceleverich: Removed the call to the `register` method

drools-core/src/main/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommand.java
lanceleverich: Added new test that provides an empty KieBase, so that the lookupKieBase gets called and returns successfully.

drools-core/src/test/java/org/drools/core/command/runtime/pmml/ApplyPmmlModelCommandTest.java
danielezonca: Are you sure? I can still see it :thinking:

lanceleverich: I missed the first one, but have now taken it out too.

PR #2433, opened 2019-07-05
[DROOLS-4292] executable-model fails with more than 5 arguments query

tkobayas:
- Unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/ca77718c7401d399ccbc672991e57aaed13710d7

PR #2436, opened 2019-07-05
[RHDM-867] force load of package-info into PMML4Compiler classloader

mariofusco:


lanceleverich (approved):
Looks good. Thanks for the note regarding the need to force the loading of the package-info in Java 9+.

PR #2439, opened 2019-07-05
DROOLS-4297 : Guided Rule Template removes quotation marks in concatenated String

Rikkola:
https://issues.jboss.org/browse/DROOLS-4297

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -1094,7 +1095,7 @@ public class RuleModelDRLPersistenceImpl
             } else {
                 addFieldRestriction(buf,
                                     constr.getConstraintValueType(),
-                                    constr.getFieldType(),
+                                    getFieldType(constr),
                                     constr.getOperator(),
                                     parameters,
                                     constr.getValue(),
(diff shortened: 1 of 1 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -4280,4 +4281,13 @@ public RuleModel getSimpleRuleModel(final String drl) {
...
         return rm;
     }
+
+    private static String getFieldType(final SingleFieldConstraint constr) {
+        final String fieldType = constr.getFieldType();
+        if (Objects.equals("this", fieldType)) {
+            return constr.getFactType();
+        } else {
+            return fieldType;
+        }
```
Rikkola, line 4291: @manstis @jomarko Making this a draft for now. The problem is the field type is "this" when the field is "this". This is how the field is given to us from the DMO. Now I am not sure why, maybe we need the information that the field is "this" some where for some reason?

Since the behaviour of DMO is a bit strange and not sure what could break if the field type of "this" was the actual fact type. I'm just changing the type of the field to the type of the fact in this class. This fixes the bug that was reported.

Let me know if you think this is enough? Or if we should change the behaviour of the DMO?

manstis: @Rikkola Should we not fix DMO?

https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/projects/BaseFactBuilder.java#L55

jomarko: As the usage of `this` can be case of multiple users, I vote for the way @manstis proposes.

Rikkola: Yes this was what I was after. Do we remote the field type "this" or not. Can't think of any location where we actually use that type, just wondering if removing it causes us more bugs.

I'll remove it and let's see what breaks. If anything.

PR #2451, opened 2019-07-17
DROOLS-4331 : Error when copying Guided Rule with Bigdecimal

Rikkola:
https://issues.jboss.org/browse/DROOLS-4331

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -16,6 +16,8 @@
 
 package org.drools.workbench.models.commons.backend.rule;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


Rikkola:
@manstis Hello. Was this what you were after with the fix proposition?

manstis (approved):
@Rikkola It is similar to what I was thinking and works the same.

I was thinking to call [this](https://github.com/kiegroup/drools/blob/e1ad62847ee8cdd402d324b99afb296a91150c4b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L4152) earlier and then use the `type` to know whether to `trim(..)` the value - if it ended with `B` or `I`... I guess now I've typed that you're doing it already.

That aside (as your implementation works too) I'd also be tempted to add a check that the `SingleFieldConstraint.getFieldType()` was correct too (I know we probably have tests for that too, but I didn't think it'd hurt to also include it in your new tests).

Approving however as my comments are only opinion.

Thanks for the fix Toni.

Rikkola:
Ok. I was not sure if I could trust the type. I'll improve the tests.

Rikkola:
@Bonuseto would you have time to review this one?

Bonuseto:

Rikkola:

Bonuseto (approved):
Tested, working perfect without issues, approved

Rikkola:
@kiegroup/gatekeepers could you please merge this.

full downstream fails, but the same setup already passed in compile downstream.
Failure in full downstream is:
 org.kie.server.integrationtests.drools.pmml.ApplyScorecardMultiModuleIntegrationTest.org.kie.server.integrationtests.drools.pmml.ApplyScorecardMultiModuleIntegrationTest

PR #2454, opened 2019-07-18
[DROOLS-4334] Type coercion with Map.get() after jitting

tkobayas:
- Unit test


mariofusco:
Repaced and fixed by https://github.com/kiegroup/drools/commit/17a6e803e551165e9802e387d5dbb053bab7f270

PR #2472, opened 2019-07-25
When is the rule evaluation performed? Is it after the insert(update) facts operation or after the fireAllRules?

alwyngo:
When is the rule evaluation performed? Is it after the insert(update) facts operation or after the fireAllRules?


ghost:

PR #2477, opened 2019-07-30
[DROOLS-4375] BigDecimal in from accumulate with MVEL with executable…

tkobayas:
…-model

- Unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/5848f70f08c3b6c0e7879b4935c2856b429604ec

PR #2478, opened 2019-07-30
[DROOLS-4376] Abbreviated comparison with Map fails in executable-model

tkobayas:
- Unit test

This test case may contain 2 factors (I was not able to split the JIRA nicely):
- Abbreviated comparison with Map
- Comparison coercion between BigDecimal and primitive type

Anyway, the customer hopes to build this kind of rule.


mariofusco:

PR #2479, opened 2019-07-30
[DROOLS-4377] Comparison error with primitive and Map in executable-m…

tkobayas:
…odel

- Unit test


mariofusco:

PR #2480, opened 2019-07-30
DROOLS-4378 DMN invoking PMML with input type intervals

tarilabs:


jiripetrlik (changes requested):
Can you add a test here. It seems that this new possibility has no test coverage.

jiripetrlik (approved):

tarilabs:
@jiripetrlik as also @rsynek explained, the SonarCloud is a tool. in this case, sonarcloud is failing to realize coverage from the other repo.
Can we please get past this coverage-threshold thing, please?

PR #2482, opened 2019-07-31
[DROOLS-4381] BigDecimal and primitive comparison error in executable…

tkobayas:
…-model build

- Unit test


mariofusco:

PR #2483, opened 2019-07-31
[DROOLS-4382] Executable-model build error with "matches" and fuction…

tkobayas:
… with a fact property as an argument

- Unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/2bfadaa37f49148fdfb130691c962b9d8364b594

PR #2485, opened 2019-08-01
RHDM-966: Added prototype test of expreession with brackets

Bonuseto:
Hi @jomarko , I add new test but looks like it's not passing 
Screenshot from 2019-08-01 16-36-23


Bonuseto:
Hi @Rikkola can you please confirm if it works as it should? Thank you in advance!
Screenshot from 2019-08-02 10-23-50
Screenshot from 2019-08-02 10-23-57
Screenshot from 2019-08-02 10-24-05

Rikkola:
@Bonuseto no, this looks to be another limitation that I might have to fix.

lucamolteni:
@Bonuseto @Rikkola can we close this?

PR #2489, opened 2019-08-06
RHDM-1015 Rule-based test scenarios that contain maps of simple types fail

lanceleverich:
* Fixed an error that was being caused by the way the workbench writes the JSON for Map objects that contain simple key/value pairs (e.g. String and String).
* Added to the tests for the expression evaluator, creating a test for a JSON string that is equivalent to one created by the workbench.

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
@@ -84,7 +84,7 @@ public abstract class AbstractExpressionEvaluator implements ExpressionEvaluator
             JsonNode jsonNode = element.getValue();
             // if is a simple value just return the parsed result
             if (numberOfFields == 1 && "value".equals(key)) {
-                return internalLiteralEvaluation(jsonNode.textValue(), genericClasses.get(0));
+                return internalLiteralEvaluation(jsonNode.textValue(), genericClasses.isEmpty() ? className : genericClasses.get(0));
             }
 
             if (jsonNode.isArray()) {
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
@@ -69,10 +69,16 @@ public abstract class AbstractExpressionEvaluator implements ExpressionEvaluator
 
     protected List<Object> createAndFillList(ArrayNode json, List<Object> toReturn, String className, List<String> genericClasses) {
         for (JsonNode node : json) {
-            String genericClassName = ScenarioSimulationSharedUtils.isMap(className) ? className : genericClasses.get(genericClasses.size() - 1);
-            Object listElement = createObject(genericClassName, genericClasses);
-            Object returnedObject = createAndFillObject((ObjectNode) node, listElement, genericClassName, genericClasses);
-            toReturn.add(returnedObject);
+            if (isSimpleTypeNode(node)) {
+                String generic = genericClasses.get(genericClasses.size() - 1);
+                Object value = internalLiteralEvaluation(getSimpleTypeNodeTextValue(node), generic);
+                toReturn.add(value);
+            } else {
+                String genericClassName = ScenarioSimulationSharedUtils.isMap(className) ? className : genericClasses.get(genericClasses.size() - 1);
+                Object listElement = createObject(genericClassName, genericClasses);
+                Object returnedObject = createAndFillObject((ObjectNode) node, listElement, genericClassName, genericClasses);
+                toReturn.add(returnedObject);
+            }
         }
         return toReturn;
     }
@@ -84,12 +90,11 @@ public abstract class AbstractExpressionEvaluator implements ExpressionEvaluator
             String key = element.getKey();
             JsonNode jsonNode = element.getValue();
             // if is a simple value just return the parsed result
-            if (isSimpleType(jsonNode)) {
-//                Map<String, Object> rawSimpleTypeMap = createRawSimpleTypeMap(key, jsonNode);
-                Map rawSimpleTypeMap = (Map) createObject(Map.class.getCanonicalName(), null);
-                Object value = internalLiteralEvaluation(getSimpleTypeTextValue(jsonNode), genericClasses.get(0));
-                rawSimpleTypeMap.put(key, value);
-                return rawSimpleTypeMap;
+            if (isSimpleTypeNode(jsonNode)) {
+                Map.Entry<String, List<String>> fieldDescriptor = getFieldClassNameAndGenerics(toReturn, key, className, genericClasses);
+                Object value = internalLiteralEvaluation(getSimpleTypeNodeTextValue(jsonNode), fieldDescriptor.getKey());
+                setField(toReturn, key, value);
+                return toReturn;
             }
 
             if (jsonNode.isArray()) {
@@ -112,14 +117,17 @@ public abstract class AbstractExpressionEvaluator implements ExpressionEvaluator
         return toReturn;
     }
 
-    protected boolean verifyResult(Object rawValue, Object resultRaw, Class<?> resultClass) {
+    protected boolean verifyResult(Object rawExpression, Object resultRaw, Class<?> resultClass) {
+        if (resultRaw == null) {
+            return false;
+        }
         if (!(resultRaw instanceof List) && !(resultRaw instanceof Map)) {
             throw new IllegalArgumentException("A list was expected");
         }
-        if (!(rawValue instanceof String)) {
+        if (!(rawExpression instanceof String)) {
             throw new IllegalArgumentException("Malformed raw data");
         }
-        String raw = (String) rawValue;
+        String raw = (String) rawExpression;
         ObjectMapper objectMapper = new ObjectMapper();
 
         try {
@@ -140,7 +148,10 @@ public abstract class AbstractExpressionEvaluator implements ExpressionEvaluator
         for (JsonNode node : json) {
             boolean success = false;
             for (Object result : resultRaw) {
-                if (verifyObject((ObjectNode) node, result, resultClass)) {
+                boolean simpleTypeNode = isSimpleTypeNode(node);
+                if (simpleTypeNode && internalUnaryEvaluation(getSimpleTypeNodeTextValue(node), result, result.getClass(), true)) {
+                    success = true;
+                } else if (!simpleTypeNode && verifyObject((ObjectNode) node, result, resultClass)) {
                     success = true;
                 }
             }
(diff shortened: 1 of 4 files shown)


drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -84,7 +84,7 @@ protected Object createAndFillObject(ObjectNode json, Object toReturn, String cl
             JsonNode jsonNode = element.getValue();
             // if is a simple value just return the parsed result
             if (numberOfFields == 1 && "value".equals(key)) {
-                return internalLiteralEvaluation(jsonNode.textValue(), genericClasses.get(0));
+                return internalLiteralEvaluation(jsonNode.textValue(), genericClasses.isEmpty() ? String.class.getCanonicalName() : genericClasses.get(0));
```
danielezonca: We need to cover all kind of `Map<String, SimpleType>` not only `Map<String, String>`. And btw if here genericClasses is empty it means that it is inconsistent, so something wrong already happened (I guess the code performed already a wrong recursion?)

danielezonca (changes requested):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
lanceleverich: Changed the above so that it now uses the passed-in simple-type's className instead of hard coded String.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -94,7 +94,14 @@ protected Object createAndFillObject(ObjectNode json, Object toReturn, String cl
                 setField(toReturn, key, returnedList);
             } else if (jsonNode.isObject()) {
                 Map.Entry<String, List<String>> fieldDescriptor = getFieldClassNameAndGenerics(toReturn, key, className, genericClasses);
-                Object nestedObject = createObject(fieldDescriptor.getKey(), fieldDescriptor.getValue());
+                Object nestedObject = null;
+                try {
+                    nestedObject = createObject(fieldDescriptor.getKey(), fieldDescriptor.getValue());
+                } catch (IllegalArgumentException iax) {
+                    if (!(iax.getCause() instanceof InstantiationException && iax.getCause().getCause() instanceof NoSuchMethodException)) {
```
danielezonca: Can you please add a comment and a test to explain why we are propagating only this specific type of exception?

danielezonca (commented):
Only one additional comment, I think if you add the test I'm asking in the comment also SonarCloud check about coverage should pass

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
lanceleverich: Update removes reliance on catching a particular type of exception, and now looks at class to see if a public default constructor is available. This functionality was moved to the BaseExpressionEvaluator, and unit tests were added.

kkufova:
I will review the PR tomorrow.

kkufova (commented):
@lanceleverich, @danielezonca, I think the PR is not ready yet.

I created a data object with a field of type `java.util.Map<java.lang.String, java.lang.String>` and a test scenario asset. The test scenario is successful only if the table is completely empty or the map is empty (`Map(0)`). Once I add any item to the map (using the collection editor popup), the test scenario fails with:

```
#1: (org.drools.scenariosimulation.backend.runner.AbstractScenarioRunner) : #1: Index: 0
```
As a result, it is not possible to test any DRL asset that works with a map. I'm attaching my project (with all the test scenarios described above) [here](https://issues.jboss.org/secure/attachment/12454277/Maps.zip).

lanceleverich:
Klara, using the project you linked in your project, I did not encounter any issues. I even added a scenario with multiple map entries (Map(2)) and it worked.

kkufova:
@lanceleverich, that is weird! It could be an error on my side. @danielezonca, could you please build the PR and double check this?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
```diff
@@ -33,6 +35,19 @@
...
+    */
+    private static List<String> SIMPLE_TYPES_LIST = Arrays.asList(Boolean.class.getCanonicalName(),
+                                                                Integer.class.getCanonicalName(),
+                                                                Long.class.getCanonicalName(),
+                                                                Double.class.getCanonicalName(),
+                                                                Float.class.getCanonicalName(),
+                                                                Short.class.getCanonicalName(),
+                                                                Byte.class.getCanonicalName(),
+                                                                Character.class.getCanonicalName(),
+                                                                LocalDate.class.getCanonicalName());
```
jomarko: Minor formatting issue:
```java
    private static List<String> SIMPLE_TYPES_LIST = Arrays.asList(Boolean.class.getCanonicalName(),
                                                                  Integer.class.getCanonicalName(),
                                                                  Long.class.getCanonicalName(),
                                                                  Double.class.getCanonicalName(),
                                                                  Float.class.getCanonicalName(),
                                                                  Short.class.getCanonicalName(),
                                                                  Byte.class.getCanonicalName(),
                                                                  Character.class.getCanonicalName(),
                                                                  LocalDate.class.getCanonicalName());
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
```diff
@@ -96,6 +111,9 @@ protected Object createObject(String className, List<String> genericClasses) {
             return new HashMap();
         } else {
             try {
+                if (SIMPLE_TYPES_LIST.contains(className)) {
+                    return null;
+                }
```
jomarko: Can be the `className` null? If not I would maybe rewrite as [1], what do you think? My point is to not nesting expression into try catch, if an exception can not be thrown.

[1]
```java
        if (ScenarioSimulationSharedUtils.isMap(className)) {
            return new HashMap();
        } else if (SIMPLE_TYPES_LIST.contains(className)) {
            return null;
        } else {
            try {
                return classLoader.loadClass(className).newInstance();
            } catch (Exception e) {
                throw new IllegalArgumentException("Impossible to instantiate " + className, e);
            }
        }
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
```diff
@@ -33,6 +35,19 @@
 public class BaseExpressionEvaluator extends AbstractExpressionEvaluator {
 
     private final ClassLoader classLoader;
+    /*
+     * Note: This list is similar to the SIMPLE_CLASSES_MAP found in the DataManagementStrategy interface
+     *
+    */
```
jomarko: Is this comment up to date? I am not able to find `SIMPLE_CLASSES_MAP` nor `DataManagementStrategy`. Is that in different repository?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -47,11 +50,62 @@ public void evaluateLiteralExpression() {
...
+    public void createObjectWithInvalidClass() {
+        String invalidClassName = "com.invalid.class.Name";
+
+        BaseExpressionEvaluator expressionEvaluator = new BaseExpressionEvaluator(classLoader);
+        Object obj = expressionEvaluator.createObject(invalidClassName, Collections.emptyList());
+
+        // This should never be reached since the invalid class name will cause the 
+        // expected exception
+        fail();
+    }
```
jomarko, line 67: We can make the test more strict (check the exception message), using assertj:
```java
    @Test
    public void createObjectWithInvalidClass() {
        String invalidClassName = "com.invalid.class.Name";

        BaseExpressionEvaluator expressionEvaluator = new BaseExpressionEvaluator(classLoader);
        Assertions.assertThatThrownBy(() -> expressionEvaluator.createObject(invalidClassName, Collections.emptyList()))
                .hasMessage("Impossible to instantiate " + invalidClassName);
    }
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -47,11 +50,62 @@ public void evaluateLiteralExpression() {
...
+
     @SuppressWarnings("unchecked")
     @Test
     public void expressionTest() {
         BaseExpressionEvaluator expressionEvaluator = new BaseExpressionEvaluator(classLoader);
 
+        String likeWorkbenchMapString = "{ \"Home\": { \"value\": \"123 Any Street\" } }";
+        List<String> genericClasses = new ArrayList<>();
+        genericClasses.add(String.class.getCanonicalName());
+        genericClasses.add(String.class.getCanonicalName());
```
jomarko, line 76: Is adding `String` twice into `genericClasses` intentional? I expected each class there just once. If that is not the case, could you please explain the purpose of the `genericClasses`?

jomarko (commented):
### Code review
Some comments to code inline.

jomarko (commented):
### Manual Review
The manual review with the original attached project **Maps.zip** passed on my machine. However I tried to update that project slightly. I changed the **ISBN** map key from **String** to the **Integer**. Updated the rule and scenario in way I expect it passes however it fails then. Maybe just other maps than **Map<String, String>** are not supported? See my updated [project](https://issues.jboss.org/secure/attachment/12454476/Maps-IntegerKey.zip)

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
lanceleverich: genericClasses is used to provide the list of expected class types in a List or Map. In a Map<String,String> there are two instances of the type String used for the generic.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
lanceleverich: Indents were added automatically by the IDE, according to the style settings that were provided.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
lanceleverich: Refactoring by Daniele has done away with this construct.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
lanceleverich: During refactoring by Daniele, this test was updated as you have suggested.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluator.java
lanceleverich: It is in a different repository/project, but is no longer relevant due to refactoring done by Daniele

danielezonca (approved):

jomarko (approved):
@danielezonca approving after our gchat discussion about supported maps. Proposing this documentation note https://github.com/kiegroup/kie-docs/pull/1837

PR #2505, opened 2019-08-15
RHDM-1049 Wrong output variable values in PMML mining model

lanceleverich:
* Changed output of the target field to be a simple value that is set at
the same time as the Sum_ version is created.
* Updated test to make sure change was effective in adding the output
value

### DIFF
--- a/kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/mining/MiningModelSumRegressionTest.java
+++ b/kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/mining/MiningModelSumRegressionTest.java
@@ -42,7 +42,8 @@ public class MiningModelSumRegressionTest {
     private static final String INPUT1_FIELD_NAME = "input1";
     private static final String INPUT2_FIELD_NAME = "input2";
     private static final String INPUT3_FIELD_NAME = "input3";
-    private static final String OUTPUT_FIELD_NAME = "Sum_Result";
+    private static final String SUM_OUTPUT_FIELD_NAME = "Sum_Result";
+    private static final String BASE_OUTPUT_FIELD_NAME = "Result";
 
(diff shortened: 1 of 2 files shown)


kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/mining/MiningModelSumRegressionTest.java
```diff
@@ -55,8 +56,8 @@
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
                 { Optional.of(10.0), Optional.of(10.0), Optional.of(10.0), 2070 },
-                { Optional.of(200.0), Optional.of(-1.0), Optional.of(2.0), -299 },
-                { Optional.of(90.0), Optional.of(2.0), Optional.of(4.0), 17040 },
+                                            {Optional.of(200.0), Optional.of(-1.0), Optional.of(2.0), -299},
```
jiripetrlik: Please revert back original indent.

jiripetrlik (changes requested):
Only minor comment. There is probably accidentally changed indent in one file. Otherwise looks good. But we need to wait for finishing of downstream PR check before merge.

jiripetrlik:

kie-pmml/src/test/java/org/kie/pmml/pmml_4_2/predictive/models/mining/MiningModelSumRegressionTest.java
lanceleverich: Fixed

jiripetrlik:
One more comment. I did not notice it before, but please change JIRA name in PR and in commit. It is RHDM-1049 instead of RHPAM-1049.

jiripetrlik (approved):

danielezonca (approved):

PR #2509, opened 2019-08-20
DROOLS-4447: Test case to reproduce java.lang.ClassCastException: org…

martinweiler:
….mvel2.optimizers.impl.refl.nodes.GetterAccessor cannot be cast to org.mvel2.optimizers.impl.refl.nodes.MethodAccessor


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/55edee1086e2bf4aaf8cc608c103dbeb1ffe7ee6

PR #2514, opened 2019-08-22
DROOLS-4454: Add test case to reproduce NPE in SegmentUtilities.creat…

martinweiler:
…eChildSegments


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/3b06d0675989615c664d13f5fc67091678269b52

PR #2519, opened 2019-08-23
DROOLS-4448: Include LocalDate into Package Data Model

Rikkola:
https://issues.jboss.org/browse/DROOLS-4448

https://github.com/kiegroup/kie-soup/pull/112
https://github.com/kiegroup/drools/pull/2519
https://github.com/kiegroup/kie-wb-common/pull/2826

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
@@ -19,6 +19,9 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


jomarko (approved):

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
```diff
@@ -465,6 +478,8 @@ static String inferDataTypeFromActionValue(final String param,
                                                final boolean isJavaDialect) {
         if (param.startsWith("sdf.parse(\"")) {
             return DataType.TYPE_DATE;
+        } else if (param.startsWith("java.time.LocalDate.parse(\"")) {
```
danielzhe, line 481: I think '"java.time.LocalDate.parse(\"' should be a constant, but the entire class doesn't use constants for cases like this. So maybe a future refactor? What do you think?

Rikkola: I think this is because it is hard to see what the actual code is if it is a constant. Best would be some sort of constant template, but I say we do this later.

Rikkola:
Let's merge this next week. The failing test needs to be fixed with proper time.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelPersistenceHelper.java
danielzhe: Ok, I agree.

danielzhe:
+1

danielzhe (approved):

Rikkola:

PR #2520, opened 2019-08-23
RHPAM-2307 Guided Scorecard can not be tested according its name

lanceleverich:
* Fixed methods that are used to find the package with a RuleUnit, so
that the model name can be in either upper or lower case.
* Updated the unit tests in kie-pmml so that they all use the
PMML4ExecutionHelper. This allowed refactoring of other classes, to
remove duplicated methods for determining the RuleUnit class to be used.
* Updated one of the test PMML resources (in kie-pmml) so that it has a
model name that starts with a lower case letter. This should make sure
that we are testing models with both upper and lower case letters at the
start of the model name.


baldimir:
@jiripetrlik can you please review this?

jiripetrlik (approved):

jiripetrlik:

PR #2521, opened 2019-08-25
DROOLS-4468: Add test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/6162ca7ea4f39d8930b7bd00788b942fa0b40c40

PR #2523, opened 2019-08-26
[DROOLS-4475] generate stable ids for field indexes across different …

mariofusco:
…executable model compilations


baldimir:

PR #2525, opened 2019-08-27
DROOLS-4467 PMML with capitalized field names does not work

lanceleverich:
* Updated the `findDataDictionaryEntry` method of `PMML4UnitImpl`. It now looks through the raw `DataDictionary` values to find the entry, if the entry isn't found using the field name as a key to the
`dataDictionaryMap`.
* Changed the `AbstractModel.getMiningFields` method (used by all models except the MiningModel) to use findDataDictionaryEntry instead of trying to use the field name as key into the `dataDictionaryMap`.
* Added a set of unit tests to make sure that the `findDataDictionaryEntry` method works as intended. This includes also adding a PMML file that contains fields, with names that start with both upper and lower cases.


tarilabs:
@jiripetrlik something you can kindly review, please?

jiripetrlik (approved):
Looks good to me.

mariofusco:
jenkins test

baldimir:
jenkins retest this

PR #2535, opened 2019-09-09
[DROOLS-4503] "Error" popup report java literal for suggested number

yesamer:
@jomarko @gitgabrio @danielezonca Can you please review and test it?

To solve the issue, I removed the "L" literal when returning a Long type object.
The reason is that parsing a String with "L" literal leads to an exception, as in the following example:

>  Long myLong = Long.parseLong("1L") // Throws NumberFormatException

Double and Float supports the parse of "D" and "F" literals.

https://issues.jboss.org/browse/DROOLS-4503


drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtilTest.java
```diff
@@ -142,6 +142,10 @@ public void convertValueTest() {
         assertEquals(1.0D, convertValue(Double.class.getCanonicalName(), "1.0", classLoader));
         assertEquals(1.0F, convertValue(float.class.getCanonicalName(), "1.0", classLoader));
         assertEquals(1.0F, convertValue(Float.class.getCanonicalName(), "1.0", classLoader));
+        assertEquals(1.0D, convertValue(double.class.getCanonicalName(), "1.0D", classLoader));
```
jomarko: Briefly checked this [page](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html) and have question, shouldn't we use `1.0d` and `1.0f` for simple data types instead of `1.0D` and `1.0F`

jomarko (commented):
Question to code before proceed to manual review.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtilTest.java
```diff
@@ -157,7 +161,7 @@ public void revertValueTest() {
         assertEquals("Test", revertValue("Test"));
         assertEquals("false", revertValue(false));
         assertEquals("1", revertValue(1));
-        assertEquals("1L", revertValue(1L));
+        assertEquals("1", revertValue(1L));
         assertEquals("1.0D", revertValue(1.0D));
         assertEquals("1.0F", revertValue(1.0F));
```
danielezonca: Can you please add a test where it performs the round-trip? Starting from real value (or expression as you prefer) do a `convertValue` and then a `revertValue` and assert the result to be the same of the original value

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtilTest.java
```diff
@@ -156,17 +169,96 @@ public void convertValueTest() {
...
+        assertNull(convertValue(String.class.getCanonicalName(), revertValue(null), classLoader));
+    }
+
+    @Test public void puppa ()  {
+
+        BaseExpressionEvaluator puppa2 = new BaseExpressionEvaluator(classLoader);
+        assertNull(puppa2.evaluateLiteralExpression(String.class.getCanonicalName(), Collections.emptyList(), "null"));
+
+
+    }
```
jomarko: Is this test intentionally here?

yesamer: @jomarko No, removed

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtilTest.java
yesamer: @danielezonca done.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtilTest.java
yesamer: @jomarko Java supports both ways. As you pointed, docs usually use the lowercase character, then I think it's fine to change it.

yesamer:
jenkins retest this please

jomarko:
@yesamer hello, I am not sure if it is related to changes I asked you for, but now I am not able to get that **Error reason** popup in format: **the expected vale is but the actual is**. Always getting just the attached one.

Screenshot from 2019-09-12 07-32-13

yesamer:
@jomarko I will check it, thank you.

yesamer:
> @yesamer hello, I am not sure if it is related to changes I asked you for, but now I am not able to get that **Error reason** popup in format: **the expected vale is but the actual is**. Always getting just the attached one.
> 
> Screenshot from 2019-09-12 07-32-13

@jomarko Regarding the reported situation:
1) It seems to me NOT related to this PR changes
2) In my understanding you have to have the same Instance/property in both Given / Expected groups, then it looks as the expected behaviour.

yesamer:
jenkins retest this please

jomarko (approved):
Approving and starting discussion outside github to understand more.

yesamer:
jenkins retest this please

yesamer:
@danielezonca or @gitgabrio can you please review the code?

danielezonca (approved):

PR #2538, opened 2019-09-11
DROOLS-4525: Add test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/3770bc5fb78a578e52525ac32406212a01ece5d1

PR #2540, opened 2019-09-15
DROOLS-4532: Test case for java.lang.RuntimeException: Unknown expres…

martinweiler:
…sion type: -1


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/79b2fda484263c587ed2aed992d58e3afd082ec3

PR #2541, opened 2019-09-15
DROOLS-4534: Test case for ClassCastException using accumulate max

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/37100dac82f6e405789f091399f4b4c5f1fd24df

PR #2542, opened 2019-09-17
[DROOLS-4522] Add an "expression" property for every complex type in Scenario Testing

yesamer:
@jomarko @gitgabrio @danielezonca can you please test and review it?

This represents the first step of development for https://issues.jboss.org/browse/BAPL-1402

Basically, with the changes introduced in this PR, every complex type object in Scenario has an additional property ("expression") where it will be possible to add expression logic to the referred object (fact). At the moment, the changes involve only front-end side.

https://issues.jboss.org/browse/DROOLS-4522

Related PR: https://github.com/kiegroup/drools-wb/pull/1232

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/ConstantsHolder.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/ConstantsHolder.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


gitgabrio (approved):
@yesamer 
Thanks! One of your most brilliant job!

jomarko:
On the first sight, PR seems really simple, however I have concern about I18N. Shouldn't we use GWT Constants as my understanding is this **expression** will appear in UI. Am I missing something? Does new scenarios hardcode everything in code and do not support any other language?

jomarko (commented):
@yesamer hello, thanks for the PR. On the first sight, PR seems really simple, however I have concern about I18N. Shouldn't we use GWT Constants as my understanding is this expression will appear in UI. Am I missing something? Does new scenarios hardcode everything in code and do not support any other language?

gitgabrio:
> @yesamer hello, thanks for the PR. On the first sight, PR seems really simple, however I have concern about I18N. Shouldn't we use GWT Constants as my understanding is this expression will appear in UI. Am I missing something? Does new scenarios hardcode everything in code and do not support any other language?

@jomarko @danielezonca 
Hi. This constants are meant to be used only inside code, to replace all the strings we have used (e.g inside switch). They have nothing to do with i18n/client - they are not shown to users. Progressively we'll move all such strings here.

jomarko (approved):

yesamer:
@jomarko @dupliaka @gitgabrio Please consider this PR on hold, because the UI introduced needs a review from UX/UI team. Thanks

PR #2550, opened 2019-09-24
[DROOLS-4558] executable-model doesn't fully parse multi-line pattern

tkobayas:
- unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/34d3f9e6c4a8306da2c2b7c3a1ec88a03d1c6438

PR #2551, opened 2019-09-24
[DROOLS-4560] executable-model wrongly calculates in eval with parent…

tkobayas:
…hesis

- Unit test


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/392ec623cc13a14f589a33b90d819fc98b71325b

PR #2553, opened 2019-09-24
[DROOLS-4513] - s2i build fails at KieServerContainerVerifier

spolti:
Signed-off-by: Filippe Spolti <fspolti@redhat.com>

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
@@ -51,6 +51,8 @@ import org.kie.internal.builder.KnowledgeBuilderConfiguration;
 import org.kie.internal.builder.ResourceChangeSet;
 import org.kie.internal.utils.ClassLoaderResolver;
 import org.kie.internal.utils.NoDepsClassLoaderResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.buildKieModule;
 import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.filterFileInKBase;
@@ -146,13 +148,14 @@ public interface InternalKieModule extends KieModule, Serializable {
 
     default CompilationCache getCompilationCache( String kbaseName) { return null; }
 
-    static InternalKieModule createKieModule( ReleaseId releaseId, File jar ) {
-        try (ZipFile zipFile = new ZipFile( jar )) {
+    static InternalKieModule createKieModule(ReleaseId releaseId, File jar) {
+        try (ZipFile zipFile = new ZipFile(jar)) {
             ZipEntry zipEntry = zipFile.getEntry(KieModuleModelImpl.KMODULE_JAR_PATH);
             KieModuleModel kieModuleModel = KieModuleModelImpl.fromXML(zipFile.getInputStream(zipEntry));
             setDefaultsforEmptyKieModule(kieModuleModel);
-            return kieModuleModel != null ? InternalKieModuleProvider.get(adapt( releaseId ), kieModuleModel, jar) : null;
-        } catch ( Exception e ) {
+            return kieModuleModel != null ? InternalKieModuleProvider.get(adapt(releaseId), kieModuleModel, jar) : null;
+        } catch (Exception e) {
+            LocalLogger.logger.error(e.getMessage(), e);
         }
         return null;
     }
@@ -193,5 +196,9 @@ public interface InternalKieModule extends KieModule, Serializable {
             this.className = className;
             this.bytecode = bytecode;
         }
+    }
+
+    final class LocalLogger {
+        private static final Logger logger = LoggerFactory.getLogger(InternalKieModule.class);
     }
 }


drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
```diff
@@ -146,13 +148,14 @@ default ProjectClassLoader createModuleClassLoader( ClassLoader parent ) {
...
+    static InternalKieModule createKieModule(ReleaseId releaseId, File jar) {
+        try (ZipFile zipFile = new ZipFile(jar)) {
             ZipEntry zipEntry = zipFile.getEntry(KieModuleModelImpl.KMODULE_JAR_PATH);
             KieModuleModel kieModuleModel = KieModuleModelImpl.fromXML(zipFile.getInputStream(zipEntry));
             setDefaultsforEmptyKieModule(kieModuleModel);
-            return kieModuleModel != null ? InternalKieModuleProvider.get(adapt( releaseId ), kieModuleModel, jar) : null;
-        } catch ( Exception e ) {
+            return kieModuleModel != null ? InternalKieModuleProvider.get(adapt(releaseId), kieModuleModel, jar) : null;
+        } catch (Exception e) {
+            LocalLogger.logger.error(e.getMessage(), e);
```
mswiderski, line 158: so in the end just logging and not throwing? what was the reason for it?

mswiderski:
@spolti would be good to link this PR with the one in droolsjbpm-knowledge repo since they refer to the same jira

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
baldimir: I am also interested in this.

spolti:
Related PR:  https://github.com/kiegroup/droolsjbpm-knowledge/pull/402

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
spolti: @mswiderski I sent you email to you and @mariofusco, I'll add it's content here for a broader discussion.

Mario had a concern about add a thrown new Exception here:
https://github.com/kiegroup/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java#L155

By adding a thrown there I got the messages I need on KieContainerVerifier, but, some tests starts to fail, like:
https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/drools-pullrequests/348/testReport/junit/org.drools.example.api.kiecontainerfromkierepo/KieContainerFromKieRepoExampleTest/testGo/

If I add a try/catch block around this line
https://github.com/kiegroup/drools/blob/master/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java#L152
I am not sure why, but the catch block is not called.


Another possible solution I found was:

    static InternalKieModule createKieModule(ReleaseId releaseId, File jar) {
        try (ZipFile zipFile = new ZipFile(jar)) {
            ZipEntry zipEntry = zipFile.getEntry(KieModuleModelImpl.KMODULE_JAR_PATH);
            KieModuleModel kieModuleModel = KieModuleModelImpl.fromXML(zipFile.getInputStream(zipEntry));
            setDefaultsforEmptyKieModule(kieModuleModel);
            return kieModuleModel != null ? InternalKieModuleProvider.get(adapt(releaseId), kieModuleModel, jar) : null;
        } catch (Exception e) {
            if (e instanceof NullPointerException) { // or another catch block
                return null;
            }
            throw new RuntimeException(e.getMessage(), e);
        }
    }


This way I still getting the error messages, the KieContainerFromKieRepoExampleTest/testGo/ stops to complain but another test failed:

https://github.com/kiegroup/drools/blob/master/kie-ci/src/test/java/org/kie/scanner/KieRepositoryScannerTest.java#L237

with the stacktrace: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/drools-pullrequests/350/testReport/junit/org.kie.scanner/KieRepositoryScannerTest/testKieScannerScopesNotRequired_Manually_wired_component_provider_false_/


With that said, it confirm the Marios' concern about use a thrown on tha try/catch.

So, for now I added a logger there, not sure if there is a better way to log stuff on interfaces:

```
} catch (Exception e) {
    LocalLogger.logger.error(e.getMessage(), e);
}

....

final class LocalLogger {
    private static final Logger logger = LoggerFactory.getLogger(InternalKieModule.class);
}

```


Besides that, we could refactor the createKieModule method to not use try with resources and deal with specific cases on different catch blocks.

What do you guys think?

spolti:
@mswiderski @mariofusco did you have time to take a look on my comment? thoughts?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/InternalKieModule.java
mswiderski: thanks for detailed explanation @spolti 

if that solves the original issue then I think we are good to go here.

mswiderski (approved):

mariofusco (approved):

PR #2556, opened 2019-09-27
[DROOLS-4572] a fact with many properties causes a problem with "not"…

tkobayas:
… rule in executable-model

- unit test


lucamolteni:
@mareknovotny Not much to review here, the fix is yet to be done :)

mareknovotny:
sure @lucamolteni, just tagged you are notified ;)

mariofusco:
Reproduce and fixed by https://github.com/kiegroup/drools/commit/cd1d9f3e65800c786c42f6602f3ca115505fcf32

PR #2558, opened 2019-09-29
DROOLS-4577: Test case to verify expired events with additional const…

martinweiler:
…raints


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/8c4a260f764c77282beb363ddaa158532a2f7f75

PR #2559, opened 2019-09-30
[DROOLS-4580] Test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/a2a3fdcc1e9b32ab5eaf4bed540accd3e1eef7a3

PR #2563, opened 2019-10-02
[DROOLS-4588] Add support for MVEL expression on simple types in Rule test scenario

danielezonca:
https://issues.jboss.org/browse/DROOLS-4588

Implementation details:
- Add `#` at the beginning of the expression to use MVEL syntax (i.e. `# 10 + 10`)
- Use `actualValue` variable to refer actual value in a EXPECT condition that use MVEL (i.e. `# java.util.Objects.equals(actualValue, 10)`)
- MVEL expression used in GIVEN cell has to return a value/object of the same type of the corresponding column
- MVEL expression in EXPECT columns should return a `boolean` value or the same type of the column and in that case it will be used `compare`/`equals` to compare with actual value (i.e. `# new MyObject()` in a EXPECT cell will be evaluated as `actualValue.compareTo(new MyObject())` if `MyObject` is `Comparable` or `actualValue.equals(new MyObject())` otherwise). This corner case has been introduced to avoid to write `# actualValue.equals(...)` in every MVEL expression in EXPECT section

@yesamer @gitgabrio @dupliaka


drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
+        if (Type.DMN.equals(type)) {
+            return getOrCreateDMNExpressionEvaluator();
+        }
+
+        Object rawValue = factMappingValue.getRawValue();
+
+        if (rawValue instanceof String && ((String) rawValue).startsWith("#")) {
```
yesamer: @danielezonca 
I suggest to always use a static variable to refer to "#" char, in any place is used, in order to easily update it in case of change request.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactoryTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void create() {
+        assertNotNull(ExpressionEvaluatorFactory.create(classLoader, ScenarioSimulationModel.Type.RULE));
+    }
+
+    @Test
+    public void getOrCreate() {
+        FactMappingValue simpleFMV = new FactMappingValue(FactIdentifier.INDEX, ExpressionIdentifier.INDEX, "10");
+        FactMappingValue objectFMV = new FactMappingValue(FactIdentifier.INDEX, ExpressionIdentifier.INDEX, "10");
+        FactMappingValue mvelFMV = new FactMappingValue(FactIdentifier.INDEX, ExpressionIdentifier.INDEX, "# 10");
```
yesamer: Same of previous comment. (For all tests)

yesamer (approved):
@danielezonca well done, just a very little suggestion.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        this.classLoader = classLoader;
+        this.type = type;
+    }
+
+    public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
+        if (Type.DMN.equals(type)) {
+            return getOrCreateDMNExpressionEvaluator();
+        }
+
+        Object rawValue = factMappingValue.getRawValue();
```
dupliaka, line 57: SuggestionL: rawValue could be final

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+
+        if (rawValue instanceof String && ((String) rawValue).startsWith("#")) {
+            return getOrCreateMVELExpressionEvaluator();
+        } else {
+            return getOrCreateBaseExpressionEvaluator();
+        }
+    }
```
dupliaka, line 64: Suggestion: better to keep only one exit point

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+
+        Map<String, Object> params = new HashMap<>();
+        params.put(ACTUAL_VALUE_IDENTIFIER, resultValue);
+
+        Object expressionResult = compileAndExecute((String) rawExpression, params);
+        if (!(expressionResult instanceof Boolean)) {
+            // try to compare via compare/equals operators
+            return BaseExpressionOperator.EQUALS.eval(expressionResult, resultValue, resultClass, classLoader);
+        }
+        return (boolean) expressionResult;
```
dupliaka, line 61: Suggestion: One exit point as well.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
```diff
@@ -57,12 +57,12 @@ public boolean evaluateUnaryExpression(Object rawExpression, Object resultValue,
     }
 
     @Override
-    public Object evaluateLiteralExpression(String className, List<String> genericClasses, Object raw) {
-        if (!(raw instanceof String)) {
-            return raw;
+    public Object evaluateLiteralExpression(String className, List<String> genericClasses, Object rawExpression) {
```
gitgabrio, line 60: @danielezonca 
It seems that this method is exactly the same as BaseExpressionEvaluator.evaluateLiteralExpression.
If so, can't we unify them in one single place ? Since it does not use any "state" - it could also be a _static_ method somewhere

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
gitgabrio: @danielezonca 
Beside @yesamer suggestion, and to recap what we discussed, I think it would be a good idea to have two "expression" indicators (not only the "#") to clearly differentiate all over the whole stack (from user interface to backend) if the expression is meant as "value generator" or "method invocator". Even if now it is not strictly needed, it would reflect more exactly the possible use cases without relying on convention/assumption, and in general this lead to better results/avoid issues in long term.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
```diff
@@ -204,20 +207,23 @@ protected void validateAssertion(List<ScenarioResult> scenarioResults, Scenario
...
         } else if (!resultValue.isSatisfied()) {
-            expectedResult.setErrorValue(expressionEvaluator.fromObjectToExpression(resultValue.getResult()));
+            try {
+                expectedResult.setErrorValue(expressionEvaluator.fromObjectToExpression(resultValue.getResult()));
+            } catch (Exception e) {
+                expectedResult.setExceptionMessage(e.getMessage());
+            }
         } else {
             expectedResult.resetStatus();
         }
```
dupliaka:
```java
if (resultValue.isSatisfied()) {
            expectedResult.resetStatus();
        } else {
            if (resultValue.getErrorMessage().isPresent()) {
                expectedResult.setExceptionMessage(resultValue.getErrorMessage().get());
            } else {
                try {
                    expectedResult.setErrorValue(expressionEvaluator.fromObjectToExpression(resultValue.getResult()));
                } catch (Exception e) {
                    expectedResult.setExceptionMessage(e.getMessage());
                }
            }
        }
```
I know it looks a bit spaghetti like, but at least it is more readable

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+public class ExpressionEvaluatorFactory {
+
+    private final ClassLoader classLoader;
+    private final Type type;
+
+    private BaseExpressionEvaluator baseExpressionEvaluator;
+    private DMNFeelExpressionEvaluator dmnFeelExpressionEvaluator;
+    private MVELExpressionEvaluator mvelExpressionEvaluator;
+
+    public static ExpressionEvaluatorFactory create(ClassLoader classLoader, Type type) {
```
gitgabrio, line 38: @danielezonca 
Is this indirection really useful? Hiding the instantiation of the **ExpressionEvaluator** behind a static _Factory_ that in turns hide its private constructor, and at the end **ExpressionEvaluator** instantiation depends on a state of the factory itself. I do not read the whole code at glance, so it is possible there is a good reason, but looking at this piece I fail to understand it.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        Object rawValue = factMappingValue.getRawValue();
+
+        if (rawValue instanceof String && ((String) rawValue).startsWith("#")) {
+            return getOrCreateMVELExpressionEvaluator();
+        } else {
+            return getOrCreateBaseExpressionEvaluator();
+        }
+    }
+
+    private ExpressionEvaluator getOrCreateBaseExpressionEvaluator() {
```
gitgabrio, line 66: @danielezonca 
Related to comment about factory, here it is clear that the code need a _state_ to works because of the original design, but I have the impression that, without the factory indirection, all this could have been implemented in more functional way

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.drools.core.util.MVELSafeHelper;
+import org.kie.soup.project.datamodel.commons.util.MVELEvaluator;
+import org.mvel2.MVEL;
+import org.mvel2.ParserConfiguration;
+import org.mvel2.ParserContext;
+
+public class MVELExpressionEvaluator implements ExpressionEvaluator {
+
+    public static final String MVEL_ESCAPE_SYMBOL = "#";
```
gitgabrio: @danielezonca 
This relates to my comment about having two -specific - mvel escape symbol, one for "value generator" and another for "method invocation"

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
```diff
@@ -164,7 +165,7 @@ public void run(KieContainer kieContainer, SimulationDescriptor simulationDescri
     protected Map<List<String>, Object> getParamsForBean(SimulationDescriptor simulationDescriptor,
                                                          FactIdentifier factIdentifier,
                                                          List<FactMappingValue> factMappingValues,
-                                                         ExpressionEvaluator expressionEvaluator) {
+                                                         ExpressionEvaluatorFactory expressionEvaluatorFactory) {
         Map<List<String>, Object> paramsForBean = new HashMap<>();
 
         for (FactMappingValue factMappingValue : factMappingValues) {
```
dupliaka, line 184: Suggestion: That is not regression case but probably we could Create one instance of IllegalStateException("Wrong expression, this should not happen")) before the loop and reuse it inside?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        this.classLoader = classLoader;
+        this.config = new ParserConfiguration();
+        config.setClassLoader(classLoader);
+    }
+
+    @Override
+    public boolean evaluateUnaryExpression(Object rawExpression, Object resultValue, Class<?> resultClass) {
+        if (!(rawExpression instanceof String)) {
+            String rawClass = rawExpression == null ? null : rawExpression.getClass().getCanonicalName();
+            throw new IllegalArgumentException("Raw value should be a String and not a '" + rawClass + "'");
```
gitgabrio: @danielezonca 
I would replace "Raw value" with "Expression"; beside that: how it is possible to get to this method with a "rawExpression" that is not a "String" ? I have the impression it should be blocked before

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        Map<String, Object> params = new HashMap<>();
+        params.put(ACTUAL_VALUE_IDENTIFIER, resultValue);
+
+        Object expressionResult = compileAndExecute((String) rawExpression, params);
+        if (!(expressionResult instanceof Boolean)) {
```
gitgabrio, line 57: @danielezonca 
This relates on the assumptions that I think we should not rely - and code - upon

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+            // try to compare via compare/equals operators
+            return BaseExpressionOperator.EQUALS.eval(expressionResult, resultValue, resultClass, classLoader);
+        }
+        return (boolean) expressionResult;
+    }
+
+    @Override
+    public Object evaluateLiteralExpression(String className, List<String> genericClasses, Object rawExpression) {
+        if (!(rawExpression instanceof String)) {
+            throw new IllegalArgumentException("Raw value should be a String and not a '" + rawExpression.getClass().getCanonicalName() + "'");
```
gitgabrio: @danielezonca 
See my previous comment

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
```diff
@@ -204,20 +207,23 @@ protected void validateAssertion(List<ScenarioResult> scenarioResults, Scenario
...
 
     protected ScenarioResult fillResult(FactMappingValue expectedResult,
-                                        FactIdentifier factIdentifier,
                                         Supplier<ResultWrapper<?>> resultSupplier,
                                         ExpressionEvaluator expressionEvaluator) {
         ResultWrapper<?> resultValue = resultSupplier.get();
 
         if (!resultValue.isSatisfied() && resultValue.getErrorMessage().isPresent()) {
             expectedResult.setExceptionMessage(resultValue.getErrorMessage().get());
         } else if (!resultValue.isSatisfied()) {
```
gitgabrio: @danielezonca 
Reverting the order of these evaluations would simplify the overall structure: the first two _if_ relates to the same main condition: _!resultValue.isSatisfied()_ and the first one (_resultValue.getErrorMessage().isPresent()_)   is a sub-condition of it

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/MigrationStrategy.java
```diff
@@ -36,37 +36,37 @@
      * Method to obtain the migration function from 1.0 to 1.1
      * @return
      */
-    ThrowingConsumer<Document>  from1_0to1_1();
+    ThrowingConsumer<Document> from1_0to1_1();
```
gitgabrio, line 39: @danielezonca 
Good job!

gitgabrio (changes requested):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
gitgabrio: @danielezonca @yesamer 
I think this static constant could be put in the class @yesamer created to store all such kind of strings, since it could potentially be used in a lot of different places

yesamer: @gitgabrio @danielezonca Unfortunately, the PR where I introduced that class is now closed, because it was part of the previous FE logic which was discarded. If we want to put this (and other constants) in a single "constants holder" class, we need to add it in this PR.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertEquals("Value", evaluator.evaluateLiteralExpression(String.class.getCanonicalName(), Collections.emptyList(), "# \"Value\""));
+
+        assertThatThrownBy(() -> evaluator.evaluateLiteralExpression(String.class.getCanonicalName(), Collections.emptyList(), "1+"))
+                .isInstanceOf(RuntimeException.class);
+
+        assertThatThrownBy(() -> evaluator.evaluateLiteralExpression(String.class.getCanonicalName(), Collections.emptyList(), new Object()))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageStartingWith("Raw value should be a String");
+
+        assertThatThrownBy(() -> evaluator.evaluateLiteralExpression(String.class.getCanonicalName(), Collections.emptyList(), new Object()))
```
dupliaka: duplicate

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
gitgabrio: @yesamer @danielezonca 
If possible, I would prefer to add it now - I think it would be useful to have it, and this could be a good time to introduce it.

jomarko: @yesamer @danielezonca @gitgabrio I am not sure, but wouldn't be the topic we discuss here, two characters for distinguish expression type, brought a backward incompatibility? Asking as currently users do not need to prefix cell inputs with any character, am I missing something?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
jomarko: We should extract this piece of code [1] as a private method and use it both in `evaluateUnaryExpression` and `evaluateLiteralExpression`

[1]
```java
if (!(rawExpression instanceof String)) {
    String rawClass = rawExpression == null ? null : rawExpression.getClass().getCanonicalName();
    throw new IllegalArgumentException("Raw value should be a String and not a '" + rawClass + "'");
}
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        return compileAndExecute((String) rawExpression, Collections.emptyMap());
+    }
+
+    @Override
+    public String fromObjectToExpression(Object value) {
+        throw new UnsupportedOperationException("The condition has not been satisfied");
```
jomarko, line 80: The message doesn't sound describing unsupported operation to me, but just double checking.

jomarko (commented):
Few questions, also a PR with coverage increase is coming.

jomarko:
Proposed coverage increase https://github.com/kiegroup/drools/pull/2566

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: @danielezonca @gitgabrio @jomarko If I correctly understood, the 2 new characters will cover new cases: one for handle value generator and another to handle method invocator. Both will be managed with the new MVEL expression evaluator introduced by Daniele. This means the old expression evaluator (without any char) will continue to work. Thus, I don't expect problem with backward incompatibility. Am I correct?

gitgabrio: @jomarko 
Hi. This is the result of discussion with UX. Anyway, point is: inside DMN editor the user is expected to insert "expressions", but until now - _inside scenario grid_ - (s)he 's expected to insert _raw values_. Now, this is the proposed way to let the user insert arbitrarily _raw value_ **OR** _expression_ - the prefixes are needed to know (server side) what is the case (I'm asking for two different prefixes because the "expression" use case has - in turns - two different usages)

jomarko: @gitgabrio @yesamer thank you, if the discussed characters are related just tho new MVEL, I am fine, just would be good to add both characters into cheat-sheet we have in right panel.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
danielezonca: This method is invoked to obtain a String expression that represent the actual value but this is not possible with MVEL expression. That's the reason of the `UnsupportedOperationException`. At the same time this message is the error message displayed in the UI so I decided to use this message instead of "It is not possible to suggest the value to use"

jomarko: @danielezonca thank you, I have no better message in mind, so fine for me.

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.drools.scenariosimulation.api.utils;
+
+/**
+ * Class which contains shared Constants in Scenario Simulation module
+ */
+public class ConstantsHolder {
+
+    /* Constants for manage MVEL expressions */
+    public static final String MVEL_ESCAPE_SYMBOL = "#";
```
gitgabrio, line 24: @danielezonca 
Please use two different constants to differentiate the use cases - beside that, probably is better to have a slightly more complex pattern - `#` is too easy and common to use

gitgabrio (changes requested):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
danielezonca: Moved symbol to a common class.
About the possibility to use two different symbols to distinguish "expression that define a value" and "expression that verify a condition" this is an ongoing discussion so I will add a comment to the PR as soon as this will be clarified (probably on Monday)

@jomarko 
@yesamer is right, this PR introduce MVEL support to Rule scenarios and we are discussing about MVEL expression only, a simple value can still be specified without MVEL (i.e. `MyString`)

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactoryTest.java
danielezonca: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
danielezonca: I don't like to specify `final` for local variable, actually most/all of local variable we are usually using could be declared as `final` (Jave 8 introduced the definition of _effectively final_ for this kind of situation) but I think this has an impact in terms of readability without a real benefit

dupliaka: Sorry I suggest this idea just for better code readability so the next user could understand the value won't change anymore. I have notice that practice several times in the code so I've made a wrong conclusion about this code style agreement. Thanks

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/MigrationStrategy.java
danielezonca: Yes, I'm very proud of this change :smile:

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
danielezonca: The method is really small and to have a single exit point I need to initialize a variable initialized as `null` and then reassign it. I prefer to have multiple returns rather than additional variables (especially with small methods)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
danielezonca: Same comment as above

jomarko:
@danielezonca hello, is this ready for a re-review?

danielezonca:
Not yet, I want to perform another couple of tests on my side first

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
danielezonca: I have created a ticket for that [DROOLS-4618](https://issues.jboss.org/browse/DROOLS-4618)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
danielezonca: As agreed added javadoc to explain the logic of this factory class

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
danielezonca: Same as above

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
danielezonca: Error renamed and created a ticket for expression evaluator refactoring [DROOLS-4618](https://issues.jboss.org/browse/DROOLS-4618)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
danielezonca: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
danielezonca: Done

danielezonca:
@gitgabrio @jomarko @dupliaka 
PR is ready for review

gitgabrio (approved):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
jomarko: @danielezonca it seems you decided to not incorporate my point about `if (!(rawExpression instanceof String))` extraction as private method, could you please at least respond why?

jomarko (approved):
Approving, however one of my comments from previous review round still not answered.

danielezonca:
@jomarko 
I have created this ticket https://issues.jboss.org/browse/DROOLS-4618 to refactor all `ExpressionEvaluator`s and change `rawExpression` from `Object` to `String` as Gabriele suggested. Doing that I plan to refactor/remove also all these `instanceof String`. Sorry but I forgot to explicitly answer you about that point, feel free to edit/add information in that ticket too.
Is it fine for you?

jomarko:
@danielezonca unfortunately, I think it doesn't cover my comment. There is no link to code lines/my github comment/this original PR. However I have no capacity to invest more time into this PR. Assume my approval still as valid.

dupliaka (commented):
Reviewed with comments

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
dupliaka: Done in later commit
https://github.com/kiegroup/drools/blob/09e30a280c914af88241ff04184a4aad4c255b0e/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java#L232-L246

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
danielezonca: I think we should stay with the code like it is now because if we create the instance of `IllegalStateException` at the beginning of the method the stack trace will refer to that line instead of the line where the proper `throw` will be performed. Moreover in the current code the exception is created lazily (only if it happens) while if we move this code at the beginning of the method it will be always created

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
danielezonca: Done, good catch :)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/AbstractRunnerHelper.java
dupliaka: All right

dupliaka (approved):
Manual check done.
Here is a small summary:

Expressions for Rule based scenarios for simple types
Starts from “#”
Runs for properties
Runs for duplicated properties
Returns the same type as property type in GIVEN
Returns boolean in EXPECTED
Receives rule processed value as a parameter in EXPECTED via keyword “actualValue”
Applies java.util.Objects.equals(actualValue, expression result) if the expression result if the EXPECTED value not a boolean
Works for Simple types only
Could run any of java function for the project
Exports and imports as any other value


Limitation for GIVEN
1. Mvel expression results not property same type - drives scenario failure, error reason and highlighted
Use case:
User inputs 1/2 to Integer property
2.Mvel execution throws exception - drives scenario execution stopping, failure, error alert
Use case:
User inputs inexistent MVEL function  in one of the properties
3.Mvel expression exception stops expression execution for all EXPECT colons
Use case:
User got one of the GIVEN fact expression excepted 
4.It cannot be mixed with simple expressions in one cell
Use cases:
For cell inputs :
[#4+4,68]
#actualvalue>1

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

PR #2565, opened 2019-10-03
DROOLS-4601 : Escape characters added to code after editing comma separated list in guided rule editor

Rikkola:
https://issues.jboss.org/browse/DROOLS-4601

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceUnmarshallingTest.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceUnmarshallingTest.java
@@ -9563,4 +9563,27 @@ public class RuleModelDRLPersistenceUnmarshallingTest extends BaseRuleModelTest 
         assertEqualsIgnoreWhitespace(drl,
                                      RuleModelDRLPersistenceImpl.getInstance().marshal(model));
     }
+
+    @Test
+    public void unmarshalStringListsCorrectly() {
+        final String drl = "package org.mortgages;\n" +
+                "rule \"aaa\"\n" +
+                "\tdialect \"mvel\"\n" +
+                "\twhen\n" +
+                "\t\tApplicant( applicantName : name in ( \"a\", \"b\", \"c\" ) )\n" +
+                "\tthen\n" +
+                "end";
+
+        when(dmo.getPackageName()).thenReturn("org.mortgages");
+
+        final RuleModel model = RuleModelDRLPersistenceImpl.getInstance().unmarshal(drl,
+                                                                                    Collections.emptyList(),
+                                                                                    dmo);
+
+        SingleFieldConstraint fieldConstraint = model.getLHSBoundField("applicantName");
+
+        assertEquals("a, b, c", fieldConstraint.getValue());
+        assertEqualsIgnoreWhitespace(drl,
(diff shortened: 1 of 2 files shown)


Rikkola:
@jomarko @manstis if you two could be kind enough to review.

jomarko:
No comments to the code, however it seems to me quotes are added in some strange way now:

- Screenshot from 2019-10-09 10-06-30
- 
Screenshot from 2019-10-09 10-06-39

Rikkola:
@jomarko that is actually how master also works without this PR. I dare to say quotes haven't worked in that setup before. I'll try to see if there if there is a way to add support for this without breaking backwards compatibility.

jomarko:
@Rikkola ok, sorry I need to clarify. Are you saying that you are fixing just case, when user put in values without double quotes and then after reopening still no double quotes added in UI? just in the source code? And are you saying that if user put in values already with double quotes it is not working correctly even without your PR?

Rikkola:
Yes so even without this change, if you used double quotes like this `a, "b,c", d` it does a broken split between b and c. It looks like quotes do not work with list in GRE and never have.

This PR fixes the problem when you have a list `a,b,c` and it is changed to `"a", "b", "c"` in source. However if you reopen the rule textbox content is `\"a\", \"b\", \"c\"` if you save for a second time a new set of quotes is added and you can keep adding quotes this way infinitely. This fix keeps the content as it is, even if it is broken.

Rikkola:
I'm not sure if that helped. The explanation is not the clearest :)

Rikkola:
@jomarko what are your feelings on us making a separate ticket for the issue you found? That is how it worked before and it looks like the " quotes are not used to group the content inside of them to produce a single "a, b" element. There is a small possiblity that we break backward compatibility for any items that use " character in as an option. Although I it is very unlikely that anyone uses it.

jomarko:
New issue reported as https://issues.jboss.org/browse/RHPAM-2457

jomarko (approved):

jomarko:
@manstis could you please review and possibly merge?

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceUnmarshallingTest.java
```diff
@@ -9563,4 +9563,27 @@ public void testNotMatchesAndNotSoundsLikeDeclaredInDsl() {
...
                                      RuleModelDRLPersistenceImpl.getInstance().marshal(model));
     }
+
+    @Test
+    public void unmarshalStringListsCorrectly() {
+        final String drl = "package org.mortgages;\n" +
+                "rule \"aaa\"\n" +
+                "\tdialect \"mvel\"\n" +
+                "\twhen\n" +
+                "\t\tApplicant( applicantName : name in ( \"a\", \"b\", \"c\" ) )\n" +
```
manstis, line 9573: @Rikkola Any chance we could have a test for strings containing commas; e.g. `"a", "b", "c,d", "e"`?

jomarko: @manstis it doesn't work, see my https://github.com/kiegroup/drools/pull/2565#issuecomment-542639075

manstis (approved):
Approving on the basis of https://github.com/kiegroup/drools/pull/2565#discussion_r335906561 (a new JIRA has been created for the test I requested be added to this PR).

PR #2566, opened 2019-10-04
DROOLS-4588: Coverage increase

jomarko:
Dependent on #2563

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
@@ -16,6 +16,7 @@
 package org.drools.scenariosimulation.backend.expression;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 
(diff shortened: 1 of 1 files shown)


drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
```diff
@@ -36,6 +39,11 @@ public void evaluateUnaryExpression() {
         assertTrue(evaluator.evaluateUnaryExpression("# 1", 1, Integer.class));
         assertFalse(evaluator.evaluateUnaryExpression("# 2", 1, Integer.class));
 
+        assertTrue(evaluator.evaluateUnaryExpression("# " + ACTUAL_VALUE_IDENTIFIER + " == 123", 123, Integer.class));
```
gitgabrio: @jomarko 
Hi. I think @danielezonca  is creating a specific class to hold the string constants used throughout the code, so probably that `#` should be replaced. Beside that, I think we are also going to have two different "expression" identifier, one for value creation (the one used in all this tests, I think) and one for method invocation. This latter case seems not covered- even if I am not sure on how to test it. @danielezonca ?

gitgabrio (changes requested):
@jomarko 
I read the very last commit and looks almost fine for me - beside a small detail.
What is not really clear to me is that in the PR made by @danielezonca something is already changed

jomarko:
@gitgabrio @dupliaka @yesamer @danielezonca may I ask you for a re-review?

danielezonca (approved):

jomarko:
@danielezonca may I ask you for help? As you see in last commit I updated the test to use proper type for particular expressions, however problem is with array expressions like **# {"Jim"}** where I expect the type **Object[]** however getting an exception from class loader that it can not be loaded. Any advice?

yesamer (approved):
@jomarko fine for me, my only suggestion is to apply the change requested by @gitgabrio during his review.

jomarko:
@yesamer yes, good point, @danielezonca could you please point me to the class with those constants?

yesamer:
> @yesamer yes, good point, @danielezonca could you please point me to the class with those constants?

It should be `ConstantHolder` class

danielezonca:
@jomarko 
I had a look to your failing test and the problem is related to load class of arrays. It is broken/not supported so I have created a ticket for that ([DROOLS-4649](https://issues.jboss.org/browse/DROOLS-4649))

dupliaka (approved):
Fine for me.

jomarko:
jenkins please retest this

gitgabrio (approved):

PR #2568, opened 2019-10-07
Added null value checker in `flushRule()`

bfrggit:
Adding such check should prevent the following issue:

At `KieBuilder::buildAll()`, a decision table file that contains no rule table triggers a `NullPointerException` in the edited method, instead of a parser exception, which makes it hard to find the cause.


ghost:

PR #2570, opened 2019-10-08
[DROOLS-4620] Add test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/4c1d578f44c748f23b504e1c41e771ef1fa20c58

PR #2571, opened 2019-10-09
[DROOLS-4623] using a negated constraint in a not pattern doesn't wor…

mariofusco:
…k with executable model


mariofusco:
Fixed by https://github.com/kiegroup/drools/pull/2674

PR #2580, opened 2019-10-15
[DROOLS-4646] Add support for ? placeholder in unary expressions in DMN scenarios

danielezonca:
This PR introduce `?` as variable to refer actual value in unary expressions (EXPECT) so it will be possible to write expression like `max(1, ?) > 1` or similar.

NOTE: `?` is a special identifier in FEEL and the only one we can use in an unary test

@yesamer @gitgabrio @dupliaka 

https://issues.jboss.org/browse/DROOLS-4646

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
@@ -25,7 +25,7 @@ public class ConstantsHolder {
     public static final String ACTUAL_VALUE_IDENTIFIER = "actualValue";
 
     /* FEEL constants */
-    public static final String UNARY_EXPRESSION_IDENTIFIER = "?";
+    public static final String UNARY_PARAMETER_IDENTIFIER = "?";
 
     private ConstantsHolder() {
         // Not instantiable
--- a/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
@@ -24,6 +24,9 @@ public class ConstantsHolder {
     public static final String MVEL_ESCAPE_SYMBOL = "#";
     public static final String ACTUAL_VALUE_IDENTIFIER = "actualValue";
 
+    /* FEEL constants */
+    public static final String UNARY_EXPRESSION_IDENTIFIER = "?";
+
     private ConstantsHolder() {
         // Not instantiable
     }
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
@@ -51,6 +51,7 @@ public class DMNFeelExpressionEvaluatorTest {
         assertFalse(expressionEvaluator.evaluateUnaryExpression("max(1, ?) < 1", BigDecimal.valueOf(2), BigDecimal.class));
         assertTrue(expressionEvaluator.evaluateUnaryExpression("? = 2", BigDecimal.valueOf(2), BigDecimal.class));
         assertFalse(expressionEvaluator.evaluateUnaryExpression("? > 2", BigDecimal.valueOf(2), BigDecimal.class));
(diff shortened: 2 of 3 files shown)


drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
```diff
@@ -45,6 +45,8 @@ public void evaluateUnaryExpression() {
         assertTrue(expressionEvaluator.evaluateUnaryExpression("not( true )", false, boolean.class));
         assertTrue(expressionEvaluator.evaluateUnaryExpression(">2, >5", BigDecimal.valueOf(6), BigDecimal.class));
         assertTrue(expressionEvaluator.evaluateUnaryExpression("abs(-1)", BigDecimal.valueOf(1), BigDecimal.class));
+        assertTrue(expressionEvaluator.evaluateUnaryExpression("max(1, ?) > 1", BigDecimal.valueOf(2), BigDecimal.class));
+        assertTrue(expressionEvaluator.evaluateUnaryExpression("? = 2", BigDecimal.valueOf(2), BigDecimal.class));
```
gitgabrio, line 52: @danielezonca 
Can not evaluate the code - anyway I think it would be safer to add some "assertFalse" or whatever test - I mean not something that throws an exception, but something that simply does not returns the expected value.

gitgabrio (changes requested):

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
danielezonca: Done :+1:

gitgabrio (approved):

yesamer (approved):

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
```diff
@@ -24,6 +24,9 @@
     public static final String MVEL_ESCAPE_SYMBOL = "#";
     public static final String ACTUAL_VALUE_IDENTIFIER = "actualValue";
 
+    /* FEEL constants */
+    public static final String UNARY_EXPRESSION_IDENTIFIER = "?";
```
dupliaka: Suggestion: we  can name it something like UNARY_PARAMETER_IDENTIFIER. UNARY_EXPRESSION_IDENTIFIER it looks that I should place ? before each unary expression

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
dupliaka: Also I m not sure, but probably it could be better to add a test where we passing a null in ?
like
assertThatThrownBy(() -> expressionEvaluator.evaluateUnaryExpression("? > 2", null, BigDecimal.class)))

dupliaka: Also we can add a test where we can use as ? as we want (at last two) to declare that it is allowed

dupliaka (approved):
Approved with suggestions

dupliaka:

dupliaka:

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluatorTest.java
dupliaka: I meant something like  : assertTrue(expressionEvaluator.evaluateUnaryExpression("? = 1*?", BigDecimal.valueOf(2), BigDecimal.class));

danielezonca: Done, added two new tests to cover these additional cases :)

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ConstantsHolder.java
danielezonca: Good suggestion, done!

PR #2584, opened 2019-10-16
DROOLS-4649: Add a reproducer test

jomarko:
Will be rebased due to util methods introduced in #2566


jomarko:
@danielezonca this is rebased now and should be ready for a review.

danielezonca (approved):

PR #2586, opened 2019-10-16
Remove the copied method

jomarko:
This PR removes a copy of an existing mehtod. The rmeoved method was used two times while the original one more than 90 times.


Rikkola (approved):
Good stuff.

jomarko:
@kiegroup/gatekeepers please merge.

PR #2587, opened 2019-10-16
DROOLS-4650 : GDST: Re-editing BRL columns causes NPE

Rikkola:
https://issues.jboss.org/browse/DROOLS-4650

### DIFF
--- a/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/ConditionCol52.java
+++ b/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/ConditionCol52.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.drools.workbench.models.datamodel.rule.BaseSingleFieldConstraint;
 import org.drools.workbench.models.datamodel.rule.HasParameterizedOperator;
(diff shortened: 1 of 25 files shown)


Rikkola:
I'll try to get green builds before review. I fear this is a risky change.

manstis:
@Rikkola I think this is also a fix for https://github.com/kiegroup/drools-wb/pull/1212?

It might be worth linking the JIRA, if that is true.

Rikkola:
@manstis This fixes that. There are two dublicate product tickets, but only linked this to one. I'll link both.

Rikkola:
All green, nice. I'll fix the code smell, but removing the diamond operator should not change behaviour.

Rikkola:
@jomarko @manstis if you two would be kind enough to share some thoughts about the changes.

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/ConditionCol52.java
```diff
@@ -33,7 +33,7 @@
         HasParameterizedOperator,
         HasBinding {
 
-    private static final long serialVersionUID = 510l;
+    private static final long serialVersionUID = 729l;
```
Rikkola, line 37: No idea if these matter, but thought that I'll update these to current version :)

jomarko: Shouldn't we change it witch each community release then?

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/DTColumnConfig52.java
```diff
@@ -171,10 +171,12 @@ public boolean equals(Object o) {
         if (this == o) {
             return true;
         }
-        if (!(o instanceof DTColumnConfig52)) {
+
+        if(o == null || o.getClass() != this.getClass()){
```
jomarko: This change is not needed, https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.20.2

jomarko (changes requested):
Screenshot from 2019-10-18 07-34-23

Still doesn't work for BRL action columns

jomarko:
@Rikkola as this is about NPE in equals so I (understand my Intelij Idea) reimplemented equals of dtables, I guess also rule model need similar update. https://github.com/Rikkola/drools/pull/6

It solved issue I reported https://github.com/kiegroup/drools/pull/2587#pullrequestreview-303673306

drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model/ConditionCol52.java
Rikkola: Probably only worth it to update when a release has changes.

Rikkola:
@jomarko Is this now ok from you side? Retested and can't figure out how to break it.

jomarko (approved):

Rikkola:
@manstis Hi, I could use your feedback on this.

manstis (approved):
Thanks @Rikkola so my [suspicion](https://github.com/kiegroup/drools-wb/pull/1212#issuecomment-520838543) about equals somewhere was correct.

Rikkola:
@manstis Yep that helped. Although debugging to figure out what exactly is the problem was PITA, but can't imagine how long it would have taken without any hints ;)

Rikkola:
@kiegroup/gatekeepers Can I get this reverted, it breaks unit tests for drools-wb, but the fix will take some time to get green.

Rikkola:
Actually not sure if github has a revert button, I'll make a PR just in case.

mareknovotny:
The Revert button is there, but it generates new PR ;)

Rikkola:
@mareknovotny managed to find it, hehe.

PR #2589, opened 2019-10-17
DROOLS-4448: Simplify if then else statement

jomarko:
During DROOLS-4448 review, complex if then else statements were spotted. This PR reduces the complexity of those if then else statements.


Rikkola (approved):

jomarko:
@kiegroup/gatekeepers please merge

PR #2544, opened 2019-09-17
DROOLS-4539 : Legacy Test Scenario is failing when using java.time.LocalDate type field in expect part.

Rikkola:
https://issues.jboss.org/browse/DROOLS-4539

@jomarko and maybe @danielezonca could you please review.

PR #2593, opened 2019-10-17
DROOLS-4539: Prevent NPE in getting scenario result

jomarko:
During getting test scenario result we need to prevent NPE as we do not use `boolean` but `Boolean`

Dependent on:
- kiegroup/drools#2544

Ensemble together with:
- kiegroup/drools-wb#1243


yesamer (approved):
@jomarko An alternative is to use Boolean.FALSE.equals(verifyField.getSuccessResult().booleanValue())

Rikkola (approved):

jomarko:
@kiegroup/gatekeepers please merge together with kiegroup/drools-wb#1243

PR #2596, opened 2019-10-17
[DROOLS-4658] Add testcase

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/a47901fa5f5519154042969affcf7d0ba402c3e5

PR #2610, opened 2019-10-23
[7.26.x] DROOLS-4539: Prevent NPE in getting scenario result (#2593)

jomarko:
During getting test scenario result we need to prevent NPE as we do not use `boolean` but `Boolean`

Cherry-pick of https://github.com/kiegroup/drools/commit/d1d891393afef5fae4f0506402326b724f3e6e33

Ensemble together with:
https://github.com/kiegroup/drools-wb/pull/1248


Rikkola (approved):

jomarko:
@kiegroup/gatekeepers please merge

PR #2639, opened 2019-11-08
DTAnalysis should skip gaps/overlaps for Collect decision tables

SimonRinguette:
@tarilabs  I believe that tables that output collections should not be checked by the DTAnalysis. Bruce original code did not check them as well.

Certainly collect tables should not be analyzed for gaps since its not a bad practice to have gaps in a collect table, it just yields an empty output and its fine. For example, a table that calculate all the rebates a customer can have access to could return no elements if the customer is not eligible to anything.

### DIFF
--- a/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/dtanalysis/DMNDTAnalyser.java
+++ b/kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/dtanalysis/DMNDTAnalyser.java
@@ -50,6 +50,7 @@ import org.kie.dmn.feel.lang.impl.UnaryTestInterpretedExecutableExpression;
 import org.kie.dmn.feel.runtime.Range.RangeBoundary;
 import org.kie.dmn.model.api.DecisionRule;
 import org.kie.dmn.model.api.DecisionTable;
+import org.kie.dmn.model.api.HitPolicy;
 import org.kie.dmn.model.api.InputClause;
 import org.kie.dmn.model.api.ItemDefinition;
 import org.kie.dmn.model.api.LiteralExpression;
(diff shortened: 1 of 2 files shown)


ghost:

tarilabs:
Thanks @SimonRinguette for this, I have some doubts however: the approach you are suggesting as code change is very drastic and would skip for `R`, `O`, `C*` tables Subsumptions, and xNF checks. Is this really what is desired? /cc @etirelli 

My proposal would be as-follows: if this change is urgent, we can merge this code asap.
It's not urgent, instead of skipping altogether any kind of dt static analysis for all these policies, I would write code and test to just skip gaps and overlaps for the mentioned hit policies that you highlighted.

Please let me know

Jenkins test again

SimonRinguette:
@tarilabs Maybe my fix a little extreme and I should just exclude as you mentioned the gaps/overlaps. let me reconsider this and come back with a "less drastic" proposal.

I was basically going with what Bruce originally had. In his original code, this is what he was doing.

SimonRinguette:
@tarilabs I went back and tested a much less drastic version of DTAnalysis that skip only gaps/overlaps for collect decision tables.

This is more in line with what bruce has in his book as well.

The rest of the checks are extra and seems to fit.

Let me know if that works for you!

SimonRinguette:
I forgot to commit the change to the unit test

tarilabs:
jenkins test again

tarilabs:
```
10:59:00 Caused by: java.io.IOException: Backing channel 'kie-psi-rhel7-large-2033' is disconnected.
10:59:00 	at hudson.remoting.RemoteInvocationHandler.channelOrFail(RemoteInvocationHandler.java:212)
```

Jenkins test again

PR #2645, opened 2019-11-15
DROOLS-4768: raising the timeout from 40000 to 600_000 as it is on master.

hellowdan:
DROOLS-4768: raising the timeout from 40000 to 600_000 as it is on master.


mareknovotny (approved):

mareknovotny:
will wait on CI to be green

PR #2657, opened 2019-11-27
DROOLS-4814 DMN v1.3 marshaller support for group artifact

tarilabs:


tarilabs:
Please note the SonarCloud report X as the hierarchy is deep, but that is by design since we follow the hierarchy from the XSD, not much we can do about that.. I vote to merge, please approve.

baldimir (approved):

PR #2667, opened 2019-12-02
[DROOLS-4829] Rule with NotNode in the middle is not fired after kses…

tkobayas:
…sion.reset()

In the original ConcurrentNodeMemories.resetAllMemories(), only one segment from PathMemory (by getSegmentMemory()) was notified to the PathMemory. So if the PathMemory has another segment with NotNode in the middle of its segments, it will not be notified so will not be linked on the next run. Note that even if a BetaMemory of NotNode was reset earlier than the PathMemory in the loop and notify the segment link to the PathMemory, later the PathMemory resets itself so its mask is overwritten.

I think there are 2 approaches to fix the issue. One is to sort the memory list and reset PathMemory first. Then other memories (including BetaMemory of NotNode) are reset so the PathMemory will be notified. This is my PR fix.

The other idea is to reset all segments in a PathMemory (by getSegmentMemories()) and let them notify. In this case, we don't need to care the order but it could be redundant. The code is: https://gist.github.com/tkobayas/8836018dd844ab672eec30c44c53083d

Feel free to share your thought on this.

### DIFF
--- a/drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
+++ b/drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
@@ -26,6 +26,9 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.assertj.core.api.Assertions;
+import org.drools.compiler.FactA;
+import org.drools.compiler.FactB;
+import org.drools.compiler.FactC;
 import org.drools.core.common.EventSupport;
 import org.drools.core.event.DefaultAgendaEventListener;
 import org.drools.core.event.DefaultRuleRuntimeEventListener;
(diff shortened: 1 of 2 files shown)


drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
```diff
@@ -254,6 +257,68 @@ public void testSegmentMemoriesReset() {
...
+                "when\n" +
+                "  $factA: FactA( field1 == \"code1\")\n" +
+                "  $factC: FactC( f1 == \"code1\")\n" +
+                "then\n" +
+                "end";
+
+        KieContainer kcontainer = new KieHelper().addContent( drl, ResourceType.DRL ).getKieContainer();
+        KieSessionsPool pool = kcontainer.newKieSessionsPool( 1 );
+
+        KieSession ksession = pool.newKieSession();
```
baldimir, line 290: This is a bit nitpicking, however please dispose the ksession in a try-finally block, so we are sure it gets disposed if anything bad happens.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
```diff
@@ -254,6 +257,68 @@ public void testSegmentMemoriesReset() {
...
+
+        createFactAndInsert(ksession);
```
baldimir: This is a very minor thing - please dispose the session in a try-finally block so we are 100% sure it gets disposed.

baldimir (commented):
Thanks for the fix! Just one comment from me. As of the solution(s) - both add some performance impact, so I will let Mario consider this as he knows much more about the internals.

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
tkobayas: Thanks! I fixed and pushed.

baldimir (approved):

drools-compiler/src/test/java/org/drools/compiler/integrationtests/session/SessionsPoolTest.java
tkobayas: Thanks! I fixed and pushed.

mariofusco:
@tkobayas I reviewed a bit what you did. Great job in finding this. My only concern is that sorting all memories could be quite expensive for large rule bases. At same time it's wasteful to reset multiple times the same segment from different paths. For these reason I developed the following variation of your 2nd solution with the only difference that it avoids to reset the same segment twice from different paths. 

Let me know if this makes sense for you and in case update your pull request with this solution. Conversely, if you see any problem or drawback in my proposal please let me know.

```
    public void resetAllMemories(StatefulKnowledgeSession session) {
        InternalKnowledgeBase kBase = (InternalKnowledgeBase) session.getKieBase();
        Set<Integer> resetSmems = new HashSet<>();

        for (int i = 0; i < memories.length(); i++) {
            Memory memory = memories.get(i);
            if (memory != null) {
                memory.reset();
                if (memory instanceof PathMemory ) {
                    SegmentMemory[] smems = ((PathMemory) memory).getSegmentMemories();
                    if (smems != null) {
                        for (SegmentMemory smem : smems) {
                            resetSegmentMemory(session, kBase, smem, resetSmems);
                        }
                    }
                } else {
                    resetSegmentMemory(session, kBase, memory.getSegmentMemory(), null);
                }
            }
        }
    }

    private void resetSegmentMemory(StatefulKnowledgeSession session, InternalKnowledgeBase kBase, SegmentMemory smem, Set<Integer> resetSmems) {
        if (smem != null && ( resetSmems == null || resetSmems.add( smem.getTipNode().getId() ) )) {
            smem.reset(kBase.getSegmentPrototype(smem));
            if (smem.isSegmentLinked()) {
                smem.notifyRuleLinkSegment((InternalWorkingMemory) session);
            }
        }
    }

```

tkobayas:
Thank you for reviewing, @mariofusco 

The "resetSmems' approach has a scenario where:

1. PathMemory R1 resets itself and its segmentMemories (S1, S2, S3). Put them into resetSmems
2. PathMemory R2 resets itself and its segmentMemories (S1, S2, S4) but it can reset only S4. So R2 is not notified by S1 nor S2.

I added a test case testSegmentMemoriesResetWithNotNodeInTheMiddle2() which reproduces the issue.

Now I updated the PR with a new approach. This collects segmentMemories from all nodes first and then reset those segmentMemories later.

WDYT?

mariofusco:
@tkobayas Great job, thank you.

PR #2668, opened 2019-12-02
[KOGITO-608] remove dependency from TypeResolver in kie-soup

mariofusco:


manstis:
Has this broken the build?

https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/daily-build-prod/job/daily-build-prod-pipeline-master/27/

I didn't see a fu1l d0wnst3am bui1d for this PR...

manstis:
```
WELD-001408: Unsatisfied dependencies for type MVELEvaluator with qualifiers @Default
```

mariofusco:
@manstis this commit https://github.com/kiegroup/drools/commit/6e6c4246e3039ae8c032888353b71be6460dcc23 should have fixed the problem. Please give it a try.

manstis:
@mariofusco Thank-you.. @Rikkola reports our problems are resolved.

gitgabrio:
@mariofusco 
Hi Mario, since I've been hit by chance by this PR... I understand putting those two classes inside drools-core to remove unwanted dependencies; I do not understand why not simply _move_ those classes (and then remove the dependencies) instead of _copy_ them. I think drools-core is one of the few modules required everywhere, so moving classes there would keep them available and - at the same time - would bring up all the places where they are used, giving the chance to fix the changed import... IMHO... mtc.....

PR #2672, opened 2019-12-04
[Drools-4761] Moved modules from openshift-drools-hacep into drools-ha [Draft status]

desmax74:
Moved hacep and kie-remote modules from openshift-drools-hacep under drools-ha with names ha-core and ha-remote


drools-ha/ha-core/src/main/java/org/kie/hacep/Config.java
```diff
@@ -0,0 +1,153 @@
...
+        //.append(",").append("my-cluster-kafka-brokers.my-kafka-project.svc").append(":9093");//tls
+        return sb.toString();
+    }
+
+    public static Properties getDefaultConfig() {
+        return getDefaultConfigFromProps(CONF);
+    }
+
+    public static Properties getConsumerConfig(String caller) {
+        if(consumerConf == null){
```
baldimir: Nitpicking: Please check your formatter config, I see a lot of places where there are no spaces around "if" clauses. We have formatter configs for Eclipse and IntelliJ IDEA available here https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration.

drools-ha/ha-core/src/main/java/org/kie/hacep/Config.java
```diff
@@ -0,0 +1,153 @@
...
+        if(snapshotProducerConf == null){
+            snapshotProducerConf = getDefaultConfigFromProps(SNAPSHOT_PRODUCER_CONF);
+        }
+        logConfig("SnapshotProducer",snapshotProducerConf);
+        return snapshotProducerConf;
+    }
+
+    public static Properties getDefaultConfigFromProps(String fileName) {
+            Properties config = new Properties();
+            InputStream in = null;
```
baldimir: You could use try with resources here like this: 
```
try (InputStream in = Config.class.getClassLoader().getResourceAsStream(fileName)) {
    config.load(in);
}
```

drools-ha/ha-core/src/main/java/org/kie/hacep/consumer/CommandHandler.java
```diff
@@ -0,0 +1,303 @@
...
+
+    private List getSerializableItemsByClassType(ListObjectsCommandClassType command) {
+        Collection<? extends Object> objects = ObjectFilterHelper.getObjectsFilterByClassType(command.getClazzType(),
+                                                                                              kieSessionContext.getKieSession());
+        return getListFromSerializableCollection(objects);
+    }
+
+    private List getListFromSerializableCollection(Collection<?> objects) {
+        List serializableItems = new ArrayList<>(objects.size());
+        Iterator<? extends Object> iterator = objects.iterator();
```
baldimir: You can use 

```
serializableItems.addAll(objects);
```
instead of iterating the collection.

drools-ha/ha-core/src/main/java/org/kie/hacep/consumer/CommandHandler.java
```diff
@@ -0,0 +1,303 @@
...
+        KieServices ks = KieServices.get();
+        UpdateKjarMessage msg = new UpdateKjarMessage(command.getId(), Boolean.FALSE);
+        if (ks != null) {
+            ReleaseId releaseId = ks.newReleaseId(command.getGroupID(), command.getArtifactID(), command.getVersion());
+            if(envConfig.isUpdatableKJar()) {
+                try {
+                    kieSessionContext.getKieContainer().updateToVersion(releaseId);
+                    msg = new UpdateKjarMessage(command.getId(), Boolean.TRUE);
+                } catch (java.lang.UnsupportedOperationException ex) {
+                    logger.info("It isn't possible update a classpath container to a new version");
```
baldimir: Just checking: I saw on more places that an exception is swallowed and just logged. Is this correct in this scenarios? Because usually that is a bad idea, because it might lead to application state inconsistency. I.e. the application would just swallow an exception which could lead to some corrupt state in later execution.

drools-ha/ha-core/src/main/java/org/kie/hacep/consumer/CommandHandler.java
```diff
@@ -0,0 +1,303 @@
...
+                }
+            }else{
```
baldimir: Formatting again - please check your formatter settings.

drools-ha/ha-core/src/main/java/org/kie/hacep/consumer/DroolsConsumerHandler.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Override
+    public void process( ItemToProcess item, State state) {
+        RemoteCommand command  = deserialize((byte[])item.getObject());
+        process( command, state );
+    }
+
+    @Override
+    public void process( RemoteCommand command, State state ) {
+        if(envConfig.isUnderTest()) {  loggerForTest.warn("DroolsConsumerHandler.process Remote command on process:{} state:{}", command, state); }
```
baldimir: Formatting again - please check your formatter settings.

drools-ha/ha-core/src/main/java/org/kie/hacep/core/Bootstrap.java
```diff
@@ -0,0 +1,118 @@
...
+        return consumerController;
+    }
+
+    private static void leaderElection() {
+        LeaderElection leadership = coreKube.getLeaderElection();
+        coreKube.getLeaderElection().addCallbacks(Arrays.asList( consumerController.getCallback()));
+        try {
+            leadership.start();
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
```
baldimir: Again I see that the exception is just logged. In this case it unfortunately catches all possible exceptions. It is suspicious in this case because I expect this method selects the leader. And if it fails the whole thing should be broken, correct? However the exception is just logged so the application continues to run.

drools-ha/ha-core/src/main/java/org/kie/hacep/core/infra/OffsetManager.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public static Map<TopicPartition, Long> getEventsTopicOffset() {
+        return eventsTopicOffset;
+    }
+
+    public static Properties load() {
+        Properties prop = null;
+        InputStream input = null;
+        try {
```
baldimir: You can use try with resources here so the input stream is closed automatically.

drools-ha/ha-core/src/main/java/org/kie/hacep/core/infra/OffsetManager.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        return prop;
+    }
+
+    public static void store(Map<TopicPartition, OffsetAndMetadata> offsetAndMetadataMap) {
+        Properties prop = new Properties();
+        OutputStream output = null;
+
+        try {
```
baldimir: You can use try with resources here, so the output stream is closed automatically.

drools-ha/ha-core/src/main/java/org/kie/hacep/core/infra/consumer/LocalConsumer.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void initConsumer( ConsumerHandler consumerHandler ) {
+        this.consumerHandler = consumerHandler;
+    }
+
+    @Override
+    public void poll() {
+        String topic = envConfig.getEventsTopicName();
+        while (true) {
```
baldimir: This while (true) could be changed to something like this to avoid the danger of while (true) hanging in the future after changes. 

```
RemoteCommand command = ( RemoteCommand ) queue.poll(topic, envConfig.getPollTimeout());
while (command != null) {
    consumerHandler.process( command, currentState );
    command = ( RemoteCommand ) queue.poll(topic, envConfig.getPollTimeout());
}
```

drools-ha/ha-remote/src/main/java/org/kie/remote/impl/ClientUtils.java
```diff
@@ -0,0 +1,46 @@
...
+public class ClientUtils {
+
+    public static final String CONSUMER_CONF = "consumer.properties";
+    public static final String PRODUCER_CONF = "producer.properties";
+    public static final String CONF = "configuration.properties";
+
+    public static Properties getConfiguration(String filename) {
+        Properties props = new Properties();
+        InputStream in = null;
+        try {
```
baldimir: You can  use try with resources here so the Input stream is closed automatically.

drools-ha/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+            <artifactId>aopalliance</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+
+      <!-- kafka -->
+      <dependency>
+        <groupId>org.apache.kafka</groupId>
+        <artifactId>kafka-clients</artifactId>
+        <version>${version.org.apache.kafka}</version>
```
baldimir, line 86: I think all these versions should be in kie-parent where we store all artifact versions. Am I right @mareknovotny?

baldimir (changes requested):
Please see my comments.

baldimir:
I see that the code has just 14 % coverage. I know there are integration tests in other repository, but still, this is pretty low number. Would you mind adding unit test coverage to the classes? Also SonarCloud reports 154 new issues with this code of which there are 27 marked as Critical. Please go through them and clean everything possible. I saw you already tried to fix some of them, but I am afraid that this many problems is a blocker for me from QE perspective in Drools (connected with nearly no unit coverage).

desmax74:
@baldimir  is it possible to enable SonarCloud on https://github.com/kiegroup/openshift-drools-hacep/ to discover these problems before move instead to move and then try to fix here during a draft-status of the PR ?

baldimir:
Sure, it should be possible. I will take a look on Friday.

drools-ha/ha-core/src/main/java/org/kie/hacep/Config.java
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                in = Config.class.getClassLoader().getResourceAsStream(fileName);
+            } catch (Exception e) {
+                logger.info(e.getMessage());
```
rsynek: Most often error messages are logged on ERROR level. (just accidentally glanced over this piece of code)

baldimir: I personally dislike more that they are just swallowed. This approach could be very error prone.

drools-ha/ha-core/src/main/java/org/kie/hacep/consumer/CommandHandler.java
desmax74: In this specific case yes because a client is trying to update a Container of the wrong type and the returned value to the client is just a boolean false

baldimir:
SonarCloud integration prepared for openshift-drools-hacep in following PRs: https://github.com/kiegroup/kie-jenkins-scripts/pull/647 and https://github.com/kiegroup/openshift-drools-hacep/pull/66

PR #2677, opened 2019-12-06
[DROOLS-3280] DMNRuntime API for evaluateByID/Name to error if empty …

tkobayas:
…array

Please review. I have two points:

1) I implemented to throw an IllegalArgumentException but if you prefer another Exception or just creating an error message with MsgUtil.reportMessage(), please let me know.

2) I added 2 unit test methods. Is it verbose? But may be better than no test...

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
@@ -188,6 +188,9 @@ public class DMNRuntimeImpl
         Objects.requireNonNull(model, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "model"));
         Objects.requireNonNull(context, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "context"));
         Objects.requireNonNull(decisionNames, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "decisionNames"));
+        if (decisionNames.length == 0) {
+            throw new IllegalArgumentException(MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_EMPTY, "decisionNames"));
+        }
         final DMNResultImpl result = createResult( model, context );
         for (String name : decisionNames) {
             evaluateByNameInternal( model, context, result, name );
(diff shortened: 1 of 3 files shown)


hellowdan (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -188,6 +188,9 @@ public DMNResult evaluateByName( DMNModel model, DMNContext context, String... d
         Objects.requireNonNull(model, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "model"));
         Objects.requireNonNull(context, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "context"));
         Objects.requireNonNull(decisionNames, () -> MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_NULL, "decisionNames"));
+        if (decisionNames.length == 0) {
+            throw new IllegalArgumentException(MsgUtil.createMessage(Msg.PARAM_CANNOT_BE_EMPTY, "decisionNames"));
```
tarilabs, line 192: 👍 

> I implemented to throw an IllegalArgumentException

it's good, as it is consistent with the RuntimeException thrown by the `Objects.requireNonNull` if the parameters are null.

tarilabs (approved):
thank you!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -2720,5 +2721,43 @@ public void testBindingContextTypeCheck() {
...
+
+        final DMNContext context = DMNFactory.newContext();
+        context.set( "Monthly Salary", 1000 );
+
+        try {
+            String[] decisionNames = new String[]{};
+            runtime.evaluateByName(dmnModel, context, decisionNames);
+            fail("Should throw an IllegalArgumentException");
+        } catch (IllegalArgumentException e) {
+            assertTrue(true);
```
baldimir: You can annotate the test method with @Test(expected = IllegalArgumentException.class). That way you can remove the try-catch block and the fail(...) call.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -2720,5 +2721,43 @@ public void testBindingContextTypeCheck() {
...
+
+        final DMNContext context = DMNFactory.newContext();
+        context.set( "Monthly Salary", 1000 );
+
+        try {
+            String[] decisionIds = new String[]{};
+            runtime.evaluateById(dmnModel, context, decisionIds);
+            fail("Should throw an IllegalArgumentException");
+        } catch (IllegalArgumentException e) {
+            assertTrue(true);
```
baldimir: Same as my previous comment.

baldimir (changes requested):

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
tkobayas: Fixed. Thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
tkobayas: Fixed. Thanks!

baldimir (approved):

tarilabs (approved):

PR #2678, opened 2019-12-06
DROOLS-4850: ModelCompiler, rm RuleUnitDescription.getRuleUnitClass()

evacchi:
follow up to DROOLS-4837: remove usages of RuleUnitDescription.getRuleUnitClass() throughout model compiler so that non-reflective implementations are allowed.

For the time being we will keep the method for backwards compatibility with PMML impl

requires https://github.com/kiegroup/droolsjbpm-knowledge/pull/415


baldimir:
SonarCloud reports some unused code. Could you please take a look? https://sonarcloud.io/project/issues?id=org.drools%3Adrools&pullRequest=2678&resolved=false&types=CODE_SMELL

PR #2686, opened 2019-12-09
DROOLS-4340 - Contains can not be used for String with the Expression editor

Rikkola:
https://issues.redhat.com/browse/DROOLS-4340

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -2744,8 +2744,9 @@ public class RuleModelDRLPersistenceImpl
                 }
             }
             int opPos = expr.indexOf(opString);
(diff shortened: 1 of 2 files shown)


Rikkola:
jenkins please retest this

Rikkola:
@manstis @jomarko I could use a review here please.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -2773,6 +2774,11 @@ private static String findOperator(String expr) {
         return null;
     }
 
+    private static boolean isNotMethodName(final String expr,
```
jomarko: I think more appropriate names are `expression` instead of `expr` and `operatorPosition` instead of `opPos`.

jomarko (approved):
One minor comment. Approving regardless your decision to incorporate/not incorporate my proposal. Thank you @Rikkola

manstis (approved):
LGTM, thanks @Rikkola

Rikkola:

PR #2688, opened 2019-12-11
NullPointer error for Memory loading class in customized classloader

ccsuzuka:
Scenario:  1. Using KieFileSystem loading DRL file and dependency resource.
                  2. Dynamic load byte class into customized classloader, and transmit the classloader into KIESession.
                  3. in update working area case.
It will raise NullPointer Exception.


mariofusco:
Please also submit a reproducer for this issue.

mariofusco:
I'm sorry but without a reproducer of the problem that you're attempting to solve I cannot merge this pull request. Moreover, provided that there is a problem, I have the feeling that this pull request is fixing its symptom instead of the root cause.

PR #2689, opened 2019-12-13
DROOLS-4871 : Complex value list doesn't work in guided rule editor

Rikkola:
https://issues.redhat.com/browse/DROOLS-4871

Part one of the fix. This fixes from file to editor change.

https://github.com/kiegroup/drools/pull/2689
https://github.com/kiegroup/drools-wb/pull/1273


Rikkola:
@jomarko @manstis Here the fix has two parts. One from file to editor model and another for situations when the same data is moved from model to file. If you could give me some feedback and review I would appreciate it.

manstis (commented):
@jomarko (as author of the original test), @Rikkola I was expecting the constraint value to be `"a", "b, something", "c()"` i.e. preserving the quotes around `"a"` and `"c()"` however I see the test expects these quotes to be removed. Is this correct?

manstis (approved):
OK, I see there's some other black magic going on with the conversion back to DRL.

Works well :-)

jomarko (approved):
Approving 1/2

PR #2690, opened 2019-12-16
[DROOLS-4618] Test Scenario: refactor ExpressionEvaluator implementations

danielezonca:
Changed `ExpressionEvaluator` input value to `String` (it should be possible to refactor `factMappingValue.getRawValue()` to `String` too but impact is big so will be done in the future)

@gitgabrio @yesamer @dupliaka 

https://issues.redhat.com/browse/DROOLS-4618

PR #2691, opened 2019-12-16
[DROOLS-4698] Managing user defined expressions for Collections

yesamer:
@dupliaka @jomarko  @danielezonca can you please review it and test it?

Collection editor currently use a JSON to send collections (Maps, Lists) data to the backend. 
Lists are wrapped on JSONArray and Maps on JSONObject
The idea I had to manage used defined expressions are to wrap them to a simple JSONTextual node (i.e. a JSON which contains only a string). This is useful to easily separate the scope of data assigning it to a very specific data structure.
Then, the changes here consist basically to handle this new case ( JSONTextual ) to the already present cases of JSONArray and JSONObject, for all three Evaluators (MVEL, Base and FEEL).
JSONTextual format is: `"expression"`, it wraps a String between quote `"` char.

https://issues.redhat.com/browse/DROOLS-4698

Related PR: https://github.com/kiegroup/drools-wb/pull/1276


jomarko:
I think review should be postponed until https://github.com/kiegroup/drools/pull/2690 is in, WDYT @yesamer ?

yesamer:
> I think review should be postponed until #2690 is in, WDYT @yesamer ?

@jomarko Yes, fine for me.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -56,13 +60,41 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
...
 
+    /**
+     * A rawValue is an MVEL expression if:
+     * - NOT COLLECTIONS CASE: It's a <code>String</code> which starts with MVEL_ESCAPE_SYMBOL ('#')
+     * - COLLECTION CASE: It's a JSON String node, which is used only when an expression is set
+     *   (in other cases it's a JSON Object (Map) or a JSON Array (List)) and it's value starts with MVEL_ESCAPE_SYMBOL ('#')
+     * @param rawValue
+     * @return
+     */
+    protected boolean isAnMVELExpression(Object rawValue) {
```
danielezonca: Please change to `String`: we just changed all ExpressionEvaluator to manage only `String`s so we can assume a string also here (with a cast on caller side)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -56,13 +60,41 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
...
+            return false;
+        }
+        /* NOT COLLECTIONS CASE: It's a <code>String</code> which starts with MVEL_ESCAPE_SYMBOL ('#') */
+        if (((String) rawValue).trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+            return true;
+        }
+        /* COLLECTION CASE: It's a JSON String node, which is used only when an expression is set
+           and it's value starts with MVEL_ESCAPE_SYMBOL ('#') */
+        Optional<JsonNode> optionalNode = JsonUtils.convertFromStringToJSONNode((String) rawValue);
+        if (optionalNode.isPresent()) {
```
danielezonca: You can use `optionalNode.ifPresent(jsonNode -> ...`

jomarko:
jenkins please retest this

jomarko:

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
```diff
@@ -141,22 +143,29 @@ protected void setField(Object toReturn, String fieldName, Object fieldValue) {
 
     /**
      * In DMN only Lists are structured result while Maps are context so "plain" FEEL expressions
+     * Maps are considered structured only when the given raw value is a JSON textual type
+     * i.e. an user defined expression (The JSON textual must be processed to take its value)
```
danielezonca: DMN doesn't support Map in the collection editor: context (Map equivalent for DMN) is managed as normal string

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
```diff
@@ -0,0 +1,71 @@
...
+     * It determines if a given json in String format, is a Json Textual node
+     * @param json
+     * @return
+     */
+    public static boolean isAJSONTextualNode(String json) {
+        if (json != null && !json.isEmpty()) {
+            try {
+                ObjectMapper objectMapper = new ObjectMapper();
+                JsonNode jsonNode = objectMapper.readTree(json);
+                return jsonNode.isTextual();
```
danielezonca: Use `convertFromStringToJSONNode` to simplify the implementation

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -76,17 +80,22 @@ protected Object convertResult(String rawString, String className, List<String>
...
+                return extractAndEvaluateExpression((TextNode) jsonNode, className);
+            /* JSON Array: User defined List */
+            } else if (jsonNode.isArray()) {
                 return createAndFillList((ArrayNode) jsonNode, new ArrayList<>(), className, genericClasses);
+            /* JSON Object: User defined Map */
             } else if (jsonNode.isObject()) {
                 return createAndFillObject((ObjectNode) jsonNode,
                                            createObject(className, genericClasses),
                                            className,
                                            genericClasses);
```
jomarko: Could we please put comments in the way bellow, it is minor thing, but that ways kiegroup code-formatting will not change the code
```java
if (jsonNode.isTextual()) {
                /* JSON String: User defined Expression */
                return extractAndEvaluateExpression((TextNode) jsonNode, className);
            } else if (jsonNode.isArray()) {
                /* JSON Array: User defined List */
                return createAndFillList((ArrayNode) jsonNode, new ArrayList<>(), className, genericClasses);
            } else if (jsonNode.isObject()) {
                /* JSON Object: User defined Map */
                return createAndFillObject((ObjectNode) jsonNode,
                                           createObject(className, genericClasses),
                                           className,
                                           genericClasses);
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -151,17 +160,49 @@ protected boolean verifyResult(String rawExpression, Object resultRaw) {
...
+            /* JSON String: User defined Expression */
+            if (jsonNode.isTextual()) {
+                return verifyExpression((TextNode) jsonNode, resultRaw, resultClass);
+            /* JSON Array: User defined List */
+            } else if (jsonNode.isArray()) {
                 return verifyList((ArrayNode) jsonNode, (List) resultRaw);
+            /* JSON Object: User defined Map */
             } else if (jsonNode.isObject()) {
                 return verifyObject((ObjectNode) jsonNode, resultRaw);
             }
```
jomarko: I have concern about repeating the conditions. There is a risk of adding one more `else if` into `convertResult` and forgetting updating `verifyResult`, what do you think?

Also putting comments inside {} blocks would satisfy kiegroup code formatting.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
```diff
@@ -141,22 +143,29 @@ protected void setField(Object toReturn, String fieldName, Object fieldValue) {
...
+     * @param raw
      * @return
      */
     @Override
-    protected boolean isStructuredInput(String className) {
-        return ScenarioSimulationSharedUtils.isList(className);
+    protected boolean isStructuredInput(String className, String raw) {
+        return ScenarioSimulationSharedUtils.isList(className) ||
+                (ScenarioSimulationSharedUtils.isMap(className) && JsonUtils.isAJSONTextualNode(raw));
     }
```
jomarko: The implementation seems to be same, I had idea about using private method with single implementation, let me kwno what do you think.
```java
    /**
     * In DMN only Lists are structured result while Maps are context so "plain" FEEL expressions
     * Maps are considered structured only when the given raw value is a JSON textual type
     * i.e. an user defined expression (The JSON textual must be processed to take its value)
     * @param resultClass
     * @return
     */
    @Override
    protected boolean isStructuredResult(Class<?> resultClass, String raw) {
        return resultClass != null && isStructured(resultClass.getCanonicalName(), raw);
    }

    /**
     * In DMN only Lists are structured input while Maps are context so "plain" FEEL expressions.
     * Maps are considered structured only when the given raw value is a JSON textual type
     * i.e. an user defined expression (The JSON textual must be processed to take its value)
     * @param className
     * @param raw
     * @return
     */
    @Override
    protected boolean isStructuredInput(String className, String raw) {
        return isStructured(className, raw);
    }

    private boolean isStructured(String className, String raw) {
        return className != null && (ScenarioSimulationSharedUtils.isList(className) ||
                (ScenarioSimulationSharedUtils.isMap(className) && JsonUtils.isAJSONTextualNode(raw)));
    }
```

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -56,13 +60,41 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
...
+    protected boolean isAnMVELExpression(Object rawValue) {
+        if (!(rawValue instanceof String)) {
+            return false;
+        }
+        /* NOT COLLECTIONS CASE: It's a <code>String</code> which starts with MVEL_ESCAPE_SYMBOL ('#') */
+        if (((String) rawValue).trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+            return true;
+        }
+        /* COLLECTION CASE: It's a JSON String node, which is used only when an expression is set
+           and it's value starts with MVEL_ESCAPE_SYMBOL ('#') */
```
jomarko: This seems to be different from what we say in method javadoc, could you please double check?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -85,10 +87,28 @@ protected Object compileAndExecute(String rawExpression, Map<String, Object> par
...
-        return rawExpression.replaceFirst(MVEL_ESCAPE_SYMBOL, "");
+        Optional<JsonNode> optionalJSONNode = JsonUtils.convertFromStringToJSONNode(rawExpression);
+        if (optionalJSONNode.isPresent()) {
+            JsonNode jsonNode = optionalJSONNode.get();
+            if (jsonNode.isTextual() && jsonNode.asText() != null && jsonNode.asText().trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+                String expression = jsonNode.asText();
+                return expression.replaceFirst(MVEL_ESCAPE_SYMBOL, "");
+            }
+        }
+        throw new IllegalArgumentException("Malformed MVEL expression '" + rawExpression + "'");
```
jomarko: Do we want to put message into `ContantHolder` class?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -200,12 +201,36 @@ public void listOfSimpleTypeTest() {
...
+    @Test(expected = UnsupportedOperationException.class)
+    public void expressionMapConvertResultTest() {
+        String expressionCollectionJsonString = new TextNode("JsonText").toString();
+        expressionEvaluator.convertResult(expressionCollectionJsonString, Map.class.getCanonicalName(), Collections.EMPTY_LIST);
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void expressionListVerifyResultTest() {
+        String expressionCollectionJsonString = new TextNode("JsonText").toString();
+        expressionEvaluator.verifyResult(expressionCollectionJsonString, new ArrayList<>(), null);
```
jomarko: Can we please use `Collections.empyList()`

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -200,12 +201,36 @@ public void listOfSimpleTypeTest() {
...
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void expressionMapVerifyResultTest() {
+        String expressionCollectionJsonString = new TextNode("JsonText").toString();
+        expressionEvaluator.verifyResult(expressionCollectionJsonString, new HashMap<>(), null);
```
jomarko: Can we please use `Collections.emptyMap()`?

jomarko (commented):
Some comments to code

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
yesamer: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
yesamer: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: @jomarko I simplified it

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/DMNFeelExpressionEvaluator.java
yesamer: @jomarko I reverted all changes on that class.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
yesamer: Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
yesamer: @jomarko Comments fixed. About your concern, it could be a risk but I think it's a Developer responsibility to handle it in the correct way. In addition, how we should impose this constraint? Probably we should completely change the architecture of these classes.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
yesamer: @jomarko I removed this test because this case is no more handled

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
yesamer: @jomarko I removed this test because this case is no more handled

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
yesamer: @jomarko I removed this test because this case is no more handled

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -76,17 +77,22 @@ protected Object convertResult(String rawString, String className, List<String>
...
             JsonNode jsonNode = objectMapper.readTree(rawString);
-            if (jsonNode.isArray()) {
+            if (jsonNode.isTextual()) {
+                /* JSON String: User defined Expression */
+                return internalLiteralEvaluation(jsonNode.asText(), className);
+            } else if (jsonNode.isArray()) {
+                /* JSON Array: User defined List */
                 return createAndFillList((ArrayNode) jsonNode, new ArrayList<>(), className, genericClasses);
             } else if (jsonNode.isObject()) {
+                /* JSON Object: User defined Map */
```
danielezonca: @yesamer 
Can you please update the comments with something like?
Text: expression manually written by the user to build a list/map
Array: list of expressions created using List collection editor
Map: map of expressions created using Map collection editor

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -151,14 +157,19 @@ protected boolean verifyResult(String rawExpression, Object resultRaw) {
...
             JsonNode jsonNode = objectMapper.readTree(rawExpression);
-            if (jsonNode.isArray()) {
+            if (jsonNode.isTextual()) {
+                /* JSON String: User defined Expression */
+                return internalUnaryEvaluation(jsonNode.asText(), resultRaw, resultClass, false);
+            } else if (jsonNode.isArray()) {
+                /* JSON Array: User defined List */
                 return verifyList((ArrayNode) jsonNode, (List) resultRaw);
             } else if (jsonNode.isObject()) {
+                /* JSON Object: User defined Map */
```
danielezonca: Same as above

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
```diff
@@ -61,7 +62,7 @@ protected boolean isStructuredResult(Class<?> resultClass) {
 
     /**
      * Check if className represents a structured input
-     * @param className
+     * @param className Used to determine if a structured input is passed
      * @return
      */
     protected boolean isStructuredInput(String className) {
```
gitgabrio: @yesamer 
This method is functionally identical to the previous one, even if written differently: please remove one of them

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -59,7 +62,7 @@ public boolean evaluateUnaryExpression(String rawExpression, Object resultValue,
 
     @Override
     public Object evaluateLiteralExpression(String rawExpression, String className, List<String> genericClasses) {
-        Object expressionResult = compileAndExecute(rawExpression, Collections.emptyMap());
+        Object expressionResult = compileAndExecute(rawExpression, new HashMap<>());
```
danielezonca: This changed has been moved to [this PR](https://github.com/kiegroup/drools/pull/2713) so we can remove it from here (with the tests of course)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -54,15 +58,34 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
             return getOrCreateDMNExpressionEvaluator();
         }
 
-        Object rawValue = factMappingValue.getRawValue();
+        String rawValue = (String) factMappingValue.getRawValue();
```
gitgabrio, line 61: @yesamer 
Not related to this PR, but we have to get rid of this `Object` _rawValue_ (is cast to String everywhere)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
```diff
@@ -0,0 +1,61 @@
...
+    public static Optional<JsonNode> convertFromStringToJSONNode(String json) {
+        if (json == null || json.isEmpty()) {
+           return Optional.empty();
+        }
+        try {
+            ObjectMapper objectMapper = new ObjectMapper();
+            JsonNode jsonNode = objectMapper.readTree(json);
+            return Optional.of(jsonNode);
+        } catch (IOException e) {
+            return Optional.empty();
```
gitgabrio, line 51: @yesamer 
Not sure about swallowing this exception that way: if we get here rawValue is expected to be a valid json, but if it throws exception, something bad has happened, IMO. 
"Given a json in String format, it try to convert it in a <code>JsonNode</code>. In case of success, i.e.
    the given string is a valid json"...: what would mean a "json in String format" but invalid json?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    /**
+     * It determines if a given json in String format, is a Json Textual node
+     * @param json
+     * @return
+     */
+    public static boolean isAJSONTextualNode(String json) {
```
gitgabrio: @yesamer 
I'm not sure where it is used, but I'm almost sure you would need to re-convert the string again and retrieve the `.isTextual` object: would not be better to directly return an Optional"<TextNode or whatever>" here, as in convertFromStringToJSONNode?

gitgabrio (changes requested):
@yesamer 
Just some minor changes, pls

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
yesamer: @gitgabrio To be honest it's not used, I removed it.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
```diff
@@ -85,10 +88,28 @@ protected Object compileAndExecute(String rawExpression, Map<String, Object> par
...
+        if (rawExpression != null && rawExpression.trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+            return rawExpression.replaceFirst(MVEL_ESCAPE_SYMBOL, "");
         }
-        return rawExpression.replaceFirst(MVEL_ESCAPE_SYMBOL, "");
+        Optional<JsonNode> optionalJSONNode = JsonUtils.convertFromStringToJSONNode(rawExpression);
+        if (optionalJSONNode.isPresent()) {
+            JsonNode jsonNode = optionalJSONNode.get();
+            if (jsonNode.isTextual() && jsonNode.asText() != null && jsonNode.asText().trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+                String expression = jsonNode.asText();
+                return expression.replaceFirst(MVEL_ESCAPE_SYMBOL, "");
```
danielezonca, line 110: I think we can use `rawExpression` instead of convert back `jsonNode` to text. Wdyt?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -162,9 +163,9 @@ public void mapOfComplexTypeTest() {
         element.setPhones(phones);
         toCheck.put("first", element);
 
-        assertTrue(expressionEvaluator.verifyResult(mapJsonString, toCheck));
+        assertTrue(expressionEvaluator.verifyResult(mapJsonString, toCheck, null));
```
danielezonca, line 167: Can you please add a test where the last parameter of `verifyResult` is used and it has to be different from `null`?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
yesamer: @gitgabrio This is a good point, I had the thought, but I didn't find a different way to validate a json. 
According to Jackson docs:

    

>  * @throws JsonParseException if underlying input contains invalid content
>      *    of type {@link JsonParser} supports (JSON for default case)
>      */
>     public JsonNode readTree(String content) throws IOException {
>         return _readTreeAndClose(_jsonFactory.createParser(content));
>     }


Then, I can improve it managing `JsonParseException` and `IOException` in different ways: in the first case, it's a malformed content i.e. not a JSON, in the second case It could be any other IOException, I can re-throw it with an other exception (which one?)
Consider the `JsonParseException` is a sub class of `IOException`
WDYT?

gitgabrio: @yesamer 
If I've understood correctly, `ObjectMapper.readTree(String content)` may throw one or the other (it declares only the latter, being the ancestor).
So, I agree both of them should be catch and managed in different way.
In either case, a specific exception should be re-thrown: I would create twos, to help calling code differentiate the actual issue.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
gitgabrio: @yesamer 
great!

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
yesamer: @gitgabrio done! Thanks!

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: @gitgabrio @danielezonca already planned a future refactor to set rawValue to `String` everywhere.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
yesamer: @gitgabrio no, a `JsonParseException` is thrown if the input is not a valid json. In this case, I return an empty `Optional` which means: the given string can't be parsed as json and then it's not a json. In all other cases, when a generic `IOException` is thrown, then I re-throw an `IllegalArgumentException`.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
yesamer: @danielezonca Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/AbstractExpressionEvaluator.java
yesamer: @danielezonca Done

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluator.java
yesamer: @danielezonca I just merged master which now contains that PR.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
yesamer: @danielezonca added :)

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/JsonUtils.java
gitgabrio: @yesamer 
I thought that method would be invoked *only* to parse an (expected) json string. If this is true, then IMO `JsonParseException` should not be hided behind an `Optional.empty()`, because it means something unexpected is happening. If - instead - this method is invoked on any possible String to verify if it is a valid json, then the javadoc should somehow clarify that

yesamer: @gitgabrio 
> If - instead - this method is invoked on any possible String to verify if it is a valid json

This is the correct case, I updated the javadoc to clarify it.

gitgabrio (approved):

danielezonca (approved):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -54,15 +58,34 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
...
 
+    /**
+     * A rawValue is an MVEL expression if:
+     * - NOT COLLECTIONS CASE: It's a <code>String</code> which starts with MVEL_ESCAPE_SYMBOL ('#')
+     * - COLLECTION CASE: It's a JSON String node, which is used only when an expression is set
+     *   (in other cases it's a JSON Object (Map) or a JSON Array (List)) and it's value starts with MVEL_ESCAPE_SYMBOL ('#')
+     * @param rawValue
+     * @return
+     */
+    protected boolean isAnMVELExpression(String rawValue) {
```
jomarko, line 78: This is standard public static final helper method, so just wondering if the placement is the best choice we have.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
```diff
@@ -54,15 +58,34 @@ public ExpressionEvaluator getOrCreate(FactMappingValue factMappingValue) {
             return getOrCreateDMNExpressionEvaluator();
         }
 
-        Object rawValue = factMappingValue.getRawValue();
+        String rawValue = (String) factMappingValue.getRawValue();
 
-        if (rawValue instanceof String && ((String) rawValue).trim().startsWith(MVEL_ESCAPE_SYMBOL)) {
+        if (isAnMVELExpression(rawValue)) {
```
jomarko, line 63: Please notice we changed the logic, we are now not checking `instanceof` in the `isAnMVELExpression`. We automatically cast to String what theoretically can cause cast exception.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
```diff
@@ -200,12 +202,26 @@ public void listOfSimpleTypeTest() {
...
+        List<BigDecimal> contextValue = Collections.singletonList(BigDecimal.valueOf(10));
+        assertTrue(expressionEvaluator.verifyResult(expressionCollectionJsonString, contextValue, List.class));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void expressionMapVerifyResultTest() {
+        String expressionCollectionJsonString = new TextNode("{key_a : 1}").toString();
+        Map<String, BigDecimal> contextValue = Collections.singletonMap("key_a", BigDecimal.valueOf(1));
+        assertTrue(expressionEvaluator.verifyResult(expressionCollectionJsonString, contextValue, Map.class));
     }
```
jomarko, line 226: Could you please help me understand this, why `IllegalArgumentException` is expected in both cases? Is it because `expressionCollectionJsonString` needs to start with '[' and ends with ']' ?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
```diff
@@ -115,6 +117,31 @@ public void evaluateLiteralExpression() {
         assertThatThrownBy(() -> evaluator.evaluateLiteralExpression("# a = 1; a+b;", Integer.class.getCanonicalName(), Collections.emptyList()))
                 .isInstanceOf(CompileException.class);
 
+        assertEquals(Arrays.asList("Bob", "Michael"), evaluator.evaluateLiteralExpression(mvelExpression("a = \"Bob\";\n" +
+                                                                                                                 "test = new java.util.ArrayList();\n" +
+                                                                                                                 "test.add(a);\n" +
+                                                                                                                 "test.add(\"Michael\");\n" +
+                                                                                                                 "test;"),
+                                                                                          ArrayList.class.getCanonicalName(),
+                                                                                          new ArrayList<>()
```
jomarko: Could we please use `Collections.emptyList()`?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
```diff
@@ -115,6 +117,31 @@ public void evaluateLiteralExpression() {
...
+        ));
+
+        HashMap<String, String> expectedMap = new HashMap<String, String>() {{
+            put("Jim", "Person");
+        }};
+
+        assertEquals(expectedMap, evaluator.evaluateLiteralExpression(mvelExpression("[\"Jim\" : \"Person\"]"),
+                                                                      HashMap.class.getCanonicalName(),
+                                                                      new ArrayList<>()
```
jomarko: Could we please use `Collections.emptyList()`?

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
```diff
@@ -115,6 +117,31 @@ public void evaluateLiteralExpression() {
...
+        ));
+        assertEquals(expectedMap, evaluator.evaluateLiteralExpression(mvelExpression("a = \"Person\";\n" +
+                                                                                             "test = new java.util.HashMap();\n" +
+                                                                                             "test.put(\"Jim\", a);\n" +
+                                                                                             "test;"),
+                                                                      HashMap.class.getCanonicalName(),
+                                                                      new ArrayList<>()
```
jomarko: Could we please use `Collections.emptyList()`?

jomarko (commented):
### Code review
See inline comments

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/BaseExpressionEvaluatorTest.java
yesamer: @jomarko Yes, the expression collection feature doesn't work with `BaseExpressionEvaluator`, but only with `MVELExpressionEvaluator` and `DMNExpressionEvaluator`. Consider that the frontend, with the current status, can never sends an expression defined collection to be evaluated with `BaseExpressionEvaluator`. If happens, it's not expected.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: @jomarko You're right BUT: rawValue is always a `String`, even if it's declared as `Object` in `FactMappingValue` DTO. If is not a String, it's an error, then it's correct to receive an exception. Consider that we planned with @danielezonca a refactor where we'll change that `Object` type in `String` -->  https://issues.redhat.com/browse/DROOLS-4948

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorFactory.java
yesamer: @jomarko It could be moved to a Util class. On the other side, this method is used only in this class and I don't see another place where it could be used. In my opinon, an `Util` class should be created if it cans hold static methods which can be shared in multiple places. This is not the case.  I vote to don't change it, if you have an alternative pls share.

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
yesamer: @jomarko Done!

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
yesamer: @jomarko Done!

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/expression/MVELExpressionEvaluatorTest.java
yesamer: @jomarko Done!

jomarko (approved):
Approving this part, but before merge I would like to:
- increase coverage https://github.com/yesamer/drools/pull/2 slightly
- Review drools-wb- part of ensemble
- Do manual check

Thank you @yesamer

PR #2695, opened 2019-12-18
DROOLS-4882 : [Guided DecisionTable] Incorrect getters and setters for fields that starts with one lowercase letter

Rikkola:
https://issues.redhat.com/browse/DROOLS-4882

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -1657,7 +1657,7 @@ public class RuleModelDRLPersistenceImpl
                 buf.append(fieldValue.getField());
             } else {
                 buf.append(".set");
-                buf.append(Character.toUpperCase(fieldValue.getField().charAt(0)));
+                buf.append(formatFieldFirstCharacterToFitDroolsCoreStandards(fieldValue.getField()));
                 buf.append(fieldValue.getField().substring(1));
             }
             buf.append("( ");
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM.

jomarko:
@Rikkola Hello, I am sorry, I am confused with your message "cherry picked from". I expect such message in cases we have the issue already solved in some stream, but neither DROOLS-4882 or RHPAM-412 seems to be resolved. Could you please explain me what are we cherry-picking?

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -1735,6 +1735,14 @@ protected void generateModifyMethodSeparator(final RHSGeneratorContext gctx,
             }
         }
 
+        private char formatFieldFirstCharacterToFitDroolsCoreStandards(final String fieldName) {
+            if (fieldName.length() > 2 && Character.isLowerCase(fieldName.charAt(0)) && Character.isUpperCase(fieldName.charAt(1))) {
```
jomarko: I think the condition needs top start as `fieldName.length() > 1` instead  of `fieldName.length() > 2 `.

I think the generated code getters/setters has same rules for fields **aField** and **aF**. Or am I wrong?

jomarko (changes requested):
I think we need to update the PR due to a corner case - two letter fields in the form [lower case letter][upper case letter].

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
Rikkola: Fixed and unit test updated to cover this.

Rikkola:
@jomarko The "cherry picked from" was due to me cherry-picking this from another branch that had other changes not related to this ticket. I removed the text from the commit. Ready for rereview now.

jomarko (approved):
Thank you @Rikkola

Rikkola:
Jenkins please execute SonarCloud Code Analysis

Rikkola:
Jenkins please execute SonarCloud Code Analysis

Rikkola:
@kiegroup/gatekeepers Can't get the Analysis green due to it being broken. Based on the repository commit history this does not stop merges. So can I have a merge please :)

PR #2696, opened 2019-12-19
DROOLS-4886: Test case, rule with updated nested fact evaluation does…

martinweiler:
… not fire after incremental update


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/21c99db14155b2b2361066cc2da2ba33198e3aca

PR #2698, opened 2019-12-19
DROOLS-4890 DMN classic comparison and range test

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
@@ -34,7 +34,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Random;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
```diff
@@ -2720,5 +2721,31 @@ public void testBindingContextTypeCheck() {
...
     }
+
+    @Test
+    public void testClassicComparisonVsRange() {
+        final DMNRuntime runtime = DMNRuntimeUtil.createRuntime("classicComparisonVsRange.dmn", this.getClass());
+        final DMNModel dmnModel = runtime.getModel("http://www.trisotech.com/definitions/_ac6efb68-08ed-43ec-b427-e99e78f51ba1", "Drawing 1");
+        assertThat(dmnModel, notNullValue());
+        assertThat(DMNRuntimeUtil.formatMessages(dmnModel.getMessages()), dmnModel.hasErrors(), is(false));
+
+        Random random = new Random();
```
baldimir: This might bring inconsistencies into the test in case of a potential bug. I mean some bug might be related to some specific value or similar.

baldimir: It might be better to test the whole potential range that can be generated. What do you think?

baldimir (approved):

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNRuntimeTest.java
tarilabs: Addressed with 1abaf57

PR #2700, opened 2019-12-20
[DROOLS-4878] Test Scenario: fix support for default stateless kieSession

danielezonca:
@yesamer @gitgabrio @dupliaka @jomarko 

Changed approach to force pseudo clock based on `KieSessionConfiguration` that is more reliable.
Removed also pseudo clock override in Stateless KieSession because it is not needed

https://issues.redhat.com/browse/DROOLS-4878

PRs:
- https://github.com/kiegroup/drools/pull/2700
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/417


yesamer (approved):

yesamer:
@danielezonca fine for me, my only suggestion is to evaluate Code Smells reported by Sonar.

danielezonca:
@yesamer 
Fixed one code smell but I cannot/don't want to fix the other one: `NewKieSessionCommand` implements `org.kie.api.command.Command` interface (root interface for all our commands) that is `Serializable` because we expose most of them also via a remote API (XML/JSON based). This command is not `Serializable` because of `customizeSessionConfiguration` field (a `BiFunction`).
I prefer to avoid a `transient` modifier to skip that field and silent serialization exception because it is correct that it will fail if someone will try to use it remotely.

jomarko:

jomarko:

danielezonca:

jomarko (commented):
[Edit] forgot to check stateless checkbox, rechecking.

jomarko (approved):
Works fine after recheck, thank you @danielezonca

PR #2701, opened 2019-12-22
module name change

wcarmon:
# Change Summary
Replacing module name with a valid one.

# Justification for change
- See https://docs.oracle.com/javase/specs/jls/se10/html/jls-7.html#jls-ModuleDeclaration
  - Specifically: `A module name consists of one or more Java identifiers (§3.8) separated by "." tokens`
- See https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleFinder.html#automatic-modules

# Example error based on current module name (Oracle jdk-11.0.5)
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Unable to derive module descriptor for /home/wcarmon/.m2/repository/org/drools/drools-canonical-model/7.30.0.Final/drools-canonical-model-7.30.0.Final.jar
Caused by: java.lang.module.FindException: Automatic-Module-Name: org.drools.canonical-model: Invalid module name: 'canonical-model' is not a Java identifier
```

### DIFF
--- a/drools-model/drools-canonical-model/pom.xml
+++ b/drools-model/drools-canonical-model/pom.xml
@@ -14,7 +14,7 @@
   <packaging>bundle</packaging><!-- bundle = jar + OSGi metadata -->
 
   <properties>
-    <java.module.name>org.drools.canonical-model</java.module.name>
+    <java.module.name>org.drools.canonicalModel</java.module.name>
   </properties>
 
   <dependencies>


wcarmon:
These also have the same module naming issue:
- https://github.com/kiegroup/drools/blob/master/drools-core-dynamic/pom.xml#L18
- https://github.com/kiegroup/drools/blob/master/drools-core-reflective/pom.xml#L18
- https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/pom.xml#L17
- https://github.com/kiegroup/drools/blob/master/drools-model/drools-mvel-compiler/pom.xml#L17
- https://github.com/kiegroup/drools/blob/master/drools-model/drools-mvel-parser/pom.xml#L14
- https://github.com/kiegroup/drools/blob/master/drools-ruleunit/pom.xml#L18

drools-model/drools-canonical-model/pom.xml
```diff
@@ -14,7 +14,7 @@
   <packaging>bundle</packaging><!-- bundle = jar + OSGi metadata -->
 
   <properties>
-    <java.module.name>org.drools.canonical-model</java.module.name>
+    <java.module.name>org.drools.canonicalModel</java.module.name>
```
wcarmon, line 17: or `<java.module.name>org.drools.canonicalmodel</java.module.name>` if it makes more sense.

The important thing is that the hyphen is removed to make it a valid java identifier (so this library can be used with java9+ code)

mariofusco:
Thanks a lot for having reported this. We are already aware of this problem and logged it here https://issues.redhat.com/browse/DROOLS-4736 Since this PR only fixes one module I will close it and we will work immediately on a more complete fix.

PR #2703, opened 2020-01-02
[DROOLS-4888] improve kmodule indexing from jar files

mariofusco:


mariofusco:

PR #2705, opened 2020-01-03
DROOLS-1084 Exclude 'provided', 'test' and 'system' scope before scanning full tree

awgroeneveld:
In a Dockerized environment with an empty local Maven repository getting all artefacts from a remote repository is an expensive operation. The same is true for very big dependency trees in a local environment.

As it seems even non-needed dependencies with _provided_, _test_ or _system_ scope are being fully traversed for transitive dependencies.

My proposition would be to exclude _'provided'_, _'test'_ and _'system'_ scope before scanning the full dependency tree.
As the results in the original code are already filtered after getting all transitive dependencies, it would just as well be filtered on the direct dependencies first as this restricts the scope already and should not be harmful therefore.

There is a report DROOLS-1084 issue describing the unwanted behaviour though it also describes some side effects likely not noticed if the scope filtering happens before getting the tree.

The original [commit](https://github.com/kiegroup/drools/commit/80a06ea4bbb542eb8b240b484145684618b05851) likely fixed the issue with respect to failing but it still traverses the tree if the top level dependency scopes can already be ignored.

The performance gain in a Dockerized Spring environment is huge, if there are no objections I would very much appreciate the code change to be merged. Thank you!

The kie-ci module was built without failing tests.


mariofusco:

PR #2706, opened 2020-01-07
[DROOLS-4870] Materialize consequences using Drools

tkobayas:
This is a draft PR to share my idea. Please let me know your thoughts.

Basically, it writes BitMask variables into the materialized LambdaConsequence as "private final" fields so execute() method can access them.

It passed all drools-model unit tests but I found that it fails with IncrementalCompilationCepTest#testIncrementalCompilationWithNewEvent in drools-test-coverage/test-compiler-integration. The failure is caused by duplication of LambdaConsequence class name during incremental compilation. I will look into it further.

One more point from code design aspect, I added "List<VariableDeclarator> bitMaskVariables" to MaterializedLambda. Actually, "bitMaskVariables" is required only by MaterializedLambdaConsequence. But if I add "List<VariableDeclarator> bitMaskVariables" to MaterializedLambdaConsequence, I would need more refactoring (e.g. MaterializedLambda.create() and ExecModelLambdaPostProcessor.extractLambdaFromMethodCall()). If you think it's worth refactoring, please let me know.

Thanks!

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
@@ -18,12 +18,19 @@
 package org.drools.modelcompiler.util.lambdareplace;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.FieldAccessExpr;
 import com.github.javaparser.ast.expr.LambdaExpr;
@@ -31,6 +38,7 @@ import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.ast.type.Type;
+import org.drools.model.BitMask;
 import org.drools.modelcompiler.builder.RuleWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -80,7 +88,10 @@ public class ExecModelLambdaPostProcessor {
                     .forEach(this::convertIndexedByCall);
 
             clone.findAll(MethodCallExpr.class, this::isExecuteNonNestedCall)
-                    .forEach(methodCallExpr -> extractLambdaFromMethodCall(methodCallExpr, () -> new MaterializedLambdaConsequence(packageName, ruleClassName)));
+                    .forEach(methodCallExpr -> {
+                        List<MaterializedLambda.BitMaskVariable> bitMaskVariables = findBitMaskFields(methodCallExpr);
+                        extractLambdaFromMethodCall(methodCallExpr, () -> new MaterializedLambdaConsequence(packageName, ruleClassName, bitMaskVariables));
+                    });
     }
 
     private boolean isExecuteNonNestedCall(MethodCallExpr mc) {
@@ -111,6 +122,40 @@ public class ExecModelLambdaPostProcessor {
 
     private Expression lambdaInstance(ClassOrInterfaceType type) {
         return new FieldAccessExpr(new NameExpr(type.asString()), "INSTANCE");
+    }
+
+    private List<MaterializedLambda.BitMaskVariable> findBitMaskFields(MethodCallExpr methodCallExpr) {
+        return optionalToStream(methodCallExpr.findAncestor(MethodDeclaration.class))
+                .flatMap(node -> node.findAll(VariableDeclarator.class).stream())
+                .filter(this::isBitMaskType)
+                .flatMap(this::findAssignExpr)
+                .map(this::toMaterializedLambdaFactory)
+                .collect(Collectors.toList());
+    }
+
+    private boolean isBitMaskType(VariableDeclarator vd) {
+        return vd.getType().asString().equals(BitMask.class.getCanonicalName());
+    }
+
+    private MaterializedLambda.BitMaskVariable toMaterializedLambdaFactory(AssignExpr ae) {
+        String maskName = ae.getTarget().asVariableDeclarationExpr().getVariables().iterator().next().getNameAsString();
+        MethodCallExpr maskInit = ae.getValue().asMethodCallExpr();
+        if (maskInit.getArguments().isEmpty()) {
+            return new MaterializedLambda.AllSetButLastBitMask(maskName);
+        } else {
+            NodeList<Expression> arguments = maskInit.getArguments();
+            String domainClassMetadata = arguments.get(0).toString();
+            List<String> fields = arguments.subList(1, arguments.size()).stream().map(Expression::toString).collect(Collectors.toList());
+            return new MaterializedLambda.BitMaskVariableWithFields(domainClassMetadata, fields, maskName);
+        }
+    }
+
+    private Stream<? extends AssignExpr> findAssignExpr(VariableDeclarator vd) {
+        return optionalToStream(vd.findAncestor(AssignExpr.class));
+    }
+
+    private <T> Stream<T> optionalToStream(Optional<T> opt) {
+        return opt.map(Stream::of).orElse(Stream.empty());
     }
 
     private void extractLambdaFromMethodCall(MethodCallExpr methodCallExpr, Supplier<MaterializedLambda> lambdaExtractor) {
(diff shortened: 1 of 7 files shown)


tkobayas:
"Linux - compile downstream" failure was caused by revapi-maven-plugin for optaplanner-core so this PR doesn't seem to be related.

lucamolteni:
This approach works perfectly, 
I didn't realize the BitMask depended only from a static DomainClassMetadata file, we can embed it perfectly in the fields of the consequence and the Lambda is not capturing anymore. Good job thank you.
I have a few suggestion on the code let me know what you think.

Please refer to this patch https://gist.github.com/lucamolteni/e898cfdd56089c8852a1384add16777d

I'm also commenting inline

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
```diff
@@ -114,12 +120,21 @@ private Expression lambdaInstance(ClassOrInterfaceType type) {
     }
 
     private void extractLambdaFromMethodCall(MethodCallExpr methodCallExpr, Supplier<MaterializedLambda> lambdaExtractor) {
+        List<VariableDeclarator> bitMaskVariables = new ArrayList<>();
```
lucamolteni: Let's not use JP types as input for the model compiler, only for output. 

This is because in the future we'd like to emit JP code at the very end, therefore during the processing we should only use types owned by us.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
```diff
@@ -114,12 +120,21 @@ private Expression lambdaInstance(ClassOrInterfaceType type) {
...
+        if (lambdaExtractor.get() instanceof MaterializedLambdaConsequence) {
```
lucamolteni: I moved this parsing outside in the patch to avoid the instanceof

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
```diff
@@ -38,6 +39,7 @@
 abstract class MaterializedLambda {
 
     final List<LambdaParameter> lambdaParameters = new ArrayList<>();
+    final List<VariableDeclarator> bitMaskVariables = new ArrayList<>();
```
lucamolteni: Superclass shouldn't be aware of implementation details of Consequences

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequence.java
```diff
@@ -65,6 +65,23 @@ void createMethodDeclaration(EnumDeclaration classDeclaration) {
         }
     }
 
+    @Override
+    protected EnumDeclaration create(CompilationUnit compilationUnit) {
+        EnumDeclaration lambdaClass = super.create(compilationUnit);
+
+        boolean hasDroolsParameter = lambdaParameters.stream().anyMatch(this::isDroolsParameter);
```
lucamolteni, line 72: In the refactor patch I changed a logic reconstructing the original initialization method here, to avoid depending on JP data on input.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequenceTest.java
```diff
@@ -32,4 +38,188 @@ public void createConsequence() {
         assertThat(aClass.getCompilationUnitAsString(), equalToIgnoringWhiteSpace(expectedResult));
 
     }
+
+    @Test
+    public void createConsequenceWithDrools() {
+        String originalJavaSource = "package defaultpkg;\n" +
+                "\n" +
```
lucamolteni: This test has too much Java code in text to be readable.

In the patch I modified the signature of the create method so that it can be tested in an easier way. Please take a look at it

lucamolteni (changes requested):
Thank you for the good work.

I made some suggestion and attached the patch to improve the code in a way that I think it's more similar to what it was before.

tkobayas:
@lucamolteni Thank you very much for the refactoring patch. It makes much sense to me. I applied it and then added a few modifications. Please see inline comments.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
```diff
@@ -113,6 +124,40 @@ private Expression lambdaInstance(ClassOrInterfaceType type) {
...
+                            })
+                            .map(ae -> {
+                                String maskName = ae.getTarget().asVariableDeclarationExpr().getVariables().iterator().next().getNameAsString();
+                                MethodCallExpr maskInit = ae.getValue().asMethodCallExpr();
+                                if(maskInit.getArguments().isEmpty()) {
+                                    return new MaterializedLambda.AllSetButLastBitMask(maskName);
+                                } else {
+                                    NodeList<Expression> arguments = maskInit.getArguments();
+                                    String domainClassMetadata = arguments.get(0).toString();
+                                    List<String> fields = arguments.subList(1, arguments.size()).stream().map(Expression::toString).collect(Collectors.toList());
```
tkobayas: Originally, it picks only one field but BitMask may take multiple fields so I changed it to a List.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
```diff
@@ -146,4 +154,53 @@ private EnumDeclaration create(CompilationUnit compilationUnit) {
             this.type = type;
         }
     }
+
+    interface BitMaskVariable {
+
+        void generateBitMaskField(NodeWithMembers<EnumDeclaration> clazz);
```
tkobayas, line 167: Originally, the argument `NodeWithMembers` was not parameterized so i see compiler warns. It's fine to have `<EnumDeclaration>` here?

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
tkobayas: Fix applied. thanks!

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
tkobayas: Fix applied. thanks!

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
tkobayas: Fix applied. thanks!

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequenceTest.java
tkobayas: Fix applied. thanks!

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequence.java
tkobayas: Fix applied. thanks!

tkobayas:
Btw, I still need to look into IncrementalCompilationCepTest.testIncrementalCompilationWithNewEvent() failure. Please allow me some more time.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
lucamolteni: I also think it's better to remove the generic type argument

tkobayas: Talked with Luca and confirmed that `NodeWithMembers<EnumDeclaration>` is better.

tkobayas:
Applied Luca's refactoring.

lucamolteni (approved):

tkobayas:
I have isolated DROOLS-4922 from the IncrementalCompilationCepTest.testIncrementalCompilationWithNewEvent() error.
I think we need to fix it before this DROOLS-4870.

lucamolteni:
> I have isolated DROOLS-4922 from the IncrementalCompilationCepTest.testIncrementalCompilationWithNewEvent() error.
> I think we need to fix it before this DROOLS-4870.

I Agree @tkobayas

tkobayas:
jenkins test again

tkobayas:
Okay, DROOLS-4922 is resolved so now all tests passed. Adding reviewers... @mariofusco @baldimir  @hellowdan  Please review, thanks!

tkobayas:

tkobayas:
Jenkins execute compile downstream build

PR #2707, opened 2020-01-07
DROOLS-4873 DMN v1.3 unary test syntax of range

tarilabs:

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
+++ b/kie-dmn/kie-dmn-feel/src/main/antlr4/org/kie/dmn/feel/parser/feel11/FEEL_1_1.g4
@@ -317,10 +317,10 @@ BooleanLiteral
 
 // #7
 simplePositiveUnaryTest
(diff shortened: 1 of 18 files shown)


hellowdan (approved):

tarilabs:
Jenkins test again

tarilabs:
no way it's 0% red coverage.

Jenkins test again

tarilabs:
no way it's 0% red coverage.

Jenkins test again

jiripetrlik (changes requested):
It seems that there are some expressions which always evaluate to "false" according to Sonar cloud report. Please check and remove these expressions if possible.

tarilabs:
The SonarCloud report I believe is dangerous.

I believe the correct solution is NOT to remove the expression which SonarCloud is suggesting to remove, but change it from
`... lowEndPoint == null || highEndPoint == null ...`
to
`... lowEndPoint == null && highEndPoint == null ...`

jiripetrlik:
@tarilabs 
Yes, I agree. Sonar cloud is definitely not always true. Can you try to change it as you proposed and run PR checks?

baldimir (approved):

PR #2714, opened 2020-01-09
[PLANNER-1770] Replace jboss annotation api with jakarta for quarkus-optaplanner extension

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-1770

Merge with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1159

Fixes banned dependency by [Quarkus](https://github.com/quarkusio/quarkus/blob/master/core/creator/src/main/java/io/quarkus/creator/curator/Curator.java#L268) when creating the Quarkus-OptaPlanner extension.

### DIFF
--- a/drools-model/drools-mvel-parser/pom.xml
+++ b/drools-model/drools-mvel-parser/pom.xml
@@ -22,8 +22,8 @@
       <artifactId>javaparser-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.jboss.spec.javax.annotation</groupId>
-      <artifactId>jboss-annotations-api_1.2_spec</artifactId>
+      <groupId>jakarta.annotation</groupId>
+      <artifactId>jakarta.annotation-api</artifactId>
     </dependency>
 
     <dependency>


mareknovotny:
I think we can't upgrade only one of the Java EE API to Jakarta API, all or none IMHO, there will be conflicts or duplicate classes due not complete migration to Jakarta APIs.

cuijulian:
@mareknovotny Noted. I will upgrade to Jakarta for the entire drools repo.

mareknovotny:
well, is that required to migrate to Jakarta now? just curious what drives that need

cuijulian:
@mareknovotny Good point. I'm doing work on the Quarkus Optaplanner extension, which brings in optaplanner-core and drools dependencies into the Quarkus repo. The problem is that Quarkus bans JBoss spec jars, and it seems to be pushing for the Jakarta spec alternative instead. The long-term goal is to keep the entire `kiegroup` consistent with these spec jars and use Jakarta rather than other outdated standards. That is my understanding, I'm happy to discuss this.

mareknovotny:
ok, thanks for explanation @cuijulian. 

@ge0ffrey this types of changes which are driven by Kogito is not really good to include in our stable kiegroup stream. The alignment of APIs here happens always on top of our platform and this is for V7 Widlfly/EAP. When we move to APIs from this platform this won't be dealbreaker, but right now it is.

cuijulian:
@mareknovotny I will set us up for a BJ meeting with @ge0ffrey and @mariofusco to further discuss the spec jars for Kie 7. We can take this discussion there.

cuijulian:
Resolved by https://github.com/kiegroup/drools/commit/04065025a48aa579c973609d3379b12ff143cddf (thanks Mario!)

PR #2715, opened 2020-01-10
[DROOLS-4922] LinkageError: attempted duplicate class definition for …

tkobayas:
…materialized Lambda during incremental compilation

- Unit Test only now

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.EnumConstantDeclaration;
 import com.github.javaparser.ast.body.EnumDeclaration;
@@ -34,13 +35,16 @@ import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.UnknownType;
+import org.drools.modelcompiler.builder.JavaParserCompiler;
+
+import static org.drools.modelcompiler.util.StringUtil.md5Hash;
 
 abstract class MaterializedLambda {
 
     final List<LambdaParameter> lambdaParameters = new ArrayList<>();
 
     protected final String packageName;
-    protected String className;
+    protected String temporaryClassName;
 
     LambdaExpr lambdaExpr;
     private String ruleClassName;
@@ -58,7 +62,7 @@ abstract class MaterializedLambda {
         }
 
         lambdaExpr = expression.asLambdaExpr();
-        className = className(expressionString);
+        temporaryClassName = className(expressionString);
 
         parseParameters();
 
@@ -68,6 +72,9 @@ abstract class MaterializedLambda {
         EnumDeclaration classDeclaration = create(compilationUnit);
 
         createMethodDeclaration(classDeclaration);
+
+        String className = className(JavaParserCompiler.getPrettyPrinter().print(compilationUnit));
+        classDeclaration.setName(className);
 
         return new CreatedClass(compilationUnit, className, packageName);
     }
@@ -107,10 +114,13 @@ abstract class MaterializedLambda {
     }
 
     private EnumDeclaration create(CompilationUnit compilationUnit) {
-        EnumDeclaration lambdaClass = compilationUnit.addEnum(className);
+        EnumDeclaration lambdaClass = compilationUnit.addEnum(temporaryClassName);
         lambdaClass.addAnnotation(org.drools.compiler.kie.builder.MaterializedLambda.class.getCanonicalName());
         lambdaClass.setImplementedTypes(createImplementedType());
         lambdaClass.addEntry(new EnumConstantDeclaration("INSTANCE"));
+
+        lambdaClass.addFieldWithInitializer(String.class, "EXPRESSION_HASH", StaticJavaParser.parseExpression("\"" + (md5Hash(lambdaExpr.toString())) + "\""),
+                                            Modifier.Keyword.PUBLIC, Modifier.Keyword.STATIC, Modifier.Keyword.FINAL);
         return lambdaClass;
     }
 
@@ -130,7 +140,7 @@ abstract class MaterializedLambda {
                 .collect(Collectors.toList());
     }
 
-    abstract String className(String expressionString);
+    abstract String className(String sourceCode);
 
     abstract ClassOrInterfaceType functionType();
 
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/LambdaIntrospector.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/LambdaIntrospector.java
@@ -19,6 +19,7 @@ package org.drools.modelcompiler.util;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.lang.invoke.SerializedLambda;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -46,7 +47,7 @@ public class LambdaIntrospector implements LambdaPrinter {
     @Override
     public String getLambdaFingerprint(Object lambda) {
         if(lambda.toString().equals("INSTANCE")) { // Materialized lambda
-            return lambda.getClass().getCanonicalName();
+            return getExpressionHash(lambda);
         }
 
         if (lambda instanceof IntrospectableLambda ) {
@@ -63,6 +64,16 @@ public class LambdaIntrospector implements LambdaPrinter {
             }
         }
         return result;
+    }
+
+    private String getExpressionHash(Object lambda) {
+        Field expressionHash;
+        try {
+            expressionHash = lambda.getClass().getDeclaredField("EXPRESSION_HASH");
+            return (String) expressionHash.get(lambda);
+        } catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
+            throw new RuntimeException( e );
+        }
     }
 
     private static SerializedLambda extractLambda( Serializable lambda ) {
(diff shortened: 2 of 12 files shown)


tkobayas:
In case of incremental compilation, say, a rule is changed:
- A lambda consequence (or a lambda predicate) has the same lambda expression
- But an "import" statement is added to the rule

the generated LambdaConsequence (LambdaPredicate) class's content will change but the class name stays the same (because the class name hash is generated from lambda expression). Then, DynamicProjectClassLoader throws the LinkageError.

Note: If the content also stays the same, CanonicalKieModule.findChanges() will detect it so the class will not be reloaded.

To fix the error:

A) Generate a class name hash from the whole source code so any change will result in a different class name

B) DynamicProjectClassLoader should properly handle reloading even in this JIRA's scenario (unload and reload the class)

Which is appropriate (or any other)? Please share your thought on this. @lucamolteni @mariofusco

lucamolteni:
I'd say A) is fine, I'm only worried about the order of the imports, but if that happens I think a new class is generated and everything is fine anyway

tkobayas:
This issue gets tricky. I implemented A) so the unit tests (testIdenticalConsequenceButImportChange/testIdenticalPredicateButImportChange) passed. But now some tests fail.

IncrementalCompilationTest.testKJarUpgradeSameAndDifferentSessions
IncrementalCompilationTest.testKJarUpgradeWithDeclaredType
MultiKieBaseTest.testHelloWorldWithPackagesAnd2KieBases

For example, with non-executable-model, IncrementalCompilationTest.testKJarUpgradeSameAndDifferentSessions() works like:

`mvn test -Dtest=IncrementalCompilationTest#testKJarUpgradeSameAndDifferentSessions[STANDARD_FROM_DRL]`

1. The first fact is inserted (matches R1).
2. fireAllRule()
  -> R1 is fired for the first fact
   -> fire count is 1
3. updateToVersion()
  -> R1 doesn't change. R2 is updated (the first fact matches R2). 
4. The second fact is inserted (matches R1 and R2).
5. fireAllRule()
  -> R2 is fired for 2 facts. R1 is fired for the second fact
    -> fire count is 3. Assertion success.
...

output
~~~
R1 : $m = 5d21202d
------------ updateToVersion
R2_2 : $m = 26586b74
R2_2 : $m = 5d21202d
R1 : $m = 26586b74
~~~

Then, with executable-model, IncrementalCompilationTest.testKJarUpgradeSameAndDifferentSessions() works like:

mvn test `-Dtest=IncrementalCompilationTest#testKJarUpgradeSameAndDifferentSessions[PATTERN_DSL]`

1. The first fact is inserted (matches R1).
2. fireAllRule()
  -> R1 is fired for the first fact
   -> fire count is 1
3. updateToVersion()
  -> R1 is updated. R2 is updated (the first fact matches R1 and R2). 
4. The second fact is inserted (matches R1 and R2).
5. fireAllRule()
  -> R1 is fired for 2 facts. R2 is fired for 2 facts.
    -> fire count is 4. Assertion failure.
...

output
~~~
R1 : $m = 5d508d22
------------ updateToVersion
R1 : $m = 782fd504
R1 : $m = 5d508d22
R2_2 : $m = 782fd504
R2_2 : $m = 5d508d22
~~~

Why R1 is updated? It's because:

1. R1's LambdaConsequence has `import static org.drools.incremental.RulesCB6ED88489B51A62DA8E5222836C1DAE.*;`
2. The Rule class name changes as the releaseId changes
3. Now LambdaConsequence class name hash is generated from whole source code. So LambdaConsequence class name changes.

Next step:

I think we need to load the new LambdaConsequence class anyway because the import statement indeed changes (e.g. The import is used for calling a function). But in order to have the same result with [STANDARD_FROM_DRL], R1 should not be re-evaluated (= should not match with the first fact again). I will look further. If you have any suggestions, feel free to let me know.

tkobayas:
@lucamolteni  @mariofusco 
I have updated the PR to share my current idea.

It fails to compile unless you apply this change to droolsjbpm-knowledge repo.

https://gist.github.com/tkobayas/9dadafbdc2f87beb9127c079bd8f8c57

The points are:

1) Materialized LambdaConsequence class contains an "import" statement which refers to a new generated Rule class (e.g. `import static org.drools.incremental.RulesCB6ED88489B51A62DA8E5222836C1DAE.*;`). So materialized LambdaConsequence class source code is always new in case of incremental compilation.
   (Note: this JIRA's original concern was a scenario where an "import" statement is added to a rule. But actually, even without adding, materialized LambdaConsequence class source code always changes)
2) So we want to load/use the new materialized LambdaConsequence class even if the rule itself is not changed. (especially because we want that old classes will be garbage-collected)
3) But if we UPDATE the rule, it will remove/add the rule (https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/CanonicalKieBaseUpdater.java#L183-L184). So the rule is re-evaluated and the result (e.g. fire count) is different from non-executable-model. We don't want to re-evaluate the rule in this case.

 -> Correct me if these assumptions are wrong

So my idea is to set the new LambdaConsequence to the old rule. Please see source code comments below.

I wonder if I'm on the right track. Please kindly review and share your thought on this.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -543,6 +544,11 @@ private void addModifiedItemsToChangeSet( ResourceChangeSet mainChangeSet, Map<S
             if ( compare == 0 ) {
                 if ( !areEqualInModel( currentOld, currentNew ) ) {
                     registerChange( mainChangeSet, changes, type, ChangeType.UPDATED, currentOld );
+                } else {
+                    // Considered as equal in model. It means LambdaHash are equal. So now check class name hash
+                    if (currentOld instanceof RuleImpl && RuleImpl.areDifferentInLambdaClassNameHash( (RuleImpl)currentOld, (RuleImpl)currentNew )) {
+                        registerChange( mainChangeSet, changes, type, ChangeType.MERGE, currentOld );
```
tkobayas: Detect the case where rules are the same (= lambda expressions are the same) but lambda class class names are different (= whole lambda class source codes are different). In this case, I set ChangeType.MERGE. To compile, you need to apply this change to droolsjbpm-knowledge repo.
https://gist.github.com/tkobayas/9dadafbdc2f87beb9127c079bd8f8c57

drools-model/drools-canonical-model/src/main/java/org/drools/model/impl/RuleImpl.java
```diff
@@ -125,4 +126,17 @@ public boolean isEqualTo( ModelComponent o ) {
...
+
+    public static boolean areDifferentInLambdaClassNameHash(RuleImpl o1, RuleImpl o2) {
+        System.out.println("areDifferentInLambdaClassNameHash : " + o1.getName());
+        String consequenceName1 = ((IntrospectableLambda) o1.getDefaultConsequence().getBlock()).getLambda().getClass().getName();
+        String consequenceName2 = ((IntrospectableLambda) o2.getDefaultConsequence().getBlock()).getLambda().getClass().getName();
+        System.out.println("  consequenceName1 = " + consequenceName1);
+        System.out.println("  consequenceName2 = " + consequenceName2);
+        if (!consequenceName1.equals(consequenceName2)) {
+            System.out.println("  Different!!!");
+            return true;
```
tkobayas: This method only checks LambdaConsequnece for now. If this is the right approach, I would need to check LambdaPredicate and LambdaExtractor as well.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/CanonicalKieBaseUpdater.java
```diff
@@ -188,6 +199,12 @@ public void run() {
         }
     }
 
+    private void mergeRuleWithLambda(RuleImpl oldRule, RuleImpl newRule) {
+        Consequence newConsequence = newRule.getConsequence();
+        oldRule.setConsequence(newConsequence);
+        System.out.println("mergeRuleWithLambda!!");
```
tkobayas: This method only replaces LambdaConsequnece for now. If this is the right approach, I would need to replace LambdaPredicate and LambdaExtractor as well. I'm not sure how much work will be required though (and there are any side effects).

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambda.java
```diff
@@ -107,10 +114,13 @@ void setMethodParameter(MethodDeclaration methodDeclaration) {
...
 
     private EnumDeclaration create(CompilationUnit compilationUnit) {
-        EnumDeclaration lambdaClass = compilationUnit.addEnum(className);
+        EnumDeclaration lambdaClass = compilationUnit.addEnum(temporaryClassName);
         lambdaClass.addAnnotation(org.drools.compiler.kie.builder.MaterializedLambda.class.getCanonicalName());
         lambdaClass.setImplementedTypes(createImplementedType());
         lambdaClass.addEntry(new EnumConstantDeclaration("INSTANCE"));
+
+        lambdaClass.addFieldWithInitializer(String.class, "EXPRESSION_HASH", StaticJavaParser.parseExpression("\"" + (md5Hash(lambdaExpr.toString())) + "\""),
+                                            Modifier.Keyword.PUBLIC, Modifier.Keyword.STATIC, Modifier.Keyword.FINAL);
```
tkobayas, line 123: I added a public static final field "EXPRESSION_HASH" to validate the equality of materialized Lambda. This field is used in LambdaIntrospector.

tkobayas:
jenkins test again

tkobayas:
Note: DO-NOT-MERGE yet. This PR is WIP and contains some System.out.println() etc.

tkobayas:
@lucamolteni @mariofusco 
Updated the PR. Talked with Mario, we removed "ChangeType.MERGE" to minimize this PR's focus.

Now, by this fix:

- Materialized Lambda class name hash is generated from its whole source code instead of its Lambda content. It means the class will have a different class name on every incremental compilation because it will have a different "import defaultPkg.RuleXXXXX.*" (regardless of user's "import" is added or not).

 -> This solves the classloader error : IncrementalCompilationTest.testIdenticalConsequenceButImportChange()/testIdenticalPredicateButImportChange().

- Materialized Lambda class has "public static final String EXPRESSION_HASH" field which is a hash from its Lambda content. This is used for equality check (ModelComponent.areEqualInModel()). So if the Lambda content is not changed, the rule will not be updated even if "import" is added/changed. In this case, the rule will keep using old materialized Lambda class.

 -> So this doesn't break IncrementalCompilationTest.testKJarUpgradeSameAndDifferentSessions() etc. So all test passed.

The remaining problem is:

- Regardless of "import" is added or not, using old materialized Lambda class doesn't sound right because "import defaultPkg.RuleXXXXX.*" is changed. I should be able to write a failing unit test by a rule with "function". But executable model + incremental compile + function seems to have a problem now. I think I will file a JIRA for that. But I think we can merge this PR independently.

PR #2721, opened 2020-01-15
Support for DecisionTable template configuration in SpringBoot

nesterenko-alexey:
In SpringBoot application classes and resources are placed into BOOT-INF/classes/ subfolder of the assembled JAR file. That is why fileName (see updated code) starts with BOOT-INF/classes/ folder name. And template.getDtable() (taken from kmodule.xml) does not contain BOOT-INF/classes/ prefix. For this reason "if-condition" is changed to "fileName.endsWith( template.getDtable() )" - it will work both for SpringBoot and non-SpringBoot application.

As well, to get a template resource, a full path (with BOOT-INF/classes/ root subfolder) should be specified. That is why original code "getResource( template.getTemplate() );" will not work for SpringBoot application. Because of this, this code was replaced by "getResource( fileName.replace(template.getDtable(), "") + template.getTemplate() );". New code will work no matter which root folder is used inside JAR file - in SpringBoot application it will always prepend template.getTemplate() value with BOOT-INF/classes/ subfolder.


ghost:

mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/4264

PR #2722, opened 2020-01-15
DROOLS-4936 : Unsaved changes dialog appears even if it just have been saved

Rikkola:
https://issues.redhat.com/browse/DROOLS-4936

### DIFF
--- a/drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
+++ b/drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.models.datamodel.rule;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+public class CEPWindowTest {
+
+    private CEPWindow cepWindowAllFieldsFilled;
+    private CEPWindow cepWindowDefaultValues;
+    private CEPWindow cepWindowNullValues;
+    private CEPWindow cepWindowNullParameters;
+    private CEPWindow cepWindowNullOperator;
+
+    @Before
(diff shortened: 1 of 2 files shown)


Rikkola:
Not sure if a test makes sense. Since after and before the test the results are the same. Unless the test is ran in GWT, but GWTUnit tests are too slow to use.

jomarko:
@Rikkola I think we need to add unit tests, at least to confirm null values of fields that are used for hashcode computation doesn't cause an exception Thank you.

Rikkola:
@jomarko I added some tests, however with or without the change in the main code they pass.

jomarko:
@Rikkola then they do not test your change.

Rikkola:
@jomarko Yes that is the problem. Only way to test the change is to use GWTTestCase, but we should not use them since they are slow.

Rikkola:
Jenkins please retest this.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Test
+    public void defaultValues() {
+        final CEPWindow cepWindow = new CEPWindow();
+
+        assertNotNull(cepWindow.hashCode());
+    }
+
+    @Test
+    public void checkNulls() {
```
dupliaka, line 114: What about case if only parameters or only operator is set to null?

Rikkola: I added more test coverage for these situations.

dupliaka (approved):
Tested manually. 
Thanks for fix!

Rikkola:
@manstis Could you also take a look at this PR please.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
```diff
@@ -0,0 +1,52 @@
+package org.drools.workbench.models.datamodel.rule;
```
manstis, line 16: License please.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
manstis: IDK if it's an overkill but (at least) one test creating two equal instances of `CEPWindow` and verifying `hashCode`s are equal may be nice....

manstis:
Otherwise works fine (given the test on the RHPAM JIRA).

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/CEPWindowTest.java
Rikkola: @manstis I improved this test to not only check the hashcode, but to compare hashcodes between other instances. I hope the other workbench objects do not get jealous :)

manstis (approved):

Rikkola:
@kiegroup/gatekeepers Can I have a merge please.

PR #2723, opened 2020-01-15
RHDM-1159: Exception while coverting spreedsheet decision table to guided decision table

manstis:
See https://issues.redhat.com/browse/RHDM-1159

The XLS attached to the JIRA now converts correctly (and the User can view _source_).

Part of an ensemble:
- https://github.com/kiegroup/drools/pull/2723
- https://github.com/kiegroup/drools-wb/pull/1282

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -2753,7 +2754,7 @@ public class RuleModelDRLPersistenceImpl
             }
             int opPos = expr.indexOf(opString);
             if (opPos >= 0 &&
-                    isNotMethodName(expr, opPos) &&
+                    isNotMethodName(expr, opString, opPos) &&
                     !isInQuote(expr, opPos) &&
                     !(Character.isLetter(opString.charAt(0)) &&
                             (expr.length() == opPos + opString.length() || Character.isLetter(expr.charAt(opPos + opString.length())) ||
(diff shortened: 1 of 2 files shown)


Rikkola (approved):

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -2783,7 +2784,16 @@ private static String findOperator(String expr) {
...
     private static boolean isNotMethodName(final String expression,
+                                           final String potentialOperator,
                                            final int operatorPosition) {
+        if (Objects.equals(potentialOperator, Operator.EQUAL.getOperatorString()) ||
+                Objects.equals(potentialOperator, Operator.NOT_EQUAL.getOperatorString()) ||
+                Objects.equals(potentialOperator, Operator.LESS.getOperatorString()) ||
+                Objects.equals(potentialOperator, Operator.LESS_OR_EQUAL.getOperatorString()) ||
+                Objects.equals(potentialOperator, Operator.GREATER.getOperatorString()) ||
+                Objects.equals(potentialOperator, Operator.GREATER_OR_EQUAL.getOperatorString())) {
+            return true;
```
jomarko, line 2795: Shouln't we use `org.drools.core.base.evaluators.Operator.getAllOperators()` ?

jomarko (commented):
### Code review

jomarko:
Manaul review postponed until comment from code review responded.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
manstis: No, this addition came about due to the XLS attached to the JIRA containing a CONDITION definition `traceId==("$param")` where there are no spaces between the operator and the value so [this](https://github.com/kiegroup/drools/blob/9ea375f2be1d54d0598f35ad571681c798fec1f9/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L2797) check failed and the parsing of the XLS lead to incorrect results. 

I only check for `==`, `!=`, `<`, `<=`, `>`, `>=` as these cannot be the prefix to a method name whereas, for example, `matches` could be in the method name `matchesCheese()`.

jomarko (approved):
1/2

manstis:
@mareknovotny Sorry to ask.. how do I trigger a "downstream compile" (for the failed check above)?

jomarko:

PR #2726, opened 2020-01-16
RHDM-1189: 'Unable to expand' exception with from condition in Guided Rule Editor with DSL option checked

manstis:
See https://issues.redhat.com/browse/RHDM-1189

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -613,14 +613,23 @@ public class RuleModelDRLPersistenceImpl
         }
 
         public void visitFactPattern(final FactPattern pattern) {
-            buf.append(indentation);
-            if (isDSLEnhanced) {
+            visitFactPattern(pattern,
+                             rootContext);
+        }
+
+        protected void visitFactPattern(final FactPattern pattern,
+                                        final LHSGeneratorContext parentContext) {
+            final LHSGeneratorContext gctx = generatorContextFactory.newPeerGeneratorContext(parentContext, pattern);
+            final boolean isSubPattern = gctx.getDepth() > 0;
+
+            if (!isSubPattern) {
+                buf.append(indentation);
+            }
+            if (!isSubPattern && isDSLEnhanced) {
                 // adding passthrough markup
                 buf.append(">");
             }
 
-            final LHSGeneratorContext gctx = generatorContextFactory.newChildGeneratorContext(rootContext,
-                                                                                              pattern);
             preGeneratePattern(gctx);
 
             generateFactPattern(pattern,
@@ -630,19 +639,25 @@ public class RuleModelDRLPersistenceImpl
             }
 
             postGeneratePattern(gctx);
-            buf.append("\n");
+            if (!isSubPattern) {
+                buf.append("\n");
+            }
         }
 
         public void visitFreeFormLine(final FreeFormLine ffl) {
             visitFreeFormLine(ffl,
-                              generatorContextFactory.getMaximumPatternDepth() > 1);
+                              rootContext);
         }
 
-        public void visitFreeFormLine(final FreeFormLine ffl,
-                                      final boolean isSubPattern) {
+        protected void visitFreeFormLine(final FreeFormLine ffl,
+                                         final LHSGeneratorContext parentContext) {
             if (ffl.getText() == null) {
                 return;
             }
+
+            final LHSGeneratorContext gctx = generatorContextFactory.newPeerGeneratorContext(parentContext, ffl);
+            final boolean isSubPattern = gctx.getDepth() > 0;
+
             String[] lines = ffl.getText().split("\\n|\\r\\n");
             for (String line : lines) {
                 this.buf.append(indentation);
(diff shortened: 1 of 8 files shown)


Rikkola (approved):

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -634,22 +634,34 @@ public void visitFactPattern(final FactPattern pattern) {
         }
 
         public void visitFreeFormLine(final FreeFormLine ffl) {
+            visitFreeFormLine(ffl,
+                              generatorContextFactory.getMaximumPatternDepth() > 1);
+        }
+
+        public void visitFreeFormLine(final FreeFormLine ffl,
```
jomarko: Please do not introduce new public api method as there is single occurence of it, in the same class.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
```diff
@@ -4763,4 +4763,160 @@ public void testRHSDateInsertActionWithoutSystemProperty() {
             }
         }
     }
+
+    @Test
+    public void testFromCompositeFactPatternWithDSLAlthoughModelHasNoDSLSentences() {
+        final RuleModel model = new RuleModel() {
+            @Override
+            public boolean hasDSLSentences() {
+                return true;
```
jomarko, line 4910: Does this `return true` means we are saving `rdslr` file? I though it means the rules combine drools and dsl sentences, however I was wrong probably. 

Can we please add test where a `from` expression and DSL sentence is present? Or do we have such test please?

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
manstis: OK.. the single parameter form of these methods (`visitXYZ`) are called by a relection-based visitor however the two parameter versions are only called _internally_ but _could_ be overridden so I've made them all `protected`. I hope that is more acceptable.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
manstis: It replicates the state of validating a model when it was created as a `rdslr` file but has no actual DSL elements (as per https://issues.redhat.com/browse/RHDM-1189). Validation is called [here](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/main/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImpl.java#L318) and the source obtained [here](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/main/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImpl.java#L341). `RuleModelWrapper` constructor set to `true` when the file is ` .rdslr`. It is used by `RuleModelWrapper` as the `hasDSLSentences` [here](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/main/java/org/drools/workbench/screens/guided/rule/backend/server/RuleModelWrapper.java#L225).

Let me check about the test (or add one)....

manstis: @jomarko New tests added.. what a can of worms that opened.. but good they were added as it highlighted a bug (not only relating to my changes but also existing tests that contained `from`).

manstis: [This](https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L696) requirement was not being adhered to in all scenarios.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -630,33 +639,54 @@ public void visitFactPattern(final FactPattern pattern) {
...
+
+            final LHSGeneratorContext gctx = generatorContextFactory.newPeerGeneratorContext(parentContext, ffl);
+            final boolean isSubPattern = gctx.getDepth() > 0;
+
             String[] lines = ffl.getText().split("\\n|\\r\\n");
             for (String line : lines) {
                 this.buf.append(indentation);
-                if (isDSLEnhanced) {
+                if (!isSubPattern && isDSLEnhanced) {
                     buf.append(">");
```
jomarko, line 665: When I grep  `RuleModelDRLPersistenceImpl` class for occurrences of `.append(">");` code snippet, I am now getting two kinds of occurrences:
- ">" is add/not added according variables `isSubPattern` and `isDSLEnhanced`
- ">" is add/not added according variable `isDSLEnhanced`

Shouldn't be ">" added/not added always according to both variables?

manstis: @jomarko You are probably correct.. I can't think of any reason for it to be different.

I've added a new commit to ensure use of `isSubPattern` is used consistently in `LHSPatternVisitor`. Some occurrences were in `RHSActionVisitor` however this has no concept of _sub pattern_ and hence can remain unchanged.

jomarko (approved):
@manstis thank you, perfect coverage!

PR #2730, opened 2020-01-21
[DROOLS-4924] LambdaConsequence without bind variable is not material…

tkobayas:
…ized

In a failing case, the ancestor is "D.execute" instead of "execute". Thus, fixed like this.

Note: I also tried to add a unit test in ExecModelLambdaPostProcessorTest but I cannot reproduce the issue because ExecModelLambdaPostProcessorTest approach (simply using StaticJavaParser.parseResource()) doesn't reproduce the problem (= "D.execute").

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
@@ -248,7 +248,7 @@ public class Consequence {
 
 
         boolean requireDrools = rewriteRHS(ruleVariablesBlock, ruleConsequence);
-        MethodCallExpr executeCall = new MethodCallExpr(onCall, onCall == null ? "D." + EXECUTE_CALL : EXECUTE_CALL);
+        MethodCallExpr executeCall = new MethodCallExpr(onCall != null ? onCall : new NameExpr("D"), EXECUTE_CALL);
         LambdaExpr executeLambda = new LambdaExpr();
         executeCall.addArgument(executeLambda);
         executeLambda.setEnclosingParameters(true);
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequence.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequence.java
@@ -49,7 +49,7 @@ public class MaterializedLambdaConsequence extends MaterializedLambda {
         }
 
         MethodDeclaration methodDeclaration = classDeclaration.addMethod("execute", Modifier.Keyword.PUBLIC);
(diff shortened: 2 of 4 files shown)


lucamolteni:
@tkobayas I see what you did there. 

When I did the lambda materialization I wondered, can we solve it once and for all and align all the code generation towards one among the twos: `execute` or `D.execute`? 

In that way we don't have to check twice. What do you think?

Also @mariofusco what do you think?

lucamolteni (commented):
See https://github.com/kiegroup/drools/pull/2730#issuecomment-576718490

tkobayas:
@lucamolteni Sure, I'll investigate code generation and look for a solution.

tkobayas:
Updated the PR. See comments in codes. WDYT? @lucamolteni

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
```diff
@@ -248,7 +248,7 @@ private MethodCallExpr executeCall(BlockStmt ruleVariablesBlock, BlockStmt ruleC
 
 
         boolean requireDrools = rewriteRHS(ruleVariablesBlock, ruleConsequence);
-        MethodCallExpr executeCall = new MethodCallExpr(onCall, onCall == null ? "D." + EXECUTE_CALL : EXECUTE_CALL);
+        MethodCallExpr executeCall = new MethodCallExpr(onCall != null ? onCall : new NameExpr("D"), EXECUTE_CALL);
```
tkobayas, line 251: This creates a standard construct (as I see in case of simple StaticJavaParser.parseResource()) of "D" and "execute" so we will see only "execute" afterwards.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequence.java
```diff
@@ -49,7 +49,7 @@ void createMethodDeclaration(EnumDeclaration classDeclaration) {
         }
 
         MethodDeclaration methodDeclaration = classDeclaration.addMethod("execute", Modifier.Keyword.PUBLIC);
-        methodDeclaration.setThrownExceptions(NodeList.nodeList(parseClassOrInterfaceType("Exception")));
+        methodDeclaration.setThrownExceptions(NodeList.nodeList(parseClassOrInterfaceType("java.lang.Exception")));
```
tkobayas, line 52: This change is required to pass AmbiguousExceptionTest in drools-test-coverage/test-suite/.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
lucamolteni: Perfect thank you @tkobayas

lucamolteni (approved):

baldimir (approved):

PR #2732, opened 2020-01-23
DROOLS-4954 DMN v1.3 instanceof new syntax support

tarilabs:
@etirelli @baldimir @jiripetrlik @hellowdan 
/cc @evacchi 

This is to support the new syntax, we can squash commits at merge time, I believe can be helpful if need to review this change "progressively"

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
@@ -71,8 +71,10 @@ import org.kie.dmn.core.util.Msg;
 import org.kie.dmn.core.util.MsgUtil;
 import org.kie.dmn.feel.lang.FEELProfile;
 import org.kie.dmn.feel.lang.Type;
+import org.kie.dmn.feel.lang.impl.MapBackedType;
 import org.kie.dmn.feel.lang.types.AliasFEELType;
 import org.kie.dmn.feel.lang.types.BuiltInType;
+import org.kie.dmn.feel.lang.types.GenListType;
 import org.kie.dmn.feel.runtime.UnaryTest;
 import org.kie.dmn.feel.util.Either;
 import org.kie.dmn.model.api.DMNElementReference;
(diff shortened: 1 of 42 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -603,11 +608,17 @@ private DMNType buildTypeDef(DMNCompilerContext ctx, DMNModelImpl dmnModel, DMNN
...
             }
+            Map<String, Type> fieldsAsFEEL = new HashMap<>();
             for (ItemDefinition fieldDef : itemDef.getItemComponent()) {
                 DMNCompilerHelper.checkVariableName(dmnModel, fieldDef, fieldDef.getName());
                 DMNType fieldType = buildTypeDef(ctx, dmnModel, node, fieldDef, false);
                 fieldType = fieldType != null ? fieldType : dmnModel.getTypeRegistry().unknown();
                 compType.addField(fieldDef.getName(), fieldType);
+                fieldsAsFEEL.put(fieldDef.getName(), ((BaseDMNTypeImpl) fieldType).getFeelType());
+            }
+            type.setFeelType(new MapBackedType(type.getName(), fieldsAsFEEL));
```
baldimir, line 619: Nitpicking - it might be better to call this in an else part of the next if statement. Currently .setFeelType can be called twice in some situations. What I mean is: 

```
if (type.isCollection()) {
    type.setFeelType(new GenListType(type.getFeelType()));
} else {
    type.setFeelType(new MapBackedType(type.getName(), fieldsAsFEEL));
}
```

baldimir (approved):
Just one minor comment, therefore approving.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
baldimir: I missed that it is getting wrapped. Ignore this comment.

tarilabs: 👍

evacchi (approved):
it seems ok to me

PR #2733, opened 2020-01-24
[DO-NOT-MERGE] RHDM-1145: [DMN Designer] Java Date field is not converted to DMN date type

danielzhe:
RHDM-1145: [DMN Designer] Java Date field is not converted to DMN date type
DROOLS-4977: [DMN Designer] Refactor DataObjectsServiceImpl to not duplicate logic from ParseHelper

Part of an ensemble. 
PR: https://github.com/kiegroup/kie-wb-common/pull/3124
PR: https://github.com/kiegroup/drools/pull/2733

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/ParserHelper.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/parser/feel11/ParserHelper.java
@@ -24,6 +24,7 @@ import java.time.OffsetDateTime;
 import java.time.OffsetTime;
 import java.time.ZonedDateTime;
 import java.time.chrono.ChronoPeriod;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Stack;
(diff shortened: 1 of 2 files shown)


tarilabs (changes requested):
Sorry but I believe just changing in the parser helper won't do.
The actual implementation of the engine must be able to support these new literal instantiations, in operations.
Feel free to either proceed for implementation, or raise a jira for this RFE.

danielzhe:
@tarilabs Hi! I didn't get it. I'm just returning 'BuiltInType.DATE_TIME' if type of class is 'Date' and 'BuiltInType.STRING' if the type is 'Character'. What else needs to be changed?

tarilabs:
@danielzhe no problem: your solution would solve Parsing stage, but actually supplying such a Java Pojo while evaluating a DMN model will break (actually for the string change might work ootb out of chance, but I'm quite realistic to believe won't work for the `java.util.Date` case).
The part of the code which you touched is in charge of parsing, compiling a DMN model.
What is entirely missing, and that won't be a trivial part, is to implement evaluation support for the new literal instantiations that you are proposing.
That's why I mentioned that if this change is required or RFE based, would be best to have a specific jira for the engine part, so it can be prioritized accordingly.

danielzhe:
@tarilabs Humm... Thanks for the clarification! So how can we proceed now, @jomarko ? This comes from a Jira that you opened (https://issues.redhat.com/browse/RHDM-1145). Wdyt @manstis ? We talked about this on the chat.

manstis:
@danielzhe The simplest thing to do is raise a JIRA for @tarilabs requesting the change to support the issues @jomarko found testing the UI's ability to convert Java classes to DMN types (i.e. support for `Character` and `Date`) and set it as a blocker for this JIRA/PR.

If you pursue the alternative approach of making the changes in our existing _mapping_ class we run the risk of the User authoring something (converting Java classes to DMN types) that is incompatible with runtime semantics).

danielzhe:
@tarilabs @manstis @jomarko Created: https://issues.redhat.com/browse/DROOLS-4983
Let me know if there's any question. Thank you!

jomarko:
@danielzhe thank you for your ensemble. Reading the comments, it sems the conversion support of DMN Date and Character is not usable without DROOLS-4983. My question is, should we move RHDM-1145 out of 7.7 target? Drools-4983 seems quite extensive to me.

danielzhe:
@jomarko You are right, but I don't think it will be necessary to move out of 7.7 since Matteo already opened the PR for DROOLS-4983: https://github.com/kiegroup/drools/pull/2735 
🚀

jomarko:
@danielzhe please resolve conflicts

danielzhe:
@jomarko Closed because this PR is not required anymore thanks to changes made by @tarilabs in [here](https://github.com/kiegroup/drools/pull/2735). Only [this other PR](https://github.com/kiegroup/kie-wb-common/pull/3124) is required now.

PR #2734, opened 2020-01-27
DROOLS-4985 : A Spreadsheet decision tables does not convert to a DRLfile correctly when a Function is being used in the RuleTable constraints

Rikkola:
https://issues.redhat.com/browse/DROOLS-4985

As a bonus this fixes situations where the is only one template key, but the "setter" also has a method call.

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -3125,24 +3125,31 @@ public class RuleModelDRLPersistenceImpl
         //identified as an explicit operation above; normally where the RHS line began with such a call. Therefore it is likely the
         // variable they are modifying was recorded as Free Format DRL and hence the "sets" need to be Free Format DRL too.
         for (Map.Entry<String, List<String>> entry : setStatements.entrySet()) {
-            if (boundParams.containsKey(entry.getKey())) {
-                ActionSetField action = new ActionSetField(entry.getKey());
-                addSettersToAction(entry.getValue(),
-                                   action,
-                                   entry.getKey(),
-                                   boundParams,
-                                   dmo,
-                                   m,
-                                   isJavaDialect);
+            if (boundParams.containsKey(entry.getKey()) && SetStatementValidator.validate(entry.getValue())) {
+                final ActionSetField action = new ActionSetField(entry.getKey());
+                final List<String> setters = entry.getValue();
+                if (setters != null) {
+                    for (final String statement : setters) {
+                        addSetterToAction(action,
+                                          entry.getKey(),
+                                          boundParams,
+                                          dmo,
+                                          m,
+                                          isJavaDialect,
+                                          statement);
+                    }
+                }
                 m.addRhsItem(action,
                              setStatementsPosition.get(entry.getKey()));
             } else {
-                FreeFormLine action = new FreeFormLine();
                 StringBuilder sb = new StringBuilder();
                 for (String setter : entry.getValue()) {
                     sb.append(setter).append("\n");
                 }
-                action.setText(sb.toString());
+                String drl = sb.toString();
+
+                FreeFormLine action = new FreeFormLine();
+                action.setText(drl);
                 m.addRhsItem(action,
                              setStatementsPosition.get(entry.getKey()));
             }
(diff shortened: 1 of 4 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/StatementValidator.java
```diff
@@ -0,0 +1,42 @@
+package org.drools.workbench.models.commons.backend.rule;
```
Rikkola: License!

Rikkola: Done!

Rikkola: Thanks!

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/StatementValidator.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.util.List;
+
+public class StatementValidator {
+
+    public static boolean validate(final String statement) {
```
jomarko: Add javadoc for public methods please

jomarko (commented):
@Rikkola the described scenario from DROOLS-4985 seems to be fixed. However If I run tests in the `drools-wb-dtable-xls-editor-backend` module with your changes -  there are some failures.

If I run `drools-wb-dtable-xls-editor-backend` with `drools` master version (without) your PR. NO test is failing.

I found out this preparing additional coverage https://github.com/kiegroup/drools-wb/pull/1295

Rikkola:
@jomarko getting the same and fixing. Strange that the full build here worked.

Rikkola:
Heh, the unit tests are failing since the tests fails to convert due to the bug that I just fixed. So I need to update the tests, but when opening converted StaticFieldsTable.xls I do get some exception. Looking into that one.

jomarko:
@Rikkola please let me know, if you open additional PR, os I can help you somehow.

Rikkola:
Ok GuidedDecisionTableGeneratorListenerTest is probably picking up actual problems, so fixing that. The StaticFieldsTable.xls we already have in our code base is failing on V&V.

Rikkola:
I think I got it. StaticFieldsTable.xls no longer converts, but the customer reported scenario does.

jomarko:

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/StatementValidatorTest.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.models.commons.backend.rule;
+
+import java.util.ArrayList;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
jomarko: not sure this will pass checkstyle validation.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/StatementValidatorTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class StatementValidatorTest {
+
+    @Test
+    public void twoParameters() {
+        final ArrayList<String> setters = new ArrayList<>();
```
jomarko: It is more standard in java declare like `final List<String> setters = new ArrayList<>();`

jomarko (approved):
Approving in advance as have PTO tomorrow and do not want to block merging.

Manually tried case reported by customer. Works nice.
Also fired my tests https://github.com/kiegroup/drools-wb/pull/1295

Assuming build will be green, +1 for merge.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/StatementValidatorTest.java
Rikkola: Changed.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/StatementValidatorTest.java
Rikkola: Changed.

manstis (approved):

Rikkola:
Just renamed StatementValidator.java  -> SetStatementValidator.java and added some Javadock

Rikkola:
@kiegroup/gatekeepers Hi! Can I get a merge please.

PR #2740, opened 2020-01-30
DROOLS-4990: Test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/24f9ff6fd7073acaa02aa88151d466f97188c836

PR #2741, opened 2020-01-31
DROOLS-4969 DMN v1.3 instanceof support itemDef resolution

tarilabs:
@etirelli @baldimir @jiripetrlik @hellowdan


tarilabs:
Jenkins test again

hellowdan (approved):
I can't see anything critical, but I believe another reviewers should review it because my knowledge about DMN is very limited.

PR #2746, opened 2020-02-04
DROOLS-5007: Test case

martinweiler:


mariofusco:
Reproduced and fixed by https://github.com/kiegroup/drools/commit/40ea8c74db9edc51c0d8635519c237a4b8dc9f34

PR #2751, opened 2020-02-07
[DROOLS-5023] executable model doesn't fire when Or with trimmed dupl…

tkobayas:
…icate constraints

- Unit Test only

testTrimmedConstraint() shows the main issue that AlphaNode is not shared.
testOrWithTrimmedConstraint() is the rule mentioned in the JIRA. It's caused by the non-sharing. But it also produces a wrong outcome (you may comment out the assertEquals for AlphaNode check and let ksession run). So it's not only a performance issue.


tkobayas:
Fixed by https://github.com/kiegroup/drools/commit/68529adf00ad2fb22dd1b9d4a95a3f8a15c72a66

PR #2756, opened 2020-02-10
RHPAM-2554 - Require the ability to define mandatory process /case va…

mswiderski:
…riables that can only be changed by Admin

depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/422


mswiderski:
Jenkins retest this

PR #2757, opened 2020-02-10
Drools 4933 - Setup PMML refactoring implementation

gitgabrio:
See https://issues.redhat.com/browse/DROOLS-4933

This PR mainly setup the overall architecture of the new module, with the planned submodules.
It also provide initial implementation for our internal models (KiePMMLModel...) and the runtime/compiler.
Inside the kie-pmml-refactoring module there are some images and file to explain the planned architecture.

It does not provide any model-specific implementation - so basically it does not do anything.
Last, I've add a couple of helping test class in the kie-test-util module.


kie-pmml-refactoring/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.drools</groupId>
+    <artifactId>drools</artifactId>
+    <version>7.33.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>kie-pmml-refactoring</artifactId>
```
jiripetrlik: I would suggest to not use word refactoring in GAV. I know this module is not part of KIE build yet. But if it is merged and added into Drools parent pom as module these GAVs will be exported to Maven central etc. I would suggest to choose something which can be published from the start.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/KiePMMLExtension.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.kie.pmml.api.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * @see <a href=http://dmg.org/pmml/v4-4/GeneralStructure.html#xsdElement_Extension>Extension</a>
+ */
+public class KiePMMLExtension implements Serializable {
```
jiripetrlik: Do we need this class to be serializable. Is there some use case which requires serialization. There is already different serialization implemented in KIE.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
```diff
@@ -0,0 +1,59 @@
...
+        super(name);
+        this.extensions = extensions;
+    }
+
+    public List<KiePMMLExtension> getExtensions() {
+        return extensions;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
jiripetrlik: Do we expect to compare objects of this class somewhere?

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLTerm.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class KiePMMLTerm extends KiePMMLNamed {
+
+    private static final long serialVersionUID = 1452775408881599004L;
+
+    public KiePMMLTerm(String name) {
+        super(name);
+    }
+
+    public abstract Number getCoefficient();
+
```
jiripetrlik: Please remove unnecessary new lines.

kie-pmml-refactoring/kie-pmml-api/src/test/java/org/kie/pmml/api/model/TestUtils.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.kie.pmml.api.model;
+
+import java.util.Optional;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class TestUtils {
```
jiripetrlik: I would suggest to do not add any test utils to API.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/KiePMMLStatusHolder.java
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.pmml.api;
+
+/**
+ * Class used inside drools. Rules are fired based on the value of status
```
danielezonca: If this is used only by Drools implementation it should be in a implementation module and not here in the common API. Wdyt?

kie-pmml-refactoring/kie-pmml-integration-tests/pom.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jiripetrlik: I think we do not need any integration tests in architecture which you proposed. I expect there will be unit tests for models in kie-pmml-models module. Integration tests maybe for example on the level of kie-server. I would suggest to stay with only unit tests in new PMML implementation.

It is also much easier to measure test coverage when there are only unit tests located in proper modules with implementation.

kie-pmml-refactoring/ToDo.md
```diff
@@ -0,0 +1,8 @@
+TO-DOs
```
danielezonca: I think it should be better to track TODOs with a Jira epic or similar instead of a MD file here. Wdyt?

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    </dependency>
+  </dependencies>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- PMML -->
+      <dependency>
+        <groupId>org.drools</groupId>
+        <artifactId>kie-pmml-api</artifactId>
+        <version>${version.org.kie}</version>
```
jiripetrlik: Manage version of this artifact in the parent or even better in kie-parent.

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+      </dependency>
+      <!-- EXTERNAL -->
+      <!-- TEST -->
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-test-util</artifactId>
+        <version>${version.org.kie}</version>
```
jiripetrlik: Isn't the version of kie-test-util managed in kie-parent. If not I think we should add it there.

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/src/test/java/org/kie/pmml/library/testutils/TestUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import javax.xml.bind.JAXBException;
+
+import org.dmg.pmml.PMML;
+import org.kie.test.util.filesystem.FileUtils;
+import org.xml.sax.SAXException;
+
```
jiripetrlik: Please remove unnecessary new lines.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/functionals/FunctionUtils.java
```diff
@@ -0,0 +1,92 @@
+/*
```
jiripetrlik: I think it is not a proper place to add definition of functional interfaces here. API should serve for methods like loadModel, verifyModel, evaluateModel etc. and not for extending Java 8 implementation. If we really need these utils and interfaces we should move it to some "*common*" module.

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+  <artifactId>kie-pmml-library-commons</artifactId>
+
+  <name>KIE :: PMML :: Library :: Commons </name>
+  <description>Kie PMML Commons</description>
+
+  <dependencies>
+    <!-- PMML -->
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>kie-pmml-library-api</artifactId>
```
jiripetrlik: Do we need a separate API for library. Can't we merge it with kie-pmml-api. It would be much easier for user to have only one API. Many API modules are usually source of mistakes and confusion.

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-compiler/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+      <dependency>
+        <groupId>org.drools</groupId>
+        <artifactId>kie-pmml-library-api</artifactId>
+        <classifier>tests</classifier>
+        <scope>test</scope>
+        <version>${version.org.kie}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
```
jiripetrlik: Remove new lines.

kie-pmml-refactoring/kie-pmml-runtime/kie-pmml-runtime-core/src/main/java/org/kie/pmml/runtime/core/utils/Converter.java
```diff
@@ -0,0 +1,44 @@
...
+     * @return
+     */
+    public static Map<String, Object> getUnwrappedParametersMap(Map<String, ParameterInfo> parameterMap) {
+        return parameterMap.entrySet()
+                .stream()
+                .collect(Collectors.toMap(Map.Entry::getKey,
+                                          e -> e.getValue().getValue()));
+
+    }
+
```
jiripetrlik: Please remove newlines.

kie-test-util/src/test/java/org/kie/test/util/filesystem/ResourceHelperTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+import static org.kie.test.util.filesystem.ResourceHelper.getResourcesByExtension;
+import static org.kie.test.util.filesystem.ResourceHelper.getResourcesFromDirectory;
+
+public class ResourceHelperTest {
+
+    private static final String TEST_FILE = "TestFile.txt";
+
```
jiripetrlik, line 37: Please remove newlines.

jiripetrlik (commented):
Hello @gitgabrio thank you for your PR with PMML architecture! I have some comments for discussion.

kie-pmml-refactoring/pom.xml
gitgabrio: @jiripetrlik 
Hi! @danielezonca asked me to name it that way. Anyway, I would prefer not to spend too much time in such kind of details right now. If the new and old modules will have to live together (that I would prefer not) we will address this. In the meantime, I did not add kie-pmml-refactoring to parent pom exactly for this reason.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
gitgabrio: @jiripetrlik 
Hi! All "equals" method of child classes have this snippet
`  if (!super.equals(o)) {
            return false;
        }`
 - so implementing "equals" on abstract/parent classes allow "equals" on children to also considering inherited fields.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/functionals/FunctionUtils.java
gitgabrio: @jiripetrlik 
Hi! I may rename the "api" module to "common" - anyway I would like to avoid proliferation of unneeded modules: if I follow your logic, in that top-level module I should create a "common" package for shared methods/utils, a "model" package" for models, and an api module for interfaces that has to be implemented - and all of them should always -anyway - be declared as dependency together.

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
gitgabrio: @jiripetrlik 
Hi! Let's consider for main group of modules
1) kie-pmml-refactoring
2) kie-pmml-library
3) kie-pmml-runtime
4) kie-pmml-models-(model_name)

Each of this need an api/common/whatever shared package to allow decoupling of implementations - e.g. kie-pmml-models-regression-api will define some "model" that must to be visible to both kie-pmml-models-regression-library and kie-pmml-models-regression-runtime, without having one the them to depend on the other. Recursively, same logic apply to all other group of modules. Does this make sense?

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
jiripetrlik: Ok, but the question is why you need to compare children of this class? What is the example or use case for it?

kie-pmml-refactoring/pom.xml
jiripetrlik: I'm ok with it. I have no strong opinion about it. We only need to make sure that we do not publish the artifacts on Maven central etc., because it may confuse users in the future.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/functionals/FunctionUtils.java
jiripetrlik: I think API usually contains only interfaces which are supposed to be used by user. It is usually very stable and there should be no breaking changes with new versions. In PMML evaluator I would expect only methods or classes related directly to evalution of PMML in API. Such as loadPMML, evaluatePMML etc.

It is not good to use API to share the "implementation" with other modules. Users may start using these shared utils methods and it is not possible to change them in the future because of breaking contract. Utils classes etc. which are not supposed to be used directly by users are usually placed in commons modules. But maybe we can find some decomposition where these "common" modules will not be necessary.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/annotations/Nullable.java
```diff
@@ -0,0 +1,45 @@
...
+ * <p>
+ * This is a marker annotation and it has no specific attributes.
+ *
+ * @paramDoc This value may be {@code null}.
+ * @returnDoc This value may be {@code null}.
+ * @hide
+ */
+@Retention(SOURCE)
+@Target({METHOD, PARAMETER, FIELD})
+public @interface Nullable {
```
danielezonca: Can you please remove it? It is not used

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
gitgabrio: @jiripetrlik 
This is part of models that will be put in set/map/list - so having the equals/hashmap will avoid clashing of instances.
Beside that, IIRC it was suggested by Sonar. Last, I think it is a good practice to add such methods to DTOs because it add predictable behavior.

kie-pmml-refactoring/pom.xml
gitgabrio: @jiripetrlik 
agreed with @danielezonca and @mariofusco to call it "kie-pmml-new"

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/KiePMMLExtension.java
gitgabrio: This class is contained in models that - in turns - will be serialized during kjar compilation (not implemented yet). At the same time, this module (kie-pmml-new) should not depend in anyway from something more "drools/kie" specific - that's why I've used java standard "Serializable" interface

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/exceptions/KiePMMLException.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.pmml.api.exceptions;
+
+public class KiePMMLException extends Exception {
```
danielezonca: Why `Exception` instead of `RuntimeException`? This change has as consequence a sort of "duplication" of many Java 8 interfaces (`Stream` and `Function` as first).
If your goal is to enforce user to manage the errors I would suggest to use one/more RuntimeException internally and then do a try/catch at public method level and generate the checked exception only there. In this way you should be able to avoid interface duplication. Wdyt?

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
danielezonca: @jiripetrlik 
All these models/pojos are defined to represent XML content and should be not changed. Implementing `equals`/`hashCode` give you the possibility to manage them as "values" easier (for example compare or add them to a `Set`).
I don't think there is a specific use case for now because this PR is just a "stub" of the design that will be extended later with multiple model implementations so it is safer to already implement a meaningful equals.
Does this answer to your question?

kie-pmml-refactoring/kie-pmml-api/src/test/java/org/kie/pmml/api/model/TestUtils.java
danielezonca: I think this module is more intended to be a -common more than API so I would suggest to rename it to -common.
@gitgabrio wdyt?

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-compiler/src/test/resources/LinearRegressionSample.xml
```diff
@@ -0,0 +1,26 @@
+<PMML xmlns="http://www.dmg.org/PMML-4_2" version="4.2">
```
danielezonca: Isn't it a copy of [this](https://github.com/kiegroup/drools/blob/1073fc4f1da4d51c1d54b7f003f189699fc62e6d/kie-pmml-refactoring/kie-pmml-integration-tests/src/test/resources/LinearRegressionSample.xml) ?

kie-pmml-refactoring/kie-pmml-runtime/kie-pmml-runtime-api/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>kie-pmml-runtime</artifactId>
+    <groupId>org.drools</groupId>
+    <version>7.33.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>kie-pmml-runtime-api</artifactId>
```
danielezonca: I don't like the name `runtime` to refer to a sort of "kie-integration" module. Wdyt?

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
jiripetrlik: I'm not sure whether this is not too complex for what we want to achieve. Which is basically PMML evaluator. It sounds like we will end up with for example 40 or more modules. Which is quite much I think.
- PMML itself defines about 20 models. If we want implementation and interface for each module it is about 40 modules.
- Other modules such as library, runtime etc. It will probably be at lease 5 modules.
If you want more implementations for modules than it will be even more modules. It will be probably impossible or very hard to maintain. Especially it might be very hard for user to get oriented in it.

kie-pmml-refactoring/kie-pmml-api/src/test/java/org/kie/pmml/api/model/TestUtils.java
gitgabrio: @danielezonca @jiripetrlik 
Renamed to kie-pmml-commons -as it is actually used for

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/functionals/FunctionUtils.java
gitgabrio: @danielezonca @jiripetrlik
Renamed to kie-pmml-commons -as it is actually used for

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
gitgabrio: @jiripetrlik 
Each model may have a different optimal way to be implemented - e.g. I'm implementing Regression in pure java, while Tree will be done with drools. As such, each model has a different dependency requirement. 
At the same time, we want to avoid to create a can of worms where one model depends to a lot of unneeded others only for transitive dependencies. 
That lead to a design that is easy to understand, even if complex:

- library  (we have to rename it) is responsible to translation of original PMML to "internal" model; 
- runtime is responsible to actual evaluation.

Every model will have its own implementation of both, with its own internal model definition.
Encapsulating every model in specific module, from my point of view, allow and enforce the developer to focus *only* in the implementation is currently doing **regardless** the "abstraction" layer - that as you see is pretty thin.
PMML is an "aggregator" of different models, and the complexity of implementation reflect that. It is -more or less - like put all other models (scorecard, guideddecisions, whatever) under the same roof.
The biggest mistake -in my opinion - of the previous implementation was exactly to have (tried to) managed all of them in one single way.
Last note: I've already implemented an archetype (not yet commit) exactly to help developer start create a new model implementation.

kie-pmml-refactoring/ToDo.md
gitgabrio: @danielezonca 
Removed

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/abstracts/KiePMMLExtensionedNamed.java
jiripetrlik: I'm probably fine with that. I still don't see your point why to define it without use case. But we can keep it there just for safety. Do you plan to use it as keys in HashMaps? Or why to store XMLs in sets? But I'm ok with it to keep it as it is.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/model/KiePMMLExtension.java
jiripetrlik: Thank you for explanation!

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
jiripetrlik: @gitgabrio 
I see your point. However consider the number of modules (40 or more) which need to be managed. Each module effectively means one artifact to publish etc. And it will be quite hard for user to figure out which artifact add to his dependencies when.

Of course we can and probably will have to do some "aggregating" module which will transitively collect all modules as dependencies, but it will effectively end up to adding all modules implementation and their transitive dependencies (for example Drools) to one classloader anyway. So we will effectively gain nothing from separate modules. Also consider that PMML need to work also as part of kie server and user will not tune jars which are present in WEB-INF/lib himself. Again we will probably end up with adding everything to classpath.

Maybe we can find some compromise between complexity and having separate modules. @gitgabrio @danielezonca @mariofusco Do you see any compromise or less complex approach?

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/KiePMMLStatusHolder.java
gitgabrio: @danielezonca 
Moved to "drooled" module

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/annotations/Nullable.java
gitgabrio: @danielezonca 
done

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-compiler/src/test/resources/LinearRegressionSample.xml
gitgabrio: @danielezonca 
Yes

kie-pmml-refactoring/kie-pmml-runtime/kie-pmml-runtime-api/pom.xml
gitgabrio: @danielezonca @jiripetrlik @mariofusco 

We have broadly two main functionalities/components:
1) translation from original PMML to internal model (to be used for actual evaluation)
2) actual execution of the evaluation inside the kie container (i.e. submit user input to retrieve predicted value).

Currently the first one is called "library", but I think "compiler" would be more appropriate; for the latter, beside "runtime", I do not have any special idea.
Anyway, let's just decide all together those names once and for all, because I have three branch to keep in sync and such renaming are pretty disruptive.
What's your suggestion for them?

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-commons/pom.xml
gitgabrio: @jiripetrlik @danielezonca @mariofusco 
It may well be that in future we'll see a lot of unneeded complexity. 
If - for ex, all modules will be implemented only in two ways, probably it will make sense to simplify collapsing them. But, IMO, all this is "optimization", that I would like to postpone. For the moment being, this "complexity" enforce some clear boundaries and avoid mixing up things - it is easy to remove such separations in future, but it is extremely hard to do the other way around

jiripetrlik: @gitgabrio 
Agree. It is probably preliminary optimization. Let's continue with more modules for now and optimize in the future.

kie-pmml-refactoring/kie-pmml-integration-tests/pom.xml
gitgabrio: @jiripetrlik 
TBH I think such tests are needed, **beside** unit tests.
One of the main reason is that actual implementations are not hardcoded, but dinamically retrieved at runtime. So, only firing up the whole environment it is possible to verify that all components are found in proper place when needed. Does this make sense?

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
gitgabrio: @jiripetrlik 
All those dependency management definitions are only transitory, since as soon as the module will be merged they will be defined in kie-parent

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
gitgabrio: @jiripetrlik 
It seems needed. See answer to above question

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
gitgabrio: @jiripetrlik 
... anyway, for the moment being I've put all of them (the management only) inside parent pom

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/exceptions/KiePMMLException.java
gitgabrio: @danielezonca 
I'm not convinced of that. I understand your point, but at the same time it is not clear where such boundaries should be drawn. Putting them too "deep" would not change that much, but putting them too "high" would vanify the scope of the checked exception

gitgabrio:
jenkins please retest this

gitgabrio:
Hi guys! I've done all the renaming/refactoring we have talked about, and managed all (but one) of your proposals. Could you please take a look again? 
Of course you will see a lot of file changed, but it is just because they have been moved

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/exceptions/KiePMMLException.java
danielezonca: I would say that the boundary should be public (= end-user) API. My proposal is to limit this checked exception at `PMMLRuntime` level. It should be enough to try/catch runtime exceptions in `PMMLRuntimeImpl` and wrap them in the checked exception. Wdyt?

EDIT: let me add that I'm not a fan of checked exception so I would like to remove them completely but I'm trying to preserve them without add too much overhead (= interface duplication)

kie-pmml-refactoring/kie-pmml-library/kie-pmml-library-api/pom.xml
jiripetrlik: @gitgabrio 
Yes agree. It is probably the best place for dependency management during refactoring.

kie-pmml-refactoring/kie-pmml-api/src/main/java/org/kie/pmml/api/exceptions/KiePMMLException.java
jiripetrlik: @gitgabrio @danielezonca 
I have no strong opinion here. But I would also vote for avoiding of checked exceptions, because I also don't like them.

For example if I recall correctly it was also not recommended in Clean Code book, because: 
Using checked exceptions comes with a cost - it breaks the open-closed principle: You have to declare the checked exception (in throws clause) in signatures of all methods between the throwing and catching the exception.

But I'm not sure. As I've said I have no strong opinion here.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/exceptions/ExternalException.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.pmml.commons.exceptions;
+
+/**
+ * Checked Exceptions wrapping all <b>not-KiePMML</b> ones at <i>customer</i> API boundaries
+ */
+public class ExternalException extends Exception {
```
jiripetrlik: Can we change also this to RuntimeException?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/exceptions/KiePMMLException.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.pmml.commons.exceptions;
+
+/**
+ * Checked Exceptions to be wrapping to <b>unchecked</b> ones at <i>customer</i> API boundaries
+ */
+public class KiePMMLException extends Exception {
```
jiripetrlik: Can we change also this to RuntimeException?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/functionals/FunctionUtils.java
```diff
@@ -0,0 +1,85 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.pmml.commons.functionals;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class FunctionUtils {
```
jiripetrlik: I'm not sure about adding these functional utils into code. Can't we use some library instead?

@mariofusco 
What do you think? Can you help us to find some library with these functional utils?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/INVALID_VALUE_TREATMENT_METHOD.java
```diff
@@ -0,0 +1,48 @@
...
+    MIXED("mixed");
+
+    private String name;
+
+    INVALID_VALUE_TREATMENT_METHOD(String name) {
+        this.name = name;
+    }
+
+    public static INVALID_VALUE_TREATMENT_METHOD byName(String name) throws KieEnumException {
+        return Arrays.stream(INVALID_VALUE_TREATMENT_METHOD.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find MINING_FUNCTION with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/FIELD_USAGE_TYPE.java
```diff
@@ -0,0 +1,49 @@
...
+    ANALYSIS_WEIGHT("analysisWeight");
+
+    private String name;
+
+    FIELD_USAGE_TYPE(String name) {
+        this.name = name;
+    }
+
+    public static FIELD_USAGE_TYPE byName(String name) throws KieEnumException {
+        return Arrays.stream(FIELD_USAGE_TYPE.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find FIELD_USAGE_TYPE with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/MINING_FUNCTION.java
```diff
@@ -0,0 +1,48 @@
...
+    MIXED("mixed");
+
+    private String name;
+
+    MINING_FUNCTION(String name) {
+        this.name = name;
+    }
+
+    public static MINING_FUNCTION byName(String name) throws KieEnumException {
+        return Arrays.stream(MINING_FUNCTION.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find MINING_FUNCTION with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/MISSING_VALUE_TREATMENT_METHOD.java
```diff
@@ -0,0 +1,48 @@
...
+    MIXED("mixed");
+
+    private String name;
+
+    MISSING_VALUE_TREATMENT_METHOD(String name) {
+        this.name = name;
+    }
+
+    public static MISSING_VALUE_TREATMENT_METHOD byName(String name) throws KieEnumException {
+        return Arrays.stream(MISSING_VALUE_TREATMENT_METHOD.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find MINING_FUNCTION with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/OP_TYPE.java
```diff
@@ -0,0 +1,44 @@
...
+    CONTINUOUS("continuous");
+
+    private String name;
+
+    OP_TYPE(String name) {
+        this.name = name;
+    }
+
+    public static OP_TYPE byName(String name) throws KieEnumException {
+        return Arrays.stream(OP_TYPE.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find MINING_FUNCTION with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/OUTLIER_TREATMENT_METHOD.java
```diff
@@ -0,0 +1,48 @@
...
+    MIXED("mixed");
+
+    private String name;
+
+    OUTLIER_TREATMENT_METHOD(String name) {
+        this.name = name;
+    }
+
+    public static OUTLIER_TREATMENT_METHOD byName(String name) throws KieEnumException {
+        return Arrays.stream(OUTLIER_TREATMENT_METHOD.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find MINING_FUNCTION with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/PMML_MODEL.java
```diff
@@ -0,0 +1,61 @@
...
+    TREE_MODEL("TreeModel");
+
+    private String name;
+
+    PMML_MODEL(String name) {
+        this.name = name;
+    }
+
+    public static PMML_MODEL byName(String name) throws KieEnumException {
+        return Arrays.stream(PMML_MODEL.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find PMMLModelType with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/VALUE_PROPERTY.java
```diff
@@ -0,0 +1,44 @@
...
+    MISSING("missing");
+
+    private String name;
+
+    VALUE_PROPERTY(String name) {
+        this.name = name;
+    }
+
+    public static VALUE_PROPERTY byName(String name) throws KieEnumException {
+        return Arrays.stream(VALUE_PROPERTY.values()).filter(value -> name.equals(value.name)).findFirst().orElseThrow(() -> new KieEnumException("Failed to find VALUE_PROPERTY with name: " + name));
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/factories/KiePMMLExtensionFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import org.dmg.pmml.Extension;
+import org.kie.pmml.commons.model.KiePMMLExtension;
+
+public class KiePMMLExtensionFactory {
+
+    private KiePMMLExtensionFactory() {
+    }
+
+    public static List<KiePMMLExtension> getKiePMMLExtensions(List<Extension> extensions) {
+        return extensions.stream().map(KiePMMLExtensionFactory::getKiePMMLExtension).collect(Collectors.toList());
```
jiripetrlik, line 30: Can we add test for this method?

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/implementations/KiePMMLModelRetriever.java
```diff
@@ -0,0 +1,56 @@
...
+     * Read the given <code>DataDictionary</code> and <code>Model</code>> to returns a <code>Optional&lt;KiePMMLModel&gt;</code>
+     * @param dataDictionary
+     * @param model
+     * @param kBuilder Using <code>Object</code> to avoid coupling with drools
+     * @return
+     * @throws KiePMMLInternalException
+     */
+    @SuppressWarnings("unchecked")
+    public static Optional<KiePMMLModel> getFromDataDictionaryAndModel(DataDictionary dataDictionary, Model model, Object kBuilder) throws KiePMMLException {
+        logger.info("getFromModel {}", model);
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/implementations/ModelImplementationProviderFinderImpl.java
```diff
@@ -0,0 +1,56 @@
...
+public class ModelImplementationProviderFinderImpl implements ModelImplementationProviderFinder {
+
+    private static final Logger logger = LoggerFactory.getLogger(ModelImplementationProviderFinderImpl.class.getName());
+
+    private ServiceLoader<ModelImplementationProvider> loader = ServiceLoader.load(ModelImplementationProvider.class);
+
+    @Override
+    @SuppressWarnings("rawtypes")
+    public List<ModelImplementationProvider> getImplementations(boolean refresh) {
+        logger.info("getImplementations {}", refresh);
```
jiripetrlik: Can we add test for this method?

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/utils/ModelUtils.java
```diff
@@ -0,0 +1,43 @@
...
+     * Return an <code>Optional</code> with the name of the field whose <b>usageType</b> is <code>TARGET</code> or <code>PREDICTED</code>
+     * <p>
+     * While the xsd schema does not strictly enforce this, it seems that <b>by convention</b> majority of models has only one target.
+     * <p>
+     * (see https://github.com/jpmml/jpmml-evaluator/issues/64 discussion)
+     * @param model
+     * @return
+     */
+    public static Optional<String> getTargetField(Model model) {
+        return model.getMiningSchema().getMiningFields().stream()
```
jiripetrlik: Can we add test for this method?

jiripetrlik (commented):
Added second round of comments. I have only few comments. But the bigger problem is that there are various util methods in this skeleton of implementation and these methods have zero test coverage. I think if we want to go from POC to ready to merge I think we need to implement tests for these methods.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/exceptions/ExternalException.java
gitgabrio: @jiripetrlik 
Hi. Nope. This is meant exactly to be a checked exception.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/exceptions/KiePMMLException.java
gitgabrio: @jiripetrlik
Hi. Nope. This is meant exactly to be a checked exception.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/INVALID_VALUE_TREATMENT_METHOD.java
gitgabrio: @jiripetrlik 
Hi Jiri. What would you like to test? Those are five basic java calls in chain:
1) ENUM.values
2) Arrays.stream
3) Stream.filter
4) Stream.findFirst
5) Optional.orElseThrow

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/implementations/ModelImplementationProviderFinderImpl.java
gitgabrio: @jiripetrlik 
This would be an "integration" test to ServiceLoader.reload () and ServiceLoader.iterator(). I think it is already covered inside integration test module

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/utils/ModelUtils.java
gitgabrio: Will do

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/implementations/KiePMMLModelRetriever.java
gitgabrio: This would be an "integration" test. Without any real implementation, hard to see what/how to test.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/INVALID_VALUE_TREATMENT_METHOD.java
jiripetrlik: @gitgabrio 
We can add some simple test whether it works as expected.
- positive example - if filter finds something
- negative - if filter does not and exception is thrown

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/functionals/FunctionUtils.java
gitgabrio: @jiripetrlik @mariofusco @danielezonca 
Fine for me if there is already such functionality in the classpath.
Anyway, a couple of things to keep in mind
1) it must not add dependency on any other drools module
2) it is a bad practice to put new libraries inside projects only to use one single feature of such libraries - duplicated code is not bad by itself, if it avoid creating/increasing coupling between otherwise unrelated code
3) Watch carefully about "functional monolith masked behind modular layout"

jiripetrlik: As I wrote. I'm not sure whether something like that exists. I've added it mostly as possible point for discussion.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/INVALID_VALUE_TREATMENT_METHOD.java
gitgabrio: @jiripetrlik 
Hi Jiri. I understand you point, but I'm not fully convinced.
Basically, behind all the "wrapping", the test you are proposing would just verify that `String.equals(String)` behave as expected. It is true that there are a series of method chained together, but each of them is a standard java invocation - the result of which is the argument for the next one. 
IMO, it is the client code that has to be tested in this situation - I mean if a "compiler" call this one, we should verify the "caller" is not invoking it with wrong value: wdyt ?
As a side note, what I'm gonna do is to replace that `name.equals` with `Objects.equals`

jiripetrlik: Agree with you. We will probably test this on different level.

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/utils/ModelUtils.java
gitgabrio: @jiripetrlik 
Hi! Just pushed tests for that class.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/exceptions/ExternalException.java
mariofusco: @gitgabrio Checked exception are an outdated concept, no JVM language other than Java support them and since Java 8 they're practically deprecated also in Java since they play horribly with lambdas. Checked exception are useless verbosity in the best case and plain evil in the worst one. Please remove ALL checked exception from this pull request. Thanks.

mariofusco (changes requested):
Pull request is fine except for abuse of checked exceptions (and yes, even one single checked exception is an abuse). Please remove them all.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/functionals/FunctionUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Returns an <code>HashMap</code> with the given key/value pairs
+     * @param k1
+     * @param v1
+     * @return
+     */
+    public static <K, V> Map<K, V> mapOf(K k1, V v1) {
+        return new HashMap<K, V>() {{
```
mariofusco: Is this useful? Even if it is please change the implementation: you're implicitly creating a subclass of HashMap at each invocation in this way.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/interfaces/FunctionalWrapperFactory.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param <R>
+     * @param <E>
+     * @return
+     * @throws E
+     */
+    public static <T, R, E extends Exception> Function<T, R> throwingFunctionWrapper(ThrowingFunction<T, R, E> throwingFunction) throws E {
+        return t -> {
+            try {
+                return throwingFunction.apply(t);
+            } catch (Exception e) {
```
mariofusco: Demonstration that checked exceptions are bogus.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/interfaces/ThrowingBiConsumer.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.kie.pmml.commons.interfaces;
+
+import java.util.Objects;
+
+/**
+ * Exception-throwing <code>BiConsumer</code>
+ */
+@FunctionalInterface
+public interface ThrowingBiConsumer<T, U, E extends Exception> {
```
mariofusco: At this point I want to die.

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/KiePMMLExtension.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * @see <a href=http://dmg.org/pmml/v4-4/GeneralStructure.html#xsdElement_Extension>Extension</a>
+ */
+public class KiePMMLExtension implements Serializable {
+
+    private static final long serialVersionUID = -5895956619311871049L;
+    private String extender;
```
mariofusco, line 27: Should this be Serializable? If so you need an empty constructor. If not make all these fields final.

gitgabrio: @mariofusco 
Fixed with last commit

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-api/src/main/java/org/kie/pmml/compiler/api/provider/ModelImplementationProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import org.dmg.pmml.Model;
+import org.kie.pmml.commons.exceptions.KiePMMLException;
+import org.kie.pmml.commons.exceptions.KiePMMLInternalException;
+import org.kie.pmml.commons.model.KiePMMLModel;
+import org.kie.pmml.commons.model.enums.PMML_MODEL;
+
+/**
+ * API for actual PMML model implementations
+ */
+public interface ModelImplementationProvider<T extends Model, E extends KiePMMLModel> {
```
mariofusco, line 28: This generic doesn't seem to be useful since you cannot use it on the call site and then obliges you to add those [at]SuppressWarning annotation.

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-commons/src/main/java/org/kie/pmml/compiler/commons/implementations/ModelImplementationProviderFinderImpl.java
```diff
@@ -0,0 +1,56 @@
...
+        logger.debug("getImplementations {}", refresh);
+        List<ModelImplementationProvider> toReturn = new ArrayList<>();
+        Iterator<ModelImplementationProvider> providers = getProviders(refresh);
+        providers.forEachRemaining(toReturn::add);
+        logger.debug("toReturn {} {}", toReturn, toReturn.size());
+        toReturn.forEach(provider -> logger.debug("{} : {}", provider.getPMMLModelType(), provider));
+        return toReturn;
+    }
+
+    @SuppressWarnings("rawtypes")
```
mariofusco, line 51: If you remove the generic on the ModelImplementationProvider as I suggested above you can get rid of this annotation.

mariofusco (approved):

jiripetrlik (approved):
@gitgabrio 
Thank you for all changes you made during the review process! It looks good to me. We will probably need to measure and check test coverage before adding to standard build and during adding new modules.

danielezonca (approved):

kie-pmml-new/kie-pmml-compiler/kie-pmml-compiler-api/src/main/java/org/kie/pmml/compiler/api/provider/ModelImplementationProvider.java
gitgabrio: @mariofusco 
Reason for the generic at interface signature is to allow inside actual implementations  (e.g.):
` public KiePMMLRegressionModel getKiePMMLModel(DataDictionary dataDictionary, RegressionModel model, Object kBuilder) ` (concrete types as returns and parameters)

I could not find a proper/clean way to achieve that putting generic definition to the method itself. Could you please suggest one, if it is doable ? Thx.

gitgabrio:

kie-pmml-new/kie-pmml-commons/src/main/java/org/kie/pmml/commons/model/enums/INVALID_VALUE_TREATMENT_METHOD.java
baldimir: It is recommended that code has sufficient unit test coverage. What Jiri proposes is exactly that. "Hi Jiri. What would you like to test? Those are five basic java calls in chain" - exactly those 5 basic Java calls in a chain should be tested. In future, someone might change this so that it doesn't work correctly. Then we will miss the test. The code should have as big unit test coverage as possible. I know that it is sometimes a pain to write them, but they are useful, although they may not look like that.

PR #2765, opened 2020-02-13
DROOLS-5018 DMN internal refactor for itemDef isCollection (bis)

tarilabs:
@etirelli @baldimir @hellowdan 
/cc @yesamer


yesamer (commented):
@tarilabs Locally tested, it works. Approved, thanks!
Anyway, restoring the behavior we still have the issue with `CompositeTypeImpl.setCollection(boolean b)` method. If called, it leads to a `ClassCastException` when you try to add a field. Considering the issues you spot regarding removing it, I suggest to keep note of the potential issue.
Thanks!

yesamer (approved):

tarilabs:
> Anyway, restoring the behavior we still have the issue with CompositeTypeImpl.setCollection(boolean b) method. If called, it leads to a ClassCastException when you try to add a field.

Not really.
The thing is that if you call `setCollection` after constructor, you are mutating the DMNType implementation, so the calling user is responsible for maintain the coherent status.
But anyway is a fair point, removing the `#setCollection()` method altogether.

hellowdan (approved):

tarilabs:
Jenkins test again

danielezonca (approved):

PR #2767, opened 2020-02-14
Update license file name

yurloc:
"Apache Software License" was the name of the license used in versions 1.x. Version 2.0 is called just the Apache License.

https://opensource.org/licenses/apachepl-1.1.php

I've made this PR to all repositories that have a license file with the obsolete ASL identifier in the file name. According to https://spdx.org/licenses/Apache-2.0.html the short identifier for the Apache License 2.0 is "Apache-2.0".


yurloc:
Jenkins retest this

delsolcrx1:
Missing Apache 2 copyright license. Line 190 https://github.com/kiegroup/drools/blob/master/LICENSE-Apache-2.0.txt

yurloc:
@delsolcrx1 If I understand correctly, you think there is a mistake on line 190 in the license file in this repo. It's not a mistake. At first sight, it seems like there's some information missing, right? It's because that's an appendix of the license file with instructions on how to apply a license notice in source files. If you check again and read the beginning of the appendix, I think you'll agree :slightly_smiling_face:.

For reference, here is the original of the license file: http://www.apache.org/licenses/LICENSE-2.0.txt.

delsolcrx1:
@yurloc if you read the appendix it is a template which you should remove the brackets and specify the owner of the copyright as tell as the year from which the copy right should exist

>       To apply the Apache License to your work, attach the following
>       boilerplate notice, with the fields enclosed by brackets "[]"
>       replaced with your own identifying information. (Don't include
>       the brackets!)

PR #2772, opened 2020-02-19

tkobayas:
…indexing

WIP implementation. Please give a quick review before going too far. Thanks!

Limitation:
- Cannot parse constrains with "&&" (e.g. "10<age && 20>age") so the constraint will be left without normalization. I may add a specific logic for "&&" case but it may have a little performance degration. If you think it's worth trying, let me know.
- Cannot parse constrains which are surrounded by parentheses (e.g. "(10 == age)") so the constraint will be left without normalization. Similarly, I may add a specific logic for that but it may have a little performance degration.

Remaining work:
- I will add more unit tests because it would affect widely.
- I will do some performance test and I may consider tuning if it's slow


tkobayas:
Sorry, missed to make it draft PR. Closing...

PR #2773, opened 2020-02-19

tkobayas:
…indexing

WIP implementation. Please give a quick review before going too far. Thanks!

Limitation:

- Cannot parse constrains with "&&" (e.g. "10<age && 20>age") so the constraint will be left without normalization. I may add a specific logic for "&&" case but it may have a little performance degration. If you think it's worth trying, let me know.
- Cannot parse constrains which are surrounded by parentheses (e.g. "(10 == age)") so the constraint will be left without normalization. Similarly, I may add a specific logic for that but it may have a little performance degration.

Remaining work:

- I will add more unit tests because this change would affect widely.
- I will do some performance test and I may consider tuning if it's slow

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtils.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtils.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.compiler.rule.builder.util;
+
+import java.beans.IntrospectionException;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ConstraintUtils {
+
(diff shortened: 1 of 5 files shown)


tkobayas:
As written in the JIRA, this PR tries to modify a constraint to move a fact property from right side to left side. 

ConstraintUtils.normalizeConstraintExpression() is called in

- PatternBuilder.processConstraintsAndBinds() ... non-executable-model case

- ConstraintExpression.createConstraintExpression() ... executable-model case

Earlier than the point, we don't have type information (so cannot identify which is a fact property). Later than the point, the expression is already parsed to RelationalExprDescr in non-executable-model so it will be more complicated.

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtils.java
```diff
@@ -0,0 +1,91 @@
...
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ConstraintUtils {
+
+    private static final Pattern RELATIONAL_CONSTRAINT = Pattern.compile("^([\\S&&[^&|\\(\\)=!><]]*)\\s*(==|!=|<=|>=|<|>)\\s*([\\S&&[^&|\\(\\)=!><]]*)$");
```
tkobayas, line 29: This regex is meant to be:
- Split into 3 parts
 (1) characters (not space, &, |, (, ), !, =, >, <)
 (2) operator (== or != or <= or >= or < or >)
 (3) characters (not space, &, |, (, ), !, =, >, <)

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtils.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public static final String DROOLS_NORMALIZE_CONSTRAINT = "drools.normalize.constraint";
+
+    private static boolean ENABLE_NORMALIZE = Boolean.valueOf(System.getProperty(DROOLS_NORMALIZE_CONSTRAINT, "true"));
```
tkobayas, line 33: I added this system property switch because this feature will affect widely so I want that users are able to revert back to the original behavior.

tkobayas:
Closing this PR replaced by https://github.com/kiegroup/drools/pull/2788

PR #2777, opened 2020-02-20
[NOISSUE] Fixed loading kjar files to modules

librucha:
If using busines central built files (DMN), which are packed into *.kjar, is not possible to load it into module, because path does contain ".jar!" sequence.


ghost:

PR #2785, opened 2020-02-25
DROOLS-4865 : Add i18n reporting for DMN V&V

Rikkola:
https://issues.redhat.com/browse/DROOLS-4865

https://github.com/kiegroup/kie-wb-common/pull/3183
https://github.com/kiegroup/drools/pull/2785


kie-dmn/kie-dmn-validation/src/test/resources/i18n/AnalyticsMsg_fi.properties
```diff
@@ -0,0 +1,26 @@
+TheValidatorWasUnableToCompileTheEmbeddedDMNValidationRulesValidationOfTheDMNModelCannotBePerformed=No niin
```
manstis, line 1: Yay, I learn some Finnish ;-)

manstis:
@Rikkola Please let me known when you're ready for review (LGTM so far..).

You'll probably want @tarilabs to take a look too when ready....

tarilabs:
Is this really the way i18n is expected to be implemented??
My understanding with @etirelli at the very beginning, is that the message class was more than fine --I believe it was discussed between several team leads.
In essence, I'm a bit surprised about these removals: https://github.com/kiegroup/drools/pull/2785/files#diff-6ead07e1bda6075556230c9b1138cf7eL126
I previously understood it would have required just to replace the "hardcoded" string, while this to me is a totally different design.
Please advise.

kie-dmn/kie-dmn-validation/src/test/java/org/kie/dmn/validation/dtanalysis/DMNDTAnalyserTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        final DecisionTable decisionTable = mock(DecisionTable.class);
+        decisionTables.add(decisionTable);
+        doReturn(HitPolicy.FIRST).when(decisionTable).getHitPolicy();
+
+        doReturn(decisionTables).when(definitions).findAllChildren(DecisionTable.class);
+        final List<DTAnalysis> analyse = dmndtAnalyser.analyse(model);
+        final List<DMNMessage> dmnMessages = analyse.get(0).asDMNMessages();
+        assertFalse(dmnMessages.isEmpty());
+        assertTrue(containsMessage(dmnMessages, "DMN: The HitPolicy for decision table '[ID: null]' should be UNIQUE (DMN Validation, Decision Table Analysis, Hit Policy Recommender) "));
```
tarilabs, line 99: do we really need mocking, especially in this case you are getting result for something which usually through public API it would not even reach I believe DT Analysis --does it make sense to have a table without requirements and without rules?
I believe there are plenty of very basic DMN example already used in tests, and if none suitable it can still be the case better to write a XSD compliant, DMN compilable, DMN model with a super simple decision table which would return the message "DT Analysis completed with nothing to say" or the likes.
Wdyt?

Rikkola:
> Is this really the way i18n is expected to be implemented??

This is still a proposal for how I would expect the **backend** i18n to work.

This can be moved to the client side if I store the necessary data for forming the text message in `DMNDTAnalysisMessage`, basically the parameters that are now passed for `Msg`. This approach has several benefits if we ever want to visualise the issues in any other way than simple text line. For example the GDST verification uses HTML.

The reason for this approach is that I actually expected the report forming to live in the backend. I'll make the alternative proposal for client side.

tarilabs:
There is no reason to limit i18n to client-side, I believe is beneficial to have i18n available anywhere. My only surprise is that this is different than the original design which was discussed 🤷‍♂ 
I'm fine if this is the blessed way to provide i18n inside of Drools, I just want to be sure this is consistent with other Drools module as I expect the effort will be to refactor all the message to this new design

manstis:
@Rikkola @tarilabs My take on this PR is that @Rikkola has moved static `MessageX` instances, with fixed descriptive text in English, to a factory `AnalyticsMsg` wherein the descriptive text is i18n'ed based on the locale. 

**IF** we only cared for i18n in the browser we _could_ have kept the code as it was and perform the translation client-side based on the `DMNMessageType` (with related change to ensure `AbstractMessage` and sub-classes etc can be marshalled from server to browser). 

I don't see what @Rikkola has done here as a change in design, just a refinement to support i18n of messages both server and client-side.

As for:
> blessed way to provide i18n inside of Drools...

I don't believe anything else in Drools (backend/engine) domain does any i18n. DMN is alone.

Rikkola:
@manstis @tarilabs So why I did this on the server side. 

Looking at the code the new way that the DMN editor uses is that you can add the "Validate" button for the editor, but then the validation for each editor only differs in the backend. My previous assumption was that the "Validate" button would work as before with the `KieEditor` base, where the validation button is recreated for each editor and each editor then calls a specific backend service for that editor. I believe the `KieEditor` approach is not used for DMN Editor due to Kogito. Now to add the i18n swap for the client side I would have to alter the client side and add some sort of pluggable filter for all the validation messages. (Probably too much UI specific info for @tarilabs, but incase @manstis and someone else from UI side is interested).  **To put it short the way the editors act on "Validate" had changed from what I used to know.**

So now the cleaner way to do this would have been to translate the messages in the DMN backend validation service, but since @tarilabs had a nice `Msg` class where I could pick the text up I thought perhaps best to make `Msg` and the core analysis code to use i18n.

We can talk next week if the i18n is welcomed there or if core is kept 100% i18n free like before. But since the `Msg` class nicely isolates all the messages there could be an opportinity to translate the entire DMN and offer a better product for the languages we officially support. I know finnish people do not mind if they see some English, but I hear the cultures for the languages we do support in the product consider any English they see as a bug and the compile messages are shown in the UI.

tarilabs:
> We can talk next week if the i18n is welcomed there or if core is kept 100% i18n free like before

I am not sure why the conversation is still about whether to i18n on drools.git/Drools core or not.
I already stated here: https://github.com/kiegroup/drools/pull/2785#issuecomment-592048803
and I will state it again.

Having i18n on drools.git/Drools core and consequently Kie DMN engine is more than welcome.

My question was about the technical mechanism.

> since the Msg class nicely isolates all the messages there could be an opportinity to translate the entire DMN and offer a better product for the languages we officially support

Exactly my point.
At the time, we isolated the messages purposefully to have a single point of translation: my understanding the original design was validated and it would have required a mechanism to override only the string values.
If the design requires the ResourceBundle that is more than OK, I just want to be sure as all the messages will make sense to have it translated, that the refactor/new design being proposed here:
 https://github.com/kiegroup/drools/pull/2785/files#diff-479c733abe4d4b64593fe951322e3c14R31
is a blessed one from project/product perspective.

In summary:

- this was never a debate wheter to i18n on drools.git/Drools core or not: i18n on Kie DMN core more than welcome
- please let's have the affected Tech Leads approve the [ResourceBundle mechanism](https://github.com/kiegroup/drools/pull/2785/files#diff-479c733abe4d4b64593fe951322e3c14R31 ) is valid: I would assume reading from @manstis this is already fine later down the line from DMN Editor perspective, so I guess we are just missing @mariofusco and @krisv to validate this is fine for drools.git/Drools Core and Kogito perspective is the blessed mechanism to load i18n
- this way, we can later proceed to extend i18n to all DMN messages, and if other Drools Core want they can adopt the same mechanism

Hope this clarifies!

jomarko:
@Rikkola hello, please what is status of this PR. Opened quite long time ago so just want to double check we are not waiting just for my review.

Rikkola:
@jomarko No I was told to give this low priority for now. So changed it back to draft. I can close it if that is better.

PR #2788, opened 2020-02-26

tkobayas:
…indexing

WIP implementation. As suggested, now I focus on executable-model.

Note that this PR has several @Ignore tests where I haven't yet implemented logic for those cases.

Please give a quick review before going too far. Thanks!

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
@@ -1,0 +1,100 @@
+package org.drools.compiler.rule.builder.util;
+
+import org.drools.compiler.lang.descr.BaseDescr;
+import org.drools.compiler.lang.descr.OperatorDescr;
+import org.drools.compiler.lang.descr.RelationalExprDescr;
+import org.drools.core.base.ClassObjectType;
+import org.drools.core.rule.Pattern;
+import org.mvel2.util.PropertyTools;
+
+public class ConstraintUtil {
+
+    /**
+     * Swap left and right operands in a constraint when a fact property is located on the right side.
+     * 
+     * e.g. Person(20 < age) should be normalized to Person(age > 20)
+     * 
+     * @param expression
+     * @param operator 
+     * @param rightValue 
+     * @param leftValue 
+     * @return Normalized <code>expression</code>
+     */
+    public static String inverseExpression(RelationalExprDescr relDescr, String expression, String leftValue, String rightValue, String operator, Pattern pattern) {
+
+        Class<?> clazz = pattern.getObjectType().getClassType();
+
+        String leftProp = getFirstProp(leftValue);
+        String rightProp = getFirstProp(rightValue);
+
+        OperatorDescr operatorDescr = relDescr.getOperatorDescr();
+
+        if (!(pattern.getObjectType() instanceof ClassObjectType)) {
+            // do not inverse
+            return expression;
+        }
+
+        if (!operator.equals(operatorDescr.getOperator())) {
+            // do not inverse
+            return expression;
+        }
+        if (canInverse(operator) && (PropertyTools.getFieldOrAccessor(clazz, leftProp) == null) && ((PropertyTools.getFieldOrAccessor(clazz, rightProp) != null) || (rightProp.equals("this")))) {
+
+            boolean negate = false;
+            if (isNagatedExpression(expression, leftValue, rightValue, operator)) {
+                if (relDescr.getOperatorDescr().isNegated()) {
+                    negate = true;
+                } else {
+                    // do not inverse
+                    return expression;
+                }
+            }
+
+            BaseDescr left = relDescr.getLeft();
+            relDescr.setLeft(relDescr.getRight());
+            relDescr.setRight(left);
+
+            String inversedOperator = inverseOperator(operator);
+
+            operatorDescr.setOperator(inversedOperator);
+
+            StringBuilder sb = new StringBuilder();
+            String inversedExpression = sb.append(rightValue).append(" ").append(inversedOperator).append(" ").append(leftValue).toString();
+            if (negate) {
+                inversedExpression = "!( " + inversedExpression + " )";
+            }
+
+            return inversedExpression;
+        }
+
+        return expression;
+    }
+
+    private static String getFirstProp(String str) {
+        int i = str.indexOf(".");
+        return (i != -1) ? str.substring(0, i) : str;
+    }
+
+    private static boolean isNagatedExpression(String expression, String leftValue, String rightValue, String operator) {
+        return expression.matches("^!\\s*\\(\\s*" + leftValue + "\\s*" + operator + "\\s*" + rightValue + "\\s*\\)$");
+    }
+
+    private static boolean canInverse(String operator) {
+        return (operator.equals("==") || operator.equals("!=") || operator.equals(">") || operator.equals("<") || operator.equals(">=") || operator.equals("<="));
+    }
+
+    private static String inverseOperator(String operator) {
+        switch (operator) {
+            case ">":
+                return "<";
+            case "<":
+                return ">";
+            case ">=":
+                return "<=";
+            case "<=":
+                return ">=";
+            default:
+                return operator;
+        }
+    }
+}
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
@@ -5,6 +5,7 @@ import org.drools.compiler.lang.descr.OperatorDescr;
 import org.drools.compiler.lang.descr.RelationalExprDescr;
 import org.drools.core.base.ClassObjectType;
 import org.drools.core.rule.Pattern;
+import org.drools.core.util.index.IndexUtil;
 import org.mvel2.util.PropertyTools;
 
 public class ConstraintUtil {
@@ -52,7 +53,7 @@ public class ConstraintUtil {
             relDescr.setLeft(relDescr.getRight());
             relDescr.setRight(left);
 
-            String inversedOperator = inverseOperator(operator);
+            String inversedOperator = IndexUtil.ConstraintType.decode(operator).inverse().getOperator();
 
             operatorDescr.setOperator(inversedOperator);
 
@@ -83,11 +84,7 @@ public class ConstraintUtil {
         if (leftProp.isEmpty() || rightProp.isEmpty()) {
             return false;
         }
-        return canInverse(operator);
-    }
-
-    private static boolean canInverse(String operator) {
-        return (operator.equals("==") || operator.equals("!=") || operator.equals(">") || operator.equals("<") || operator.equals(">=") || operator.equals("<="));
+        return IndexUtil.ConstraintType.decode(operator).canInverse();
     }
 
     private static String getFirstProp(String str) {
@@ -104,20 +101,5 @@ public class ConstraintUtil {
 
     private static boolean isNagatedExpression(String expression, String leftValue, String rightValue, String operator) {
         return expression.matches("^!\\s*\\(\\s*" + leftValue + "\\s*" + operator + "\\s*" + rightValue + "\\s*\\)$");
-    }
-
-    private static String inverseOperator(String operator) {
-        switch (operator) {
-            case ">":
-                return "<";
-            case "<":
-                return ">";
-            case ">=":
-                return "<=";
-            case "<=":
-                return ">=";
-            default:
-                return operator;
-        }
     }
 }
--- a/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
@@ -34,24 +34,10 @@ public class ConstraintUtil {
 
         String leftProp = getFirstProp(leftValue);
         String rightProp = getFirstProp(rightValue);
-        if (leftProp.isEmpty() || rightProp.isEmpty()) {
-            // do not inverse
-            return expression;
-        }
 
         OperatorDescr operatorDescr = relDescr.getOperatorDescr();
 
-        if (!(pattern.getObjectType() instanceof ClassObjectType)) {
-            // do not inverse
-            return expression;
-        }
-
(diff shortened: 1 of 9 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
```diff
@@ -0,0 +1,149 @@
...
+import org.drools.modelcompiler.builder.generator.drlxparse.DrlxParseResult;
+import org.drools.modelcompiler.builder.generator.drlxparse.SingleDrlxParseSuccess;
+import org.drools.modelcompiler.builder.generator.visitor.pattern.PatternConstraintParseResult;
+import org.drools.modelcompiler.util.EvaluationUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.THIS_PLACEHOLDER;
+
+public class ConstraintUtil {
```
lucamolteni, line 21: I like how everything related to this logic is in just one class

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
```diff
@@ -0,0 +1,149 @@
...
+                    inversedOperator = ConstraintType.EQUAL;
+                    break;
+                case NOT_EQUAL:
+                    inversedOperator = ConstraintType.NOT_EQUAL;
+                    break;
+                case GREATER_THAN:
+                    inversedOperator = ConstraintType.LESS_THAN;
+                    methodName = replaceMethodName(methodName, "greaterThan", "lessThan");
+                    break;
+                case GREATER_OR_EQUAL:
```
lucamolteni: Please double check the inversion because I'm not sure but maybe > gets inverted to ≤ and < to ≥ (greater than => less than equal) (less than => greater than equal)

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ConstraintNormalizationTest.java
```diff
@@ -0,0 +1,420 @@
...
+                           " Person(20 < age, 30 > age)\n" +
+                           " Person(30 <= age, 40 >= age)\n" +
+                           "then\n" +
+                           "end";
+
+        final KieSession ksession = getKieSession(str);
+
+        if (testRunType == RUN_TYPE.FLOW_DSL || testRunType == RUN_TYPE.PATTERN_DSL) {
+            String exprId1 = new DRLIdGenerator().getExprId(Person.class, "org.drools.modelcompiler.util.EvaluationUtil." +
+                                                                          "greaterThanNumbers(_this.getAge(), 20)");
```
lucamolteni: Here maybe instead of the definition of the rule we should test the behaviour. So we insert all kind of people and only the one that matches are in the result. What do you think?

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
tkobayas: This normalization positionally reverses a constraint while keeping the same logical meaning. e.g. from "20 > age" to "age < 20". So a simple inversion ('>' to '<', '>=' to '<=') is correct :)

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ConstraintNormalizationTest.java
tkobayas: This normalization changes the constraint expression but keeps the same rule behaviour. So verifying rule outcome doesn't prove that the normalization works expectedly ("insert all kind of people and only the one matches" should happen with and without normalization). But I assert the rule outcome to confirm that it doesn't break the rule anyway.

Differences can be observed by:

- PropertyReactivity (in reality, executable-model works correctly even without normalization, though.)
- AlphaIndexing
- NodeSharing

So I added those tests first (testNormalizationForPropertyReactivity, testNormalizationForAlphaIndexing, testNormalizationForNodeSharing).

Then, I added other tests (testOperators, testNestedProperty ...) for coverage of various expressions by comparing its exprId.

But, yeah, I may use NodeSharing check instead of exprId check. It would make sense better and be easier to read/write. I'll try and update. Thanks!

lucamolteni: Understood, thank you for the explanation

tkobayas:
Rebased and pushed to consume DROOLS-5115 fix. Please review, thx!

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
```diff
@@ -0,0 +1,219 @@
...
+import org.drools.modelcompiler.builder.generator.drlxparse.DrlxParseResult;
+import org.drools.modelcompiler.builder.generator.drlxparse.SingleDrlxParseSuccess;
+import org.drools.modelcompiler.util.EvaluationUtil;
+
+import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.THIS_PLACEHOLDER;
+
+public class ConstraintUtil {
+
+    private static final String CLASS_NAME = EvaluationUtil.class.getCanonicalName() + ".";
+    private static final String TO_BIG_DECIMAL = EvaluationUtil.class.getCanonicalName() + ".toBigDecimal";
```
mariofusco, line 26: I understand why you're doing this, but having a so special case only for BigDecimal is not ideal. I think this is still manageable but we need to avoid having tons of special cases like this. Regarding this specific fix, I'm afraid this is not avoidable, or anyway I also don't see a better way to cover this case. For now please just add a comment to explain why you need to do this otherwise we will have to "rediscover" it again when somebody else will bump into this in 6 months.

mariofusco (approved):
Great job as usual. The only problem of this is that I don't think we could merge it until we won't have a similar solution for plain drl (that will be probably harder to implement). I don't think we can afford to have drl and execmodel behaving in 2 different ways.
@tkobayas @lucamolteni My advice is to implement something similar also for drl and merge everything when it will be ready, but I'm open to discuss this. Feedback is welcome.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
tkobayas: Thanks. Added a comment.

tkobayas:
@mariofusco @lucamolteni  Thank you for reviewing. Sure, I'm going to start implementing for standard drl. At first, I thought of filing another JIRA but they will share the same unit tests anyway so I will work on it with the same DROOLS-4956 and this PR.

lucamolteni:
> @mariofusco @lucamolteni Thank you for reviewing. Sure, I'm going to start implementing for standard drl. At first, I thought of filing another JIRA but they will share the same unit tests anyway so I will work on it with the same DROOLS-4956 and this PR.

Any ideas on the implementation? I think the one with the regexp is too limiting, and one that came is too actually parse the constraint using the same parser we use in the executable model and use the same `ConstraintUtil` class you created here but there might be some problems:

1) `drools-core` should depend on `drools-mvel-parser`
2) we should bring the minimal amount of code and SingleDrlxParseSuccess it's just too big...

What do you think?

tkobayas:
@lucamolteni When I developed the Util with regex, I was looking at the earlier stage of pattern building in order to share the Util with both standard-drl and executable-model. This time, I will develop a new Util only for standard-drl. So I can work at the later stage of pattern building of standard-drl. I hope that the new Util doesn't need to depend on drools-mvel-parse or SingleDrlxParseSuccess. Anyway, I will share my thought after some experiments :)

lucamolteni:
> ... I will share my thought after some experiments :)

Perfect thank you :)

tkobayas:
@lucamolteni @mariofusco Updated the PR with implementation for standard-drl. At PatternBuilder.normalizeExpression(), constraint expressions are well split and concise. Also I can retrieve left/right/operator so the inversion is easier.

This is not yet ready to merge. See code comments below:

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
```diff
@@ -0,0 +1,100 @@
...
+
+        String leftProp = getFirstProp(leftValue);
+        String rightProp = getFirstProp(rightValue);
+
+        OperatorDescr operatorDescr = relDescr.getOperatorDescr();
+
+        if (!(pattern.getObjectType() instanceof ClassObjectType)) {
+            // do not inverse
+            return expression;
+        }
```
tkobayas: I wonder if I should implement the normalization for FactTemplateObjectType (If yes, need to detect its properties). Is it still used (I only hit this in Misc2Test#testFactTemplates)? According to https://issues.redhat.com/browse/DROOLS-600 , it will likely be removed...

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ConstraintNormalizationTest.java
```diff
@@ -0,0 +1,547 @@
...
+
+        // Check NodeSharing to verify if normalization works expectedly
+        assertEquals(1, ReteDumper.collectNodes(ksession).stream().filter(AlphaNode.class::isInstance).count());
+
+        final Person p = new Person("Toshiya");
+        p.getItemsString().put("Key", "Value");
+
+        ksession.insert(p);
+        assertEquals(2, ksession.fireAllRules());
+    }
```
tkobayas: There seems to be an issue with Map access and executable-model. I'll check next week.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
```diff
@@ -0,0 +1,229 @@
...
+     * @return Normalized <code>DrlxParseResult</code>
+     */
+    public static DrlxParseResult normalizeConstraint(DrlxParseResult drlxParseResult) {
+        if (!ENABLE_NORMALIZE) {
+            return drlxParseResult;
+        }
+
+        if (drlxParseResult instanceof SingleDrlxParseSuccess) {
+            // Create a copy
+            SingleDrlxParseSuccess s = new SingleDrlxParseSuccess((SingleDrlxParseSuccess) drlxParseResult);
```
hellowdan: I think that avoiding naming variables with a single character makes the code more readable. There are more places like this in the PR.

hellowdan (approved):

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
tkobayas: Confirmed with @mariofusco , we don't support FactTemplates so it's okay to "do not inverse" here.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ConstraintNormalizationTest.java
tkobayas: Filed https://issues.redhat.com/browse/DROOLS-5136 . This issue needs to be fixed/merged first.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/ConstraintUtil.java
tkobayas: Thank you for pointing this out! I have fixed it.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ConstraintNormalizationTest.java
tkobayas: DROOLS-5136 has been fixed and merged. Now this PR is rebased to consume the fix so Map access normalization is implemented.

tkobayas:
Jenkins test again

tkobayas:
Jenkins test again

tkobayas:
@mariofusco @lucamolteni Updated with standard-drl normalization including Map access use case. You have already approved this PR before but I think it's better to ask you to check again. Thanks!

Note that the SonarCloud "Bug" suggests potential NullPointerException but the "simple" flag means "relDescr is not null" so I think we can ignore.

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
```diff
@@ -0,0 +1,123 @@
...
+            return false;
+        }
+        if (leftProp.isEmpty() || rightProp.isEmpty()) {
+            return false;
+        }
+        return canInverse(operator);
+    }
+
+    private static boolean canInverse(String operator) {
+        return (operator.equals("==") || operator.equals("!=") || operator.equals(">") || operator.equals("<") || operator.equals(">=") || operator.equals("<="));
```
mariofusco: Minor thing: we already have part of this analysis for operator inside the enum `IndexUtil.ConstraintType`. If possible I'd keep all the similar code there. For instance you can add this `canInverse` method in that enum and then testing it with `IndexUtil.ConstraintType.decode(operator).canInverse()`

tkobayas:
Jenkins test again

drools-compiler/src/main/java/org/drools/compiler/rule/builder/util/ConstraintUtil.java
tkobayas: Thank you for the suggestion. I did the refactoring for `org.drools.core.util.index.IndexUtil.ConstraintType` in standard-drl and did similar for `org.drools.model.Index.ConstraintType` in executable-model. Finally, `IndexUtil.ConstraintType` and `Index.ConstraintType` have duplicate logic but maybe it's okay? (If `Index.ConstraintType` wants to use `IndexUtil.ConstraintType` method, it will require a conversion from `Index.ConstraintType` to `IndexUtil.ConstraintType` so it would be a bit redundant)

PR #2791, opened 2020-03-02
[DROOLS-5115] executable model fails with negation and BigDecimal

tkobayas:


tkobayas:
Converted UnaryExpr's inner expression using an approach similar to parseBinaryExpr(). Please review, thanks!

lucamolteni (approved):

PR #2802, opened 2020-03-04
[FAI-58] Extend DMNRuntimeEventListener

kostola:
Add methods + events to DMNRuntimeEventListener to be able to monitor when API method calls are invoked on the DMNRuntime (evaluateAll, evaluateDS, etc.)

### DIFF
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
@@ -1,0 +1,9 @@
+package org.kie.dmn.api.core.event;
+
+public interface AfterEvaluateAllEvent extends DMNEvent {
+
+    String getModelNamespace();
+
+    String getModelName();
+
+}
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.kie.dmn.api.core.event;
 
 public interface AfterEvaluateAllEvent extends DMNEvent {
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
@@ -1,0 +1,9 @@
+package org.kie.dmn.api.core.event;
+
+public interface BeforeEvaluateAllEvent extends DMNEvent {
+
+    String getModelNamespace();
+
+    String getModelName();
+
+}
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.kie.dmn.api.core.event;
 
 public interface BeforeEvaluateAllEvent extends DMNEvent {
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/AfterEvaluateAllEventImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/AfterEvaluateAllEventImpl.java
@@ -1,0 +1,33 @@
+package org.kie.dmn.core.impl;
+
+import org.kie.dmn.api.core.DMNResult;
+import org.kie.dmn.api.core.event.AfterEvaluateAllEvent;
+
+public class AfterEvaluateAllEventImpl implements AfterEvaluateAllEvent {
+
+    private String modelNamespace;
+    private String modelName;
+    private DMNResult result;
+
+    public AfterEvaluateAllEventImpl(String modelNamespace, String modelName, DMNResult result) {
+        this.modelNamespace = modelNamespace;
+        this.modelName = modelName;
+        this.result = result;
+    }
+
+    @Override
+    public String getModelNamespace() {
+        return modelNamespace;
+    }
+
+    @Override
+    public String getModelName() {
+        return modelName;
+    }
+
+    @Override
+    public DMNResult getResult() {
+        return result;
+    }
+
+}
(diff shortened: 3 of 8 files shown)


kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
```diff
@@ -0,0 +1,9 @@
+package org.kie.dmn.api.core.event;
+
+public interface AfterEvaluateAllEvent extends DMNEvent {
```
kostola: @tarilabs please check that the structure of `AfterEvaluateAllEvent` makes sense to you

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
```diff
@@ -0,0 +1,9 @@
+package org.kie.dmn.api.core.event;
+
+public interface BeforeEvaluateAllEvent extends DMNEvent {
```
kostola: @tarilabs please check that the structure of `BeforeEvaluateAllEvent` makes sense to you

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeEvaluateAllEvent.java
```diff
@@ -0,0 +1,9 @@
...
+
+    String getModelNamespace();
+
+    String getModelName();
+
+}
```
kostola, line 25: @tarilabs please check that the structure of `BeforeEvaluateAllEvent` makes sense to you

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/AfterEvaluateAllEvent.java
```diff
@@ -0,0 +1,9 @@
...
+
+    String getModelNamespace();
+
+    String getModelName();
+
+}
```
kostola, line 25: @tarilabs please check that the structure of `AfterEvaluateAllEvent` makes sense to you

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerUtils.java
```diff
@@ -32,85 +47,114 @@
...
-    private static void notifyListeners(DMNRuntimeEventManager eventManager, Consumer<DMNRuntimeEventListener> consumer) {
+    public static AfterEvaluateAllEvent fireAfterEvaluateAll(DMNRuntimeEventManager eventManager, DMNModel model, DMNResult result) {
+        return notifyListeners(
+                eventManager,
+                () -> new AfterEvaluateAllEventImpl(model.getNamespace(), model.getName(), result),
+                DMNRuntimeEventListener::afterEvaluateAll
+        );
+    }
+
+    private static <T> T notifyListeners(DMNRuntimeEventManager eventManager, Supplier<T> eventSupplier, BiConsumer<DMNRuntimeEventListener, T> eventConsumer) {
```
tarilabs: I am missing why this change is needed, please?

tarilabs: Nevermind, I believe I got it, because to align with this: https://github.com/kiegroup/drools/commit/8116db88ff1acc028f306686ed02aaa658ea4e3f#diff-9e8ea9e754fcefdbad2891d584eaa112R34

tarilabs: please confirm.

tarilabs:
Still considering this: https://github.com/kiegroup/drools/commit/8116db88ff1acc028f306686ed02aaa658ea4e3f#diff-9e8ea9e754fcefdbad2891d584eaa112R34
might be worthy to evaluate if isn't it better to use similar approach rather then using ID/UUID in the context?

r00ta:
Any test? :)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerUtils.java
r00ta: What about removing the Supplier with
```
    private static <T extends DMNEvent> T notifyListeners(DMNRuntimeEventManager eventManager, T event, BiConsumer<DMNRuntimeEventListener, T> eventConsumer) {
        if (!eventManager.hasListeners()) {
            return null;
        }

        for( DMNRuntimeEventListener listener : eventManager.getListeners() ) {
            try {
                eventConsumer.accept(listener, event);
```
?

danielezonca: Please verify performance implication too because I expect this new `notifyListeners` method might have a negative impact

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/AfterEvaluateAllEventImpl.java
```diff
@@ -0,0 +1,33 @@
+package org.kie.dmn.core.impl;
+
+import org.kie.dmn.api.core.DMNResult;
+import org.kie.dmn.api.core.event.AfterEvaluateAllEvent;
+
+public class AfterEvaluateAllEventImpl implements AfterEvaluateAllEvent {
+
+    private String modelNamespace;
+    private String modelName;
+    private DMNResult result;
```
danielezonca, line 26: What about `final`?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BeforeEvaluateAllEventImpl.java
```diff
@@ -0,0 +1,33 @@
+package org.kie.dmn.core.impl;
+
+import org.kie.dmn.api.core.DMNResult;
+import org.kie.dmn.api.core.event.BeforeEvaluateAllEvent;
+
+public class BeforeEvaluateAllEventImpl implements BeforeEvaluateAllEvent {
+
+    private String modelNamespace;
+    private String modelName;
+    private DMNResult result;
```
danielezonca, line 26: What about `final`?

danielezonca (commented):
Minor comments

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeEventManagerUtils.java
kostola: I just reverted it back to the original one just adding the two methods I need

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/AfterEvaluateAllEventImpl.java
kostola: I looked at classes implementing other DMN events and none of them has `final` private attributes, so I assumed there may be some reason I don't know yet.
I can mark them as `final` if there's no such reason :wink:

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/BeforeEvaluateAllEventImpl.java
kostola: Same as [here](https://github.com/kiegroup/drools/pull/2802#discussion_r388801864)

kostola:
@tarilabs @danielezonca @jiripetrlik this is ready for review now, please take a look :wink:

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/classloader/DMNRuntimeListenerTest.java
```diff
@@ -139,7 +142,11 @@ public void testListenerWithBKM() throws Exception {
         final DMNRuntime runtime = DMNRuntimeUtil.createRuntime("org/kie/dmn/core/say_for_hello.dmn", this.getClass());
         TestEventListener listener = new TestEventListener();
         runtime.addListener(listener);
-        final DMNModel dmnModel = runtime.getModel("http://www.trisotech.com/dmn/definitions/_b6f2a9ca-a246-4f27-896a-e8ef04ea439c", "say for hello");
+
```
tarilabs: wouldn't be easier to just have a separate test for the 2 new events?
you can even use an anonymous class listener just tailored to the 2 new events.
This way we guarantee non regression more easily

tarilabs (approved):
minor comment if you want to consider, the rest is fine thank you

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/classloader/DMNRuntimeListenerTest.java
kostola: I expanded the already existing test because I thought its purpose was to test the whole sequence of events, since it includes also `BeforeEvaluateDecisionEvent` and `AfterEvaluateDecisionEvent`.
I can easily move my logic to a separate one... @danielezonca what do you think?

danielezonca: It is the same for me, probably if we move this additional check to another test it could be easier to read

kostola: Ok I moved it in a separate test :+1:

danielezonca (approved):

jiripetrlik (approved):
Looks good to me.

kostola:
@kiegroup/gatekeepers could you please merge this?

PR #2821, opened 2020-03-17
RHDM-1292 DMN DT Analysis message when InputEntry unary test null value

tarilabs:
Validation yields a confusing message when a null is used in a decision
table cell


manstis:
Hi @tarilabs I just wanted to check whether this highlights any deficiencies in the editor?

(Like https://issues.redhat.com/browse/RHDM-1291)?

tarilabs:
nope @manstis in this case simply the gaps/overlaps analysis has no sense, given the table per se is already out of good design (using null). So just making the error message more explicit.

hellowdan (approved):

PR #2824, opened 2020-03-19
[DROOLS-5028] Support incremental update for function with executable…

tkobayas:
… model

Works with https://github.com/kiegroup/droolsjbpm-knowledge/pull/433


tkobayas:
Note: This approach (= Use internalChanges to detect a rule which should not be updated) doesn't work for DSL/RDSLR case (IncrementalCompilationTest#testKJarUpgradeWithDSL) because we cannot connect the dslr file to its rule name.

tkobayas:
Closing this PR as replaced by https://github.com/kiegroup/drools/pull/2829

PR #2830, opened 2020-03-25
Address Thread Sanitizer Warnings in drools-core

rwils:
I realize the omission of synchronization here might be intentional, but I was encouraged to send these upstream before creating a patch in our internal mirror.


ghost:

PR #2839, opened 2020-04-03
[DROOLS-4562] DMN validation semantic rules for DMNDI

tkobayas:


tkobayas:

tkobayas:

kie-dmn/kie-dmn-backend/src/main/java/org/kie/dmn/backend/marshalling/v1_2/xstream/MarshallingUtils.java
```diff
@@ -58,6 +60,8 @@ public static String formatQName(QName qname, DMNModelInstrumentedBase parent) {
             String nsForPrefix = parent.getNamespaceURI(qname.getPrefix());
             if (parent.getURIFEEL().equals(nsForPrefix)) {
                 return qname.getLocalPart(); // DMN v1.2 feel comes without a prefix.
+            } else if (parent instanceof DMNShape || parent instanceof DMNEdge) {
```
tarilabs, line 63: I believe the same modifications will be needed as well for the same class in the v1.3 package
```
org.kie.dmn.backend.marshalling.v1_3.xstream. MarshallingUtils
```

tkobayas: Thank you for pointing that out! Fixed and updated.

tkobayas:

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/ValidatorUtil.java
```diff
@@ -45,6 +50,19 @@ public static String formatMessages(final List<DMNMessage> messages) {
         return messages.stream().map(Object::toString).collect( Collectors.joining( System.lineSeparator() ) );
     }
 
+    public static boolean doesDefinitionsContainIdForDMNEdge(Definitions definitions, String id) {
```
tarilabs, line 53: For a next iteration, we could think how to go beyond the need of this function.
I understand the need of it currently, and why it's invoked in an `eval()`.
I am wondering IF we had an utility method on all the model's object which returned their parent `Definitions` if that would have solved.
What do you think?

tarilabs (approved):
A food for thought comment, otherwise is good from my side.
Thanks!

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/ValidatorUtil.java
tkobayas: Thank you for pointing this out. The problem was that ksession has only Definitions object from "DMNImports" entry-point. The rule wants to evaluate all InformationRequirement, KnowledgeRequirement etc. objects inside the Definitions. I wasn't able to write a nice rule with nested "from" so I ended up to write the function. So rather than 

- an utility method on all the model's object which returned their parent Definitions

, inversely, 

- an utility method on Definitions to return all descendant model objects.

would help to write the rule easier, I think.

karreiro (commented):
Hi @tkobayas @tarilabs, thanks for adding me as a reviewer.

I just debugged/tested it locally, and I can confirm that it breaks the Included Models feature on DMN editor :-(

I will need to update the client-side next week to ensure that everything will continue working.

Thanks again. I'll handle it as a high-priority task to not block you.

manstis:
Why would an additional validation (especially for DMNDI) break the editor?!

Just curious.

tarilabs:
@manstis it is not the validation check.

Toshiya spotted a gap in the `kie-dmn-backend` which he is addressing in order to correctly compute lookups, during the actual checks.

Allegedly, for some reasons, there might be a dependency on the behavior caused by this gap that is breaking the editor.

I leave Guilherme to indicate where this is happening on the editor side; I just wanted to clarify that is not the validation check the root cause, is the gap in the -backend code.

While not a full answer, I hope this still helps.

tkobayas:
@manstis 

As a part of this PR, I fixed DMN marshalling for DMNShape and DMNEdge so its "dmnElementRef" is properly unmarshalled to a QName with prefix and localPart.

Before the fix, dmnElementRef with prefix (e.g. "included1:_1361F7E7-846E-408D-B1AB-D4F7D4BD247F") is unmarshalled to QName with single localPart "included1:_1361F7E7-846E-408D-B1AB-D4F7D4BD247F".

After the fix, such a dmnElementRef is unmarshalled to QName with prefix "included1" and localPart "_1361F7E7-846E-408D-B1AB-D4F7D4BD247F".

example dmn with such a dmnElementRef)
https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-validation/src/test/resources/org/kie/dmn/validation/import/DROOLS-4187b.dmn#L32

Such a dmnElementRef can be created when I include a DMN and re-use a node in 7.35.0 workbench DMN Designer.

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/ValidatorUtil.java
tkobayas: @tarilabs  Shall I file a JIRA for the improvement? Or you meant that it can be improved during this PR work?

tarilabs: let's file for improvement in another JIRA, imho.

Another viable alternative, is a static function which does as you say, would take as input an `Definitions` and will produce as output all children. This way, it can be used a source in a `from` node in the DRL rule. What do you think? :)

manstis:
Thank-you for the explanation @tarilabs and @tkobayas.

I understand now how that'd affect the editor's use of imports.

kie-dmn/kie-dmn-validation/src/main/java/org/kie/dmn/validation/ValidatorUtil.java
tkobayas: Thanks, yes, it sounds good. I filed a JIRA for that https://issues.redhat.com/browse/DROOLS-5232

danielzhe:
@karreiro Since Kogito is using this "new" behavior and I'm working in the Included Models to Kogito, maybe this is the time to make this change in the DMN editor and merge everything.

karreiro:
Hi @tkobayas @tarilabs , with https://github.com/kiegroup/kie-wb-common/pull/3287 merged, this PR doesn't break Included Models feature on DMN editor anymore :-)

tkobayas:
@karreiro Thank you!

tkobayas:

tkobayas:

tkobayas:

tkobayas:
The failures in downstream don't seem to be related to this PR.

```
Test Result (3 failures / -3)

org.kie.server.integrationtests.jbpm.UserTaskServiceIntegrationTest.testCompleteAlreadyCompletedUserTask[0: JAXB KieServicesConfiguration{transport=REST, serverUrl='http://localhost:44712/kie-server-services/services/rest/server'}]
    org.kie.server.integrationtests.jbpm.UserTaskServiceIntegrationTest.testCompleteAlreadyCompletedUserTask[1: JSON KieServicesConfiguration{transport=REST, serverUrl='http://localhost:44712/kie-server-services/services/rest/server'}]
    org.kie.server.integrationtests.jbpm.UserTaskServiceIntegrationTest.testCompleteAlreadyCompletedUserTask[2: XSTREAM KieServicesConfiguration{transport=REST, serverUrl='http://localhost:44712/kie-server-services/services/rest/server'}]

Error Message

expected:<403> but was:<404>

Stacktrace

java.lang.AssertionError: expected:<403> but was:<404>
	at org.kie.server.integrationtests.jbpm.UserTaskServiceIntegrationTest.testCompleteAlreadyCompletedUserTask(UserTaskServiceIntegrationTest.java:155)
```

hellowdan (approved):

PR #2841, opened 2020-04-06
DROOLS-5222 DMN BKM / FEEL closure and currying support

tarilabs:
@etirelli @hellowdan 
/cc @evacchi @lucamolteni


hellowdan (approved):

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELFunctionDefinitionTest.java
```diff
@@ -87,7 +87,9 @@
                 {"{ myimport : { f1 : function() \"Hi\", f2 : function() f1() + \" \" , x1 : function(name) f2() + name }, r1 : myimport.x1(\"John\") }.r1", "Hi John", null },
                 {"{ m : { n : { o : { f1 : function() \"Hi\", f2 : function() f1() + \" \" , x1 : function(name) f2() + name }}}, r1 : m.n.o.x1(\"John\") }.r1", "Hi John", null },
                 {"{ m : { n : { f1 : function() \"Hi\", f2 : function() f1() + \" \" , o : { x1 : function(name) f2() + name }}}, r1 : m.n.o.x1(\"John\") }.r1", "Hi John", null },
-                {"{ m : { n : { f1 : function() \"Hi\", f2 : function() f1() + \" \" , o : { x1 : function(name) f2() + name }}}, r2 : m.n.f1() }.r2", "Hi", null }
+                {"{ m : { n : { f1 : function() \"Hi\", f2 : function() f1() + \" \" , o : { x1 : function(name) f2() + name }}}, r2 : m.n.f1() }.r2", "Hi", null },
```
lucamolteni, line 90: Food for thought: wouldn't it be better to put this test's assertions in a resource file rather that in a java file? In that way we could avoid quoting and it could be more readable

lucamolteni (approved):

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELFunctionDefinitionTest.java
tarilabs: yes, ideally these tests will be refactored to be defined in a "protocol", "spec" document and JUnit Java code generated out of that. With the advantage, also, to be more practically run JUnit methods from IDEs as well.

PR #2861, opened 2020-04-20
7.30.x

Yongquan-He:
Because of the kie.conf problem, it will return null pointer in the jar mode.
I know this problem can be solved by the use of appendtransformer.
But my project is springboot + other dependencies, if I use the shade plugin.
Several problems will occur, and I can not find way to solve.
So I wonther are there some method without using shade plugin when use mvn package?
<plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>repackage</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
Thank you very much!


ghost:

PR #2864, opened 2020-04-21
DROOLS-5260 DMN internal DMNRuntimeBuilder custom compiler

tarilabs:


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/internal/utils/DMNRuntimeBuilder.java
```diff
@@ -105,16 +110,19 @@ public static DMNRuntimeBuilderConfigured usingStrict() {
     }
 
     public DMNRuntimeBuilderConfigured buildConfiguration() {
-        DMNCompilerImpl dmnCompiler = new DMNCompilerImpl(ctx.cc);
-        return new DMNRuntimeBuilderConfigured(ctx, dmnCompiler);
+        return buildConfigurationUsingCustomCompiler((config) -> new DMNCompilerImpl(config));
+    }
+
+    public DMNRuntimeBuilderConfigured buildConfigurationUsingCustomCompiler(Function<DMNCompilerConfiguration, DMNCompiler> dmnCompilerFn) {
```
tarilabs, line 116: @mswiderski I sent an email to better explain, line to be noticed is this one ^

hellowdan (approved):

mswiderski:
@tarilabs compiler configuration works like a charm, thanks a lot.

one additional thing for consistency would be to expose listeners via the builder so it can be easily returned from here https://github.com/kiegroup/drools/blob/34a286c5dbc4bde4089d61c764cf6cae35a011f9/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/internal/utils/DMNRuntimeBuilder.java#L212

that allows to not register them manually once the runtime is created since the DMNRuntimeKB already exposes listeners and currently returns empty list. Wdyt?

tarilabs:
> compiler configuration works like a charm, thanks a lot

Glad to hear!

Concerning your other comment, that is not the list of active listeners, but are the "default" ones coming from kjar configuration.

You can use this method:
https://github.com/kiegroup/drools/blob/f55767758dc38600ae299b32785b07ae920781f0/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/DMNRuntimeEventManager.java#L33

which is exposed also directly on the `DMNRuntime` interface to add desired listeners (`DMNRuntime` interface extends `DMNRuntimeEventManager`)

Or am I missing something?

P.S. I shall rename that method you linked as `getDefaultListeners()` I think to avoid confusion.

mswiderski:
fair enough, I actually used the `dmnRuntime.addListener` so that works well, but thought that listeners could already be used during creation of dmnRuntime. But if you say that is for other needs then I am fine with that.

tarilabs:
@mswiderski 
> thought that listeners could already be used during creation of dmnRuntime

indeed eventually we can add it to a "runtime configuration" of sort. However currently this `DMNRuntimeBuilder` is an internal utility (mainly for Kogito based work) so I would await a bit more all requirements emerge there before putting inside too much.

Different approach is for blocking gaps, like the one which is addressed by this PR, which indeed make sense to fill immediately is possible ;)

PR #2869, opened 2020-04-24
[DROOLS-5271] RHS statements ordering changes with modify block in ex…

tkobayas:
…ecutable model

https://issues.redhat.com/browse/DROOLS-5271

- Keep the statement order in RHS
- Add parentheses to surround the logic generated for the modify statement. This makes the result similar to standard-drl.


lucamolteni (approved):
thanks

hellowdan (approved):

PR #2876, opened 2020-04-29
Unexpected behavior of double groupBy()

triceo:


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/AccumulateTest.java
```diff
@@ -1920,15 +1953,28 @@ public void testGroupByRegrouped() {
...
         System.out.println("----");
 
         ksession.delete( meFH );
         ksession.fireAllRules();
 
+        Assertions.assertThat(result)
+                .contains(new GroupByAcc.Pair<>("M", 81));
+        Assertions.assertThat(result)
+                .doesNotContain(new GroupByAcc.Pair<>("G", 0),
+                        new GroupByAcc.Pair<>("E", 0));
```
triceo, line 1969: The test fails here.

mariofusco:
I investigated this and unfortunately this is how the FromNode works. During the first round the accumulate produces the pairs from M, E and G and feeds the second accumulate through the From. Then you remove "Mario" and the first accumulate produces another List with only the modified pair the "M" one. This goes through the FromNode again which this time sees only the "M", but since it remembers where were also a E and G before, it believes that they have been removed so propagates the update of M and the deletion of E and G then producing those spurious firing. As I said this is the expected behaviour of the FromNode and I cannot change it even though I admit that in this situation is misleasding. More in general this demonstrates that it is a bad idea (also for performance reasons) to have 2 accumulates one after the other connected by a from and in fact you can obtain a correct result in a faster way by using one single accumulate.

PR #2890, opened 2020-05-11
[DROOLS-5316] Test Scenario: wrong management of null values with simple types in DMN scenarios

danielezonca:
https://issues.redhat.com/browse/DROOLS-5316

Code changes:
- `AbstractScenarioRunnerHelper.createObject`: use `ValueWrapper` instead of `Optional` to distinguish `null` value and missing value (see `DMNScenarioRunnerHelper.createObject` impl)
- Refactor `ResultWrapper` to `ValueWrapper`: removed `getOptional` method (it was just wrong :) ) and renamed to be used as generic value wrapper
- Added additional tests to `DMNScenarioRunnerHelperTest`, `RuleScenarioRunnerHelperTest` and `ScenarioBeanUtilTest` to improve the coverage

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ValueWrapper.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ValueWrapper.java
@@ -21,7 +21,7 @@ import java.util.function.Supplier;
 
 /**
  * Utility class to wrap a value with the possibility to specify error message or propose valid value.
- * Note: null can be used a value.
+ * Note: null can be used as value.
  * @param <T>
  */
 public class ValueWrapper<T> {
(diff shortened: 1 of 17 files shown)


drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ValueWrapper.java
```diff
@@ -20,42 +20,46 @@
 import java.util.function.Supplier;
 
 /**
- * java.util.Optional clone to have the null result
+ * Utility class to wrap a value with the possibility to specify error message or propose valid value
```
dupliaka: Could you please mention here null behavior as well?

yesamer (approved):
ok by me!

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/runner/model/ValueWrapper.java
danielezonca: Done [link](https://github.com/kiegroup/drools/pull/2890/commits/4019c88789e66aa99a165ffad625005ed2b8c5fa)

dupliaka:
- [x] Null values as a keywords of Simple String custom Type does not interpret as unexisting values.
- [x] If values in given was not matching any Object then null in expected interprets as a passed test scenario.
- [x] If no values interprets as a cause for a decision then the scenario with empty GIVEN and value in EXPECTED passes

dupliaka (approved):

PR #2893, opened 2020-05-12
DROOLS-5322 DMN Strongly fix codegen for FEEL id being Java keywords

tarilabs:


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/declaredtype/api/FieldDefinition.java
```diff
@@ -40,4 +41,12 @@
     boolean isStatic();
 
     boolean isFinal();
+
+    default Optional<String> ovverrideGetterName() {
```
lucamolteni: I'd use `overridden` past principle rather than present

lucamolteni (approved):
Just one minor thing

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/declaredtype/api/FieldDefinition.java
```diff
@@ -40,4 +41,12 @@
     boolean isStatic();
 
     boolean isFinal();
+
+    default Optional<String> overriddenGetterName() {
+        return Optional.empty();
```
lucamolteni, line 46: Can we make a default implementation for overriden to return the field name, therefore removing the ifPresent from GeneratedClassDeclaration and the optional here?

lucamolteni (approved):

hellowdan (approved):

tarilabs:
As discussed offline, the naming of the getter/setter is currently delegated to JavaParser, so transforming the `Optional<String>` to avoid the optional would have the side effect of always overriding the JavaParser name.
Hence we agreed offline we can leave as-is, hence https://github.com/kiegroup/drools/pull/2893#pullrequestreview-410005002

will ask merge

PR #2900, opened 2020-05-15
[BAPL-1259] Add logging for rule performance analysis

tkobayas:
WIP. Only implemented for PhreakJoinNode.


drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(PerfLogUtils.class);
+
+    public static final String PERF_LOGGER_ENABLED = "drools.performance.logger.enabled";
+    private static boolean enabled = Boolean.parseBoolean(System.getProperty(PERF_LOGGER_ENABLED, "false"));
+
+    public static final String PERF_LOGGER_THRESHOLD = "drools.performance.logger.threshold";
+    private static int threshold = Integer.parseInt(System.getProperty(PERF_LOGGER_THRESHOLD, "500")); // microseconds
+
+    private static final ThreadLocal<Boolean> started = new ThreadLocal<>();
```
franz1981: Instead of 4 different ThreadLocals would be better (for perf reasons too) to encapsulate the statistics collected per `BaseNode` into a `BaseNodeStats` class that could be used into a single `ThreadLocal`

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final ThreadLocal<AtomicInteger> evalCount = new ThreadLocal<>();
```
franz1981: There is no need to use an `AtomicInteger` here, but a single `int` (or `long`?) field into the `BaseNodeStats` (see prev comment) would be enough, given that is being used in a single threaded fashion

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final ThreadLocal<Long> startTime = new ThreadLocal<>();
```
franz1981: There is no need to use a boxed type here, but just a single `long` field into `BaseNodeStats` (see previous comments) is enough: it would save both pointer chasing and garbage produced

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final ThreadLocal<BaseNode> node = new ThreadLocal<>();
```
franz1981: Ditto as before: could be part of a single thread local instance (ie `BaseNodeStats`), but it worries me that I cannot see any code path that set it to `null`..it risks to became an actual memory leak: I suggest to rethink if it worths to make this a weak reference field instead...

franz1981 (changes requested):

tkobayas:
@franz1981 Thank you for reviewing, much appreciated! I applied the suggestions. I will keep working on this.

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -55,33 +50,29 @@ public static void setEnabled(boolean enabled) {
...
-        if (enabled && started.get().booleanValue()) {
-            long elapsedTime = (System.nanoTime() - startTime.get());
-            int count = evalCount.get().intValue();
-            if (count > 0 && (elapsedTime / 1000) > threshold) {
-                logger.trace("{}, evalCount:{}, elapsed:{}", node.get(), count, elapsedTime / 1000); // microseconds
+        if (enabled) {
+            NodeStats stats = nodeStats.get();
+            if (stats != null && stats.isStarted()) {
+                long elapsedTime = (System.nanoTime() - stats.getStartTime());
+                if (stats.getEvalCount() > 0 && (elapsedTime / 1000) > threshold) {
```
franz1981: I would save this expensive computation ie  elapsedTime / 1000 to perform it just once and I would handle the case of `stats == null` with a log or something similar: at this point this is not supposed to be `null` right?

franz1981 (changes requested):

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
tkobayas: Applied the suggestion. Thanks!

tkobayas:
This is a draft implementation of https://issues.redhat.com/browse/BAPL-1259

Motivation is to find a bottle-neck in rules.

- Measure count of constraint evaluation per PhreakNode
- Measure elapsed time per PhreakNode

_The idea was provided by Tomohiro Kamata, thanks!_

Please have a look at the use case example to see how this feature helps.

https://github.com/tkobayas/PerfLogUtils/wiki/How-to-use-PerfLogUtils

Regarding performance impact, I confirmed with a benchmark:

- If PerfLogUtils is disabled (default), this commit doesn't cause a performance degrade (comparing to 7.40.0-SNAPSHOT without the commit)
- If PerfLogUtils is enalbed, it causes a slight performance degrade around 1-3%. It wouldn't likely interfere the performance analysis. Of course, you shouldn't enable PerfLogUtils in production.

https://github.com/tkobayas/PerfLogUtils/wiki/Does-performance-downgrade-if-I-enable-PerfLogUtils%3F

I also moved ReteDumper from "src/test" to "src/main" because ReteDumper is used for analysis with PerfLogUtils by users. Do you have any concern about it? (I can create another util class like ReteUtils instead. But ReteDumper is a nice tool which deserves "src/main" anyway?)

Please review. Any suggestions/thoughts would be appreciated! (Recommend to check "Hide whitespace changes" (-w) for diff)

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,84 @@
...
+    public static void setThreshold(int threshold) {
+        PerfLogUtils.threshold = threshold;
+    }
+
+    public static boolean isEnabled() {
+        return enabled;
+    }
+
+    public static void setEnabled(boolean enabled) {
+        PerfLogUtils.enabled = enabled;
```
franz1981: Given that's globally visible...maybe it should use a volatile here: you expect this to change while the application is running?

franz1981 (changes requested):

lucamolteni:
Toshiya, I see a few problem with this

1) I personally don't like the fact that you used the metrics code directly in the Phreak algorithm. I think it could have been better to have an interface (or an AbstractClass) with specific domain name such as (`AccumulateBeforeEval`, `AccumulateAfterEval`) and then registered the metrics, similarly to what we have done with listeners.

2) Regarding 1) I think it could better to use instance methods instead of static one. They should be inlined by the JVM 

3) I wonder if it's possible to have such metrics without changing the actual Phreak algorithm. 
To be honest I'm not sure how we could do this, I'm going to think in the next days

mariofusco:
>     3\. I wonder if it's possible to have such metrics without changing the actual Phreak algorithm.
>         To be honest I'm not sure how we could do this, I'm going to think in the next days

Imo this is the most interesting point/question. I believe that we don't want/need to toggle the metrics while running: either you enable them when the engine starts or you don't. If we agree with this premise I believe that we could implement a second version of the phreak nodes factory that generates nodes wrapping or decorating the original ones and emitting metrics at their boundaries.

franz1981:
@mariofusco 

>  I believe that we could implement a second version of the phreak nodes factory that generates nodes wrapping or decorating the original ones and emitting metrics at their boundaries.

That's exactly what we've talked about with @lucamolteni today +100 :)

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
tkobayas: I don't expect this to be changed while the application is running, I added this for test convenience but maybe I can make it package private (or even remove)... Thanks for pointing this out :)

tkobayas:
@lucamolteni @mariofusco @franz1981 
Thank you very much for the suggestions. I agree that we want to keep the original phreak node classes clean. Introducing phreak nodes factory, I can write startMetrics()/logAndEndMetrics() in wrapper classes. How about incrementEvalCount()? I spread the method calls right before constraints.isAllowedCachedLeft()/isAllowedCachedRight in Phreak node classes. It's not a phreak node's method boundary so we cannot count it in the wrapper class. Ideally, I should write a constraint class factory to wrap the constraint classes so we can count in the constraint wrapper class. One concern is that constraint classes are widely used in the code base so I cannot forecast how much impact the change will have. Anyway, I will try to implement this approach and am open to any further suggestions. Thanks!

tkobayas:
Jenkins test again

tkobayas:
@lucamolteni @mariofusco @franz1981 
Introduced Factory for PhreakNodes and BetaConstaints (plus EvalCondition). So changes are isolated to those wrapper classes.

I implemented wrapper classes (XXXMetric) as simple sub classes for now. I wonder if I should implement them as a decorator pattern. If so, I think I need to change SingleBetaConstraints to an Interface and rename the current SingleBetaConstraints to SingleBetaConstraintsImpl, and then implement SingleBetaConstraintsMetric (and do it for other BetaConstaints etc.). Maybe too large changes...? Feel free to let me know your thoughts. Thanks!

decorator02

(will revisit sonarcloud warnings anyway...)

hellowdan (approved):
The code is quite understandable and I couldn't see any outstanding issue. I asked @jiripetrlik a review for a double check from QE side. Thanks!

tkobayas:
Jenkins test again

tkobayas:
Regarding Sonarcloud warnings,

- **1 duplicated blocks of code must be removed.**
  - Duplication among SingleBetaConstraintsMetric, DoubleBetaConstraintsMetric ... I don't have much to do because they are subclasses of different classes.
- **Remove this "clone" implementation; use a copy constructor or copy factory instead.**
  - Needed to implement clone() to follow the super class.
- **Use super.clone() to create and seed the cloned instance to be returned.**
  - Because EvalCondition.clone() doesn't call super.clone() (-> eventually Object.clone()), I cannot follow the guidance. So I implemented the similar logic as EvalCondition.clone(). (Note: I cannot make "cloned" protected because of another sonarcloud warning "non-serializable collection cannot be protected")
- **2 duplicated blocks of code must be removed.**
  - Original code of ReteDumper. Looks okay to me.
- **Replace this use of System.out or System.err by a logger**
  - ReteDumper uses System.out for convenience

tkobayas:
Note for test coverage.

As RuleNetworkEvaluator holds PhreakNodes as static final, I cannot write a unit test to enable/disable PerfLogUtils. So the current CI just ensures that "PerfLogUtils disabled (default)" doesn't break anything.

Instead, I tested whole drools repo with "PerfLogUtils enabled" on my end and confirmed no test failure.

Maybe it's worth discussing a strategy to test this kind of configuration/property switch which is not regularly tested. Probably alphanetworkCompiler is a good reference?

drools-core/src/main/java/org/drools/core/common/metric/DefaultBetaConstraintsMetric.java
```diff
@@ -0,0 +1,78 @@
...
+import org.drools.core.spi.BetaNodeFieldConstraint;
+import org.drools.core.spi.Tuple;
+import org.drools.core.util.PerfLogUtils;
+
+public class DefaultBetaConstraintsMetric extends DefaultBetaConstraints {
+
+    private static final long serialVersionUID = 510l;
+
+    public DefaultBetaConstraintsMetric() {
+        super();
```
jiripetrlik: I think that this constructor is called by default. Any reason for explicitly adding it here?

drools-core/src/main/java/org/drools/core/common/metric/DefaultBetaConstraintsMetric.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright 2005 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: I know it is a detail. But maybe please update 2005 to 2020.

drools-core/src/main/java/org/drools/core/common/metric/DoubleBetaConstraintsMetric.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2005 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: I know it is a detail. But maybe please update 2005 to 2020.

drools-core/src/main/java/org/drools/core/common/metric/QuadroupleBetaConstraintsMetric.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2005 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: I know it is a detail. But maybe please update 2005 to 2020.

drools-core/src/main/java/org/drools/core/common/metric/DefaultBetaConstraintsMetric.java
jiripetrlik: Isn't this called by default?

drools-core/src/main/java/org/drools/core/common/metric/QuadroupleBetaConstraintsMetric.java
```diff
@@ -0,0 +1,79 @@
...
+import org.drools.core.rule.ContextEntry;
+import org.drools.core.rule.MutableTypeConstraint;
+import org.drools.core.spi.BetaNodeFieldConstraint;
+import org.drools.core.spi.Tuple;
+import org.drools.core.util.PerfLogUtils;
+import org.kie.internal.conf.IndexPrecedenceOption;
+
+public class QuadroupleBetaConstraintsMetric extends QuadroupleBetaConstraints {
+
+    private static final long             serialVersionUID = 510l;
```
jiripetrlik: Is this indent intentional?

drools-core/src/main/java/org/drools/core/common/metric/SingleBetaConstraintsMetric.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2005 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: Please update date.

drools-core/src/main/java/org/drools/core/common/metric/TripleBetaConstraintsMetric.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2005 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: Please fix date.

drools-core/src/main/java/org/drools/core/phreak/PhreakNetworkNodeFactory.java
```diff
@@ -0,0 +1,135 @@
+package org.drools.core.phreak;
```
jiripetrlik, line 17: Missing license header.

drools-core/src/main/java/org/drools/core/phreak/PhreakNetworkNodeFactory.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private static final PhreakNetworkNodeFactory INSTANCE = new PhreakNetworkNodeFactory();
+
+    public static PhreakNetworkNodeFactory getInstance() {
+        return INSTANCE;
+    }
+
+    private PhreakNetworkNodeFactory() {}
+
+    PhreakJoinNode createPhreakJoinNode() {
```
jiripetrlik: Is missing access modifier intentional? Shouldn't we rather use public?

drools-core/src/main/java/org/drools/core/phreak/metric/PhreakAccumulateNodeMetric.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
```
jiripetrlik: Please fix date.

drools-core/src/main/java/org/drools/core/reteoo/ReteDumper.java
```diff
@@ -0,0 +1,291 @@
...
+    private void dumpNode( BaseNode node, String ident ) {
+        sb.append(ident + formatNode(node));
+        if (!nodeInfoOnly) {
+            sb.append(" on " + node.getPartitionId());
+            try {
+                Object declaredMask = node.getClass().getMethod("getDeclaredMask").invoke(node);
+                Object inferreddMask = node.getClass().getMethod("getInferredMask").invoke(node);
+                sb.append(" d "+declaredMask + " i " + inferreddMask);
+            } catch (NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+                // do nothing.
```
jiripetrlik: Is it ok to do nothing for all these exception types? For example Security or Illegal argument exceptions? Maybe it would be good to at least log something at debug or trace level. What do you think?

drools-core/src/main/java/org/drools/core/reteoo/ReteDumper.java
```diff
@@ -0,0 +1,291 @@
...
+            }
+            try {
+                Object declaredMask = node.getClass().getMethod("getLeftDeclaredMask").invoke(node);
+                Object inferreddMask = node.getClass().getMethod("getLeftInferredMask").invoke(node);
+                sb.append(" Ld "+declaredMask + " Li " + inferreddMask);
+            } catch (NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+                // do nothing.
```
jiripetrlik: Same as for previous try catch block.

drools-core/src/main/java/org/drools/core/reteoo/builder/BetaNodeConstraintFactory.java
```diff
@@ -0,0 +1,76 @@
+package org.drools.core.reteoo.builder;
```
jiripetrlik, line 17: Missing license header.

drools-core/src/main/java/org/drools/core/rule/EvalCondition.java
```diff
@@ -153,7 +153,7 @@ public boolean equals(final Object object) {
             return true;
         }
 
-        if ( object == null || object.getClass() != EvalCondition.class ) {
+        if ( object == null || object.getClass() != this.getClass() ) {
```
jiripetrlik, line 156: Only for curiosity. Is there any reason for this change? It seems both do the same.

drools-core/src/main/java/org/drools/core/rule/EvalConditionFactory.java
```diff
@@ -0,0 +1,23 @@
+package org.drools.core.rule;
```
jiripetrlik, line 17: Missing license header.

drools-core/src/main/java/org/drools/core/rule/metric/EvalConditionMetric.java
```diff
@@ -0,0 +1,51 @@
...
+import org.drools.core.rule.Declaration;
+import org.drools.core.rule.EvalCondition;
+import org.drools.core.spi.EvalExpression;
+import org.drools.core.spi.Tuple;
+import org.drools.core.util.PerfLogUtils;
+
+public class EvalConditionMetric extends EvalCondition {
+
+    public EvalConditionMetric() {
+        super();
```
jiripetrlik: Any reason for calling super here?

jiripetrlik (commented):
Hello, I have few minor comments or suggestions.

drools-core/src/main/java/org/drools/core/common/metric/DefaultBetaConstraintsMetric.java
tkobayas: Thank you for pointing this! Fixed

drools-core/src/main/java/org/drools/core/common/metric/DoubleBetaConstraintsMetric.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/common/metric/QuadroupleBetaConstraintsMetric.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/common/metric/QuadroupleBetaConstraintsMetric.java
tkobayas: Thanks! Formatted

drools-core/src/main/java/org/drools/core/common/metric/SingleBetaConstraintsMetric.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/common/metric/DefaultBetaConstraintsMetric.java
tkobayas: Thank you for pointing this! Fixed

drools-core/src/main/java/org/drools/core/common/metric/TripleBetaConstraintsMetric.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/phreak/PhreakNetworkNodeFactory.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/phreak/PhreakNetworkNodeFactory.java
tkobayas: Yes, it's intentional to be package private. Thanks!

drools-core/src/main/java/org/drools/core/phreak/metric/PhreakAccumulateNodeMetric.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/reteoo/builder/BetaNodeConstraintFactory.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/rule/EvalConditionFactory.java
tkobayas: Thanks! Fixed

drools-core/src/main/java/org/drools/core/rule/metric/EvalConditionMetric.java
tkobayas: Thank you for pointing this! Fixed

drools-core/src/main/java/org/drools/core/rule/EvalCondition.java
tkobayas: I wanted to compare with the right class (= if "this" is a sub class, e.g. EvalConditionMetric, I need to use "this.getClass()").

drools-core/src/main/java/org/drools/core/reteoo/ReteDumper.java
tkobayas: Hmm, I think that this part is used only for debugging purpose by core developers so it may not need to handle Exception. But anyway, I added trace logging as suggested so that other people will not need to worry in the future :) Thanks!

drools-core/src/main/java/org/drools/core/reteoo/ReteDumper.java
tkobayas: Added logging. Thanks!

mariofusco (approved):

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(PerfLogUtils.class);
+
+    public static final String PERF_LOGGER_ENABLED = "drools.performance.logger.enabled";
+    private boolean enabled = Boolean.parseBoolean(System.getProperty(PERF_LOGGER_ENABLED, "false"));
+
+    public static final String PERF_LOGGER_THRESHOLD = "drools.performance.logger.threshold";
+    private int threshold = Integer.parseInt(System.getProperty(PERF_LOGGER_THRESHOLD, "500")); // microseconds
+
+    private final ThreadLocal<NodeStats> nodeStats = new ThreadLocal<>();
```
mariofusco: I believe that having a ThreadLocal here is only necessary if the engine is running in parallel evaluation mode (which is not the default). Probably you could have a single plain NodeStats if parallel evaluation is not enabled.

tkobayas: @mariofusco I used ThreadLocal considering the case where a user executes multiple ksessions concurrently (e.g. in kie-server). But of course, we can limit the use of this feature only for single thread execution (= Raise an error log when PerfLogUtils is concurrently accessed). If you think it's better, I'll change. Please let me know your thought, thanks!

lucamolteni:
> @lucamolteni @mariofusco @franz1981
> Introduced Factory for PhreakNodes and BetaConstaints (plus EvalCondition). So changes are isolated to those wrapper classes.
> 
> I implemented wrapper classes (XXXMetric) as simple sub classes for now. I wonder if I should implement them as a decorator pattern. If so, I think I need to change SingleBetaConstraints to an Interface and rename the current SingleBetaConstraints to SingleBetaConstraintsImpl, and then implement SingleBetaConstraintsMetric (and do it for other BetaConstaints etc.). Maybe too large changes...? Feel free to let me know your thoughts. Thanks!

Regarding this, I think subclassing is perfectly fine. 
It's also what I did when changing the nodes in the descoping of the traits. Is the decorator pattern used elsewhere?

Thanks

lucamolteni:
> Instead, I tested whole drools repo with "PerfLogUtils enabled" on my end and confirmed no test failure.
> 
> Maybe it's worth discussing a strategy to test this kind of configuration/property switch which is not regularly tested. Probably alphanetworkCompiler is a good reference?

Probably, but AFAIK the alpha network compiler is never tested. Or maybe it's tested in a specific job. @hellowdan can you please check this? I think @baldimir did that.

lucamolteni (approved):
Toshiya, I liked how you used subclassing and factories to keep the original node unchanged. 

Do you think it could be possible to put the new nodes in a specific module? 
I think it could make sense to have it separate as this is needed only in specific cases.

Thanks!

tkobayas:
Thank you for the review, @lucamolteni 

> Is the decorator pattern used elsewhere?

No, I just thought that it could be another possible approach.

> Do you think it could be possible to put the new nodes in a specific module?

Sure, let me try...

drools-core/src/main/java/org/drools/core/util/PerfLogUtils.java
tkobayas: As discussed with Mario, we keep ThreadLocal for now

tkobayas:
Hi @lucamolteni 

I created a new module `drools-metric` and moved relevant codes into it. Does this meet your suggestion?

And I'd like to know what else I should do for a new module: 
 - Add to drools-bom (droolsjbpm-build-bootstrap), right?
 - Anything needed for Jenkins?
 - Anything needed for public maven repo?

Also please kindly check the line comments.

Any advises would be appreciated!

drools-metric/src/main/resources/META-INF/kie.conf
```diff
@@ -0,0 +1,3 @@
+org.drools.core.phreak.PhreakNetworkNodeFactory=org.drools.metric.phreak.MetricPhreakNetworkNodeFactoryImpl
+org.drools.core.reteoo.builder.BetaNodeConstraintFactory=org.drools.metric.reteoo.builder.MetricBetaNodeConstraintFactoryImpl
+org.drools.core.rule.EvalConditionFactory=org.drools.metric.rule.MetricEvalConditionFactoryImpl
```
tkobayas, line 3: kie.conf has 3 entries to switch Factory classes. Should I encapsulate them into another new top layer Factory in order to reduce the number of entries in kie.conf? Or I wondered if I should let KieComponentFactory have these factories but KieComponentFactory is not singleton so doesn't seem to fit (I think it's technically possible though).

drools-metric/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+            </Export-Package>
+            <Import-Package>
+              *
+            </Import-Package>
+            <Bundle-Activator>org.drools.ruleunit.osgi.Activator</Bundle-Activator>
+            <_removeheaders>Private-Package</_removeheaders>
+          </instructions>
+        </configuration>
+      </plugin>
+    </plugins>
```
tkobayas: I'm not confident if I wrote `maven-bundle-plugin` here correctly (This is the first time I wrote this).

tkobayas:
Jenkins please retest this

drools-metric/pom.xml
lucamolteni: FYI I didn't implement OSGi for `drools-traits` as @mariofusco told me we're actually kind of deprecating it

lucamolteni (approved):
Perfect job thank you for moving it to another module

mariofusco (approved):
great job!

jiripetrlik:
@lucamolteni @mariofusco 
What does it exactly mean for the engine running in OSGI. Does it mean that the engine capabilities will be somehow limited under OSGI? Can you be more specific. We should probably mention it in some docs etc.

tkobayas:
@lucamolteni I see that drools-traits creates bundle.

https://github.com/kiegroup/drools/blob/master/drools-traits/pom.xml#L13

Will you change it to `<packaging>jar<packaging>` and remove `maven-bundle-plugin`?

For drools-metric, I think it's okay to remove osgi bundle assuming osgi platform users can test their rules with standalone java. WDYT? @mariofusco

tkobayas:

tkobayas:

lucamolteni:
@jiripetrlik we discuss it yesterday with @etirelli and we're thinking of deprecating OSGi support. Supporting new modules only when needed seems like a good start doesn't it?

tkobayas:
FDB failed with one test. Quite likely not related to this PR.
```
org.jbpm.kie.services.test.FilteredKModuleDeploymentServiceTest.testMultipleRemotableInPojoJar

Failing for the past 4 builds (Since Failed#10 )

Error Message
com.sample.MyPojo

Stacktrace
java.lang.AssertionError: com.sample.MyPojo
	at org.junit.Assert.fail(Assert.java:88)
	at org.jbpm.kie.services.test.FilteredKModuleDeploymentServiceTest.testMultipleRemotableInPojoJar(FilteredKModuleDeploymentServiceTest.java:477)
        ...
```

drools-metric/pom.xml
tkobayas: As discussed, removed OSGi bundle for now.

jiripetrlik:
@lucamolteni Thank you for info about OSGI. I think it is in general good news.

drools-core/src/main/java/org/drools/core/reteoo/ReteDumper.java
```diff
@@ -0,0 +1,295 @@
...
+            }
+        }
+    }
+
+    private void dumpNode( BaseNode node, String ident ) {
+        sb.append(ident + formatNode(node));
+        if (!nodeInfoOnly) {
+            sb.append(" on " + node.getPartitionId());
+            try {
+                Object declaredMask = node.getClass().getMethod("getDeclaredMask").invoke(node);
```
franz1981, line 216: @lucamolteni I strongly suggest to look into `MethodHandle::invokeExact` that disturb much less the security manager while producing less garbage too

tkobayas: @franz1981 Thank you for the suggestion. I filed https://issues.redhat.com/browse/DROOLS-5772

PR #2905, opened 2020-05-20
[DROOLS-5346] Enhance unit test to detect non-externalized lambda widely

tkobayas:
This commit doesn't affect runtime or unit tests by default. If you enable both of 2 properties to 'true',
- drools.externaliseCanonicalModelLambda
- drools.check.nonExternalisedLambda

executable model throws NonExternalisedLambdaFoundException when generated source codes contain non-externalized lambdas.
 
For example, if you add the following code to BaseModelTest, you can easily check if unit tests generate non-externalized lambda or not.

```
public abstract class BaseModelTest {
    
    static {
        System.setProperty("drools.externaliseCanonicalModelLambda", "true");
        System.setProperty("drools.check.nonExternalisedLambda", "true");
    }
```

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/RuleWriter.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/RuleWriter.java
@@ -22,16 +22,21 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


tkobayas:
I have confirmed many non-externalized lambda cases by using this option locally. DROOLS-5347 and DROOLS-5348 will solve some of them but I will likely file more JIRAs.

lucamolteni (approved):

lucamolteni (approved):

hellowdan (approved):

PR #2908, opened 2020-05-21
[DROOLS-5348] Lambda predicate is not externalized when a binding var…

tkobayas:
…iable is involved

https://issues.redhat.com/browse/DROOLS-5348

The main fix is to add Types to lambda parameters because these Types are required for lambda externalization.

I also moved some tests from CompilerTest to ExternalisedLambdaTest for easier maintenance.

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
@@ -517,7 +517,7 @@ public class DrlxParseUtil {
     }
 
     private static boolean canResolveAllParameterTypes(Collection<String> usedDeclarations, boolean skipFirstParamAsThis, Optional<Class<?>> patternClass, RuleContext ruleContext) {
-        if (!skipFirstParamAsThis && (!patternClass.isPresent() || patternClass.get().equals(Object.class))) {
+        if (!skipFirstParamAsThis && !patternClass.isPresent()) {
             return false;
         }
         if (usedDeclarations.isEmpty()) {
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
@@ -477,24 +477,62 @@ public class DrlxParseUtil {
                                                              Expression expr,
                                                              boolean skipFirstParamAsThis,
                                                              Optional<Class<?>> patternClass) {
+        return generateLambdaWithoutParameters(usedDeclarations, expr, skipFirstParamAsThis, patternClass, null);
+    }
+    public static Expression generateLambdaWithoutParameters(Collection<String> usedDeclarations,
+                                                             Expression expr,
+                                                             boolean skipFirstParamAsThis,
+                                                             Optional<Class<?>> patternClass,
+                                                             RuleContext ruleContext) {
         DrlxParseUtil.transformDrlNameExprToNameExpr(expr);
         if (skipFirstParamAsThis && usedDeclarations.isEmpty()) {
             return expr;
         }
         LambdaExpr lambdaExpr = new LambdaExpr();
-        lambdaExpr.setEnclosingParameters( true );
+        lambdaExpr.setEnclosingParameters(true);
+
+        // Only when we can resolve all parameter types, do it
+        boolean canResolve = canResolveAllParameterTypes(usedDeclarations, skipFirstParamAsThis, patternClass, ruleContext);
         if (!skipFirstParamAsThis) {
             Type type;
-            if(patternClass.isPresent() && usedDeclarations.isEmpty() && patternClass.filter(c -> !Object.class.equals(c)).isPresent()) {
+            if (canResolve) {
                 type = StaticJavaParser.parseClassOrInterfaceType(patternClass.get().getCanonicalName());
             } else {
                 type = new UnknownType();
             }
             lambdaExpr.addParameter(new Parameter(type, THIS_PLACEHOLDER));
         }
-        usedDeclarations.stream().map( s -> new Parameter( new UnknownType(), s ) ).forEach( lambdaExpr::addParameter );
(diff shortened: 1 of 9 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
```diff
@@ -477,26 +477,64 @@ public static Expression generateLambdaWithoutParameters(Collection<String> used
...
+                            }
+                        })
+                        .forEach(lambdaExpr::addParameter);
+
+        lambdaExpr.setBody(new ExpressionStmt(expr));
         return lambdaExpr;
     }
 
+    private static boolean canResolveAllParameterTypes(Collection<String> usedDeclarations, boolean skipFirstParamAsThis, Optional<Class<?>> patternClass, RuleContext ruleContext) {
+        if (!skipFirstParamAsThis && (!patternClass.isPresent() || patternClass.get().equals(Object.class))) {
```
tkobayas: @lucamolteni This check (if patternClass is Object.class, don't use it. So the type of "_this" will be UnknownType) comes from your commit. Do you remember why this check is required? I tested without the check but unit tests seemed to work.
https://github.com/kiegroup/drools/commit/0ac6d1d8#diff-d29bcd88c3edecef8fa6351facea3f55R473

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -271,6 +271,13 @@ private void methodCallExprParameter(AccumulateDescr.AccumulateFunctionCallDescr
         final Class accumulateFunctionResultType = accumulateFunction.getResultType();
         final String bindExpressionVariable = context.getExprId(accumulateFunctionResultType, typedExpression.toString());
 
+        if (!decl.isPresent() && input instanceof PatternDescr) {
+            Optional<DeclarationSpec> accumulateClassDeclOpt = context.getDeclarationById(((PatternDescr)input).getIdentifier());
+            if (accumulateClassDeclOpt.isPresent()) {
+                clazz = accumulateClassDeclOpt.get().getDeclarationClass();
+            }
+        }
+
```
tkobayas: @mariofusco I added this code because AccumulateTest.testAccumulateOnStaticMethod failed with my PR. Without this code, SingleDrlxParseSuccess.patternType is "java.time.Duration" in the test. My PR expects "patternType" to be a Type of "_this" (Or this expectation is wrong?). So in the test, the type should be "org.drools.modelcompiler.AccumulateTest$Interval". I'm not very sure if this is the right way to fix. Please share your thought on this.

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -271,6 +271,13 @@ private void methodCallExprParameter(AccumulateDescr.AccumulateFunctionCallDescr
         final Class accumulateFunctionResultType = accumulateFunction.getResultType();
         final String bindExpressionVariable = context.getExprId(accumulateFunctionResultType, typedExpression.toString());
 
+        if (!decl.isPresent() && input instanceof PatternDescr) {
```
mariofusco: Sorry, but I don't understand why you needed to do this. Can you please clarify a bit? Also maybe adding a comment in the code could be helpful so we won't forget in future..

tkobayas: `!decl.isPresent()` means `rootNodeName` was not resolved to a declaration. It means that `rootNodeName` is not a variable but a class with its static method. I apply this logic (= change `clazz`) only for the static method case.

Because, for example, in `AccumulateTest#testAccumulateWithMax()`, `rootNodeName` "$time" is `Date`. It's not the pattern type `StockTick`. So we shouldn't change `clazz`. (In this case, the "_this" is eventually converted to `StockTick` by `AccumulateVisitorFlowDSL.composeTwoBindings`)

tkobayas: @mariofusco If you have any doubt, feel free to let me know. I'm not fully sure about the ramification of changing `clazz` (= patternType for drlxParse()). But in this case (= class with static method), the method is properly resolved even if I change the `clazz` from `Duration` to `AccumulateTest$Interval`

hellowdan (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
lucamolteni: It's probably unused code then, you can remove it thanks!

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
tkobayas: Thanks, removed.

tkobayas:
The downstream failure doesn't seem to be related to this PR.
```
11:28:48 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (ban-duplicated-classes) on project jbpm-wb-kie-server-api: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
11:28:48 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (ban-duplicated-classes) on project jbpm-wb-kie-server-api: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed.
```

PR #2916, opened 2020-05-25
[RHPAM-2960] Case instance update fails with java.io.NotSerializableE…

elguardian:
…xception: org.jbpm.services.task.persistence.JPATaskModelFactory for cases containing multiple SLAs

### DIFF
--- a/drools-core/src/main/java/org/drools/core/time/impl/DefaultTimerJobInstance.java
+++ b/drools-core/src/main/java/org/drools/core/time/impl/DefaultTimerJobInstance.java
@@ -62,6 +62,7 @@ public class DefaultTimerJobInstance
 
     public Void call() throws Exception {
         try {
+            this.trigger.initialize(this.getJobContext().getInternalKnowledgeRuntime().orElse(null));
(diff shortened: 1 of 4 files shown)


drools-core/src/main/java/org/drools/core/time/impl/DefaultTimerJobInstance.java
```diff
@@ -62,6 +62,7 @@ public int compareTo(DefaultTimerJobInstance o) {
 
     public Void call() throws Exception {
         try {
+            this.trigger.initialize(this.getJobContext().getWorkingMemory().getKnowledgeRuntime());
```
afalhambra: Not sure if there is any chance that a trigger has no context or context=null. In that case, this will throw a NPE. wdyt?

afalhambra (commented):
Just one comment in my review. Please have a look. Otherwise it looks good to me.

afalhambra:

gmunozfe (approved):
LGTM

afalhambra (approved):
Looks good to me. Nice job!

PR #2921, opened 2020-05-28
DROOLS-5331 & DROOLS-5078: Support for Enums in Rule Test Scenario

yesamer:
@danielezonca @dupliaka Can you please review and test it?

Description of the PR will be provided into DROOLS-WB related PR
Related PR: https://github.com/kiegroup/drools-wb/pull/1377

This PR fixes this ticket as well: https://issues.redhat.com/browse/DROOLS-5078
I changed the error message: `Impossible to parse  ..... is not natively supported. Please use an MVEL expression to use it`

https://issues.redhat.com/browse/BAPL-1617
https://issues.redhat.com/browse/DROOLS-5331


drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ScenarioSimulationSharedUtils.java
```diff
@@ -56,4 +60,13 @@ public static boolean isMap(String className) {
...
     }
+
+    /**
+     * Returns true if given string equals to canonical name of <code>java.lang.Enum</code>
+     * @param className
+     * @return
+     */
+    public static boolean isEnum(String className) {
+        return Enum.class.getCanonicalName().equals(className);
+    }
```
danielezonca: I think this method is quite misleading compared to similar n this utils (`isList`, `isMap` etc): all the others accept "target" class while this method works only if the parameter is the parent class. What about at least change the name?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
```diff
@@ -178,13 +178,18 @@ public static Object convertValue(String className, Object cleanValue, ClassLoad
...
-            throw new IllegalArgumentException(new StringBuilder().append("Impossible to parse '").append(value).append("' as ").append(className).append(" [").append(e.getMessage()).append("]").toString());
+            throw new IllegalArgumentException(new StringBuilder().append("Impossible to parse '")
+                                                       .append(value).append("' as ").append(className).append(" [")
+                                                       .append(e.getMessage()).append("]").toString());
         }
 
         throw new IllegalArgumentException(new StringBuilder().append("Class ").append(className)
-                                                   .append(" is not supported").toString());
+                                                   .append(" is not natively supported. Please use an MVEL expression" +
+                                                                   " to use it.").toString());
```
danielezonca, line 192: Can you please add `DROOLS-5078` to the title of this PR :)

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ScenarioSimulationSharedUtils.java
yesamer: @danielezonca I don't understand your point. In my opinion is not misleading: it checks if a given `className` is an `Enum`, disregarding if is called on a target / parent class.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
yesamer: @danielezonca Done!

drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/utils/ScenarioSimulationSharedUtils.java
danielezonca: My comment is mainly related comparing this method with the others in the same class:
`isList("java.uti.ArrayList")` -> `true`
`isMap("java.util.HashMap")` -> `true`
`isEnum("my.Enum")` -> `false`
Do you get what I mean?
Btw I don't have a better idea in terms of naming, I just find it a bit confusing

yesamer: @danielezonca Yes, now I got your point. Probably the correct name should be something similar to `isEnumCanonicalName()` but is very bad. 
On the others side, I think the scope of the 3 method are the same: you compared 2 case where you passed the exact `ArrayList` and `HashMap` canonical names, while in the third you passed a subclass of the `Enum`. 
Yes, you can argue that is not usual to subclass `ArrayList` or `HashMap` and it's common to use subclasses of `Enum` but all 3 methods check the canonical name in the same way.
Other names I can propose are:
`isEnumClass()`
`isExactelyEnum()`

yesamer: `isEnumCanonicalName()` maybe it's the clearest option, even if ugly: it clarifies which check will be applied.

yesamer: @danielezonca Change applied.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
```diff
@@ -178,13 +178,18 @@ public static Object convertValue(String className, Object cleanValue, ClassLoad
...
         } catch (RuntimeException e) {
-            throw new IllegalArgumentException(new StringBuilder().append("Impossible to parse '").append(value).append("' as ").append(className).append(" [").append(e.getMessage()).append("]").toString());
+            throw new IllegalArgumentException(new StringBuilder().append("Impossible to parse '")
+                                                       .append(value).append("' as ").append(className).append(" [")
+                                                       .append(e.getMessage()).append("]").toString());
         }
 
         throw new IllegalArgumentException(new StringBuilder().append("Class ").append(className)
-                                                   .append(" is not supported").toString());
+                                                   .append(" is not natively supported. Please use an MVEL expression" +
```
dupliaka, line 191: Just an idea: maybe we can provide a link for MVEL expression? BCS well as I understand that not every illegal argument could be replaced by MVEL.

yesamer: @dupliaka What do you have in mind? Consider the message is shown inside that popover: https://issues.redhat.com/secure/attachment/12465830/image-2020-02-19-10-13-01-440.png

drools-scenario-simulation/drools-scenario-simulation-api/src/test/java/org/drools/scenariosimulation/api/utils/ScenarioSimulationSharedUtilsTest.java
```diff
@@ -52,4 +52,9 @@ public void isList() {
     public void isMap() {
         assertTrue(mapValues.stream().allMatch(ScenarioSimulationSharedUtils::isMap));
     }
+
+    @Test
+    public void isEnumCanonicalName() {
+        assertTrue(ScenarioSimulationSharedUtils.isEnumCanonicalName(Enum.class.getCanonicalName()));
+    }
```
dupliaka, line 61: Suggestion:
From my perspective that looks a bit like code repetition.

Would you please tell me why cant we use 

`        assertTrue(ScenarioSimulationSharedUtils.isEnumCanonicalName("java.lang.Enum"));
        assertFalse(ScenarioSimulationSharedUtils.isEnumCanonicalName("Enum"));`

yesamer: @dupliaka Of course we can use it :)

dupliaka (approved):
Approve, with two suggestions

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioBeanUtil.java
dupliaka: I meant to wrap http://mvel.documentnode.com/ link around MVEL expression if that possible?

danielezonca: Please consider we have users with offline configurations so a link to an external website is not ideal. MVEL is used in the whole platform (for example process editor) so we can assume user knows it or that in general we don't need to solve in a specific editor a problem that is at platform level

dupliaka: @danielezonca thanks for mention it, I was unavare of this.

dupliaka:
"Error message: Class *** is not natively supported. Please use an MVEL expression to use it." shows
- [x] Date
- [x] LocalTime
- [x] LocalDateTime
- [x] OffsetDateTime
- [x] BigInteger
- [x] BigDecimal 

It does not appear for:
boolean, byte, char, float, double, int, short, long, Boolean, Character, Double, Float, Integer, LocalDate, Long, Short, String

danielezonca (approved):

yesamer:
@danielezonca Can you please merge it?

yesamer:
@danielezonca Green light :)

PR #2925, opened 2020-06-01
[DROOLS-5386] Accumulate with static method with pettern bind variable

tkobayas:
https://issues.redhat.com/browse/DROOLS-5386

This PR uses pattern input identifier ("$p") and its type (Person) when parsing the expression with drlxParse(). So $p is properly resolved as "_this" so it will not result in duplication.

I thought that this approach could be applied to all cases (not only static method) but if I applied, I got some failures in FLOW_DSL (I will file a JIRA and revisit). So I decided to apply only for a static method case to minimize the effect for now.


tkobayas:
downstream failure doesn't seem to be related to this PR.
```
05:17:03 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (ban-duplicated-classes) on project jbpm-wb-kie-server-api: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
05:17:03 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (ban-duplicated-classes) on project jbpm-wb-kie-server-api: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed.
```

lucamolteni (approved):

hellowdan (approved):

PR #2926, opened 2020-06-01
[DROOLS-5324] refactor building of non-native resources from the exec…

mariofusco:
…utable model

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
@@ -61,8 +61,6 @@ public abstract class AbstractKieProject implements KieProject {
 
     private static final Predicate<String> BUILD_ALL = s -> true;
 
-    private Predicate<String> buildFilter = BUILD_ALL;
-
     public ResultsImpl verify() {
         ResultsImpl messages = new ResultsImpl();
         verify(messages);
@@ -212,8 +210,11 @@ public abstract class AbstractKieProject implements KieProject {
         }
     }
 
-    public KnowledgeBuilder buildKnowledgePackages( KieBaseModelImpl kBaseModel,
-                                                    ResultsImpl messages ) {
+    public KnowledgeBuilder buildKnowledgePackages( KieBaseModelImpl kBaseModel, ResultsImpl messages ) {
+        return buildKnowledgePackages( kBaseModel, messages, BUILD_ALL );
+    }
+
+    public KnowledgeBuilder buildKnowledgePackages( KieBaseModelImpl kBaseModel, ResultsImpl messages, Predicate<String> buildFilter ) {
         InternalKieModule kModule = getKieModuleForKBase(kBaseModel.getName());
         KnowledgeBuilderImpl kbuilder = ( KnowledgeBuilderImpl ) createKnowledgeBuilder( kBaseModel, kModule );
         if (kbuilder == null) {
@@ -239,7 +240,7 @@ public abstract class AbstractKieProject implements KieProject {
                 continue;
             }
             if (compileIncludedKieBases()) {
-                addFiles( assets, getKieBaseModel( include ), includeModule, useFolders );
+                addFiles( buildFilter, assets, getKieBaseModel( include ), includeModule, useFolders );
             }
         }
 
@@ -247,14 +248,16 @@ public abstract class AbstractKieProject implements KieProject {
             return null;
         }
 
-        addFiles( assets, kBaseModel, kModule, useFolders );
+        addFiles( buildFilter, assets, kBaseModel, kModule, useFolders );
 
         CompositeKnowledgeBuilder ckbuilder = kbuilder.batch();
 
         // cache KnowledgeBuilder and results
-        kModule.cacheKnowledgeBuilderForKieBase(kBaseModel.getName(), kbuilder);
-        kModule.cacheResultsForKieBase(kBaseModel.getName(), messages);
-
+        if (buildFilter == BUILD_ALL) {
+            kModule.cacheKnowledgeBuilderForKieBase( kBaseModel.getName(), kbuilder );
+            kModule.cacheResultsForKieBase( kBaseModel.getName(), messages );
+        }
+        
         if (assets.isEmpty()) {
             if (buildFilter == BUILD_ALL) {
                 log.warn( "No files found for KieBase " + kBaseModel.getName() +
@@ -292,11 +295,7 @@ public abstract class AbstractKieProject implements KieProject {
         return KnowledgeBuilderFactory.newKnowledgeBuilder( getBuilderConfiguration( kBaseModel, kModule ) );
     }
 
-    public void setBuildFilter( Predicate<String> buildFilter ) {
-        this.buildFilter = buildFilter;
-    }
-
-    private void addFiles( Set<Asset> assets, KieBaseModel kieBaseModel,
+    private void addFiles( Predicate<String> buildFilter, Set<Asset> assets, KieBaseModel kieBaseModel,
                            InternalKieModule kieModule, boolean useFolders) {
         for (String fileName : kieModule.getFileNames()) {
             if (buildFilter.test( fileName ) && !fileName.startsWith(".") && !fileName.endsWith(".properties") &&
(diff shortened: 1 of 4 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -205,30 +213,26 @@ public InternalKnowledgeBase createKieBase(KieBaseModelImpl kBaseModel, KieProje
...
-                final InternalKnowledgePackage originalPackage = kieBase.getPackage(pk.getName());
-                if (originalPackage != null && isInternalKnowldgePackage && ((InternalKnowledgePackage) pk).getResourceTypePackages().get(ResourceType.PMML) != null) {
-                    originalPackage.getResourceTypePackages().put(ResourceType.PMML, ((InternalKnowledgePackage) pk).getResourceTypePackages().get(ResourceType.PMML));
-                } else if (originalPackage == null) {
-                    kieBase.addPackages(pkgs);
+                final InternalKnowledgePackage originalPackage = kieBase.getPackage( pk.getName() );
+                if ( originalPackage != null && isInternalKnowldgePackage && (( InternalKnowledgePackage ) pk).getResourceTypePackages().get( ResourceType.PMML ) != null ) {
+                    originalPackage.getResourceTypePackages().put( ResourceType.PMML, (( InternalKnowledgePackage ) pk).getResourceTypePackages().get( ResourceType.PMML ) );
+                } else if ( originalPackage == null ) {
+                    kieBase.addPackage( pk );
```
danielezonca, line 230: Is it possible to remove this workaround too hardcoded on `ResourceType.PMML` and use the generic new `isNative` condition?

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -205,30 +213,26 @@ public InternalKnowledgeBase createKieBase(KieBaseModelImpl kBaseModel, KieProje
...
         InternalKnowledgeBase kieBase = new KieBaseBuilder(kBaseModel, kBaseConf).createKieBase(kpkgs);
 
-        if (hasNonModelResources(kBaseModel, kieProject)) {
-            KnowledgePackagesBuildResult knowledgePackagesBuildResult = ((AbstractKieModule) internalKieModule).buildKnowledgePackages(kBaseModel, kieProject, messages);
-            if (knowledgePackagesBuildResult.hasErrors()) {
-                return null;
-            }
-            Collection<KiePackage> pkgs = knowledgePackagesBuildResult.getPkgs();
-            for (KiePackage pk : pkgs) {
+        buildNonNativeResources( kBaseModel, kieProject, messages, kieBase );
```
danielezonca, line 215: After this change this additional build will be always invoked even if there are no DMN/PMML resources to build.
Internally it will become a no-op build in this case but I guess if we are adding visible overhead or not.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
```diff
@@ -234,30 +240,35 @@ public KnowledgeBuilder buildKnowledgePackages( KieBaseModelImpl kBaseModel,
...
-        addFiles( assets, kBaseModel, kModule, useFolders );
+        addFiles( buildFilter, assets, kBaseModel, kModule, useFolders );
 
         CompositeKnowledgeBuilder ckbuilder = kbuilder.batch();
 
+        // cache KnowledgeBuilder and results
+        if (buildFilter == BUILD_ALL) {
+            kModule.cacheKnowledgeBuilderForKieBase( kBaseModel.getName(), kbuilder );
+            kModule.cacheResultsForKieBase( kBaseModel.getName(), messages );
+        }
```
danielezonca, line 259: Why this is required only when we are in `BUILD_ALL` execution?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
```diff
@@ -58,6 +59,8 @@
 
     private final Map<String, KieSessionModel>   kSessionModels             = new HashMap<>();
 
+    private static final Predicate<String> BUILD_ALL = s -> true;
```
danielezonca, line 62: I think `BUILD_ALL` is misleading as name because if I have understood correctly, this means build all (only) canonical model related resources. Am I right?

danielezonca:
@mariofusco 
Please remember to forward port this PR to Kogito too when merged

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
mariofusco: Before the if condition traversed all the files in the project and if there was any dmn or pmml file it launched the build. Now the build is lunched regardless, it internally collects the dmn and pmml files to be built and if there isn't any it returns immediately, so I believe that the performances of the 2 implementations are very similar and this second one is more readable.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -93,7 +93,14 @@
     public static final String MODEL_FILE_DIRECTORY = "META-INF/kie/";
     public static final String MODEL_FILE_NAME = "drools-model";
     public static final String MODEL_VERSION = "Drools-Model-Version:";
+
     private static final String PROJECT_MODEL_RESOURCE_CLASS = PROJECT_MODEL_CLASS.replace('.', '/') + ".class";
+
+    private static final Predicate<String> NON_MODEL_RESOURCES = res -> {
```
gitgabrio, line 99: @mariofusco 
Would not be easier to just list PMML and DMN here ?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
mariofusco: You want to cache and reuse the kbuilder (that's necessary for incremental compilation) only if you did a complete build.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
mariofusco: @gitgabrio I don't think so, because at that point you will have another condition buried in our implementation (which is exactly the thing that we wanted to avoid) instead of being visible in the resource type to be built. What will happen when somebody else will add a third resource with same characteristic of pmml and dmn?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
danielezonca: Yes but if I have understood correctly, this doesn't really mean "complete" because then an additional build for not canonical model resources is needed. Am I right?

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
gitgabrio: @mariofusco 
TBH I've the impression this PR does not really address the main issue.
The flow (as I can understand) is:
`CanonicalKieModule.createKieBase` -> 

1. build CanonicalKiePackages out of "drl" resources
2. invoke `CanonicalKieModule.buildNonNativeResources` (that currently build DMN and PMML)

`CanonicalKieModule.buildNonNativeResources` invokes `AbstractKieProject.buildKnowledgePackages` (line 217) with the `Predicate` as last parameter. 
But that same method is also invoked by `AbstractKieProject.buildKnowledgePackages` (line 213) with a `BUILD_ALL` predicate.
So on one side it seems to be *possible* to invoke AbstractKieProject.buildKnowledgePackages just once, to build *all* the resources, and on the other side we are keeping this separated inside the CanonicalKieModule.
All this is very confusing IMO. For example, inside PMML I've a lot of different "models" - some of them are built as drools, some not - but this is decided by the implementation. 
So, I think that `CanonicalKieModule` should just ask AbstractKieProject to build every resource it finds, invoking the appropriate _builder_, and let the latter manage how  the package is built and what it is put inside.

PR #2931, opened 2020-06-04
[JBPM-9178] Immutable SpringBoot Deployment

elguardian:


elguardian:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/443 
https://github.com/kiegroup/kie-soup/pull/142
https://github.com/kiegroup/droolsjbpm-knowledge/pull/443

mattdkerr (approved):

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
```diff
@@ -98,17 +108,25 @@ public long getCreationTimestamp() {
...
             try {
-                e = classLoader.getResources(configFile );
+                ClassLoader currentClassLoader = classLoader;
+                while (currentClassLoader != null) {
+                    Enumeration<URL> list = currentClassLoader.getResources(configFile);
+                    while (list.hasMoreElements()) {
+                        resources.add(list.nextElement());
+                    }
+                    currentClassLoader = currentClassLoader.getParent();
+                }
```
afalhambra, line 121: Not sure why we want to access to all parents for the classloader? We might get a SecurityException is some cases.

drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
```diff
@@ -490,16 +491,23 @@ private void writeJarEntries(Folder f,
...
                 byte[] contents = getFileContents( (MemoryFile) rs );
                 if (contents == null) {
                     IOException e = new IOException("No content found for: " + rs);
                     log.error(e.getMessage(), e);
                     throw e;
                 }
-                out.write( contents );
+
+
+                CRC32 crc = new CRC32();
```
afalhambra: Why using CRC32 now? not sure about this change? Is this related to the issue itself or it's just an improvement of the same? can you please clarify?

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -30,50 +32,63 @@
...
+    private static final Logger log = LoggerFactory.getLogger(KieModuleMetaDataImpl.class);
+
     private final Map<String, Collection<String>> classes = new HashMap<>();
 
     private final Map<String, String> processes = new HashMap<>();
 
     private final Map<String, String> forms = new HashMap<>();
 
-    private final Map<URI, File> jars = new HashMap<>();
+    private final Set<URL> jars = new HashSet<>();
```
afalhambra, line 71: Should be changed to `Set<URI>` instead as per Sonar complain, just in case

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,207 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
 
+        List<AFReleaseId> releasesId = new ArrayList<>();
         if (releaseId != null) {
-            addArtifact(artifactResolver.resolveArtifact(releaseId));
+            releasesId.add(releaseId);
         }
         if ( kieModule != null && kieModule.getPomModel() != null ) {
             for ( AFReleaseId releaseId : kieModule.getPomModel().getDependencies(dependencyFilter) ) {
-                addArtifact( artifactResolver.resolveArtifact( releaseId ) );
+                releasesId.add(releaseId);
```
afalhambra: Loop can be replaced with `Collections.addAll`

```suggestion
releasesId.addAll(kieModule.getPomModel().getDependencies(dependencyFilter));
```

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,207 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
         packages.addAll(classes.keySet());
         packages.addAll(rulesByPackage.keySet());
     }
 
-    private void addArtifact(Artifact artifact) {
-        if (artifact != null && artifact.getExtension() != null && artifact.getExtension().equals("jar")) {
-            File jarFile = artifact.getFile();
-            addJar( jarFile, jarFile.toURI() );
+    private void addArtifact(ArtifactLocation artifactLocation) {
+        if (artifactLocation != null && artifactLocation.getArtifact().getExtension() != null && artifactLocation.getArtifact().getExtension().equals("jar")) {
```
afalhambra, line 294: `artifactLocation.getArtifact()` might be `null` due to:

https://github.com/kiegroup/kie-soup/blob/6a307c900c067c8fb4d312a8b9cc03f4e243bd2f/kie-soup-maven-utils/kie-soup-maven-integration/src/main/java/org/appformer/maven/integration/DefaultArtifactResolver.java#L54

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,207 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
             for ( DependencyDescriptor dep : artifactResolver.getAllDependecies( dependencyFilter ) ) {
-                addArtifact( artifactResolver.resolveArtifact( dep.getReleaseId() ) );
+                releasesId.add(dep.getReleaseId());
             }
         }
 
+        classesMap = new HashMap<>();
+        for (AFReleaseId rId : releasesId) {
+            ArtifactLocation artifactLocation = artifactResolver.resolveArtifactLocation(rId);
+            if (artifactLocation.isClassPath()) {
```
afalhambra: `artifactLocation` might be `null` due to:

https://github.com/kiegroup/kie-soup/blob/6a307c900c067c8fb4d312a8b9cc03f4e243bd2f/kie-soup-maven-utils/kie-soup-maven-integration/src/main/java/org/appformer/maven/integration/DefaultArtifactResolver.java#L54

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,207 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
+
+                int read = 0;
+                while ((read = zipFile.read(buffer)) > 0) {
+                    out.write(buffer, 0, read);
+                }
+
+                byte[] blob = out.toByteArray();
+                if (blob.length == 0) {
+                    continue;
+                }
```
afalhambra, line 328: Why not using the previous method to do this? 

https://github.com/kiegroup/drools/blob/8037cfba371213e74c3656228422079d09b1d9cb/drools-core/src/main/java/org/drools/core/util/IoUtils.java#L209

Is it only to keep on reading zip entries if those additional validations are met? why do we want this?

afalhambra (changes requested):
Overall looks good to me. I left some comments for you to review.

Also, test code coverage should reach 80% at least. Specially for the two following classes:

https://github.com/kiegroup/drools/blob/9a837e9d6293e8803416b08623f71f48e0ea3577/kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java

and

https://github.com/kiegroup/drools/blob/9a837e9d6293e8803416b08623f71f48e0ea3577/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieServicesImpl.java

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: when you are reading within a file that method does not work. You need to support jar within a jar... The older method was not supporting that.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/ClasspathKieProject.java
elguardian: you need a security manager installed. Much of the code would not work in that case. We would need to setup properly in that case.

drools-compiler/src/main/java/org/drools/compiler/compiler/io/memory/MemoryFileSystem.java
elguardian: removed

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: ok

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: cool

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: not really, it is more convenient to set it as URL (we are using that anyway.

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: done

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,206 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
-            for ( AFReleaseId releaseId : kieModule.getPomModel().getDependencies(dependencyFilter) ) {
-                addArtifact( artifactResolver.resolveArtifact( releaseId ) );
-            }
+            releasesId.addAll(kieModule.getPomModel().getDependencies(dependencyFilter));
+
         } else {
-            for ( DependencyDescriptor dep : artifactResolver.getAllDependecies( dependencyFilter ) ) {
-                addArtifact( artifactResolver.resolveArtifact( dep.getReleaseId() ) );
+            List<AFReleaseId> dependencies = artifactResolver.getAllDependecies(dependencyFilter).stream().map(e -> e.getReleaseId()).collect(toList());
+            releasesId.addAll(dependencies);
```
afalhambra, line 253: Do we really want to add all transitive dependencies? I guess user should carefully add them at their own risk, but we could end up having duplicate deps in different places and locations (BOOT-INF/lib and BOOT-INF/classes/KIE-INF/lib)

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
```diff
@@ -133,96 +148,206 @@ public TypeMetaInfo getTypeMetaInfo(Class<?> clazz) {
...
+            addJar(artifactLocation.toURL());
         }
     }
 
-    private void addJar( File jarFile, URI uri ) {
-        if (!jars.containsKey(uri)) {
-            jars.put(uri, jarFile);
-            scanJar(jarFile);
+    private void addJar(URL location) {
+        log.info("ADD Artifact {} to {}", location, jars);
```
afalhambra: I would remove this log. It's better with the log used below, when the jar is actually added to the HashSet jars. Also it is not an artifact but an artifact location.

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
afalhambra: sorry I cannot see the change?

afalhambra (changes requested):
Some additional comments.

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: yep you need them. Otherwise won't be available at classloader

afalhambra: just wondering - is it not possible to use the same location for all deps? like `BOOT-INF/lib` folder instead of having two different locations?

kie-ci/src/main/java/org/kie/scanner/KieModuleMetaDataImpl.java
elguardian: ok. set as debug.

elguardian:
Jenkins retest this please.

ederign:
@elguardian looks like this commit is the root cause of https://issues.redhat.com/browse/DROOLS-5574 that is actually blocking BC. Did the CI pass before merge it?

PR #2936, opened 2020-06-04
[Drools-5373] Refactor PMMLAssembler with delegates 

gitgabrio:
@danielezonca @jiripetrlik @mariofusco 

see https://issues.redhat.com/browse/DROOLS-5373

This PR simply separate code from PMMLAssemblerService in two delegates
1) PMMLCompilerService is meant to invoke the PMMLCompiler
2) PMMLLoaderService is meant to load resources from classpath

Since it is not exactly known, yet, how they will work inside Kogito, this PR is just a structure refactoring - when requirements for Kogito will be clearer code will be adapted if/where needed

### DIFF
--- a/kie-pmml-new/kie-pmml-evaluator/kie-pmml-evaluator-assembler/src/main/java/org/kie/pmml/evaluator/assembler/service/PMMLAssemblerService.java
+++ b/kie-pmml-new/kie-pmml-evaluator/kie-pmml-evaluator-assembler/src/main/java/org/kie/pmml/evaluator/assembler/service/PMMLAssemblerService.java
@@ -15,11 +15,9 @@
  */
 package org.kie.pmml.evaluator.assembler.service;
 
-import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
 import org.drools.compiler.compiler.PackageRegistry;
@@ -31,12 +29,7 @@ import org.kie.api.io.Resource;
 import org.kie.api.io.ResourceConfiguration;
 import org.kie.api.io.ResourceType;
 import org.kie.api.io.ResourceWithConfiguration;
-import org.kie.pmml.commons.exceptions.ExternalException;
-import org.kie.pmml.commons.exceptions.KiePMMLException;
 import org.kie.pmml.commons.model.KiePMMLModel;
-import org.kie.pmml.compiler.commons.factories.KiePMMLModelFactory;
-import org.kie.pmml.compiler.executor.PMMLCompiler;
-import org.kie.pmml.compiler.executor.PMMLCompilerImpl;
 import org.kie.pmml.evaluator.api.container.PMMLPackage;
 import org.kie.pmml.evaluator.assembler.container.PMMLPackageImpl;
 import org.slf4j.Logger;
@@ -44,6 +37,12 @@ import org.slf4j.LoggerFactory;
 
 import static org.kie.pmml.commons.utils.KiePMMLModelUtils.getSanitizedClassName;
 import static org.kie.pmml.commons.utils.KiePMMLModelUtils.getSanitizedPackageName;
+import static org.kie.pmml.evaluator.assembler.service.PMMLCompilerService.getKiePMMLModelsCompiledFromResource;
+import static org.kie.pmml.evaluator.assembler.service.PMMLCompilerService.getKiePMMLModelsCompiledFromResourcesWithConfigurations;
+import static org.kie.pmml.evaluator.assembler.service.PMMLCompilerService.getKiePMMLModelsFromResourceFromPlugin;
+import static org.kie.pmml.evaluator.assembler.service.PMMLCompilerService.getKiePMMLModelsFromResourcesWithConfigurationsFromPlugin;
+import static org.kie.pmml.evaluator.assembler.service.PMMLLoaderService.getKiePMMLModelsLoadedFromResource;
+import static org.kie.pmml.evaluator.assembler.service.PMMLLoaderService.getKiePMMLModelsLoadedFromResourcesWithConfigurations;
 
 public class PMMLAssemblerService implements KieAssemblerService {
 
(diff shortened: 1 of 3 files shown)


kie-pmml-new/kie-pmml-evaluator/kie-pmml-evaluator-assembler/src/main/java/org/kie/pmml/evaluator/assembler/service/PMMLAssemblerService.java
```diff
@@ -66,7 +80,11 @@ public void addResources(Object kbuilder, Collection<ResourceWithConfiguration>
         if (isBuildFromMaven()) {
             addModels(kbuilderImpl, getKiePMMLModelsFromResourcesWithConfigurationsFromPlugin(kbuilderImpl, resources));
         } else {
-            addModels(kbuilderImpl, getKiePMMLModelsFromResourcesWithConfigurations(kbuilderImpl, resources));
+            List<KiePMMLModel> toAdd = getKiePMMLModelsLoadedFromResourcesWithConfigurations(kbuilderImpl, resources);
+            if (toAdd.isEmpty()) {
+                toAdd = getKiePMMLModelsCompiledFromResourcesWithConfigurations(kbuilderImpl, resources);
+            }
+            addModels(kbuilderImpl, toAdd);
```
danielezonca, line 87: Can you please clarify the logic?
Does it try to load (aka parse and compile) from resources and only if there are no file it tries to load compiled?
I think we should identify if it is a compiled kjar or not and decide what to do.
@mariofusco 
How does it work with Drools core?
Does it try both load from resource and compiled or just one of them?

kie-pmml-new/kie-pmml-evaluator/kie-pmml-evaluator-assembler/src/main/java/org/kie/pmml/evaluator/assembler/service/PMMLAssemblerService.java
```diff
@@ -77,7 +95,11 @@ public void addResource(Object kbuilder, Resource resource, ResourceType type, R
         if (isBuildFromMaven()) {
             addModels(kbuilderImpl, getKiePMMLModelsFromResourceFromPlugin(kbuilderImpl, resource));
         } else {
-            addModels(kbuilderImpl, getKiePMMLModelsFromResource(kbuilderImpl, resource));
+            List<KiePMMLModel> toAdd = getKiePMMLModelsLoadedFromResource(kbuilderImpl, resource);
+            if (toAdd.isEmpty()) {
+                toAdd = getKiePMMLModelsCompiledFromResource(kbuilderImpl, resource);
+            }
+            addModels(kbuilderImpl, toAdd);
```
danielezonca, line 102: Same as above

kie-pmml-new/kie-pmml-evaluator/kie-pmml-evaluator-assembler/src/main/java/org/kie/pmml/evaluator/assembler/service/PMMLAssemblerService.java
gitgabrio: @danielezonca 
Scope of the "Assembler" is to put  "KiePMMLModels" inside "Packages";  such Packages are put inside a KieBuilderImpl that - in turns, it is used to build and populate a "KieBase".
How this KieBase is used depends on the original caller

1.  in case of "plugin" - models with sources are put inside the package and then the plugin manage such sources to have them compiled.
2. If there is already a generated factory class (i.e. loaded from a kjar) a new instance is created out of it (no compilation) and then such class is used to retrieve the actual kiepmmlmodels that are put inside the package.
3. last, if none of the above is true (not invoked by plugin and not found in kjar), the PMML model is compiled and the result (KiePMMLModels) are put in the package.

The only thing that this PR add is the split of logic between "compilation" and "loading" - that before was mixed in the PMMLAssembler.

In the present context "load" and "compile" are the two opposite operations.

danielezonca: Thanks for the clarification, I'm not sure about point 3: it should be possible to recognize if the kjar has been already compiled or not to decide if it has to compile it. With the current logic I am worried that in case of error with the plugin or similar, the error will be hidden by this `getOrCreate` logic.
It could be done with a different ticket of course, I'm just trying to understand the whole flow :)

gitgabrio: @danielezonca 
Short answer: I think this is the cleanest way
Long answer
There are the following things to consider
1) kjar may have any name (we can't force user to give a specific name to his/her kjar)
2) there can be multiple kjars (and that enforce the previous issue)
3) given that, only way to know if a given kjar is in the current classpath is to look for something that is expected to be *inside* the kjar itself
4) this could be for example a text file, or whatever, loaded with a `getResource` or similar
5) after all that, the _actual_ logic would be exactly the same, only more contrived

The other possible solution I see could be to implement a _declarative_ approach: the user defines _somewhere_ the kjar(s) or class(es) is expecting to find in the classpath; this "variable" is read by the PMMLAssembler and then this latter try to figure out if such artifact(s)/class(es) are actually present. Again, I can't see any *real* benefit, without an _apparence_ of control: the need is to rise an Exception when the user try to evaluate a PMML that is not present in the classloader, and this is already in place. With this _declarative_ approach we would only create a more complex matrix of possible error conditions to manage.

Last note about the `getOrCreate` logic:
this works _only_ if the users put the PMML file inside the project . If there is not such PMML file, the model won't be compiled, and then the issue won't be hided. On the other side, if such file is present in the project, I'm expecting user to have put it willingly.
If the user put *both* a kjar *and* a *PMML* file in the project, there is anyway a log telling him/her if the model has been _loaded_ or _compiled_, and this is done on a per-file base

danielezonca (approved):

mariofusco (approved):

jiripetrlik (approved):
Looks good to me.

PR #2937, opened 2020-06-05
[DROOLS-5345] fix class loading issue in Eclipse Equinox OSGi container

mariofusco:


mariofusco:

PR #2941, opened 2020-06-10
[DROOLS-5350] Externalize lambda for "from" clause

tkobayas:


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
```diff
@@ -148,6 +149,8 @@
     private Map<String, CreatedClass> lambdaClasses = new HashMap<>();
     private Set<RuleUnitDescription> ruleUnits = new HashSet<>();
 
+    private Map<LambdaExpr, java.lang.reflect.Type> lambdaReturnTypes = new HashMap<>();
```
tkobayas, line 152: In case of "`from`", we cannot infer the return type in `ExecModelLambdaPostProcessor` like other cases. So I introduced `lambdaReturnTypes` to hold return type of lambda. `lambdaReturnTypes` is populated during parsing (`FromVisitor`) and then we can use it in `ExecModelLambdaPostProcessor` later.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
```diff
@@ -496,7 +496,7 @@ public static Expression generateLambdaWithoutParameters(Collection<String> used
         if (!skipFirstParamAsThis) {
             Type type;
             if (canResolve) {
-                type = StaticJavaParser.parseClassOrInterfaceType(patternClass.get().getCanonicalName());
+                type = StaticJavaParser.parseType(patternClass.get().getCanonicalName());
```
tkobayas, line 499: Changed to `parseType()` so it can handle array types as well.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
```diff
@@ -178,6 +187,52 @@ private void convertBindCallForFlowDSL(MethodCallExpr methodCallExpr) {
...
+            if (argType instanceof Class) {
+                // java.util.List<org.drools.FromTest$MyPerson> has to be resolved to canonical name java.util.List<org.drools.FromTest.MyPerson>
+                returnTypeStr = canonicalNameParameterizedType(parameterizedType, (Class<?>)argType);
+            } else {
+                return; // e.g. java.util.Collection<V> (V is TypeVariable), nested ParameterizedType, GenericArrayType etc.
+            }
+        } else {
+            return; // e.g. GenericArrayType etc.
+        }
+
```
tkobayas, line 236: This conditional branching is to handle cases where return types are Class (e.g. `Person`) or collection of classes (e.g. `List<Person>`). Special cases like `java.util.Collection<V>` will be left non-externalized so `FromTest.testFromMapValues()` is not yet externalized. I'll file a JIRA with minor priority.

tkobayas:
Jenkins test again

tkobayas:
Jenkins test again

tkobayas:
Main addition is `PackageModel.lambdaReturnTypes` to hold return type of lambda. Please see line comments. Thanks!

hellowdan (approved):

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/RuleWriter.java
```diff
@@ -85,6 +85,7 @@ public String getMainSource() {
                                 pkgModel.getRulesFileNameWithPackage(),
                                 pkgModel.getImports(),
                                 pkgModel.getStaticImports(),
+                                pkgModel.getLambdaReturnTypes(),
```
mariofusco: Just a suggestion: what about refactoring ExecModelLambdaPostProcessor and to it a constructor that simply takes the whole PackageModel?

tkobayas: Thank you for the suggestion. I added a new constructor. (The old constructor remains because it is used by a unit test)

lucamolteni: There was a reason why the constructor didn't take the whole `PackageModel` that is the PackageModel class is gigantic and to couple directly was a bad idea IMHO.

Nevertheless, I can see why having many parameters in a constructor is strange

PR #2945, opened 2020-06-11
[JBPM-9184] Subprocess excepcion handling tx rollback causes main process to fail

elguardian:
The error happens when a subprocess is invoked sync and therefore is not the owner of the tx.
In this case we need to let the main process handle the rollback

### DIFF
--- a/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
+++ b/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
@@ -63,6 +63,9 @@ import org.kie.api.runtime.rule.EntryPoint;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.lang.Boolean.TRUE;
+import static org.drools.persistence.api.TransactionManager.STATUS_ACTIVE;
+
 public class PersistableRunner implements SingleSessionCommandService {
 
     private static Logger              logger           = LoggerFactory.getLogger( PersistableRunner.class );
@@ -578,10 +581,17 @@ public class PersistableRunner implements SingleSessionCommandService {
 
             // Open the entity manager before the transaction begins.
             PersistenceContext persistenceContext = jpm.getApplicationScopedPersistenceContext();
-
+            // We flag the current persistence runner
+            final String DROOLS_PARENT_RUNNER = "DROOLS_PARENT_RUNNER";
+            boolean isParentRunner = txm.getResource(DROOLS_PARENT_RUNNER) == null;
             boolean transactionOwner = false;
+
             try {
+                if (isParentRunner && txm.getStatus() == STATUS_ACTIVE) {
+                    txm.putResource(DROOLS_PARENT_RUNNER, TRUE);
+                }
                 transactionOwner = txm.begin();
+
 
                 persistenceContext.joinTransaction();
 
@@ -600,14 +610,15 @@ public class PersistableRunner implements SingleSessionCommandService {
                 txm.commit( transactionOwner );
 
             } catch ( RuntimeException re ) {
-                rollbackTransaction( re,
-                        transactionOwner );
+                if (isParentRunner) {
+                    rollbackTransaction(re, transactionOwner);
+                }
                 throw re;
             } catch ( Exception t1 ) {
-                rollbackTransaction( t1,
-                        transactionOwner );
-                throw new RuntimeException( "Wrapped exception see cause",
-                        t1 );
+                if (isParentRunner) {
+                    rollbackTransaction(t1, transactionOwner);
+                }
+                throw new RuntimeException("Wrapped exception see cause", t1);
             }
 
             return context;


gmunozfe (approved):
Looks good to me, awesome work @elguardian !

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
```diff
@@ -600,14 +600,15 @@ public RequestContext execute( Executable executable, RequestContext context ) {
                 txm.commit( transactionOwner );
 
             } catch ( RuntimeException re ) {
-                rollbackTransaction( re,
-                        transactionOwner );
+                if (transactionOwner) {
```
afalhambra: is not already handled in the following code?

https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-api/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java#L264-L275

If not, would it make more sense to add this condition to the rollbackTransaction method?

https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java#L406

elguardian: Actually that code block is that cause of the problem. I removed from the exeuction flow. 
When you are executing a workflow you don't want the subprocess to control the transaction (mark it as rollback).

The only doubt about this one is checking about tx control in a EJB container managed. As I am not allowing the tx rolling back it will happen during ejb tx interceptor as it will catch the tx problem.

I am kind unsure about that
@gmunozfe @afalhambra  wdyt ?

afalhambra: As discussed already, it works fine in transaction managed container based like springboot, though as you said I'm not sure about EJBs behavior. Maybe it is worth to give it a try in some other app server?

elguardian:
Jenkins retest this please.

krisv:
@elguardian imho the expectation is that if an exception happens during the execution of a sub-process (that isn't caught properly), we should roll back the transaction (to avoid inconsistent state).  This means that everything in that transaction will be rolled back, also the parent process if that was executing in the same transaction.  I don't think it's possible for the main process to still catch this error and execute some logic to handle the exception, the only thing we can practically do is roll back.  A parent process can only respond to sub-process errors if the error is caught properly in the sub-process, and turned into an error signal.  This signal can then be propagated to the main process.  But an exception not handled should always be rolled back and cannot be propagated.
I'm afraid that this code [1] where we add a try-catch around start process instance actually shouldn't happen (at least it shouldn't handle the exception, it should rethrow), as we can't guarantee consistent state in that case.
[1] https://github.com/kiegroup/jbpm/commit/50a61c9018edab159a7d8f8ae1bdf658270ed5cb#diff-fc08d1a4b352f812beb2605174f5291bR228

elguardian:
@krisv I dissagree. First because of the logic of how it works. What you are suggesting would cause a different beheaviour depending when the error is caused (it would be some sort of race condition). At present the handle error of a subprocess happens when it does asyncrhonously and in this case it is not happening because starting a subprocess should start a new transaction (which is not possible because the way it is designed). 
The only possibility is to do it a this level (as the other logic does in here).
The exception handling is reusing the one when the problem arise
It is not possible to rethrow the exception as you ar handling the exception so it would make no sense from the logical point of view.
trying to catch the excpetion within the subprocess has no means to work as you need to move the compensation to the subprocess, we have no means to do it right now because of the design of the workflow. 
Regarding your observation about if the code should be there or not. There is no difference between this and other node instance which has compensation logic which does exactly the same thing.

afalhambra (approved):
Nicely done. Great work

krisv:
@elguardian I will take a closer look, as I agree a difference in accepted behavior is indeed a warning sign (I would expect differences on how critical errors are rolled back but not whether some error is accepted as triggering some error handling and accepted).  I would expect the exception to also cause a rollback when the sub-process is executing in a different transaction and the error is not properly caught in the sub-process.  But we can't just catch an exception higher up the stack and handle it there (and still commit the transaction) unless we can guarantee we know the status is 100% guaranteed to still be consistent in every possible case.

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java
```diff
@@ -578,11 +578,18 @@ public RequestContext execute( Executable executable, RequestContext context ) {
...
             PersistenceContext persistenceContext = jpm.getApplicationScopedPersistenceContext();
-
+            // We flag the current persistence runner
+            final String DROOLS_PARENT_RUNNER = "DROOLS_PARENT_RUNNER";
+            boolean isParentRunner = txm.getResource(DROOLS_PARENT_RUNNER) == null;
             boolean transactionOwner = false;
+
             try {
+                if (isParentRunner) {
+                    txm.putResource(DROOLS_PARENT_RUNNER, new Object());
```
afalhambra: Just a silly suggestion. Maybe this is a little clearer

```suggestion
                    txm.putResource(DROOLS_PARENT_RUNNER, Boolean.TRUE);
```

afalhambra (commented):
While building and running all jbpm tests, I noticed that there are many new logs entries which make it quite annoying when debugging/logging. For instance, you will see the following output in most of the test cases:

2020-06-26 16:21:52,484 [main] WARN  Unable to put resource DROOLS_PARENT_RUNNER value java.lang.Object@6c8871f2 due to No transaction is running
2020-06-26 16:21:52,484 [main] WARN  Unable to put resource DROOLS_PARENT_RUNNER value java.lang.Object@6a699703 due to No transaction is running
2020-06-26 16:21:52,488 [main] WARN  Unable to put resource DROOLS_PARENT_RUNNER value java.lang.Object@40a6cdc7 due to No transaction is running

Maybe we can increase set log level to DEBUG in: 

https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-api/src/main/java/org/drools/persistence/api/TransactionSynchronizationRegistryHelper.java#L32-L45

wdyt?

mareknovotny:

mareknovotny:
just to not forget @mbiarnes notice that I triggered FDB, but here is additionally triggered also simple PR job too

mariofusco (approved):

mareknovotny:
Jenkins run compile downstream build

mareknovotny:

krisv:
Added more detailed description of alternative implementation, as discussed with @elguardian 
https://issues.redhat.com/browse/JBPM-9184?focusedCommentId=14248356&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14248356

mareknovotny:
all failed tests from FDB are known except org.jbpm.test.functional.subprocess.ReusableSubProcessTest.testErrorHandlingActivity  which i am not sure

elguardian:
trying another fix.

elguardian:
Jenkins retest this please.

elguardian:

elguardian:

mareknovotny:
Jenkins retest this please

elguardian:
failures unrealted.

PR #2948, opened 2020-06-15
Drools 5372 - Implement PMML coexistence

gitgabrio:
@danielezonca @mariofusco @jiripetrlik @tarilabs 
See https://issues.redhat.com/browse/DROOLS-5372

Logic behind this PR
 

1. a "NO_OP" ResourceType has been implemented
2. a "kie-pmml-implementation" System property has been defined - possible values are "legacy" and "new"
3. if (and only if) **BOTH** PMMLAssembelrs are present in the classpath, the above property is evaluated; the PMMLAssembler that **does not** match the requested version will be mapped to NO_OP resource type
4. Similar logic has been implemented inside DMNPMMLEvaluator - depending on the aforementioned property, one "evaluator" implementation or the other will be instantiated
5. To have DMNPMML tests working, a workaround has been implemented to force cleanup/population of stored KieAssemblers, since this happen - normally - only at startup


tarilabs:
@jiripetrlik how about externalising the DMN+PMML test you wrote in 2 separate integration modules? This way @gitgabrio doesn't have to resort to step5 and actually the test would match real world use-cases

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractPMMLInvocationEvaluator.java
```diff
@@ -143,9 +146,68 @@ public static AbstractPMMLInvocationEvaluator newInstance(DMNModelImpl model, Cl
...
+    }
+
+    private static boolean isLegacyPMMLRequired() {
+        final String property = System.getProperty(KIE_PMML_IMPLEMENTATION.getName(), LEGACY.getName());
+        return property.equals(LEGACY.getName());
+    }
+
+    private static boolean isLegacyImplementationPresent() {
+        try {
+            Thread.currentThread().getContextClassLoader().loadClass("org.kie.pmml.pmml_4_2.PMML4ExecutionHelper");
```
mariofusco: Is there any reason why you're attempting to load the class from the thread context class loader? Not saying it's wrong, just wondering if a plain Class.forName() wouldn't be enough in this case.

gitgabrio: @mariofusco 
Replaced with Class.forName. Seems to work.
Thanks for suggestion

gitgabrio:

gitgabrio:

gitgabrio:

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractDMNKiePMMLInvocationEvaluator.java
```diff
@@ -0,0 +1,140 @@
...
+        PMML4Result resultHolder = getPMML4Result(eventManager, dmnr);
+
+        Map<String, Object> resultVariables = resultHolder.getResultVariables();
+        Map<String, Object> result = getOutputFieldValues(resultHolder, resultVariables, dmnr);
+        if (result.isEmpty()) {
+            result = getPredictedValues(resultHolder, dmnr);
+        }
+        if (result.isEmpty()) {
+            // TO FIX
+            MsgUtil.reportMessage(LOG,
```
danielezonca: Is this wrong? What is missing?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractPMMLInvocationEvaluator.java
```diff
@@ -143,9 +146,68 @@ public static AbstractPMMLInvocationEvaluator newInstance(DMNModelImpl model, Cl
...
+            } else {
+                return getDMNKiePMMLNewInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+            }
+        } else if (legacyImplementationPresent) {
+            return getDMNKiePMMLInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+        } else if (newImplementationPresent) {
+            return getDMNKiePMMLNewInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+        } else {
+            return null;
+        }
```
danielezonca: What about wrap decision logic to a method like `checkPMMLavailability` that just returns `LEGACY` or `NEW`?
I think it could improve readability

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/util/KieHelper.java
```diff
@@ -30,11 +30,16 @@
...
+        // Constructing instances is not allowed for this class
+    }
+
     public static KieContainer getKieContainer(ReleaseId releaseId,
-                                               Resource... resources ) {
+                                               Resource... resources) {
         KieServices ks = KieServices.Factory.get();
-        createAndDeployJar( ks, releaseId, resources );
-        return ks.newKieContainer( releaseId );
+//        createAndDeployJar( ks, releaseId, resources );
```
danielezonca: To be removed

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/pmml/DMNNaiveBayesPMMLTest.java
```diff
@@ -30,12 +31,22 @@
...
 
     private DMNRuntime runtime;
     private DMNModel dmnModel;
 
+    @Before
+    public void resetEnvironment() {
+        LOG.debug("resetEnvironment");
+        resetEnvironment(LEGACY.getName());
+    }
+
```
danielezonca, line 49: Is it possible to move this code to `AbstractDMNPMMLTest`? It seems to me always the same

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private PMMLRuntime getPMMLRuntime(DMNRuntimeEventManager eventManager, String sanitizedKieBase) {
+        KnowledgeBaseImpl knowledgeBase = ((KnowledgeBaseImpl) ((DMNRuntimeImpl) eventManager.getRuntime()).getInternalKnowledgeBase());
+        KieContainer kieContainer = knowledgeBase.getKieContainer();
+        KieBase kieBase;
+        if (kieContainer.getKieBaseNames().contains(sanitizedKieBase)) {
+            kieBase = kieContainer.getKieBase(sanitizedKieBase);
+        } else {
+            kieBase = kieContainer.getKieBase();
+        }
```
danielezonca: Can you please add logging in both branches (`debug` should be fine)?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            Object r = kv.getValue();
+            populateWithObject(toReturn, kv.getKey(), r, dmnr);
+        }
+        return toReturn;
+    }
+
+    @Override
+    protected Map<String, Object> getPredictedValues(PMML4Result pmml4Result, DMNResult dmnr) {
+        LOG.debug("pmml4Result");
```
danielezonca: Can you please remove them or move to `trace` level?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
```diff
@@ -0,0 +1,134 @@
...
+        LOG.debug("getPMML4Result");
+        PMMLContext pmmlContext = getPMMLPMMLContext(UUID.randomUUID().toString(), model, dmnr);
+        String sanitizedKieBase = getSanitizedPackageName(model);
+        PMMLRuntime pmmlRuntime = getPMMLRuntime(eventManager, sanitizedKieBase);
+        return pmmlRuntime.evaluate(model, pmmlContext);
+    }
+
+    @Override
+    protected Map<String, Object> getOutputFieldValues(PMML4Result pmml4Result, Map<String, Object> resultVariables, DMNResult dmnr) {
+        LOG.debug("getOutputFieldValues");
```
danielezonca: Can you please remove them or move to `trace` level?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(DMNKiePMMLNewInvocationEvaluator.class);
+
+    public DMNKiePMMLNewInvocationEvaluator(String dmnNS, DMNElement node, Resource pmmlResource, String model, PMMLInfo<?> pmmlInfo) {
+        super(dmnNS, node, pmmlResource, model, pmmlInfo);
+    }
+
+    @Override
+    protected PMML4Result getPMML4Result(DMNRuntimeEventManager eventManager, DMNResult dmnr) {
+        LOG.debug("getPMML4Result");
```
danielezonca: Can you please remove them or move to `trace` level?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLInvocationEvaluator.java
```diff
@@ -76,70 +74,90 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult d
...
+
+    /**
+     * Returns <code>true</code> if <b>containingCollection</b> contains <b>searchedString</b> OR <b>searchedString.toLowerCase()</b>
+     * @param containingCollection
+     * @param searchedString
+     * @return
+     */
+    private boolean containsIgnoreCase(Collection<String> containingCollection, String searchedString) {
+        return containingCollection.contains(searchedString) || containingCollection.contains(searchedString.toLowerCase());
+    }
```
danielezonca: This method is a bit misleading because it is not a proper ignore case. For example using these values `List("sAmple")` and `sample` will fail

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/util/DMNRuntimeUtil.java
```diff
@@ -94,6 +103,21 @@ public static DMNRuntime createRuntime(final String resourceName, final Class te
...
+        final ServiceRegistryImpl instance = (ServiceRegistryImpl)ServiceRegistry.getInstance();
+        instance.reload();
+//
+//
+//        Map<String, Object> resetServices = serviceDiscovery.getServices();
+//        KieAssemblersImpl resetKieAssemblers = (KieAssemblersImpl)resetServices.get("org.kie.api.internal.assembler.KieAssemblers");
+//        KieAssemblersImpl kieAssemblers = (KieAssemblersImpl)instance.get(KieAssemblers.class);
+//        Map<ResourceType, KieAssemblerService> registeredKieAssemblers = kieAssemblers.getAssemblers();
+//        resetKieAssemblers.getAssemblers()
+//                .forEach(registeredKieAssemblers::put);
```
danielezonca: To be removed

kie-pmml-new/kie-pmml-models/kie-pmml-models-drools/kie-pmml-models-drools-tree/kie-pmml-models-drools-tree-compiler/src/main/java/org/kie/pmml/models/drools/tree/compiler/factories/KiePMMLTreeModelNodeASTFactory.java
```diff
@@ -98,10 +98,8 @@ protected void declareFinalRuleFromNode(final Node node,
             return;
         }
         String currentRule = String.format(PATH_PATTERN, parentPath, node.getScore());
-        if (!(predicate instanceof True)) {
```
danielezonca, line 101: Is this always true?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/util/DMNRuntimeUtil.java
```diff
@@ -94,6 +103,21 @@ public static DMNRuntime createRuntime(final String resourceName, final Class te
...
     }
 
+    public static void resetServices() {
+        final ServiceDiscoveryImpl serviceDiscovery = ServiceDiscoveryImpl.getInstance();
+        serviceDiscovery.reset();
+        final ServiceRegistryImpl instance = (ServiceRegistryImpl)ServiceRegistry.getInstance();
+        instance.reload();
+//
+//
+//        Map<String, Object> resetServices = serviceDiscovery.getServices();
```
jiripetrlik: Is there any reason to keep commented code? If yes please add description into the comment. Otherwise please remove.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractPMMLInvocationEvaluator.java
```diff
@@ -143,9 +146,68 @@ public static AbstractPMMLInvocationEvaluator newInstance(DMNModelImpl model, Cl
...
+                return getDMNKiePMMLInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+            } else {
+                return getDMNKiePMMLNewInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+            }
+        } else if (legacyImplementationPresent) {
+            return getDMNKiePMMLInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+        } else if (newImplementationPresent) {
+            return getDMNKiePMMLNewInvocationEvaluator(nameSpace, funcDef, pmmlResource, pmmlModel, pmmlInfo);
+        } else {
+            return null;
```
jiripetrlik: Maybe it would be better to throw exception.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/pmml/AbstractDMNPMMLTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static org.kie.api.pmml.PMMLConstants.KIE_PMML_IMPLEMENTATION;
+import static org.kie.dmn.core.util.DMNRuntimeUtil.resetServices;
+
+public abstract class AbstractDMNPMMLTest {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractDMNPMMLTest.class);
+
+    protected void resetEnvironment(String pmmlVersion) {
+        LOG.debug("resetEnvironment {}", pmmlVersion);
+        System.setProperty(KIE_PMML_IMPLEMENTATION.getName(), pmmlVersion);
```
jiripetrlik, line 30: I'm not sure about this. I think it would be much more clear to create two separate modules as @tarilabs suggested with different jars on classpath. We can never be sure that it DMN PMML works without logic in resetService() method.

kie-pmml-new/kie-pmml-models-archetype/src/main/resources/archetype-resources/__rootArtifactId__-compiler/pom.xml
```diff
@@ -40,6 +40,16 @@
       <groupId>org.kie</groupId>
       <artifactId>kie-test-util</artifactId>
       <scope>test</scope>
+      <exclusions>
```
jiripetrlik, line 43: Can you please explain why do we need to exclude these libs now?

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/pmml/KiePMMLNewTree.dmn
```diff
@@ -0,0 +1,148 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
jiripetrlik: Is there any difference between old dmn and new dmn? Can you please explain reason for it.

jiripetrlik (commented):
@gitgabrio 
Hello, thank you for your PR. I'm not sure about approach which you use for tests. I would suggest to use the same approach as we take with JPMML: https://github.com/kiegroup/kie-jpmml-integration/tree/master/kie-dmn-jpmml/src/test/java/org/kie/dmn/jpmml 

Using this approach we can simply create a new module and inherit all tests. The PMML implementation is chosen according to what is present on the classpath. What do you think about it?

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/pmml/KiePMMLNewTree.dmn
gitgabrio: @jiripetrlik 
Removed

kie-pmml-new/kie-pmml-models-archetype/src/main/resources/archetype-resources/__rootArtifactId__-compiler/pom.xml
gitgabrio: @jiripetrlik 
Enforcer rules (ban-duplicated-classes)

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/pmml/AbstractDMNPMMLTest.java
gitgabrio: @jiripetrlik 
This workaround is needed because all those DMN-PMML tests are using "utils" (like KieHelper) that by themselves are "by-passing" the _normal_ usage; as they currently are they do not test the real runtime.
Rewriting all those tests is out of the scope of this PR.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractPMMLInvocationEvaluator.java
gitgabrio: @jiripetrlik 
That "null" is expected since - in this case - the code is written to return a "DummyPMMLInvocationEvaluator"

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/util/DMNRuntimeUtil.java
gitgabrio: @jiripetrlik 
The issue here is different. Here we have *both* kie-pmml implementations in the classpath (legacy and new one) and unfortunatley the "utils" used in the tests does not clear the runtime. What happens is that all the KieServices are populated (normally) only at application startup (see https://github.com/kiegroup/droolsjbpm-knowledge/blob/master/kie-internal/src/main/java/org/kie/internal/services/KieAssemblersImpl.java#L67).
/---/
Removed commented code

kie-pmml-new/kie-pmml-models/kie-pmml-models-drools/kie-pmml-models-drools-tree/kie-pmml-models-drools-tree-compiler/src/main/java/org/kie/pmml/models/drools/tree/compiler/factories/KiePMMLTreeModelNodeASTFactory.java
gitgabrio: @danielezonca 
This is out of the scope of this PR - I will check in another ticket

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/util/DMNRuntimeUtil.java
gitgabrio: @danielezonca 
done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLInvocationEvaluator.java
gitgabrio: @danielezonca 
Renamed a little bit.
BTW - that is one of the reason of javadoc -  sometimes it is hard to explain a specific logic in one camel-cased word

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
gitgabrio: @danielezonca 
done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
gitgabrio: @danielezonca 
done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
gitgabrio: @danielezonca 
done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/DMNKiePMMLNewInvocationEvaluator.java
gitgabrio: @danielezonca 
done

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/pmml/DMNNaiveBayesPMMLTest.java
gitgabrio: @danielezonca 
The parameter is different - tests for new kiepmml should invoke it with "NEW.getName()", while the other tests should use "LEGACY.getName()"; (fixed inside DMNTreePMMLNewImplTest)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/util/KieHelper.java
gitgabrio: @danielezonca 
Done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractPMMLInvocationEvaluator.java
gitgabrio: @danielezonca 
Done

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/pmml/AbstractDMNKiePMMLInvocationEvaluator.java
gitgabrio: @danielezonca 
leftover

danielezonca (approved):
Thanks for your changes, approving :+1:

gitgabrio:
@jiripetrlik @tarilabs @mariofusco 

Could you please look at this again and, eventually, approve it ? Thanks

tarilabs:
we must ensure this works with jpmml , executing FDB only for jpmml interest

tarilabs (approved):
for me is okay, thanks @gitgabrio 

But I would appreciate if QE / @jiripetrlik would consider my comment here, https://github.com/kiegroup/drools/pull/2948#issuecomment-644236984,
in other words, put on the roadmap move all DMN+PMML test in a separate module of integration
(on this git repo, drools.git).

I believe placing the original test of DMN+PMML tests was only an ad-interim and we now grown the support, this is now a consistent integration aspect which imho should be addressed by a dedicated module in the build

gitgabrio:

gitgabrio:

jiripetrlik (approved):
@gitgabrio Thank you for your changes.

@tarilabs 
Yes, I agree with 2 new modules for DMN-PMML tests and I've wrote it in one of my comments in review. I also created JIRA for it today DROOLS-5443.What I do not understand is why we can't distinguish between implementations using classpath. Why we need to have always both implementations on classpath even if we use separate modules for tests? @gitgabrio Can you please explain more? But this is not related to this PR and we can merge now.

gitgabrio:
> What I do not understand is why we can't distinguish between implementations using classpath. Why we need to have always both implementations on classpath even if we use separate modules for tests? <
@jiripetrlik I think that with two separated test modules it could be possible to avoid my "workaround"  for tests. 
The needs for the System property (LEGACY vs NEW) is that in some situations (e.g. BC ? Built images ?) *both* implementations (legacy and new one) are put inside the final project - so in the classpath -  and the user has no way to exclude one or the other dependency. Hope to have been clear.

jiripetrlik:
@gitgabrio 
Understand. I think there is no engine running in the Business Central, but you are probably right for the kie-server. Makes sense for me now.

Question is whether it is good to use system property for this? What if user wants to use both new and old kjars in the same kie server? To allow user to use both can even help him with transition to the new implementation. What do you think? Maybe we can use something in kie-session like optional global to define which implementation should be used. It will also greatly simplify test implementation, because in case of global we can simply parameterize existing tests. @tarilabs @mariofusco  What is your opinion? Or do you see any other option how to switch between PMML implementations. Any suggestions ideas?

mariofusco (approved):

mariofusco:
If possible I'd rollback the changes to KieContainerImpl and AbstractKieProject that are only changed in white spaces and formatting. For the rest, ok for me.

gitgabrio:
> What if user wants to use both new and old kjars in the same kie server?

@jiripetrlik It is not possible to have both *running* at the same time - and this is already _by design_ looking at how the jpmml implementation is managed. The System property allow to - eventually - have *both* on the project classpath (for whatever reason) and choose  - at project startup, which one activate. I think this will be important especially during transition from LEGACY to NEW one. 
About parameterizing tests, this could be done also with the system property (as you see in my modification, I set it soon before a test, but the actual value of the property may be injected with test' parameter)

jiripetrlik:
@gitgabrio 
Understand, thank you for your explanation!

gitgabrio:
> If possible I'd rollback the changes to KieContainerImpl and AbstractKieProject that are only changed in white spaces and formatting. For the rest, ok for me.

@mariofusco 
Done

danielezonca:

danielezonca:
Build #33
```
org.kie.server.integrationtests.controller.WebSocketKieControllerRuleCapabilitiesIntegrationTest.testStartAndStopScanner[0: JSON KieServicesConfiguration{transport=REST, serverUrl='http://localhost:43384/kie-server-services/services/rest/server'}]
org.kie.server.integrationtests.controller.WebSocketKieControllerRuleCapabilitiesIntegrationTest.testScanNow[0: JSON KieServicesConfiguration{transport=REST, serverUrl='http://localhost:43384/kie-server-services/services/rest/server'}]
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testReadinessCheck
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testMigration
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testLivenessCheck
```

Build 35
```
org.kie.server.integrationtests.router.rest.KieServerRouterControllerRecoveryTest.testRecoveryOfCreatedContainerOnController
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testReadinessCheck
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testMigration
org.kie.processmigration.it.ProcessMigrationIntegrationTest.testLivenessCheck
```

Process migration integration tests are broken even in other FDB ( see for example https://github.com/kiegroup/kie-wb-common/pull/3329 ) so they are not related ( see [JBPM-9196](https://issues.redhat.com/browse/JBPM-9196) )

PR #2949, opened 2020-06-16
BAQE-1425 - Remove DummyProcessRuntime from session benchmarks

hellowdan:
On Drools benchmarks, the DummyProcessRuntime class is being used only on org.drools.benchmarks.session.sessionpool.DisposeSessionWithProcessRuntimeBenchmark and it is being affected by changes in the Interface InternalProcessRuntime (which is implemented by DummyProcessRuntime).

### DIFF
--- a/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
+++ b/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
@@ -2204,7 +2204,7 @@ public class StatefulKnowledgeSessionImpl extends AbstractRuntime
 
         @Override
         public ProcessInstance getProcessInstance( long processInstanceId ) {
-            throw new UnsupportedOperationException( );
+            return null;
         }
 
         @Override


hellowdan:
Jenkins test again

tarilabs:
mmm... I think this is more a question for @mariofusco but I am afraid there was a good reason why the UnsupportedOpException

MarianMacik (commented):
Not sure this is a good idea as I believe this is really just a dummy implementation in case no process runtime is needed/used. Does the drools benchmark make use of this method in any way?

hellowdan:
@MarianMacik getProcessInstance() is called to force the creation of a process runtime and I removed the exception because of this (just like it was in the Dummy class which I removed from kie-benchmarks project to avoid compilation error there).

MarianMacik:
@hellowdan So the drools benchmarks actually call this method somewhere? I think this shouldn't happen unless you really use processes. As right now it is designed in a way that it will throw an exception.

baldimir:
@hellowdan as @tarilabs mentioned, this is a question for @mariofusco. I expect the exception was there for some reason. Please consult with Mario.

hellowdan:
Considering your feedbacks, this is not a good solution. Closing the PR, will think another approach. Thank you all.

PR #2951, opened 2020-06-17
[RHDM-1376] org.kie.dmn.validation fails on master with ibm jdk

tkobayas:
I was comparing QName to String before. Coercion happened to help in case of other JDKs but fixed now.


tarilabs (approved):
Thank you!

hellowdan (approved):

PR #2970, opened 2020-07-02
[DROOLS-5412] Externalize lambda for OOPath

tkobayas:


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/OOPathExprGenerator.java
```diff
@@ -97,7 +99,10 @@ public void visit(Class<?> originalClass, String originalBind, DrlxParseSuccess
                 bindingId = context.getOOPathId(fieldType, chunkKey);
             }
             final Expression accessorLambda = generateLambdaWithoutParameters(Collections.emptySortedSet(),
-                                                                              prepend(new NameExpr(THIS_PLACEHOLDER), callExpr.getExpression()));
+                                                                              prepend(new NameExpr(THIS_PLACEHOLDER), callExpr.getExpression()), false, Optional.ofNullable(previousClass), context);
+            if (accessorLambda instanceof LambdaExpr) {
+                context.getPackageModel().getLambdaReturnTypes().put((LambdaExpr)accessorLambda, callExpr.getType());
+            }
```
tkobayas, line 105: Similar to "from" clause, we need to store the return type so that ExecModelLambdaPostProcessor can use it later.

lucamolteni (approved):

hellowdan (approved):

PR #2974, opened 2020-07-03
DROOLS-5481 DMN emit error DMNMessage when exception in contextentry

tarilabs:


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNContextEvaluator.java
```diff
@@ -129,7 +128,17 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult d
                         return new EvaluatorResultImpl( results, ResultType.FAILURE );
                     }
                 } catch ( Exception e ) {
-                    logger.error( "Error invoking expression for node '" + name + "'.", e );
+                    logger.error("Error invoking expression for node '" + name + "'.", e);
+                    MsgUtil.reportMessage(logger,
```
tarilabs, line 132: @etirelli the code was only registering to the (slf4j) logger and not building a `DMNMessage` for the error.

I am also continuing to analyse the problem from different angles to be sure is well covered, but this is already something which we need to have anyway.

etirelli: Good catch @tarilabs

hellowdan (approved):

PR #2977, opened 2020-07-06
[DROOLS-5411] Externalize lambda for CEP

tkobayas:


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/util/lambdareplace/ExecModelLambdaPostProcessor.java
```diff
@@ -137,6 +144,31 @@ public void convertLambdas() {
...
+                if (!extractorOpt.isPresent()) {
+                    logger.debug("Unable to create MaterializedLambdaExtractor for {}", lambdaExpr);
+                } else {
+                    MaterializedLambdaExtractor extractor = extractorOpt.get();
+                    replaceLambda(lambdaExpr, () -> extractor);
+                }
+            }
+        });
+    }
+
```
tkobayas, line 162: In case of temporal constraints, the generated code is like this:
```
expr("40AE484C3A9BF23BE65C0B135D4B1E33",
     (org.drools.modelcompiler.domain.StockTick _this) -> _this.getTimeFieldAsLong(),
     var_$a,
     (org.drools.modelcompiler.domain.StockTick $a) -> $a.getTimeFieldAsLong(),
     D.after(5L, java.util.concurrent.TimeUnit.MILLISECONDS, 8L, java.util.concurrent.TimeUnit.MILLISECONDS)
)
```
So I externalize lambdas one-by-one calling replaceLambda() instead of calling extractLambdaFromMethodCall().

hellowdan (approved):

lucamolteni (approved):

PR #2979, opened 2020-07-07
Drools 5482 - Extending integration tests to cover transformations

gitgabrio:

see https://issues.redhat.com/browse/DROOLS-5482

This PR

1. Implement ConvertTypeUtil to convert an object of a given type to a different type one
2. fix management of DefineFunctions
3. implement new tests that makes use of DerivedField/DefineFunction


kie-pmml-trusty/kie-pmml-commons/src/main/java/org/kie/pmml/commons/utils/ConverterTypeUtil.java
```diff
@@ -0,0 +1,103 @@
...
+        } else if (expectedClass.isAssignableFrom(Character.class) || expectedClass.isAssignableFrom(char.class)) {
+            return parseChar(originalObject);
+        } else if (expectedClass.isAssignableFrom(Byte.class) || expectedClass.isAssignableFrom(byte.class)) {
+            return Byte.parseByte(originalObject);
+        } else if (expectedClass.isAssignableFrom(Short.class) || expectedClass.isAssignableFrom(short.class)) {
+            return Short.parseShort(originalObject);
+        } else {
+            throw new KiePMMLException(String.format(FAILED_CONVERSION, originalObject,
+                                                     expectedClass.getName()));
+        }
```
danielezonca: Isn't `String` a supported type?

jiripetrlik (approved):

kie-pmml-trusty/kie-pmml-commons/src/main/java/org/kie/pmml/commons/utils/ConverterTypeUtil.java
gitgabrio: @danielezonca 
That method is "convertFromString" - it is invoked when the original object is a String and it has to be converted to a different type

kie-pmml-trusty/kie-pmml-commons/src/main/java/org/kie/pmml/commons/utils/ConverterTypeUtil.java
```diff
@@ -0,0 +1,120 @@
...
+        Object toReturn;
+        if (currentClass.equals(String.class)) {
+            try {
+                toReturn = convertFromString(expectedClass, (String) originalObject);
+            } catch (Exception e) {
+                throw new KiePMMLException(String.format(FAILED_CONVERSION, originalObject,
+                                                         expectedClass.getName()), e);
+            }
+        } else if (expectedClass == String.class) {
+            toReturn = originalObject.toString();
```
danielezonca, line 73: I would suggest `String.valueOf` to avoid a NPE

danielezonca (approved):

PR #2987, opened 2020-07-13
[DROOLS-5505] AfterEvaluateDecisionTableEvent and BeforeEvaluateDecisionTableEvent are not extending DMNEvent

danielezonca:


tarilabs (approved):
makes sense, thank you

hellowdan (approved):

PR #2990, opened 2020-07-14
DROOLS-5507 : Converting excel spreadsheet to guided decision table causes an exception

Rikkola:
https://issues.redhat.com/browse/DROOLS-5507

Failures that I ran into:
* `(age != null) == "@{param}" )` was not regognized as a predicate ( affects GRE, Templated rules and GDST BRL)
* Variables from single field predicate conditions were not found ( affects GRE, Templated rules and GDST BRL)

### DIFF
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
@@ -1077,10 +1077,8 @@ public class RuleModelDRLPersistenceImpl
         private void generateSingleFieldConstraint(final SingleFieldConstraint constr,
                                                    final LHSGeneratorContext gctx) {
             if (constr.getConstraintValueType() == BaseSingleFieldConstraint.TYPE_PREDICATE) {
-                buf.append("eval( ");
-                buf.append(constr.getValue());
-                buf.append(" )");
-                gctx.setHasOutput(true);
+                generatePredicateSingleFieldConstraint(constr,
+                                                       gctx);
             } else {
                 if (constr.isBound()) {
                     bindingsFields.put(constr.getFieldBinding(),
@@ -1122,6 +1120,14 @@ public class RuleModelDRLPersistenceImpl
                     gctx.setHasOutput(true);
                 }
             }
+        }
+
+        protected void generatePredicateSingleFieldConstraint(final SingleFieldConstraint constr,
+                                                              final LHSGeneratorContext gctx) {
+            buf.append("eval( ");
+            buf.append(constr.getValue());
+            buf.append(" )");
+            gctx.setHasOutput(true);
         }
 
         private void generateNormalFieldRestriction(final SingleFieldConstraint constr,
@@ -3543,7 +3549,8 @@ public class RuleModelDRLPersistenceImpl
             }
         }
         if (isCompositeFieldConstraint(splittedExpr)) {
-            ComplexExpr complexExpr = new ComplexExpr(splittedExpr.get(1));
+            ComplexExpr complexExpr = new ComplexExpr(splittedExpr.get(1),
+                                                      expr);
             for (int i = 0; i < splittedExpr.size(); i += 2) {
                 complexExpr.subExprs.add(parseExpr(splittedExpr.get(i),
                                                    isJavaDialect,
@@ -4328,21 +4335,29 @@ public class RuleModelDRLPersistenceImpl
 
         private final List<Expr> subExprs = new ArrayList<Expr>();
         private final String connector;
+        private final String expr;
 
-        private ComplexExpr(final String connector) {
+        private ComplexExpr(final String connector,
+                            final String expr) {
             this.connector = connector;
+            this.expr = expr;
         }
 
         public FieldConstraint asFieldConstraint(final RuleModel m,
                                                  final FactPattern factPattern) {
-            final CompositeFieldConstraint comp = new CompositeFieldConstraint();
-            comp.setCompositeJunctionType(connector.equals("&&") ? CompositeFieldConstraint.COMPOSITE_TYPE_AND : CompositeFieldConstraint.COMPOSITE_TYPE_OR);
-            for (final Expr expr : subExprs) {
-                comp.addConstraint(expr.asFieldConstraint(m,
-                                                          factPattern));
+            if (!connector.equals("&&") && !connector.equals("||")) {
+                return new EvalExpr(expr).asFieldConstraint(m, factPattern);
+            } else {
+
+                final CompositeFieldConstraint comp = new CompositeFieldConstraint();
+                comp.setCompositeJunctionType(connector.equals("&&") ? CompositeFieldConstraint.COMPOSITE_TYPE_AND : CompositeFieldConstraint.COMPOSITE_TYPE_OR);
+                for (final Expr subExpr : subExprs) {
+                    comp.addConstraint(subExpr.asFieldConstraint(m,
+                                                                 factPattern));
+                }
+                convertLegacyMatchesToNewFormat(comp);
+                return comp;
             }
-            convertLegacyMatchesToNewFormat(comp);
-            return comp;
         }
 
         /**
--- a/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/context/GeneratorContextRuleModelVisitor.java
+++ b/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/context/GeneratorContextRuleModelVisitor.java
@@ -133,45 +133,33 @@ public class GeneratorContextRuleModelVisitor {
     private void visitActionFieldList(final ActionInsertFact afl) {
         String factType = afl.getFactType();
         for (ActionFieldValue afv : afl.getFieldValues()) {
-            InterpolationVariable var = new InterpolationVariable(afv.getValue(),
-                                                                  afv.getType(),
-                                                                  factType,
-                                                                  afv.getField());
-            if (afv.getNature() == FieldNatureType.TYPE_TEMPLATE && !vars.contains(var)) {
-                vars.add(var);
-            } else {
-                hasNonTemplateOutput = true;
-            }
+            manageTemplateVariable(new InterpolationVariable(afv.getValue(),
+                                                             afv.getType(),
+                                                             factType,
+                                                             afv.getField()),
+                                   afv.getNature());
         }
     }
 
     private void visitActionFieldList(final ActionSetField afl) {
         String factType = model.getLHSBindingType(afl.getVariable());
         for (ActionFieldValue afv : afl.getFieldValues()) {
-            InterpolationVariable var = new InterpolationVariable(afv.getValue(),
-                                                                  afv.getType(),
-                                                                  factType,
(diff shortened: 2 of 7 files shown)


drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -4328,21 +4335,33 @@ private static ExpressionPart getExpressionPart(String expressionPart,
...
                                                  final FactPattern factPattern) {
-            final CompositeFieldConstraint comp = new CompositeFieldConstraint();
-            comp.setCompositeJunctionType(connector.equals("&&") ? CompositeFieldConstraint.COMPOSITE_TYPE_AND : CompositeFieldConstraint.COMPOSITE_TYPE_OR);
-            for (final Expr expr : subExprs) {
-                comp.addConstraint(expr.asFieldConstraint(m,
-                                                          factPattern));
-            }
-            convertLegacyMatchesToNewFormat(comp);
-            return comp;
+            if (!connector.equals("&&") && !connector.equals("||")) {
```
Rikkola, line 4348: If the connector is not one that the Guided Rule Model supports, this defaults to predicate.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -1124,6 +1122,14 @@ private void generateSingleFieldConstraint(final SingleFieldConstraint constr,
             }
         }
 
+        protected void generatePredicateSingleFieldConstraint(final SingleFieldConstraint constr,
```
Rikkola, line 1125: Separated this method so it can be overwritten.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/context/GeneratorContextRuleModelVisitor.java
```diff
@@ -249,6 +249,11 @@ private void visitRuleModel(final RuleModel model) {
     }
 
     private void visitSingleFieldConstraint(final SingleFieldConstraint sfc) {
+        if (BaseSingleFieldConstraint.TYPE_PREDICATE == sfc.getConstraintValueType()){
+            parseStringPattern(sfc.getValue());
+            return;
+        }
+
```
Rikkola, line 244: This fix is not only needed by this, but fixes the behaviour for GRE and BRL condition columns in GDST.

drools-workbench-models/drools-workbench-models-datamodel-api/src/main/java/org/drools/workbench/models/datamodel/rule/visitors/RuleModelVisitor.java
```diff
@@ -255,6 +255,11 @@ private void visitRuleModel(RuleModel model) {
     }
 
     private void visitSingleFieldConstraint(SingleFieldConstraint sfc) {
+        if (BaseSingleFieldConstraint.TYPE_PREDICATE == sfc.getConstraintValueType()){
+            parseStringPattern(sfc.getValue());
+            return;
+        }
+
```
Rikkola, line 262: This fix is not only needed by this, but fixes the behaviour for GRE and BRL condition columns in GDST.

A bit of redundancy here, but getting rid of it is likely a huge and risky task.

Rikkola:
@jomarko @vpellegrino If you two could please take a look.

vpellegrino (approved):
Thanks @Rikkola

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
```diff
@@ -4328,21 +4335,33 @@ private static ExpressionPart getExpressionPart(String expressionPart,
...
-                                                          factPattern));
-            }
-            convertLegacyMatchesToNewFormat(comp);
-            return comp;
+            if (!connector.equals("&&") && !connector.equals("||")) {
+
+                final SingleFieldConstraint con = new SingleFieldConstraint();
+                con.setConstraintValueType(SingleFieldConstraint.TYPE_PREDICATE);
+                con.setValue(expr);
+                return con;
```
jomarko: Shouldn't we implement as `return new EvalExpr(expr).asFieldConstraint(m, factPattern);` ?

Asking because that `EvalExpr` seems be implemented exactly in same way.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
```diff
@@ -5104,4 +5103,32 @@ public void testMatchesLegacyToNewVersion() {
         assertEquals("matches", right.getOperator());
         assertEquals("P.*", right.getValue());
     }
+
+    @Test
+    /**
+     * The GRE can not produce this, but the Persistence class is also used by XLS->GDST->XLS conversions.
+     */
+    public void testMoreComplexExpression() {
```
jomarko, line 5111: I saw in some `if` we were checking `&&` and `||` operators. I am wondering if test like below should pass then.

```java
    @Test
    public void testMoreComplexExpressionLogicAnd() {

        final String drl = "rule \"r0\"\n" +
                "dialect \"mvel\"\n" +
                "when\n" +
                "Person( ( age != null ) == true && ( name == null ) == false )\n" +
                "then\n" +
                "end\n";

        final PackageDataModelOracle dmo = mock(PackageDataModelOracle.class);
        final RuleModel m = ruleModelPersistence.unmarshal(drl,
                                                           Collections.EMPTY_LIST,
                                                           dmo);
        final String resultDrl = ruleModelPersistence.marshal(m);

        final String expectedDrl = "rule \"r0\"\n" +
                "dialect \"mvel\"\n" +
                "when\n" +
                "Person( eval( ( age != null ) == true ) && eval( ( name == null ) == false ) )\n" +
                "then\n" +
                "end\n";
        assertEqualsIgnoreWhitespace(expectedDrl, resultDrl);
    }
```

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/visitors/RuleModelVisitorTest.java
```diff
@@ -71,6 +72,27 @@ public void visitSingleFieldConstraint() {
...
 
+    @Test
+    public void testSingleFieldConstraintPredicate() {
+
+        Map<InterpolationVariable, Integer> variableMap = new HashMap<>();
+        RuleModelVisitor visitor = new RuleModelVisitor(variableMap);
+
+        SingleFieldConstraint singleFieldConstraint = new SingleFieldConstraint();
+        singleFieldConstraint.setConstraintValueType(BaseSingleFieldConstraint.TYPE_PREDICATE);
+        singleFieldConstraint.setValue("(age != null) == \"@{param1}\"");
```
jomarko, line 83: If a boolean is expected as `param1` value, shouldn't this line be rather `        singleFieldConstraint.setValue("(age != null) == @{param1}");` ?

drools-workbench-models/drools-workbench-models-guided-dtable/src/test/java/org/drools/workbench/models/guided/dtable/backend/GuidedDTBRDRLPersistenceTest.java
```diff
@@ -21,19 +21,51 @@
...
+        final GuidedDTBRDRLPersistence persistence = new GuidedDTBRDRLPersistence((key) -> "true");
+
+        final RuleModel ruleModel = new RuleModel();
+        ruleModel.name = "Template aware";
+
+        ruleModel.addRhsItem(new TemplateAwareIAction("initialValue"));
+        final FactPattern factPattern = new FactPattern("Person");
+        final SingleFieldConstraint constraint = new SingleFieldConstraint();
+        constraint.setConstraintValueType(BaseSingleFieldConstraint.TYPE_PREDICATE);
+        constraint.setValue("(age != null) == \"@{param1}\"");
```
jomarko, line 51: similar question as above

jomarko (commented):
Code review. See inline. Manual test pending.

jomarko:
My manual check failed. 
- I created a project with `org.redhat.reproducer` package
- Added dependency for jar with classes `Account` and `AccountInformation`
- Imported `product-name-reproducer2` into `org.redhat.reproducer` package
- product build failed, getting error [1]
- Converted to gdst, successfully, however even when I removed original xlsx still not able to build project

[1]
```
09:33:36,002 ERROR [org.drools.compiler.kie.builder.impl.KieProject] (default task-19) Unable to build KieBaseModel:defaultKieBase
[3,48]: [ERR 107] Line 3:48 mismatched input ';' expecting one of the following tokens: '[package, unit, import, global, declare, function, rule, query]'.
[0,0]: Parser returned a null Package
```

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
Rikkola: I'll try this out.

drools-workbench-models/drools-workbench-models-commons/src/test/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceTest.java
Rikkola: This and several more DRL snippets will fail. What we support should probably be limited to what the Guider Rule Editor components can produce, and like in this case, what we want to be able to migrate from XLS to GDST.

drools-workbench-models/drools-workbench-models-datamodel-api/src/test/java/org/drools/workbench/models/datamodel/rule/visitors/RuleModelVisitorTest.java
Rikkola: I would keep it with quotes. The GRE components tend to add the quotes since they rely on the fact that MVEL is used.

Rikkola (commented):
@jomarko The original has the ; issue too if you try to compile it. If you remove the ; it should work.

drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java
Rikkola: @jomarko Well spotted. Works and less redundancy.

jomarko (approved):
Thank you for answers @Rikkola . Approving. Just merge this coverage increase before please. https://github.com/Rikkola/drools/pull/7

Rikkola:

PR #2992, opened 2020-07-15
[DROOLS-5483] clear transactionResources ThreadLocal only when it is …

mariofusco:
…the transaction owner


mariofusco:

PR #2993, opened 2020-07-16
[DROOLS-5502] DMN Generate Strongly Typed Output Set

tkobayas:
Hi @lucamolteni @tarilabs 

This is a draft implementation of DROOLS-5502. It's still work in progress but I'd like you to check if I'm on the right track. Now we can generate `OutputSet` class and populate it by the map from `DMNResult.getContext().getAll()`. As you can see in DMNRuntimeTypesTest.testFieldCapitalization(), populating `OutputSet` is done outside of DMNRuntime. For example, we will do it in Kogito REST endpoint. Am I right? Feel free to let me know your thoughts.

(If I'm on the right track,) remaining TODO:
- Check the failing test DMNStronglyTypedSupportTest.testDateAndTime (This is interesting. I may ask you about this issue later)
- Refactor DMNOutputSetType/DMNInputSetType to share a super class
- Add more `OutputSet` assertion to existing tests (e.g. currently I assert in only last 3 test methods of DMNRuntimeTypesTest)
- (Maybe) Add another implementation of FieldDefinition for DMNFunction

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
@@ -16,11 +16,8 @@
 
 package org.kie.dmn.typesafe;
 
-import java.time.Duration;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.Period;
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalAmount;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -105,11 +102,11 @@ public class DMNAllTypesIndex {
             case UNKNOWN:
                 return Object.class;
             case DATE:
-                return LocalDate.class;
+                return Temporal.class;
             case TIME:
-                return LocalTime.class;
+                return Temporal.class;
             case DATE_TIME:
-                return LocalDateTime.class;
+                return Temporal.class;
             case BOOLEAN:
                 return Boolean.class;
             case NUMBER:
@@ -126,10 +123,10 @@ public class DMNAllTypesIndex {
         switch (expectedFEELType.getName()) {
             case SimpleType.YEARS_AND_MONTHS_DURATION:
             case "yearMonthDuration":
-                return Period.class;
+                return TemporalAmount.class;
             case SimpleType.DAYS_AND_TIME_DURATION:
             case "dayTimeDuration":
-                return Duration.class;
+                return TemporalAmount.class;
             default:
                 throw new IllegalArgumentException();
         }
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNOutputSetType.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNOutputSetType.java
@@ -16,14 +16,14 @@
 
 package org.kie.dmn.typesafe;
 
-class DMNInputSetType extends AbstractDMNSetType {
+class DMNOutputSetType extends AbstractDMNSetType {
 
-    DMNInputSetType(DMNAllTypesIndex index, DMNStronglyCodeGenConfig codeGenConfig) {
+    DMNOutputSetType(DMNAllTypesIndex index, DMNStronglyCodeGenConfig codeGenConfig) {
         super(index, codeGenConfig);
     }
 
     @Override
     public String getTypeName() {
-        return "InputSet";
+        return "OutputSet";
     }
 }
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNTypeSafeTypeGenerator.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNTypeSafeTypeGenerator.java
@@ -18,6 +18,7 @@ package org.kie.dmn.typesafe;
 
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -30,6 +31,9 @@ import org.drools.modelcompiler.builder.generator.declaredtype.api.TypeDefinitio
 import org.drools.modelcompiler.builder.generator.declaredtype.generator.GeneratedClassDeclaration;
 import org.kie.dmn.api.core.DMNModel;
 import org.kie.dmn.api.core.DMNType;
+import org.kie.dmn.api.core.ast.BusinessKnowledgeModelNode;
+import org.kie.dmn.api.core.ast.DecisionNode;
+import org.kie.dmn.api.core.ast.DecisionServiceNode;
 import org.kie.dmn.api.core.ast.InputDataNode;
 import org.kie.dmn.core.impl.DMNModelImpl;
 import org.slf4j.Logger;
@@ -102,6 +106,29 @@ public class DMNTypeSafeTypeGenerator {
         inputSetType.setJavadoc(postfixToJavadoc(new StringBuilder("A representation of all the InputData and other DRG Requirement of the whole DMN '").append(dmnModel.getName()).append("' inputs.").toString(), dmnModel));
 
         types.add(inputSetType);
+
+        Set<DecisionNode> decisions = dmnModel.getDecisions();
+        Collection<DecisionServiceNode> decisionServices = dmnModel.getDecisionServices();
+        Set<BusinessKnowledgeModelNode> bkms = dmnModel.getBusinessKnowledgeModels();
+
+        DMNOutputSetType outputSetType = new DMNOutputSetType(index, codeGenConfig);
+        for (InputDataNode i : inputs) {
+            outputSetType.addField(i.getName(), i.getType()); // OutputSet also contains inputs
+        }
+        for (DecisionNode d : decisions) {
+            outputSetType.addField(d.getName(), d.getResultType());
+        }
+        for (DecisionServiceNode ds : decisionServices) {
+            outputSetType.addField(ds.getName(), dmnModel.getTypeRegistry().unknown());
+        }
+        for (BusinessKnowledgeModelNode bkm : bkms) {
+            outputSetType.addField(bkm.getName(), dmnModel.getTypeRegistry().unknown());
+        }
+
+        outputSetType.initFields();
+        outputSetType.setJavadoc(postfixToJavadoc(new StringBuilder("A representation of all the OutputData of the whole DMN '").append(dmnModel.getName()).append("' outputs.").toString(), dmnModel));
+
+        types.add(outputSetType);
 
         for (DMNType type : index.typesToGenerateByNS(dmnModel.getNamespace())) { // this generator shall only be concerned with the types belonging to this generator dmnModel.
             DMNDeclaredType dmnDeclaredType = new DMNDeclaredType(index, type, codeGenConfig);
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/FEELPropertyAccessibleImplementation.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/FEELPropertyAccessibleImplementation.java
@@ -186,6 +186,8 @@ public class FEELPropertyAccessibleImplementation {
         MethodWithStringBody setFeelProperty = new MethodWithStringBody("fromMap", "void", body.toString());
         setFeelProperty.addParameter("java.util.Map<String, Object>", "values");
 
+        addOverrideAnnotation(setFeelProperty);
+
         return setFeelProperty;
     }
 
--- a/kie-dmn/kie-dmn-core/src/main/resources/org/kie/dmn/core/impl/DMNTypeSafeTypeTemplate.java
+++ b/kie-dmn/kie-dmn-core/src/main/resources/org/kie/dmn/core/impl/DMNTypeSafeTypeTemplate.java
@@ -43,8 +43,12 @@ public interface DMNTypeSafeTypeTemplate {
         {
             Object propertyValues = values.get("$property$");
             if(propertyValues != null) {
-                $property$ = new PropertyType();
-                $property$.fromMap((java.util.Map<String, Object>) propertyValues);
+                if (propertyValues instanceof PropertyType) {
+                    $property$ = (PropertyType) propertyValues;
+                } else {
+                    $property$ = new PropertyType();
+                    $property$.fromMap((java.util.Map<String, Object>) propertyValues);
+                }
             }
         }
 
@@ -53,13 +57,18 @@ public interface DMNTypeSafeTypeTemplate {
             Object propertyValues = values.get("$property$");
             if(propertyValues != null) {
                 $property$ = new java.util.ArrayList<>();
-                for (java.util.Map<String, Object> v : (java.util.Collection<java.util.Map<String, Object>>) propertyValues) {
-                    PropertyType item = new PropertyType();
-                    item.fromMap(v);
-                    $property$.add(item);
+                for (Object v : (java.util.Collection<Object>) propertyValues) {
+                    if (v instanceof PropertyType) {
+                        $property$.add((PropertyType)v);
+                    } else {
+                        PropertyType item = new PropertyType();
+                        item.fromMap((java.util.Map<String, Object>)v);
+                        $property$.add(item);
+                    }
                 }
             }
         }
+
         // Collections of basic fields
         {
             Object propertyValues = values.get("$property$");
@@ -71,4 +80,5 @@ public interface DMNTypeSafeTypeTemplate {
             }
         }
     }
+
 }
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/BaseVariantTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/BaseVariantTest.java
@@ -155,6 +155,10 @@ public abstract class BaseVariantTest {
         this.testConfig = testConfig;
     }
 
+    public boolean isTypeSafe() {
+        return testConfig.isTypeSafe();
+    }
+
     public DMNRuntime createRuntime(final Class testClass) {
         DMNRuntime runtime = DMNRuntimeUtil.createRuntime(testClass);
         if (testConfig.isTypeSafe()) {
@@ -211,6 +215,14 @@ public abstract class BaseVariantTest {
         }
     }
 
+    protected DMNResult evaluateDecisionService(DMNRuntime runtime, DMNModel dmnModel, DMNContext context, String decisionServiceName) {
+        if (testConfig.isTypeSafe()) {
+            return evaluateDecisionServiceTypeSafe(runtime, dmnModel, context, decisionServiceName);
+        } else {
+            return runtime.evaluateDecisionService(dmnModel, context, decisionServiceName);
+        }
+    }
+
     private DMNResult evaluateTypeSafe(DMNRuntime runtime, DMNModel dmnModel, DMNContext context) {
         Map<String, Object> inputMap = context.getAll();
         FEELPropertyAccessible inputSet;
@@ -229,6 +241,30 @@ public abstract class BaseVariantTest {
     protected Class<?> getStronglyClassByName(DMNModel dmnModel, String className) {
         String fqn = factory.create(dmnModel).appendPackage(className);
         return allCompiledClasses.get(fqn);
+    }
+
+    private DMNResult evaluateDecisionServiceTypeSafe(DMNRuntime runtime, DMNModel dmnModel, DMNContext context, String decisionServiceName) {
+        Map<String, Object> inputMap = context.getAll();
+        FEELPropertyAccessible inputSet;
+        try {
+            inputSet = createInstanceFromCompiledClasses(allCompiledClasses, factory.create(dmnModel), "InputSet");
+            inputSet.fromMap(inputMap);
+            return runtime.evaluateDecisionService(dmnModel, new DMNContextFPAImpl(inputSet), decisionServiceName);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public FEELPropertyAccessible convertToOutputSet(DMNModel dmnModel, DMNResult dmnResult) {
+        Map<String, Object> contextMap = dmnResult.getContext().getAll();
+        FEELPropertyAccessible outputSet;
+        try {
+            outputSet = createInstanceFromCompiledClasses(allCompiledClasses, factory.create(dmnModel), "OutputSet");
+            outputSet.fromMap(contextMap);
+            return outputSet;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
     }
 
     protected FEELPropertyAccessible createInstanceFromCompiledClasses(Map<String, Class<?>> compile, DMNTypeSafePackageName packageName, String className) throws Exception {
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/BaseVariantTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/BaseVariantTest.java
@@ -186,13 +186,15 @@ public abstract class BaseVariantTest {
 
     protected Map<String, Class<?>> allCompiledClasses;
 
+    protected Map<String, String> allSources;
+
     protected String testName = "";
 
     private void createTypeSafeInput(DMNRuntime runtime) {
         String prefix = String.format("%s%s", testName, testConfig.name());
         factory = new DMNTypeSafePackageName.ModelFactory(prefix);
         DMNAllTypesIndex index = new DMNAllTypesIndex(factory, runtime.getModels().toArray(new DMNModel[]{}));
-        Map<String, String> allSources = new HashMap<>();
+        allSources = new HashMap<>();
 
         for (DMNModel m : runtime.getModels()) {
             Map<String, String> allTypesSourceCode = new DMNTypeSafeTypeGenerator(m, index, factory)
@@ -235,14 +237,6 @@ public abstract class BaseVariantTest {
         }
     }
 
-    /**
-     * Utility test method, for class inspection.
-     */
-    protected Class<?> getStronglyClassByName(DMNModel dmnModel, String className) {
-        String fqn = factory.create(dmnModel).appendPackage(className);
-        return allCompiledClasses.get(fqn);
-    }
-
     private DMNResult evaluateDecisionServiceTypeSafe(DMNRuntime runtime, DMNModel dmnModel, DMNContext context, String decisionServiceName) {
         Map<String, Object> inputMap = context.getAll();
         FEELPropertyAccessible inputSet;
@@ -265,6 +259,14 @@ public abstract class BaseVariantTest {
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
+    }
+
+    /**
+     * Utility test method, for class inspection.
+     */
+    protected Class<?> getStronglyClassByName(DMNModel dmnModel, String className) {
+        String fqn = factory.create(dmnModel).appendPackage(className);
+        return allCompiledClasses.get(fqn);
     }
 
     protected FEELPropertyAccessible createInstanceFromCompiledClasses(Map<String, Class<?>> compile, DMNTypeSafePackageName packageName, String className) throws Exception {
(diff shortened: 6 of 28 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNOutputSetType.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Optional;
+
+import org.drools.modelcompiler.builder.generator.declaredtype.api.AnnotationDefinition;
+import org.drools.modelcompiler.builder.generator.declaredtype.api.FieldDefinition;
+import org.drools.modelcompiler.builder.generator.declaredtype.api.MethodDefinition;
+import org.drools.modelcompiler.builder.generator.declaredtype.api.TypeDefinition;
+import org.kie.dmn.api.core.DMNType;
+import org.kie.dmn.api.core.FEELPropertyAccessible;
+
+class DMNOutputSetType implements TypeDefinition {
```
tkobayas: I'll create a super class for DMNInputSetType and DMNOutputSetType to pull up methods/fields after confirming if I'm no the right track.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNTypeSafeTypeGenerator.java
```diff
@@ -97,6 +101,29 @@ public DMNTypeSafeTypeGenerator processTypes() {
...
+        }
+        for (DecisionNode d : decisions) {
+            outputSetType.addField(d.getName(), d.getResultType());
+        }
+        for (DecisionServiceNode ds : decisionServices) {
+            outputSetType.addField(ds.getName(), dmnModel.getTypeRegistry().unknown());
+        }
+        for (BusinessKnowledgeModelNode bkm : bkms) {
+            outputSetType.addField(bkm.getName(), dmnModel.getTypeRegistry().unknown());
+        }
```
tkobayas, line 126: Here, the actual field types are DMNDecisionServiceFunctionDefinitionEvaluator$DMNDSFunction or DMNFunctionWithReturnType. I use `dmnModel.getTypeRegistry().unknown()` (which results in Object type) for now but it may be an abuse of DMNType. Probably I should write another implementation of FieldDefinition to generate these fields.

kie-dmn/kie-dmn-core/src/main/resources/org/kie/dmn/core/impl/DMNTypeSafeTypeTemplate.java
```diff
@@ -43,8 +43,12 @@ void fromMap(Map<String, Object> values) {
...
             Object propertyValues = values.get("$property$");
             if(propertyValues != null) {
-                $property$ = new PropertyType();
-                $property$.fromMap((java.util.Map<String, Object>) propertyValues);
+                if (propertyValues instanceof PropertyType) {
+                    $property$ = (PropertyType) propertyValues;
+                } else {
+                    $property$ = new PropertyType();
+                    $property$.fromMap((java.util.Map<String, Object>) propertyValues);
+                }
```
tkobayas, line 51: When `OutputSet.fromMap()` takes DMNResult.getContext().getAll() as an argument, the context map may contain both Map based data and Typed objects (= generated by DMNTypeSafeTypeGenerator). So I enhanced the fromMap method like this. If you want keep fromMap() simple as before, I may write it as a different method (e.g. fromContextMap()). Feel free to let me know.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/BaseVariantTest.java
```diff
@@ -222,6 +234,30 @@ private DMNResult evaluateTypeSafe(DMNRuntime runtime, DMNModel dmnModel, DMNCon
...
+        Map<String, Object> contextMap = dmnResult.getContext().getAll();
+        FEELPropertyAccessible outputSet;
+        try {
+            outputSet = createInstanceFromCompiledClasses(allCompiledClasses, factory.create(dmnModel), "OutputSet");
+            outputSet.fromMap(contextMap);
+            return outputSet;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
```
tkobayas, line 262: Now we can generate `OutputSet` class and populate it by the map from `DMNResult.getContext().getAll()`. This is the goal of DROOLS-5502 (in kie-dmn-core), right? Then we will work on a KOGITO JIRA to generate REST endpoint which return type is `OutputSet`.

kie-dmn/kie-dmn-core/src/main/resources/org/kie/dmn/core/impl/DMNTypeSafeTypeTemplate.java
lucamolteni: Yes maybe we can keep this method a little bit simpler I agree

lucamolteni (approved):
Impressive speed Toshiya :)

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
```diff
@@ -318,6 +326,200 @@ public void testFieldCapitalization() {
...
+
+        final DMNContext context = DMNFactory.newContext();
+        context.set("myPerson", myPerson);
+        context.set("MyPerson", myPersonCapital);
+
+        final DMNResult dmnResult = runtime.evaluateAll(dmnModel, context);
+        assertThat(dmnResult.hasErrors(), is(false));
+        assertThat(dmnResult.getContext().get("myDecision"), is("myDecision is John"));
+        assertThat(dmnResult.getContext().get("MyDecision"), is("MyDecision is Paul"));
+
```
tkobayas: @lucamolteni @tarilabs 
Hey, I added an interesting test case. If we have node names where the difference is only "First letter is capitalized or not" (e.g. "myPerson" and "MyPerson"), it will cause a compile error for InputSet and OutputSet. Note that it works fine without typesafe.

https://github.com/tkobayas/drools/blob/DROOLS-5502/kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/stronglytyped/capitalLetterConflict.dmn

for example) Generated InputSet.java

https://gist.github.com/tkobayas/bde48ea2ef76ab2f829fe0c66cd26fd2

I think... we want to explicitly raise a build error in this case (if typesafe is enabled) even though the DMN itself is valid. Or do you have any other ideas? (hmm, adding an arbitrary suffix? but such a suffix may cause another conflict)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNTypeSafeTypeGenerator.java
tarilabs: I believe it should work if you set them in such a way for those later are set hardcoded to `org.kie.dmn.feel.runtime.FEELFunction` in the _resulting_ code-generated java class.

But it's okay for now to leave them here as FEEL : any as you are doing, the above can be an improvement for later.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/FEELPropertyAccessibleImplementation.java
```diff
@@ -186,9 +187,39 @@ private MethodDefinition fromMap() {
         MethodWithStringBody setFeelProperty = new MethodWithStringBody("fromMap", "void", body.toString());
         setFeelProperty.addParameter("java.util.Map<String, Object>", "values");
 
+        addOverrideAnnotation(setFeelProperty);
+
         return setFeelProperty;
     }
 
+    private MethodDefinition fromContextMap() {
```
tarilabs: I am not sure what is the goal of this codegen method. Could you clarify, please?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
```diff
@@ -318,6 +326,200 @@ public void testFieldCapitalization() {
...
+            assertThat(myDecision, is("myDecision is John"));
+            Object myDecisionCapital = (String) allProperties.get("MyDecision");
+            assertThat(myDecisionCapital, is("MyDecision is Paul"));
+        }
+    }
+
+    @Test
+    public void testDuplicateName() {
+        // Note: DMNValidator raises errors for duplicate names
+        final DMNRuntime runtime = createRuntime("duplicateName.dmn", this.getClass());
```
tarilabs: Yes, I'm not sure however there is value in testing this execution; the model, as you stated, is semantically incorrect due to the duplicate names, so I don't think we have to guarantee any behaviour here.

Could you further elaborate on the goal of this test?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
tarilabs: Ooh yes, this is an interesting point indeed! Thanks for spotting this!!

This is clearly a limitation of the DMN -> Java codegen strategy.

The DMN looks valid to me, and it's comforting it's executing fine with the default behaviour.

We should fix this problem when using strongly-typed, but I believe this is a separate issue.
Could you raise a separate JIRA, and leave this test as Ignored/Disabled ?
I believe whatever the strategy (for instance instead of playing with the first letter capitalization, we leave as-is, or we introduce a fixed `_`, or ... something else) this can be handled separately.

tarilabs (commented):
Thank you so far Toshiya!

I left some comments and sorry I couldn't review this right away.

Besides of what you find as comments below, I also believe there is something which we should evaluate in more details. Unless I am mistaken, the conversion of the output set is _external_ to the `DMNRuntime.evaluate` phase. I believe that is the reason why in your test, you first run the evaluation of the model (eventually using strongly typed input set) and after evaluation is done you take the dmncontext from the output and you convert it to the strongly typed variant.

I believe this can be a valid first step, and we can even consider this as a first iteration on the matter and eventually merge it with this approach, if it ease the work.
But the longer term however it should be that when the `DMNRuntime.evaluate` methods return, they return already a `DMNResult` having already the strongly typed variant of the context as the resulting context.

IIRC internally the dmnruntime performs a `DMNContext.clone()` from the input context / input set, to create the result's dmn context. 
I believe what will be needed is that IFF the input context is a strongly typed variant, in the DMNResult it will be created the corresponding output set for the context of the result --while keeping having the current, map-backed generic `DMNContextImpl` for all the other cases

What do you think?

As I mentioned, we can look into this separately if that helps and I can support further in the specific details if I was not able to fully explain this point

Thanks again so far!!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
tkobayas: Sure, filed https://issues.redhat.com/browse/DROOLS-5518

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
tkobayas: Thanks, yeah, I didn't need to have this test. Removing :)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/FEELPropertyAccessibleImplementation.java
tkobayas: `fromMap()` is to convert simple Map data to a strongly typed object.

However, `DMNResult.getContext().getAll()` may contain both simple Map data (given from DMN runtime execution) and strongly typed objects (given from `InputSet`).

So I added `fromContextMap()` which can convert both data to `OutputSet`. I was able to combine those logic in one `fromMap()` method but Luca preferred to keep `fromMap()` simple so I split to 2 methods.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNTypeSafeTypeGenerator.java
tkobayas: Thanks, filed https://issues.redhat.com/browse/DROOLS-5519

tkobayas:
Thank you for the review, @tarilabs !

> Unless I am mistaken, the conversion of the output set is external to the DMNRuntime.evaluate phase. 
> ...
> But the longer term however it should be that when the DMNRuntime.evaluate methods return, they return already a DMNResult having already the strongly typed variant of the context as the resulting context.
> ...
> As I mentioned, we can look into this separately if that helps and I can support further in the specific details if I was not able to fully explain this point

Yes, currently the conversion is external to DMNRuntime. Thanks, separating will help :)  I filed https://issues.redhat.com/browse/DROOLS-5517 and will work on it after this PR.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
```diff
@@ -107,9 +110,9 @@ private String convertBuiltin(DMNType expectedFEELType) {
             case DATE:
                 return LocalDate.class;
             case TIME:
-                return LocalTime.class;
+                return Temporal.class;
```
tarilabs, line 107: not sure about this, as the role of this method IIRC is to suggest the Class to be used in the destination Java code generated. The engine will not support a _generic_ temporal for a time, but it does support `LocalTime`. So I believe it was correct before.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
```diff
@@ -107,9 +110,9 @@ private String convertBuiltin(DMNType expectedFEELType) {
...
             case DATE_TIME:
-                return LocalDateTime.class;
+                return Temporal.class;
```
tarilabs, line 109: as above

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
```diff
@@ -126,10 +129,10 @@ private String convertBuiltin(DMNType expectedFEELType) {
         switch (expectedFEELType.getName()) {
             case SimpleType.YEARS_AND_MONTHS_DURATION:
             case "yearMonthDuration":
-                return Period.class;
+                return ChronoPeriod.class;
```
tarilabs: as above

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
```diff
@@ -126,10 +129,10 @@ private String convertBuiltin(DMNType expectedFEELType) {
...
             case SimpleType.DAYS_AND_TIME_DURATION:
             case "dayTimeDuration":
-                return Duration.class;
+                return TemporalAmount.class;
```
tarilabs, line 129: as above

tarilabs (commented):
not sure about modifications to DMNAllTypesIndex.java, could you check please?

tkobayas:
> not sure about modifications to DMNAllTypesIndex.java, could you check please?

Hi @tarilabs , Indeed, this is another point which I'd like to hear your thoughts. In some test cases, I hit failures:

DMNRuntimeTypesTest.testOneOfEachType()
```
assertThat(dmnResult.getDecisionResultByName("DecisionYMDuration").getResult(), is(ComparablePeriod.parse("P2M")));
```
  -> ComparablePeriod is put in the context by runtime. So ClassCastException is thrown for java.time.Period field during conversion.

DMNStronglyTypedSupportTest.testDateAndTime()
```
assertThat(ctx.get("Date-Time"), is(ZonedDateTime.of(2016, 12, 24, 23, 59, 0, 0, ZoneOffset.ofHours(-5))));
```
  -> ZonedDateTime is put in the context by runtime. So ClassCastException is thrown for java.time.LocalDateTime field during conversion.

```
assertThat(ctx.get("Time"), is(OffsetTime.of(0, 0, 1, 0, ZoneOffset.ofHours(-1))));
```
  -> OffsetTime is put in the context by runtime. So ClassCastException is thrown for java.time.LocalTime field during conversion.

```
assertThat(ctx.get("ymDuration2"), is(ComparablePeriod.parse("P1Y")));
```
  -> ComparablePeriod is put in the context by runtime. So ClassCastException is thrown for java.time.Duration field during conversion.

I think solutions would be:

A) Have generated class field types to be Interfaces like ChronoPeriod(or TemporalAmount), Temporal, TemporalAmount so that they can accept above objects. (as done in this update)
 -> Downside: It's not very strong typed :)

B) Convert those above objects to expected types (e.g. if ComparablePeriod, ComparablePeriod.asPeriod()) in fromContextMap() method

If you think B) is expected, I'll implement it. Or if you think of other approaches, please let me know. Thanks!

kie-dmn/kie-dmn-core/src/test/resources/org/kie/dmn/core/0007-date-time.dmn
```diff
@@ -291,8 +291,8 @@
       <text>ymDuration2.years</text>
     </literalExpression>
   </decision>
-  <decision id="_c2f1a1ae-1403-43a2-b176-dedf3da51e6b" name="seconds">
-    <variable name="seconds" typeRef="feel:number"/>
+  <decision id="_c2f1a1ae-1403-43a2-b176-dedf3da51e6b" name="d1seconds">
+    <variable name="d1seconds" typeRef="feel:number"/>
```
tkobayas, line 295: I have changed the name because it hits capitalized/non-capitalized name issue (https://issues.redhat.com/browse/DROOLS-5518) when typesafe is enabled. Instead, I created DMNRuntimeTypesTest.testCapitalLetterConflictWithInputAndDecision() to clarify the issue and will be addressed in  DROOLS-5518

tarilabs:
Hi @tkobayas re https://github.com/kiegroup/drools/pull/2993#issuecomment-660862246 I believe we should try with a third approach :) 

I see the problem and you are right.

I believe the method `DMNAllTypesIndex.convertBuiltin` is originally thought for the expected equivalent Java classes from an input point of view only. Your comments are relevant, and I believe is a _symptom_ we need a different strategy for output classes (not corresponding to InputData nodes, but to Decision, etc.)

Let me try with an example.
Let's say there is DMN for FEEL: YEARS_AND_MONTHS_DURATION with 1 InputData and 1 Decision, both of this type.
In Java, the equivalent for the InputData type is definitely `java.time.Period`
the equivalent for the Decision type is likely `org.kie.dmn.feel.lang.types.impl.ComparablePeriod`
(we do use internally `ComparablePeriod` since the JVM one is missing relevant methods for operations on periods such as `P1Y2M` etc)

IFF we generify to the common interface as you are proposing in the code change, I believe we will need however to bring reference to the original FEEL type, such that we will be able for Swagger/OpenAPI purposes, to distinguish a `Temporal` for a `FEEL:date and time` vs `FEEL:time`.
I believe if you find a way to capture the original type up to the generated code (say for now to put it as a generated Javadoc comment on the corresponding field) it would be the preferred approach.

Alternatively, we could keep the mapping to the specific Java classes as-it-was for the InputData only (so `Period` for the `FEEL:years and months duration`) and use the proposed mapping to the generic Java interfaces as-you-propose-here for the Decision nodes only.

wdyt ?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/FEELPropertyAccessibleImplementation.java
tarilabs: Yeah I noticed, but sorry I don't concur with this additional method, for the reason it impacts the `FEELPropertyAccessible` interface.
I don't think it make sense to have two methods with the same semantic signature, the user will find difficult to know when to invoke one or the other, or if implementing the pojo manually (ref scaffolding) it will find difficult to understand what's the difference between the two to implement.

Unless I am mistaken, there are two differences.

## Difference1
is relevant to the `// Composite fields`
image

I believe is perfectly fine for this to be incorporated in the original `fromMap()`, that is ~saying: if I'm being passed already the target type I will assign it directly, otherwise I will create myself the target type using the `fromMap()`.

## Difference2
is relevant to the `// Collections of composite fields`
image

I am not sure this is the modification you want. 
I believe we could leave the enhanced for loop where it was originally, can be generic on `Object` instead of `Map<>` and inside of the loop you do similarly as to what you did above for the "Composite fields" case.

Wdyt?
This way, we have only slight modifications to the original `fromMap()` which made sense any way, since the target object could have been used in a scaffolding approach, and we won't need an additional method in the `FEELPropertyAccessible` interface.

lucamolteni: I haven't thought about affecting the fromMap() method, I thought we were creating a new private method. Any way is fine

tkobayas: Thank you for the suggestion. I removed fromContextMap() and combined to fromMap().

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
```diff
@@ -318,6 +406,358 @@ public void testFieldCapitalization() {
...
+            assertThat(myPersonOut.getFEELProperty("age").toOptional().get(), is(new BigDecimal(20)));
+            assertThat(myPersonOut.getFEELProperty("employmentPeriod").toOptional().get(), is(ComparablePeriod.of(1, 3, 1)));
+        }
+    }
+
+    @Ignore
+    @Test
+    public void testCollectionType() {
+        // To be fixed by DROOLS-5538
+        final DMNRuntime runtime = createRuntime("PersonListHelloBKM2.dmn", this.getClass());
```
tkobayas, line 647: While adding test variations, I found that itemDefinition with isCollection="true" is not properly handled (itemComponent with isCollection="true" works fine). I filed another JIRA and set `@Ignore` at the moment.
https://issues.redhat.com/browse/DROOLS-5538

tkobayas:
I have updated this PR just about removing fromConxtetMap() and adding test cases.

Regarding this topic @tarilabs :

> IFF we generify to the common interface as you are proposing in the code change, I believe we will need however to bring reference to the original FEEL type, such that we will be able for Swagger/OpenAPI purposes, to distinguish a Temporal for a FEEL:date and time vs FEEL:time.
I believe if you find a way to capture the original type up to the generated code (say for now to put it as a generated Javadoc comment on the corresponding field) it would be the preferred approach.
>
> Alternatively, we could keep the mapping to the specific Java classes as-it-was for the InputData only (so Period for the FEEL:years and months duration) and use the proposed mapping to the generic Java interfaces as-you-propose-here for the Decision nodes only.

It's difficult for me to decide which is better. As you prefer the former, I'm going to try the approach and will update again.

Thanks!
Toshiya

tkobayas:
Rebased to solve conflicts. FEELType Javadoc is not yet implemented.

tkobayas:
@tarilabs 
Now generated Java classes have a javadoc comment with DMNType like this:
https://gist.github.com/tkobayas/1f275706e68871072b062a04d9bb1666

We may tailor the generated source codes further for kogito REST endpoint swagger/OpenAPI support (maybe adding ApiModelProperty annotation?). Shall I file a JIRA for that?

tarilabs:
Jenkins execute compile downstream build

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/JavadocTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertJavadoc(cu, "fromYearMonthDay", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : date }");
+
+        String tDateTimeComponents = sourceMap.get("TDateTimeComponents");
+        cu = StaticJavaParser.parse(tDateTimeComponents);
+        assertJavadoc(cu, "year", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
+        assertJavadoc(cu, "month", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
+        assertJavadoc(cu, "day", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
+        assertJavadoc(cu, "hour", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
+        assertJavadoc(cu, "minute", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
+        assertJavadoc(cu, "second", "DMNType{ http://www.omg.org/spec/FEEL/20140401 : number }");
```
tarilabs, line 132: this is awesome, thanks @tkobayas !

Just a side note however, let's limit Javadoc assertions at this model for now, the reason is, I believe this Javadoc format will change to make it more "business friendly", so I would like to avoid having to update later too many assertions! 😆 
Also, I believe we can use Eclipse MP annotation now so they will indeed properly reflect in the generated swagger/OpenApi

tarilabs (approved):

tarilabs:
> We may tailor the generated source codes further for kogito REST endpoint swagger/OpenAPI support (maybe adding ApiModelProperty annotation?). Shall I file a JIRA for that?

I don't know about this annotation, but I successfully used `@Schema` from Eclipse MP.
Anyway yes please raise a JIRA @tkobayas so we can track this and we discuss it at due time, thanks again! :)

tarilabs:
One last comment, I see some DMN models are copied into `stronglytyped` directory, but this should not be necessary, since the models could be loaded ~using : 

```
createRuntime("0007-date-time.dmn", original.package.TestClass.class);
```

I believe avoiding model file duplication could be addressed as part of this PR, or can also be done as a separate task just after, if we want to prioritize this PR merge! :) Thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/JavadocTest.java
tkobayas: Sure, I'll keep in mind :)

tkobayas:
> I don't know about this annotation, but I successfully used `@Schema` from Eclipse MP.
> Anyway yes please raise a JIRA @tkobayas so we can track this and we discuss it at due time, thanks again! :)

`ApiModelProperty` is Swagger annotation but Eclipse MP `@Schema` sounds like the right direction. Thank you for letting me know. I filed a JIRA. https://issues.redhat.com/browse/DROOLS-5554

tkobayas:
> I believe avoiding model file duplication could be addressed as part of this PR, or can also be done as a separate task just after, if we want to prioritize this PR merge! :) Thanks!

Thank you for the suggestion. I'd like to do that as a separate task because this PR is getting longish... :) I filed https://issues.redhat.com/browse/DROOLS-5555 Thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNAllTypesIndex.java
hellowdan: @tkobayas could you take a look on this comment, please?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
hellowdan: Just asking: wouldn't it be better to have separate tests for TypeSafe variations?

hellowdan (approved):

tkobayas:
Hi @hellowdan 

> @tkobayas could you take a look on this comment, please?

@tarilabs  and I discussed it in these comments:

https://github.com/kiegroup/drools/pull/2993#issuecomment-660862246
https://github.com/kiegroup/drools/pull/2993#issuecomment-660898266

DMN engine may output different Java classes for those DMN types (time related) so we use interfaces like Temporal or TemporalAmount. In order to help REST endpoint users, we plan to add annotations to describe DMN types https://issues.redhat.com/browse/DROOLS-5554

tkobayas:
@hellowdan 
> Just asking: wouldn't it be better to have separate tests for TypeSafe variations?

Thank you for pointing that out. I will consider separating tests in the next JIRA : https://issues.redhat.com/browse/DROOLS-5538

PR #3005, opened 2020-07-24
[DROOLS-5522] raise an error when trying to create a KieModule from a…

mariofusco:
… malformed kmodule.xml


mariofusco:

PR #3017, opened 2020-07-30
[DROOLS-5546] fix accumulate with more than one binding in executable…

mariofusco:
… model


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DeclarationSpec.java
```diff
@@ -111,6 +113,14 @@ public void setBoundVariable( String boundVariable ) {
         this.boundVariable = boundVariable;
     }
 
+    public MethodCallExpr getBidingExpr() {
```
lucamolteni: Typo

lucamolteni (approved):

PR #3018, opened 2020-07-30
DROOLS-5553 - fix getMetadata in DMNContextFPAImpl

r00ta:
Jira: https://issues.redhat.com/browse/DROOLS-5553

Hi!
I've found that when `kogito.decisions.stronglytyped=true` then `DMNContextFPAImpl` is used. The problem is that the kogito-tracing-addon is storing some information in the metadata of such class, but the `get` always return a new object -> the tracing addon is then broken in such case.

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
@@ -30,10 +30,12 @@ import org.kie.dmn.core.impl.DMNContextImpl.ScopeReference;
 public class DMNContextFPAImpl implements DMNContext {
 
     private final FEELPropertyAccessible fpa;
-    private Deque<ScopeReference> stack    = new LinkedList<>();
+    private Deque<ScopeReference> stack = new LinkedList<>();
+    private DMNMetadataImpl metadata;
 
     public DMNContextFPAImpl(FEELPropertyAccessible bean) {
         this.fpa = bean;
+        this.metadata = new DMNMetadataImpl();
     }
 
     @Override
@@ -86,16 +88,15 @@ public class DMNContextFPAImpl implements DMNContext {
 
     @Override
     public DMNMetadata getMetadata() {
-        return new DMNMetadataImpl();
+        return this.metadata;
     }
 
     @Override
     public DMNContext clone() {
-        DMNContextImpl newCtx = new DMNContextImpl(fpa.allFEELProperties());
+        DMNContextImpl newCtx = new DMNContextImpl(fpa.allFEELProperties(), metadata.asMap());
         for (ScopeReference e : stack) {
             newCtx.pushScope(e.getName(), e.getNamespace());
         }
         return newCtx;
     }
-
 }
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
```diff
@@ -86,16 +88,15 @@ public boolean isDefined(String name) {
...
     @Override
     public DMNMetadata getMetadata() {
-        return new DMNMetadataImpl();
+        return this.metadata;
     }
 
     @Override
     public DMNContext clone() {
-        DMNContextImpl newCtx = new DMNContextImpl(fpa.allFEELProperties());
+        DMNContextImpl newCtx = new DMNContextImpl(fpa.allFEELProperties(), metadata.asMap());
```
r00ta, line 96: Side question: Is it correct that the `clone` method returns `DMNContextImpl` instead of `DMNContextFPAImpl`?

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNTypeSafeTest.java
```diff
@@ -158,6 +159,16 @@ public void testDynamic() throws Exception {
         assertThat(d.get("Hello"), is("Hello Mr. x"));
     }
 
+    @Test
+    public void testMetadata(){
```
danielezonca: Can you please also clone the `DMNContext` and verify that `"value"` is still there?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
danielezonca: Good question :thinking: 
@lucamolteni can you help?

danielezonca (approved):

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNTypeSafeTest.java
r00ta: done :+1:

r00ta:
Hi, can anybody restart the github action?

danielezonca:
@r00ta 
GitHub Action builds are still under development, don't worry for them :)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
lucamolteni: @r00ta @danielezonca 

Ok short answer: because it doesn't work in another way :)

Slightly longer answer: the current implementation of the DMN evaluation clones the context at least once, and always using the type safe context in the internals wasn't the original purpose of this implementation. 
In other words: the idea so far was to expose a typesafe inputset as an API, not to change internals. 
I can't say it's not doable, but to me the advantages are not obvious. 

@tarilabs can you confirm this?

Thanks

lucamolteni (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
tarilabs: yes confirmed as currently `DMNRuntime` for evaluation uses input context `.clone()` result, as the context where to store the outputs, and `DMNContextImpl` can accept any additional variables, while `DMNContextFPAImpl` atm is only accepting the fields defined in the wrapped `FEELPropertyAccessible`

tarilabs (commented):
commented as requested here https://github.com/kiegroup/drools/pull/3018#discussion_r463468389

tkobayas (approved):

r00ta:
Jenkins retest this please

PR #2986, opened 2020-07-13
[DROOLS-5503] avoid creating a KnowledgeBuilder where there is no ass…

mariofusco:
…et to be built

PR #3021, opened 2020-07-31
[DROOLS-5503][RHDM-1406] avoid creating a KnowledgeBuilder where ther…

mariofusco:
…e is no asset to be built (#2986)


mareknovotny:
Jenkins retest this please

PR #3030, opened 2020-08-05
[DROOLS-5538] DMN strongly typed class compile errors for collection …

tkobayas:
…types

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNModelTypesIndex.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNModelTypesIndex.java
@@ -61,7 +61,8 @@ class DMNModelTypesIndex {
     }
 
     private boolean shouldIndex(DMNType dmnType) {
-        return !dmnType.getNamespace().equals(model.getDefinitions().getURIFEEL()) && dmnType.isComposite();
+        // Don't index a collection type of a declared type. e.g. tPersonList which is a collection of tPerson
+        return !dmnType.getNamespace().equals(model.getDefinitions().getURIFEEL()) && dmnType.isComposite() && !(dmnType.isCollection() && dmnType.getBaseType() != null);
     }
 
     private void index(DMNType innerType) {
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
@@ -642,10 +642,8 @@ public class DMNRuntimeTypesTest extends BaseVariantTest {
         }
     }
 
-    @Ignore
     @Test
-    public void testCollectionType() {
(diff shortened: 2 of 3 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNModelTypesIndex.java
```diff
@@ -61,7 +61,8 @@ public void createIndex() {
     }
 
     private boolean shouldIndex(DMNType dmnType) {
-        return !dmnType.getNamespace().equals(model.getDefinitions().getURIFEEL()) && dmnType.isComposite();
+        // Don't index a collection type of a declared type. e.g. tPersonList which is a collection of tPerson
+        return !dmnType.getNamespace().equals(model.getDefinitions().getURIFEEL()) && dmnType.isComposite() && !(dmnType.isCollection() && dmnType.getBaseType() != null);
```
tkobayas, line 65: Don't index a collection type of a declared type (e.g. `tPersonList` which is a collection of `tPerson` in `testTopLevelTypeCollection`) because it will be a `Collection<TPerson>` field in `InputSet`.

On the other hand, a collection of a composite structure (e.g. `tPairList` which is a collection of composite of `string` and `number` in `testTopLevelCompositeCollection`) will generate a Java class of the composite structure. The collection will be a `Collection<TPairList>` field in `InputSet`. (btw, in this case, we may want to name the type "tPair" rather than "tPairList". But I chose "tPairList" because it looks natural in the DMN itself)

lucamolteni (approved):

tkobayas:
@hellowdan , per your comment, I considered separating some tests (including testTopLevelTypeCollection) to a TypeSafe specific test class. However, validating these tests with non-typesafe is useful to confirm that typesafe behaves the same as non-typesafe so I didn't separate them eventually. If you have other thoughts, feel free to let me know. Thanks!

https://github.com/kiegroup/drools/pull/2993/#discussion_r463500865

hellowdan (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNModelTypesIndex.java
tarilabs: Makes sense, thank you for this fix @tkobayas !
We might need to further check indexing when dealing with collection of collection (of collection)* later
But it's comforting to know it now works in the most common use cases thanks to this fix!

tarilabs (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/DMNModelTypesIndex.java
tkobayas: > We might need to further check indexing when dealing with collection of collection (of collection)* later

Sure, filed a JIRA https://issues.redhat.com/browse/DROOLS-5561

mareknovotny:
Jenkins do compile downstream build

mbiarnes:
Jenkins do compile downstream build

PR #3032, opened 2020-08-05
[DROOLS-5517] Enhance DMNResult/DMNContext to be able to return stron…

tkobayas:
…gly typed OutputSet

WIP: This is a draft PR. Feel free to share your thoughts/ideas. @tarilabs @lucamolteni 

* I created DMNResultFPAImpl which has `getOutputSet()` method. But it means you need to cast the result in order to use the method. Should I move the method to Interface DMNResult?
* In this draft, DMNResultFPAImpl receives an OutputSet instance from DMNContextFPAImpl. It means that you need to instantiate InputSet/OutputSet instances before executing DMNRuntime. (as we do in BaseVariantTest.createTypeSafeInput/evaluateTypeSafe). Or do we want to encapsulate those logic inside DMNRuntime?

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
@@ -33,6 +33,8 @@ public class DMNContextFPAImpl implements DMNContext {
     private Deque<ScopeReference> stack = new LinkedList<>();
     private DMNMetadataImpl metadata;
 
+    public static final String STRONGLY_TYPED_FPA = "STRONGLY_TYPED_FPA"; // metadata key
+
     public DMNContextFPAImpl(FEELPropertyAccessible bean) {
         this.fpa = bean;
         this.metadata = new DMNMetadataImpl();
@@ -46,6 +48,15 @@ public class DMNContextFPAImpl implements DMNContext {
     @Override
     public Object get(String name) {
         return fpa.getFEELProperty(name).toOptional().orElse(null);
+    }
+
+    /**
+     * Internal utility method
+     * 
+     * @return FEELPropertyAccessible which represents strongly typed context
+     */
+    public FEELPropertyAccessible getFpa() {
+        return fpa;
     }
 
     private Map<String, Object> getCurrentEntries() {
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
@@ -33,8 +33,6 @@ public class DMNContextFPAImpl implements DMNContext {
     private Deque<ScopeReference> stack = new LinkedList<>();
     private DMNMetadataImpl metadata;
 
-    public static final String STRONGLY_TYPED_FPA = "STRONGLY_TYPED_FPA"; // metadata key
-
     public DMNContextFPAImpl(FEELPropertyAccessible bean) {
         this.fpa = bean;
         this.metadata = new DMNMetadataImpl();
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/AbstractDMNSetType.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/AbstractDMNSetType.java
@@ -32,7 +32,8 @@ import org.kie.dmn.api.core.FEELPropertyAccessible;
 
 abstract class AbstractDMNSetType implements TypeDefinition {
 
-    List<DMNDeclaredField> fields = new ArrayList<>();
+    List<DMNDeclaredField> dmnFields = new ArrayList<>();
+    List<FieldDefinition> extraFields = new ArrayList<>();
 
     Map<String, DMNType> fieldsKey = new HashMap<>();
 
@@ -53,15 +54,19 @@ abstract class AbstractDMNSetType implements TypeDefinition {
 
     @Override
     public List<? extends FieldDefinition> getFields() {
-        return fields;
+        List<FieldDefinition> combinedFields = new ArrayList<>();
+        combinedFields.addAll(dmnFields);
+        combinedFields.addAll(extraFields);
+        return combinedFields;
     }
 
     public void initFields() {
         FieldGenStrategy fieldGenStrategy = FieldGenStrategy.getFieldGenStrategy(fieldsKey.keySet(), getTypeName());
         for (Map.Entry<String, DMNType> f : fieldsKey.entrySet()) {
             DMNDeclaredField dmnDeclaredField = new DMNDeclaredField(index, f, codeGenConfig, fieldGenStrategy);
-            fields.add(dmnDeclaredField);
+            dmnFields.add(dmnDeclaredField);
         }
+        extraFields.add(new DefinedKeySetField());
     }
 
     @Override
@@ -81,7 +86,7 @@ abstract class AbstractDMNSetType implements TypeDefinition {
 
     @Override
     public List<MethodDefinition> getMethods() {
-        return new FEELPropertyAccessibleImplementation(fields).getMethods();
+        return new FEELPropertyAccessibleImplementation(dmnFields, this).getMethods();
     }
 
     @Override
@@ -101,5 +106,43 @@ abstract class AbstractDMNSetType implements TypeDefinition {
     @Override
     public Optional<String> getJavadoc() {
         return Optional.of(this.javadoc);
+    }
+
+    class DefinedKeySetField implements FieldDefinition {
+
+        @Override
+        public String getFieldName() {
+            return "definedKeySet";
+        }
+
+        @Override
+        public String getObjectType() {
+            return "java.util.Set<String>";
+        }
+
+        @Override
+        public String getInitExpr() {
+            return "new java.util.HashSet<>()";
+        }
+
+        @Override
+        public boolean isKeyField() {
+            return false;
+        }
+
+        @Override
+        public boolean createAccessors() {
+            return false;
+        }
+
+        @Override
+        public boolean isStatic() {
+            return false;
+        }
+
+        @Override
+        public boolean isFinal() {
+            return false;
+        }
     }
 }
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
@@ -29,7 +29,6 @@ import java.util.Map;
 
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNDecisionResult.DecisionEvaluationStatus;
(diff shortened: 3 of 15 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
```diff
@@ -29,12 +29,20 @@
 
 public class DMNContextFPAImpl implements DMNContext {
 
-    private final FEELPropertyAccessible fpa;
+    private final FEELPropertyAccessible inputSet;
+    private final FEELPropertyAccessible outputSet;
```
tarilabs: I am not sure we need for a context to hold reference for an input Vs output, given this would lead to a different design and semantic.

I believe it should be possible to keep `DMNContextFPAImpl` as is.

I also believe it should be possible to keep `DMNResultImpl` without any need for an additional specialization, but this less of a concern to me.

When the `DMNRuntimeImpl` creates the `DMNResultImpl`, it should be possible to supply a `DMNContext` which has for the _wrapped_ `FEELPropertyAccessible fpa` the full list of inputs,decisions,bkms, etc.

What do you think?

tkobayas:
Thank you for the review, @tarilabs 

> I believe it should be possible to keep DMNContextFPAImpl as is.
> 
> I also believe it should be possible to keep DMNResultImpl without any need for an additional specialization, but this less of a concern to me.
>
> When the DMNRuntimeImpl creates the DMNResultImpl, it should be possible to supply a DMNContext which has for the wrapped FEELPropertyAccessible fpa the full list of inputs,decisions,bkms, etc.

I updated the PR. This time, I keep `DMNContextFPAImpl` and `DMNResultImpl` unchanged. Instead, I introduced `DMNContextTypeSafeImpl` which works like a usual `DMNContextImpl` during runtime execution and also holds `fpa` (= `OutputSet`).

One main concern is how `DMNContextTypeSafeImpl` can know `OutputSet` class in order to convert the result map to the `fpa`. At the moment, `InputSet` and `OutputSet` are compiled/instantiated outside of `DMNRuntime` (e.g. `BaseVariantTest.evaluateTypeSafe()`). So in this PR, I supplied `OutputSet` class from the test case after `runtime.evaluateAll()` (I demonstrate it only in `BaseVariantTest.testOneOfEachType()` just for explanation). But probably it's not what you expect. Do you want to supply the `OutputSet` class when we call `evaluateAll()`? Then, how to supply it... (maybe using `DMNContextFPAImpl.metadata`?). I may still be misunderstanding your points. Feel free to share your thoughts. Thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/stronglytyped/DMNRuntimeTypesTest.java
```diff
@@ -98,7 +99,8 @@ public void testOneOfEachType() {
         assertThat(dmnResult.getDecisionResultByName("DecisionTime").getResult(), is(LocalTime.of(10, 0)));
 
         if (isTypeSafe()) {
-            FEELPropertyAccessible outputSet = convertToOutputSet(dmnModel, dmnResult);
+            DMNContextTypeSafeImpl resultContext = (DMNContextTypeSafeImpl)dmnResult.getContext();
+            FEELPropertyAccessible outputSet = resultContext.getFpa(getOutputSetClass(dmnModel));
```
tkobayas: Here I supply `OutputSet` class.

tkobayas:
Jenkins trigger cdb

tkobayas:
Rebased because of conflict

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNContextFPAImpl.java
```diff
@@ -93,7 +149,15 @@ public DMNMetadata getMetadata() {
...
-        DMNContextImpl newCtx = new DMNContextImpl(fpa.allFEELProperties(), metadata.asMap());
+        FEELPropertyAccessible newFpa;
+        try {
+            newFpa = fpa.getClass().getDeclaredConstructor().newInstance();
+        } catch (InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException
+                | NoSuchMethodException | SecurityException e) {
+                    throw new DMNTypeSafeException("Exception while instantiating " + fpa.getClass(), e);
+        }
+        newFpa.fromMap(fpa.allFEELProperties());
+        DMNContextFPAImpl newCtx = new DMNContextFPAImpl(newFpa, metadata.asMap(), fallbackMap);
```
tarilabs:
```suggestion
        DMNContextFPAImpl newCtx = new DMNContextFPAImpl(newFpa, metadata.asMap(), fallbackMap.clone());
```

I assume?

tkobayas:
Rebased because of conflict. Also squashed commits because they contain back-and-forth commits which make it difficult to understand.

Implemented as:

1. A client creates `DMNContextImpl` with an `OutputSet` object which contains input data. Execute `DMNRuntime` with the context.
2. `DMNRuntimeImpl` creates a `DMNContextImpl` from `DMNContextFPAImpl`
3. `DMNRuntimeImpl` evaluation goes through with `DMNContextImpl`
4. Before responding to the client, create `DMNContextFPAImpl` with `OutputSet` and return

Filed a new JIRA to use `DMNContextFPAImpl` for `DMNRuntime` evaluation to split from this JIRA.

https://issues.redhat.com/browse/DROOLS-5611

Key points:
* In order to achieve the above sequence, the `OutputSet` object is internally conveyed through `DMNContext.metadata` (key=`DMNContextFPAImpl.STRONGLY_TYPED_FPA`). Users don't need to mind.
* Introduced `DMNResultImpl.stronglyTyped` flag to know if it should convert from `DMNContextImpl` to `DMNContextFPAImpl` in step 4. I chose `DMNResult` for the flag because there could be nested evaluations. Users also don't need to mind.

  These 2 points could be removed once we implement DROOLS-5611

* Added `DefinedKeySetField` for `AbstractDMNSetType` (= `InputSet`/`OutputSet`). It is used to collect "explicitly added property even if the value is null". So it will adjust the result Map which is returned by `allFEELProperties()` (= put the property with null value). This is required to make `DMNContextFPAImpl.isDefined()` work correctly.

tkobayas:
CDB error doesn't seem to be related:
```
17:36:00  [ERROR] Failed to execute goal on project google-calendar-workitem: Could not resolve dependencies for project org.jbpm.contrib:google-calendar-workitem:jar:7.43.0-SNAPSHOT: Could not transfer artifact com.google.oauth-client:google-oauth-client:jar:1.23.0 from/to mirror-central (https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kie-all/): /home/jenkins/.m2/repository/com/google/oauth-client/google-oauth-client/1.23.0/google-oauth-client-1.23.0.jar.part (No such file or directory) -> [Help 1]
17:36:00  org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project google-calendar-workitem: Could not resolve dependencies for project org.jbpm.contrib:google-calendar-workitem:jar:7.43.0-SNAPSHOT: Could not transfer artifact com.google.oauth-client:google-oauth-client:jar:1.23.0 from/to mirror-central (https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kie-all/): /home/jenkins/.m2/repository/com/google/oauth-client/google-oauth-client/1.23.0/google-oauth-client-1.23.0.jar.part (No such file or directory)
```

tarilabs (commented):
@tkobayas thank you!

Overall is fine, but I believe we should seize the chance to fully update as discussed in email thread, so that `DMNRuntimeImpl` is not responsible in having to deal with the classname for outputset.

I believe for now, we should let `DMNRuntimeImpl` just use internally `DMNContextImpl` for evaluation, and just return that out of `evaluate...` methods.

Then in the tests, since we are in full control of the name of the outputset class, we return that to the concrete test method. In other words the transformation from `DMNContext` based on map, to `DMNContext` based on the outputset FPA, is handled by the test infrastructure.

This way, we don't have to use metadata for the outputset classname, and `DMNRuntime` does not need any modification for that transformation.
I believe this will reflect more closely what we are meaning to do for REST enpoint on Kogito anyway.

What do you think?

tkobayas:
Jenkins retest this

tkobayas:
Thank you for reviewing, @tarilabs 

Yes, sure. I moved the conversion to the test code. Now DMNRuntimeImpl is much simpler.

Is this good to merge? :)

tkobayas:
Jenkins run fdb

hellowdan (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/AbstractDMNSetType.java
```diff
@@ -32,7 +32,8 @@
 
 abstract class AbstractDMNSetType implements TypeDefinition {
 
-    List<DMNDeclaredField> fields = new ArrayList<>();
+    List<DMNDeclaredField> dmnFields = new ArrayList<>();
+    List<FieldDefinition> extraFields = new ArrayList<>();
```
tarilabs, line 36: just to be sure I understood the latest modifications. This `extraFields` is used to store the `DefinedKeySetField` which is a `Set<String>` of the properties which have been set on the Input/Output Set pojo. However, this is not really "used" so far, since DMNRuntimeImpl uses the Map-backed DMNContextImpl. Correct?
It's fine I just wanted to check the intention :)

tarilabs (approved):
Looks good to me, thanks!
Just a question to check I understood correctly.

tarilabs:
image

🤷‍♂️ 

Jenkins run cdb

lucamolteni:
> image
> 
> 🤷‍♂️
> 
> Jenkins run cdb

AFAIK 11 is borken

tarilabs:
@lucamolteni as you can see also in the screenshot, I re-triggered the C-D-B

tarilabs:
as most CDB are unrelated also on other PRs I would just leave as-is
```
11:14:14  [ERROR] Failed to execute goal on project google-drive-workitem: Could not resolve dependencies for project org.jbpm.contrib:google-drive-workitem:jar:7.44.0-SNAPSHOT: Could not transfer artifact com.google.oauth-client:google-oauth-client-java6:jar:1.23.0 from/to mirror-central (https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kie-all/): /home/jenkins/.m2/repository/com/google/oauth-client/google-oauth-client-java6/1.23.0/google-oauth-client-java6-1.23.0.jar.part (No such file or directory) -> [Help 1]
11:14:14  org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project google-drive-workitem: Could not resolve dependencies for project org.jbpm.contrib:google-drive-workitem:jar:7.44.0-SNAPSHOT: Could not transfer artifact com.google.oauth-client:google-oauth-client-java6:jar:1.23.0 from/to mirror-central (https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kie-all/): /home/jenkins/.m2/repository/com/google/oauth-client/google-oauth-client-java6/1.23.0/google-oauth-client-java6-1.23.0.jar.part (No such file or directory)
```

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/typesafe/AbstractDMNSetType.java
tkobayas: @tarilabs 

`DefinedKeySetField` is actually used now. When we start strongly typed DMN, we instantiate `DMNContextFPAImpl` with an `OutputSet` instance. Then, `DMNContextFPAImpl` is converted to `DMNContextImpl` for evaluation. At this point, we call `fpa.allFEELProperties()` inside `DMNContextImpl.clone()` to convert the FPA fields to a Map. Here, we need to distinguish between "undefined" (= don't put in the Map) and "explicit null" (= put the key with null value). So `DefinedKeySetField` mechanism is required.

If we use `InputSet` for the first `DMNContextFPAImpl`, we don't need this mechanism because `InputSet` doesn't have Decision node fields so we can always assume "null" is "explicit null". So... if we decide back to the approach "Instantiate `DMNContextFPAImpl` with an `InputSet` instance first. Then convert the result to `OutputSet`", I can remove the `DefinedKeySetField`. WDYT?

tarilabs: oh I see now, thanks @tkobayas for the explanation, makes sense.

I would say please leave this as-is.
This can be helpful for supporting the use case when for instance on Kogito, we want to supply pre-evaluated-decision via kogito scaffolding.

Given inputset and outputset are infrastructural classes and emerging by the design of the DMN model itself, I believe it is good we have this utility method to distinguish the cases, as you pointed out. I just wanted to be sure if it was actually used, and now I see how :) 

Thanks!

mariofusco (approved):

PR #3037, opened 2020-08-12
Add missing .gitignore in child Maven modules kie-dmn-pmml-tests-parent

tarilabs:
our `drools.git` repo setup requires to copy the necessary `.gitignore` files in each Maven module.

In this case the file was copied from `/kie-dmn`

### DIFF
--- a/kie-dmn/kie-dmn-pmml-tests-parent/.gitignore
+++ b/kie-dmn/kie-dmn-pmml-tests-parent/.gitignore
@@ -1,0 +1,11 @@
+/target
+/local
+/bin
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
+!.gitignore
+/nbproject
+/*.ipr
+/*.iws
+/*.iml
--- a/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-legacy/.gitignore
+++ b/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-legacy/.gitignore
@@ -1,0 +1,11 @@
+/target
+/local
+/bin
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
+!.gitignore
+/nbproject
+/*.ipr
+/*.iws
+/*.iml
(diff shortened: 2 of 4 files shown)


hellowdan (approved):

tarilabs:
can a @kiegroup/gatekeepers kindly merge this with squash, please?

mareknovotny:
just curious why the top level .gitignore at  kie-dmn/kie-dmn-pmml-tests-parent doesn't work? from quick view they all are the same, right? @tarilabs

tarilabs:
> they all are the same, right?

they are @mareknovotny but the instruction `/.*` works in git only for the sibling of the `.gitignore` file, hence as per setup in `drools.git` repo, I always found you must cascade it down in every module.

I believe this is not experienced by IDEA users, since IDEA allegedly does NOT generate the `.classpath` files for instance, but if you use VSCode or Eclipse you experience this

Please notice it is a GIT behavior, not related to IDEs.

Hope this clarifies

mareknovotny:
I am Eclipse and VSCode user and the pattern used in `.gitignore` is  which is not general one.
in case of here added files, starting with slash is applied only to directory on that level and if you remove it it will be applied for all subdirectories, therefore i asked why duplicating all same .gitignore files

https://git-scm.com/docs/gitignore says
```
- The slash / is used as the directory separator. Separators may occur at the beginning, middle or end of the .gitignore search pattern.
- If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.
```

I don't mind too much if you would like to duplicate unnecessary these files, but my curiosity was raised as I am always for less files with more general pattern in `.gitignore` files means less maintenance.

kie-dmn/kie-dmn-pmml-tests-parent/.gitignore
```diff
@@ -0,0 +1,11 @@
+/target
+/local
+/bin
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
```
mareknovotny, line 6: if you omit starting slash `.*` will be used for all directory levels of dot files and other .gitignore files won't be needed.
The same is for 
/*.ipr
/*.iws
/*.iml

tarilabs: Yes I very well know this, but since the mandate was to use a single format for `.gitignore` this is the situation

PR #3081, opened 2020-09-07
Unexpected rule firing

triceo:
I think that the difference from the other tests is that:

- `exists` checks for other fact type (`Child`) than the first pattern (`Parent`).
- `groupBy` uses  the other fact type (`Child`) as key (`Parent::getChild`).
- and I then remove one instance of the `Child` fact.

Therefore the `exists` should remove the match altogether.
I have not seen a test like that in there.

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GroupByTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GroupByTest.java
@@ -165,6 +165,7 @@ public class GroupByTest {
         final Variable<Person> var_$p = DSL.declarationOf(Person.class);
         final Variable<Integer> var_$age = DSL.declarationOf(Integer.class);
         final Variable<Integer> var_$sumOfAges = DSL.declarationOf(Integer.class);
+        final Variable<Integer> var_$derivedFromKey = DSL.declarationOf(Integer.class);
 
         final Rule rule1 = PatternDSL.rule("R1").build(
                 DSL.groupBy(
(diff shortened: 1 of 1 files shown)


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GroupByTest.java
```diff
@@ -447,4 +454,77 @@ public void testGroupBy2Vars() throws Exception {
         assertEquals( 40, results.get("G5") );
         assertEquals( 119, results.get("M5") );
     }
+
+    @Test
+    public void testUnexpectedRuleMatch() {
```
triceo, line 459: This is where the test starts.
The other changes is just my IDEA "cleaning up" the code.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GroupByTest.java
```diff
@@ -152,40 +159,43 @@ public void testSumPersonAgeGroupByInitialWithAcc() throws Exception {
...
                         // Grouping Function
                         var_$p, var_$key, person -> person.getName().substring(0, 1),
                         // Accumulate Result (can be more than one)
-                        D.accFunction(org.drools.core.base.accumulators.IntegerSumAccumulateFunction::new, var_$age).as(var_$sumOfAges)),
+                        DSL.accFunction(IntegerSumAccumulateFunction::new, var_$age).as(var_$sumOfAges)),
                 // Filter
-                D.pattern(var_$sumOfAges).expr($sumOfAges -> EvaluationUtil.greaterThanNumbers($sumOfAges, 10)),
+                PatternDSL.pattern(var_$sumOfAges)
+                        .expr($sumOfAges -> EvaluationUtil.greaterThanNumbers($sumOfAges, 10))
+                        .bind(var_$derivedFromKey, var_$key, k -> 1),
```
triceo, line 180: @mariofusco I modified this test to show another problem I just uncovered. Binding a var on the final pattern and then using it on the conseqence will throw an NPE.

Let me know if you prefer a different test for this.

triceo: To explain why I need this: 

Sometimes a Constraint Stream has two key mapping functions. I solve that with the current API as follows:

    groupBy(..., inVar, groupKeyVar, in -> {
        Object key = keyMapper.apply(in);
        Object key2 = keyMapper2.apply(in);
        return new CompositeKey(key, key2);
    }, ...)

However, that means that at the end, I have to have a way of extracting `key1` and `key2` from the `groupKeyVar`. Which is what I use the final binding for.

mariofusco:
incorporated and fixed by https://github.com/kiegroup/drools/pull/3085

PR #3087, opened 2020-09-09
NPE with from() after exists()

triceo:


triceo:
Fixed with 26e24005d78b9752fd5e8928efd24d3cb79658e0.

PR #3088, opened 2020-09-10
[DROOLS-5486] CEP doesn't evaluate correctly when a bind variable is …

tkobayas:
…used as the first temporal parameter in executable model

**JIRA**: https://issues.redhat.com/browse/DROOLS-5486

<pre>
How to retest a PR or trigger a specific build:

</pre>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/TemporalConstraintEvaluator.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/TemporalConstraintEvaluator.java
@@ -51,7 +51,12 @@ public class TemporalConstraintEvaluator extends ConstraintEvaluator {
         long start2 = getStartTimestamp( fhs[1], workingMemory, getDeclarations()[1], temporalConstraint.getF2() );
         long duration2 = getDuration( fhs[1] );
         long end2 = start2 + duration2;
-        return temporalConstraint.getTemporalPredicate().evaluate( start1, duration1, end1, start2, duration2, end2);
+        TemporalPredicate temporalPredicate = temporalConstraint.getTemporalPredicate();
+        if (temporalPredicate.isThisOnRight()) {
+            return temporalPredicate.evaluate( start2, duration2, end2, start1, duration1, end1);
+        } else {
+            return temporalPredicate.evaluate( start1, duration1, end1, start2, duration2, end2);
+        }
     }
 
     private long getDuration( InternalFactHandle fh ) {
@@ -76,7 +81,12 @@ public class TemporalConstraintEvaluator extends ConstraintEvaluator {
         long start2 = getNonEventTimestamp(temporalConstraint, handle, workingMemory);
         long duration2 = 0;
         long end2 = start2 + duration2;
-        return temporalConstraint.getTemporalPredicate().evaluate( start1, duration1, end1, start2, duration2, end2);
+        TemporalPredicate temporalPredicate = temporalConstraint.getTemporalPredicate();
+        if (temporalPredicate.isThisOnRight()) {
+            return temporalPredicate.evaluate( start2, duration2, end2, start1, duration1, end1);
+        } else {
+            return temporalPredicate.evaluate( start1, duration1, end1, start2, duration2, end2);
+        }
     }
 
     private long getNonEventTimestamp(TemporalConstraint temporalConstraint, InternalFactHandle handle, InternalWorkingMemory workingMemory) {
--- a/drools-model/drools-canonical-model/src/main/java/org/drools/model/PatternDSL.java
+++ b/drools-model/drools-canonical-model/src/main/java/org/drools/model/PatternDSL.java
@@ -204,14 +204,14 @@ public class PatternDSL extends DSL {
                                                                 Variable<G> var8, Variable<H> var9, Variable<I> var10, Variable<J> var11, Variable<K> var12, Variable<L> var13,
                                                                 Predicate13<T, A, B, C, D, E, F, G, H, I, J, K, L> predicate, ReactOn reactOn );
 
-        PatternDef<T> expr( String exprId, Variable<T> thisVar, long value, TemporalPredicate temporalPredicate );
-        PatternDef<T> expr( String exprId, Variable<T> thisVar, Function1<T, ?> f, long value, TemporalPredicate temporalPredicate );
+        <U> PatternDef<T> expr( String exprId, Variable<U> var1, TemporalPredicate temporalPredicate );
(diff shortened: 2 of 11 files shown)


tkobayas:
@mariofusco  @lucamolteni 

Could you review this to check if I'm on the right track?

The main problem is that the current Cep related PatternDSL expr() methods assume that "this" is always on the left side of the temporal function (e.g. this after[1s,2s] $a). So tests failed if I put "this" on the right side.

So I changed the expr() signatures to accept 2 Variables. Also I changed PatternExpressionBuilder/FlowExpressionBuilder.buildExpression() to always add "this" argument (= actually, its bound variable name) to the DSL. So now we can have "this" on left or right side.

Any suggestions/opinions would be welcome. Thanks!

lucamolteni (approved):

mariofusco:
@tkobayas I'm pretty sure that I'm missing something, but I don't understand why you need to add the `thisVar` to the methods' signature and more important how this is helping you to fix the problem. The `thisVar` is always already defined when you create the `PatternDef`, so having to specify it again in the expr definition seems redundant (and this is why we didn't need to do this so far). 

You mentioned that you're doing it for the situations when `this` is used on the right. Can you please clarify why we don't have this problem for plain predicates but only for temporal ones? I'd prefer to avoid having to add the `thisVar` only for temporal expressions even because this will create an asymmetry with other exprs.

tkobayas:
Hi @mariofusco 

For plain predicates, the lambda predicate (e.g. `Predicate2$Impl`) contains the whole statement including the arguments order.

```
D.expr("8ADC3A9D4F73F1246AEBB7775E290BDB", var_$olderV, var_$markV, (org.drools.modelcompiler.domain.Person _this, org.drools.modelcompiler.domain.Person $markV) -> org.drools.modelcompiler.util.EvaluationUtil.greaterThanNumbers(_this.getAge(), $markV.getAge()))
```

On the other hand, `TemporalPredicate` (e.g. `AfterPredicate`) has only its operator (e.g. `after[5000, 8000]`). So `TemporalConstraintEvaluator` manages the order of operands:

https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/TemporalConstraintEvaluator.java#L45-L55

The order of operands comes from `VariableTemporalConstraint.getVariables()`. So we need to be able to control the order in `PatternDSL$TemporalPatternExpr`.

For example, when we create a `VariableTemporalConstraint` for `$a after[5s,8s] this`, the current `expr()` signature doesn't have a clue whether `this` is on left or on right. (-> so always considers that `this` is on left)

https://github.com/kiegroup/drools/blob/master/drools-model/drools-canonical-model/src/main/java/org/drools/model/PatternDSL.java#L508-L511

The reason why I introduced 2 Variables in the `expr()` signature is to clarify whether `this` exists on left or right. I agree that we can get `this` by `patternDef.getFirstVariable()` so it may look redundant.

Other possible approach would be to add a flag to indicate whether `this` is on left or right:

```
        public <U> PatternDef<T> expr( String exprId, Variable<U> var1, TemporalPredicate temporalPredicate, boolean isThisOnLeft) {
            items.add( new TemporalPatternExpr<>( exprId, null, var1, null, temporalPredicate, isThisOnLeft) );
            return this;
        }
```

Based on the flag, we can created an expected `VariableTemporalConstraint` in `TemporalPatternExpr.asConstraint()`. This approach may be less intrusive. WDYT?

Btw, below 2 methods are exceptional cases because they use `FixedTemporalPatternExpr`. I just ignore `thisVar` here and just use the current implementation at the moment. In order to make "this on right" work, it would require some more changes. 

```
        @Override
        public PatternDef<T> expr( String exprId, Variable<T> thisVar, long value, TemporalPredicate temporalPredicate ) {
            items.add( new FixedTemporalPatternExpr<>( exprId, null, value, temporalPredicate) );
            return this;
        }

        @Override
        public PatternDef<T> expr( String exprId, Variable<T> thisVar, Function1<T, ?> f, long value, TemporalPredicate temporalPredicate ) {
            items.add( new FixedTemporalPatternExpr<>( exprId, new Function1.Impl<>( f ), value, temporalPredicate) );
            return this;
        }
```

mariofusco:
Hi @tkobayas and thanks for your explanation. Now I see what you mean. So to recap the problem is in the fact that with plain predicate you have specified if the `this` is on the left or on the right implicitly, given how the lambda is written, but for temporal predicate you don't have this information. The fact that `this` is on the left or on the right is in indeed a property of the predicate (regardless if it is temporal or not) and I'd leave it there.

In other words I'd leave the signature of `expr` methods for temporal expressions exactly as they are now and I'd add a `boolean thisOnRight` on the `AbstractTemporalPredicate` class similar to the already existing `negated` one. The only difference from the current `negated` impl is that I would allow to set this flag also in the `TemporalPredicate` interface.

In this way the constraint `$a after this` could be written with the expression `expr(var_$a, after().thisOnRight())`. Doing so will also have the twofold advantage of not polluting the dsl when `this` is on the left (which is by far the most common case) and keeping all existing generated executable models compatible with this new feature.

Please let me know if you agree with this solution or can find any drawback with it.

tkobayas:
@mariofusco  Thank you very much for the suggestion. It looks much better. Updated the PR.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/TemporalConstraintEvaluator.java
```diff
@@ -51,7 +51,12 @@ public boolean evaluate( InternalFactHandle handle, Tuple tuple, InternalWorking
         long start2 = getStartTimestamp( fhs[1], workingMemory, getDeclarations()[1], temporalConstraint.getF2() );
         long duration2 = getDuration( fhs[1] );
         long end2 = start2 + duration2;
-        return temporalConstraint.getTemporalPredicate().evaluate( start1, duration1, end1, start2, duration2, end2);
+        TemporalPredicate temporalPredicate = temporalConstraint.getTemporalPredicate();
+        if (temporalPredicate.isThisOnRight()) {
+            return temporalPredicate.evaluate( start2, duration2, end2, start1, duration1, end1);
+        } else {
+            return temporalPredicate.evaluate( start1, duration1, end1, start2, duration2, end2);
+        }
```
tkobayas, line 59: I inverse the arguments here rather than doing it inside each temporalPredicate `evaluate` method because this approach is concise and less error-prone, I think.

tkobayas:
@mariofusco @hellowdan Could you kindly review again? Thanks!

hellowdan (approved):

mariofusco (approved):

PR #3093, opened 2020-09-11
[DROOLS-5635] fix executable model when using an existental pattern i…

mariofusco:


triceo (approved):
Confirming that this fixes my problem.

PR #3094, opened 2020-09-12
NPE when joining three patterns

triceo:


triceo:
Mario and I have come to the conclusion that this is something not currently supported by the engine.
I'll work to give Mario some benchmarks to toy with, so that maybe one day we could make this possible.

PR #3120, opened 2020-09-24
[DROOLS-5668] ignore no longer needed node memories serialized by old…

mariofusco:


slem1:
Hi Mario, I wrote it yesterday and finally deleted it (as I thought It must be irrelevant for you but may it finally has some usefulness...), but as I tried the same approach as you (report the filling of nodemeory for (RIA/ACCUMULATE/FROM) on deserialization which was previously delete). But what I found is that now as activationkey for dormant activation on "has memory node" use tuple objects which where previously not serialized, the following piece of code in accept method

```
     RuleImpl rule = activation.getRule();
                ActivationKey activationKey = PersisterHelper.hasNodeMemory( rtn ) ?
                        PersisterHelper.createActivationKey( rule.getPackageName(), rule.getName(), activation.getTuple().toObjects(true)) :
                        PersisterHelper.createActivationKey( rule.getPackageName(), rule.getName(), activation.getTuple() );

                this.tuplesCache.put( activationKey, activation.getTuple() );

                return !dormantActivations.contains(activationKey);
```
could not evict dormantActivations, thus leading to retrigger already fired rules in old-version serialized kieSession. Must admit it was a dead end for my current understanding of drools internal mechanics.

mariofusco:
@slem1 You're right, my last commit should address the problem and process correctly dormant activation in case of old style marshalling with serialized node memories. I tested it modifying your reproducer and it works for me. Please give a look if possible.

PR #3126, opened 2020-09-28
[DROOLS-5677] make property reactivity to recognize multiple properti…

mariofusco:


drools-core/src/main/java/org/drools/core/rule/constraint/MvelConstraint.java
```diff
@@ -455,38 +464,73 @@ private String findBoundVariable(String variable) {
...
+        }
+        return skipOperator( expression, cursor );
+    }
+
+    private int skipOperator( String expression, int cursor ) {
+        if (cursor < expression.length() && expression.charAt(cursor) == '.') {
+            while (cursor < expression.length() && Character.isWhitespace(expression.charAt(++cursor)));
+        }
+
+        boolean namedOperator = false;
```
lucamolteni, line 519: Why the manual parsing here? Is it necessary? 

I’m pretty sure we removed all the occurrences of manual parsing

lucamolteni (changes requested):

PR #3136, opened 2020-10-02
[DROOLS-5687] propagate old kbuilders to updated modules during incre…

mariofusco:


mariofusco:
jenkins execute cdb

mariofusco:
jenkins test

PR #3140, opened 2020-10-05
[DROOLS-5662] creation of Drools wrapper modules

mariofusco:


drools-engine-classic/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    </dependency>
+   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+           <skip>true</skip>
```
danielezonca: Why is enforcer disabled?

drools-engine/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+            <skip>true</skip>
```
danielezonca: Same as above

drools-engine-classic/pom.xml
mariofusco: @danielezonca I'm not sure if that is correct indeed. I took it from kie-ci-osgi which is also a pure wrapper module, but I guess that there for necessary because it does shading and here this is not the case. I'll remove it.

danielezonca (approved):
LGTM

PR #3144, opened 2020-10-06
[RHDM-1461] allow to define an AgendaFilter when starting a process

mariofusco:


elguardian:
depends on
https://github.com/kiegroup/droolsjbpm-knowledge/pull/470

elguardian (approved):

PR #3149, opened 2020-10-08
[KOGITO-3505] avoid to uselessly generate the rete network during a p…

mariofusco:


drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -900,22 +900,32 @@ private boolean isSwappable(Resource original,
...
     @Override
     public void addPackage(final PackageDescr packageDescr) {
-        PackageRegistry pkgRegistry = getOrCreatePackageRegistry(packageDescr);
+        if ( addPackageWithoutRete( packageDescr ) ) {
+            compileRete(packageDescr);
+        }
+    }
+
+    /**
+     * Ads a package from a Descr/AST also triggering its compilation
```
gitgabrio: @mariofusco 
Typo ? "Ads"

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -900,22 +900,32 @@ private boolean isSwappable(Resource original,
     }
 
     /**
-     * This adds a package from a Descr/AST This will also trigger a compile, if
-     * there are any generated classes to compile of course.
+     * Ads a package from a Descr/AST also triggering its compilation
```
gitgabrio: @mariofusco 
Typo ? "Ads"

gitgabrio (approved):
@mariofusco 
LGTM. Just two minor comments
1) could you add ac ouple of tests to verify/enforce the different behavior ?
2) would it be possible, in another PR maybe, to change a little bit the behavior when "getOrCreatePackageRegistry" fails, to throw an exception, so that calling code may be notified of that ?

PR #3170, opened 2020-10-15
DROOLS-5733 prevent removing trailing zeros for numbers in scientific…

TommyNeubert:
… notation

**JIRA**:

[link](https://issues.redhat.com/browse/DROOLS-5733)

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
@@ -711,6 +711,9 @@ public class EvalHelper {
     }
 
     private static String removeTrailingZeros(final String stringNumber) {
+        if(stringNumber.contains("E")) {
+            return stringNumber;
+        }
         final String stringWithoutZeros = stringNumber.replaceAll("0*$", "");
         if (Character.isDigit(stringWithoutZeros.charAt(stringWithoutZeros.length() - 1))) {
             return stringWithoutZeros;
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/util/EvalHelperTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/util/EvalHelperTest.java
@@ -16,10 +16,13 @@
 
 package org.kie.dmn.feel.util;
 
+import java.math.BigDecimal;
+
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.kie.dmn.feel.util.EvalHelper.getBigDecimalOrNull;
 import static org.kie.dmn.feel.util.EvalHelper.normalizeVariableName;
 
 public class EvalHelperTest {
@@ -49,4 +52,12 @@ public class EvalHelperTest {
         assertEquals("ab c", normalizeVariableName("ab c  "));
         assertEquals("a b", normalizeVariableName("a\u00A0b"));
     }
+    
+    @Test
+    public void testGetBigDecimalOrNull() {
+        assertEquals(new BigDecimal("10"), getBigDecimalOrNull(10d));
+        assertEquals(new BigDecimal("10"), getBigDecimalOrNull(10.00000000D));
+        assertEquals(new BigDecimal("10000000000.5"), getBigDecimalOrNull(10000000000.5D));
+    }
+    
 }


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
```diff
@@ -711,6 +711,9 @@ private static Boolean isEqual(Object l, Object r) {
     }
 
     private static String removeTrailingZeros(final String stringNumber) {
+        if(stringNumber.contains("E")) {
+            return stringNumber;
+        }
```
j-beyer, line 716: @tarilabs seems I was a bit too quick, we found another one :sweat_smile:

There's probably a number of other solutions, such as using
```java
String.format("%f", value)
```
instead of 
```java
value.toString()
```
If you'd prefer a different solution, let us know.

Overall, it's probably not a huge issue - normally, when dealing with large numbers with fractions, one should provide a BigDecimal as input value, instead of a float/double, anyways.
We just wanted to inform you that this bug exists. We're also fine with you closing this PR if you don't consider it imporant :slightly_smiling_face:

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/util/EvalHelperTest.java
```diff
@@ -49,4 +52,12 @@ public void testNormalizeSpace() {
         assertEquals("ab c", normalizeVariableName("ab c  "));
         assertEquals("a b", normalizeVariableName("a\u00A0b"));
     }
+    
+    @Test
+    public void testGetBigDecimalOrNull() {
+        assertEquals(new BigDecimal("10"), getBigDecimalOrNull(10d));
+        assertEquals(new BigDecimal("10"), getBigDecimalOrNull(10.00000000D));
+        assertEquals(new BigDecimal("10000000000.5"), getBigDecimalOrNull(10000000000.5D));
```
j-beyer, line 60: Note: this assertion would break without the change.
```java
value.toString()
```
will produce `1.00000000005E10`, and `removeTrailingZeros()` would change this to `1.00000000005E1`

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
tarilabs: > when dealing with large numbers with fractions, one should provide a BigDecimal as input value, instead of a float/double, anyways.

exactly, the recommendation is to supply a BigDecimal, as part of input DMNContext.

However, thank-you for mentioning this, and I concur if there is a way to improve the behaviour also when (mis)using the doubles anyway, why not!

Allow me some time on this one since I recall there was specific reasons we opted for the double->string->BigDecimal in the first place, so I just want to double check with that in mind.

Many thanks again for letting us know!

tarilabs:
ok to test

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
TommyNeubert: In my opinion the easiest way of doing the double -> string -> BigDecimal conversion is to use:
```
BigDecimal result = new BigDecimal(value.toString()).stripTrailingZeros();
```
and remove the self-written method:
```
private static String removeTrailingZeros(final String stringNumber) {
   ...
}
```

tarilabs: Unfortunately the solution of the comment
would be less ideal, since a call to
```
getBigDecimalOrNull(10.00000000D)
```

would cause to return a
```
new BigDecimal("1E+1")
```

rather then the most naturally expected:
```
new BigDecimal("10")
```

So far, the current PR proposal is the most preferable, but I am always open to discuss additional options! :)

tarilabs (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
j-beyer: If I'm not mistaken, the current PR proposal would show the same behavior that you described.
```
getBigDecimalOrNull(10.00000000D)
```
would also just internally convert the double to `"1E+1"` using `toString()`, and would pass that to the BigDecimal constructor as well. We just wouldn't strip any trailing zeroes after the `E` anymore.

Maybe the above metioned 
```
String.format("%f", value)
```
would change this. We haven't checked the details of that yet, though.

hellowdan (approved):

mariofusco (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
tarilabs: the current PR proposal works fine for me, let's merge this as-is and if later there are suggestion for further improvement we can always revise it. For now the most important part was to avoid the bug when the `toString` results in something ending with `...E10` or similar exponent ending with zero 👍

PR #3172, opened 2020-10-19
DROOLS-5740 Fix CMEs when building KieBase

triceo:
**JIRA**: [DROOLS-5740](https://issues.redhat.com/browse/DROOLS-5740)

**Referenced Pull Requests**: None.

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
@@ -31,6 +31,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -116,7 +117,7 @@
     };
 
     private final InternalKieModule internalKieModule;
-    private final Map<String, CanonicalKiePackages> pkgsInKbase = new HashMap<>();
+    private final ConcurrentMap<String, CanonicalKiePackages> pkgsInKbase = new ConcurrentHashMap<>();
```
mariofusco, line 120: @triceo I understand that you need the other Map to be a concurrent one since it is static and than the same for all threads, but if you're building different projects in parallel you should have different instances of the CanonicalKieModule and then of this Map. Am I missing something? Did you experience concurrency problems also here?

triceo: Yes, I have, see the JIRA for stack trace.

mariofusco:
jenkins execute cdb

mariofusco:
jenkins test

mariofusco (approved):

PR #3187, opened 2020-10-26
[RHDM-1468] allow to set the min inflate ratio for decision table exc…

mariofusco:


Rikkola (approved):
In my this is probably the best option. Everything else gets too complicated. For Business Central we have an UI for setting this up.

PR #3189, opened 2020-10-26
DROOLS-5760 Use named tokens in TokenTypes

evacchi:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5760

Minor annoyance: the generated code of the MVEL parser moves numeric IDs around when a token is added to the grammar. Manually written code in TypeTokens break because some of these integer values are referred directly, because their corresponding token is unnamed.
This tiny PR should fix the issue by assigning a name to these anonymous tokens.


lucamolteni (approved):

evacchi:
jenkins run cdb

mariofusco (approved):

PR #3198, opened 2020-10-29
[DROOLS-5697] example of 2 combined groupBy

mariofusco:


mariofusco:
Replaced by GroupBy node implementation in https://github.com/kiegroup/drools/pull/3298

PR #3208, opened 2020-11-03
[DROOLS-5785] when a condition cell value in a decision table is termi…

mariofusco:


manstis (approved):
If I was being really pedantic I think `value == "$param"?` would break - you'd get `name == "x"` generated. Ideally you might want to change `SnippetBuilder` to be aware of a _template_ that ends with `?` and change it's `build(..)` method to skip value construction.... but, as it stands, this PR seems a reasonable quick fix. I guess it should also be documented?

Rikkola (approved):

tarilabs (approved):

PR #3212, opened 2020-11-04
DROOLS-5786 Reproducer

triceo:
This reproducer:

- Is in the wrong place.
- Can possibly be further simplified once Luca or Toshiya understand precisely what's going on here.


triceo:
Superseded by https://github.com/kiegroup/drools/pull/3212.

PR #3214, opened 2020-11-05
[DROOLS-5786] NPE with a double groupBy

tkobayas:
**JIRA**:

https://issues.redhat.com/browse/DROOLS-5786

**referenced Pull Requests**:

The test is from @tricero 's https://github.com/kiegroup/drools/pull/3212

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
@@ -115,11 +115,7 @@ public class SegmentPropagator {
             AccumulateMemory am = ((AccumulateMemory)wm.getNodeMemory(accNode));
             BetaMemory bm = am.getBetaMemory();
             TupleMemory ltm = bm.getLeftTupleMemory();
-            if (ltm.contains(peer)) {
-                return false;
-            } else {
-                return true;
-            }
+            return !ltm.contains(peer);
         }
         return false;
     }
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
@@ -16,12 +16,17 @@
 package org.drools.core.phreak;
 
 import org.drools.core.common.InternalWorkingMemory;
+import org.drools.core.common.NetworkNode;
 import org.drools.core.common.TupleSets;
+import org.drools.core.reteoo.AccumulateNode;
+import org.drools.core.reteoo.AccumulateNode.AccumulateMemory;
+import org.drools.core.reteoo.BetaMemory;
 import org.drools.core.reteoo.LeftTuple;
 import org.drools.core.reteoo.LeftTupleSink;
 import org.drools.core.reteoo.LeftTupleSource;
 import org.drools.core.reteoo.PathMemory;
 import org.drools.core.reteoo.SegmentMemory;
+import org.drools.core.reteoo.TupleMemory;
 
 import static org.drools.core.phreak.AddRemoveRule.forceFlushLeftTuple;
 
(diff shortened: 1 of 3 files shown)


tkobayas:
The issue sequence:

0) Initial fireAllRules() : No MrProcessAssignment is propagated because all "isMoved() == false". Only InitialFactImpl is inserted to the first AccmulateNode. a FromNodeLeftTuple is created for accctx.resultLeftTuple.

https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java#L626

1) 1st switchMachinesInAssignments() : assignment1 and assignment2 are inserted as rightTuples to the first AccmulateNode. The accumulate result is Collection[assignment1,assignment2]. FromNodeLeftTuple (= accctx.resultLeftTuple) holds the result and is inserted to the trgLeftTuples.

   A peer of the FromNodeLeftTuple is created for the next segment (= rootNode is the second AccmulateNode).

   https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java#L80 

   The original FromNodeLeftTuple is inserted to FromNode. The peer FromNodeLeftTuple is inserted to the second AccmulateNode (as leftTuple) when inserted, contextObject is created and set to the peer FromNodeLeftTuple.

   https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java#L144

   Also facts are inserted as rightTuples (came from FromNode) in the second AccmulateNode. Finally the accumulate produces the result (Long 2).

2) 2nd switchMachinesInAssignments() : assignment1 and assignment2 are swapped. It means all assignments are "isMoved() == false". So 2 rightTuples are deleted in the first AccmulateNode. The accumulate result is an empty Collection. Because empty collection is not allowed by the rule, the original FromNodeLeftTuple (= accctx.resultLeftTuple) will be deleted (= trgLeftTuples.addDelete(childLeftTuple)) in FromNode.

   The peer FromNodeLeftTuple is also deleted in the second AccmulateNode. (designated by SegmentPropagator.processPeerDeletes()). PhreakAccumulateNode.doLeftDeletes() sets contextObject to null.

   https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java#L535

   No result (expected).

3) 3rd switchMachinesInAssignments() : assignment3 and assignment4 are inserted to as rightTuples the first AccmulateNode. The accumulate result is Collection[assignment3,assignment4]. FromNodeLeftTuple (= accctx.resultLeftTuple) holds the result and is inserted to the trgLeftTuples.

   The peer FromNodeLeftTuple is already exists. It was deleted and cleared so its stagedType is LeftTuple.NONE. In this case, the peer FromNodeLeftTuple will be **updated** (instead of inserted).

   https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakJoinNode.java#L466-L467

   During PhreakAccumulateNode.doLeftUpdates(), contextObject is not given to the peer FromNodeLeftTuple so it's still null.

   Finally, at PhreakAccumulateNode.evaluateResultConstraints(), failed with NPE (accctx = null).

   https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java#L606

drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
```diff
@@ -98,6 +108,22 @@ private static void processPeers(SegmentMemory sourceSegment, TupleSets<LeftTupl
...
+            BetaMemory bm = am.getBetaMemory();
+            TupleMemory ltm = bm.getLeftTupleMemory();
+            if (ltm.contains(peer)) {
+                return false;
+            } else {
+                return true;
+            }
+        }
+        return false;
+    }
```
tkobayas: My fix is to insert the peer LeftTuple only when:
- stagedType is LeftTuple.NONE
- The next node is AccumulateNode
- The tuple doesn't exist in leftTupleMemory (So we guess it was deleted)
If there is a better way to verify "it was deleted", it would be great.

tkobayas:
The other possible approach is to create and set a new AccumulateContext in doLeftUpdates() when it's null. (Ah, but probably updating a tuple which doesn't exist in leftTupleMemory is wrong anyway...)

lucamolteni (approved):

triceo (approved):
I have confirmed that this fixes the downstream issue.
As to the implementation choices, I can not say - I am not a domain expert here.

hellowdan (approved):

mariofusco:
> The other possible approach is to create and set a new AccumulateContext in doLeftUpdates() when it's null. (Ah, but probably updating a tuple which doesn't exist in leftTupleMemory is wrong anyway...)

@tkobayas as I discussed with @triceo I believe that the fix in this PR is the most formally correct, but I'm also afraid that it could have a performance impact (even if minor) especially for this linear search https://github.com/kiegroup/drools/pull/3214/files#diff-9352bb251c8442e0443f57eaa946ee36d1cc74b43e8096e1c81bbe8770e2c969R118 and moreover to cover something that can be considered an extreme edge case.

For this reason I decided to take the other approach that you suggested in the comment that I'm quoting. I don't think it will be a problem if the tuple is propagated as an update instead of an insert because it will normalized downstream. In reality I must admit that I'm not 100% sure of this but until we won't have a test that could prove this wrong I prefer to assume that it works and bring in the less performance impacting fix.

I'm closing this PR and replacing it with this one https://github.com/kiegroup/drools/pull/3221

PR #3221, opened 2020-11-05
[DROOLS-5786] fix NPE with a double accumulate

mariofusco:


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/AccumulateTest.java
```diff
@@ -2593,4 +2594,114 @@ public void testAccumulateNumberFromSum() {
...
+        session.update(rightHandle, right);
+        return session.fireAllRules();
+    }
+
+    @Test
+    public void testDoubleAccumulateNPE() {
+        // Prepare reproducing data.
+        MrMachine machine2 = new MrMachine();
+        MrMachine machine3 = new MrMachine();
+        MrProcessAssignment assignment1 = new MrProcessAssignment(new MrProcess(), machine3, machine3);
```
triceo, line 2615: Looking at the code, the entire `process` arg incl. the inner class can be removed.
It is not necessary in this latest version of the test.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/AccumulateTest.java
```diff
@@ -2593,4 +2594,114 @@ public void testAccumulateNumberFromSum() {
...
+        MrProcessAssignment assignment2 = new MrProcessAssignment(new MrProcess(), machine2, machine2);
+        MrProcessAssignment assignment3 = new MrProcessAssignment(new MrProcess(), machine2, machine2);
+        MrProcessAssignment assignment4 = new MrProcessAssignment(new MrProcess(), machine3, machine3);
+
+        String rule = "import " + MrProcessAssignment.class.getCanonicalName() + ";\n" +
+                "import " + List.class.getCanonicalName() + ";\n" +
+                "rule R1\n" +
+                "when\n" +
+                "   $assignments: List(size > 0) from accumulate(\n" +
+                "        $a: MrProcessAssignment(machine != null, this.isMoved() == true),\n" +
```
triceo, line 2625:
```suggestion
                "        $a: MrProcessAssignment(machine != null, machine != originalMachine),\n" +
```

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/AccumulateTest.java
```diff
@@ -2593,4 +2594,114 @@ public void testAccumulateNumberFromSum() {
...
+        // ************************************************************************
+        // Complex methods
+        // ************************************************************************
+
+        public boolean isMoved() {
+            if (machine == null) {
+                return false;
+            }
+            return !Objects.equals(originalMachine, machine);
+        }
```
triceo, line 2699: May be removed if the suggestion above is accepted.

triceo (approved):
The fix works. 
I have just noticed that the test could be simplified - use if you want, no need to.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/AccumulateTest.java
```diff
@@ -2593,4 +2594,114 @@ public void testAccumulateNumberFromSum() {
...
+        kieSession.dispose();
+    }
+
+    public static class MrProcess {
+
+    }
+
+    public static class MrProcessAssignment {
+
+        private MrProcess process;
```
triceo, line 2664: The field is not necessary in this latest version of the test.

PR #3222, opened 2020-11-06
[DROOLS-5799] Second accumulate produces a wrong result

tkobayas:
**JIRA**: 

https://issues.redhat.com/browse/DROOLS-5799

DROOLS-5786 solved the NPE but the second accumulate (in AccumulateTest.testDoubleAccumulateNPE()) produces a wrong result (= 0 instead of 2). 2 points to be fixed:

1. RuleNetworkEvaluator.doUpdatesReorderLeftMemory() causes wrong LeftTupleMemory size when calling TupleList.remove() for a tuple which doesn't exist in the list. (When removed, size becomes -1. Then added, size becomes 0. so the leftTuple is not evaluated in doRightInserts() : https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java#L215 )

2. Creating AccumulateContext right before evaluateResultConstraints() is too late. It is used by addMatch() so it needs to be created earlier (e.g. in doLeftUpdates())

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
@@ -255,7 +255,6 @@ public class PhreakAccumulateNode {
             AccumulateContext accctx = (AccumulateContext) leftTuple.getContextObject();
             if (accctx == null) {
                 accctx = initAccumulateContextOnLeftTuple( am, wm, accumulate, leftTuple );
-                leftTuple.setContextObject(accctx);
             }
 
             constraints.updateFromTuple(contextEntry,
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
@@ -104,9 +104,6 @@ public class PhreakAccumulateNode {
         for (LeftTuple leftTuple = tempLeftTuples.getUpdateFirst(); leftTuple != null; ) {
             LeftTuple next = leftTuple.getStagedNext();
             AccumulateContext accCtx = (AccumulateContext) leftTuple.getContextObject();
-            if (accCtx == null) {
-                accCtx = initAccumulateContextOnLeftTuple( am, wm, accumulate, leftTuple );
-            }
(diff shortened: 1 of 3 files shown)


drools-core/src/main/java/org/drools/core/phreak/RuleNetworkEvaluator.java
```diff
@@ -804,7 +804,9 @@ public static void doUpdatesReorderLeftMemory(BetaMemory bm,
 
         // sides must first be re-ordered, to ensure iteration integrity
         for (LeftTuple leftTuple = srcLeftTuples.getUpdateFirst(); leftTuple != null; leftTuple = leftTuple.getStagedNext()) {
-            ltm.remove(leftTuple);
+            if (leftTuple.getMemory() != null) {
```
tkobayas, line 807: `if (leftTuple.getMemory() == ltm) {`  may look more accurate but I followed the same approach as other methods in RuleNetworkEvaluator.

drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
```diff
@@ -255,7 +252,11 @@ public void doLeftUpdates(AccumulateNode accNode,
 
         for (LeftTuple leftTuple = srcLeftTuples.getUpdateFirst(); leftTuple != null; ) {
             LeftTuple next = leftTuple.getStagedNext();
-            final AccumulateContext accctx = (AccumulateContext) leftTuple.getContextObject();
+            AccumulateContext accctx = (AccumulateContext) leftTuple.getContextObject();
+            if (accctx == null) {
+                accctx = initAccumulateContextOnLeftTuple( am, wm, accumulate, leftTuple );
+                leftTuple.setContextObject(accctx);
```
mariofusco: Minor: the method `initAccumulateContextOnLeftTuple` already sets the `contextObject` on the `leftTuple`, I believe you can remove this statement.

tkobayas: Thanks, removed.

tkobayas:
Jenkins retest this

mariofusco (approved):

triceo (approved):
Verified.

PR #3245, opened 2020-11-18
[DROOLS-5833] executable-model doesn't provide line number of DRL par…

tkobayas:
…se error

**JIRA**:

https://issues.redhat.com/browse/DROOLS-5833

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
@@ -147,7 +147,7 @@ public abstract class AccumulateVisitor {
                 addNonExistingFunctionError(context, e.function);
                 return;
             } catch (AccumulateParsingFailedException e) {
-                context.addCompilationError(new InvalidExpressionErrorResult(e.getMessage()));
+                context.addCompilationError(new InvalidExpressionErrorResult(e.getMessage(), Optional.of(context.getRuleDescr())));
                 return;
             }
             processNewBinding(accumulateDSL);
@@ -209,7 +209,7 @@ public abstract class AccumulateVisitor {
         } else if (accumulateFunctionParameter instanceof LiteralExpr) {
             literalExprParameter(function, functionDSL, bindingId, accumulateFunctionParameter);
         } else {
-            context.addCompilationError(new InvalidExpressionErrorResult("Invalid expression " + accumulateFunctionParameterStr));
+            context.addCompilationError(new InvalidExpressionErrorResult("Invalid expression " + accumulateFunctionParameterStr, Optional.of(context.getRuleDescr())));
             throw new AccumulateParsingFailedException();
         }
     }
@@ -444,7 +444,7 @@ public abstract class AccumulateVisitor {
                             "Pattern of type: '[ClassObjectType class=%s]' " +
                                     "on rule '%s' " +
                                     "is not compatible with type %s returned by accumulate function."
-                            , expectedResultType.getCanonicalName(), context.getRuleName(), actualResultType.getCanonicalName())));
+                            , expectedResultType.getCanonicalName(), context.getRuleName(), actualResultType.getCanonicalName()), Optional.of(context.getRuleDescr())));
         }
     }
 
@@ -455,7 +455,7 @@ public abstract class AccumulateVisitor {
     }
 
     private void addNonExistingFunctionError(RuleContext context, AccumulateDescr.AccumulateFunctionCallDescr function) {
-        context.addCompilationError(new InvalidExpressionErrorResult(String.format("Unknown accumulate function: '%s' on rule '%s'.", function.getFunction(), context.getRuleDescr().getName())));
+        context.addCompilationError(new InvalidExpressionErrorResult(String.format("Unknown accumulate function: '%s' on rule '%s'.", function.getFunction(), context.getRuleDescr().getName()), Optional.of(context.getRuleDescr())));
     }
 
     private void addBindingAsDeclaration(RuleContext context, String bindingId, AccumulateFunction accumulateFunction) {
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
@@ -133,7 +133,13 @@ public abstract class PatternDSL implements DSLNode {
 
             ConstraintExpression constraintExpression = ConstraintExpression.createConstraintExpression(patternType, constraint, isPositional);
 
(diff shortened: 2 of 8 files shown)


tkobayas:
The basic idea is to convey Descr to Exception constructors just like standard_drl does. So the Exception can get the line number.

There are more places to apply this constructor but at the moment, I implemented it only to satisfy existing test cases in CompilationFailuresTest. I will file a JIRA to cover other cases.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -455,7 +455,7 @@ private boolean isCollectFunction(AccumulateFunction accumulateFunction) {
     }
 
     private void addNonExistingFunctionError(RuleContext context, AccumulateDescr.AccumulateFunctionCallDescr function) {
-        context.addCompilationError(new InvalidExpressionErrorResult(String.format("Unknown accumulate function: '%s' on rule '%s'.", function.getFunction(), context.getRuleDescr().getName())));
+        context.addCompilationError(new InvalidExpressionErrorResult(String.format("Unknown accumulate function: '%s' on rule '%s'.", function.getFunction(), context.getRuleDescr().getName()), Optional.of(context.getRuleDescr())));
```
tkobayas, line 458: Actually, we can provide other Descr than RuleDescr to give a better line number. But at the moment, I follow the same approach as standard_drl (so use line number from RuleDescr) in order to have the same assertion value.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -133,7 +133,13 @@ private void generatePatternIdentifierIfMissing() {
...
             ConstraintExpression constraintExpression = ConstraintExpression.createConstraintExpression(patternType, constraint, isPositional);
 
-            DrlxParseResult drlxParseResult = constraintParser.drlxParse(patternType, patternIdentifier, constraintExpression, isPositional);
+            DrlxParseResult drlxParseResult;
+            try {
+                context.setCurrentConstraintDescr(Optional.of(constraint));
+                drlxParseResult = constraintParser.drlxParse(patternType, patternIdentifier, constraintExpression, isPositional);
+            } finally {
+                context.resetCurrentConstraintDescr();
+            }
```
tkobayas, line 142: In order to convey the Descr which is being processed during drlxParse(), I enhanced RuleContext here.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -133,7 +133,13 @@ private void generatePatternIdentifierIfMissing() {
 
             ConstraintExpression constraintExpression = ConstraintExpression.createConstraintExpression(patternType, constraint, isPositional);
 
-            DrlxParseResult drlxParseResult = constraintParser.drlxParse(patternType, patternIdentifier, constraintExpression, isPositional);
+            DrlxParseResult drlxParseResult;
+            try {
+                context.setCurrentConstraintDescr(Optional.of(constraint));
```
lucamolteni, line 138: So the RuleContext.currentConstraintDescr is used as a global variable right? 

I'm not sure about this type, maybe we could use a `Map` with the `constraintId` as a string and the `Descr` as a value, WDYT? 

In that way we don't have way to reset the global.

P.S. I know there are already some global fields in the Context such as RuleDescr but I'd like to remove them eventually

lucamolteni (changes requested):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
tkobayas: Thank you for the suggestion, @lucamolteni 

Hmm, I can add a `Map<String constraintId, BaseDescr descr>` but at the place where an ErrorResult is created,

   e.g. https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expressiontyper/ExpressionTyper.java#L972

we don't know the `constraintId`. Do you mean that we will convey `constraintId` as a method parameter inside drlxParse() process? Sorry if I'm missing your point. (The reason why I introduced `currentConstraintDescr` as "current context" was to avoid changing method signatures. But of course, I'm open for discussion)

mariofusco (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
lucamolteni: Sometime we generate an id with the exprId (sorry for mixing up the terms, they're really similar)

https://github.com/kiegroup/drools/blob/277ef612becffbb8ea370e47d97d0cee8f1c8def/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java#L368

Take a look at it, anyway even in the current form it's fine by me

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
tkobayas: Thanks @lucamolteni  for pointing that.

IIUC, we can put the Descr with exprId (patternType+drlConstraintString hash) in a Map. Then we can get the Descr later with the exprId if we know the patternType and the drlConstraintString.

For example, in `CompilationFailuresTest.testUseNotExistingEnum()`, 

when we put the Descr at https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java#L134

constraintExpression.getExpression() is `valueType == Field.INT || == Field.DOUBLE`

but when we get the Descr at https://github.com/kiegroup/drools/blob/master/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expressiontyper/ExpressionTyper.java#L972 ,

n.toString() is `valueType == Field.INT || valueType == Field.DOUBLE`

So we don't have access to the original drlConstraintString at the point.

This may be resolved by some tweaks (e.g. enhance ExpressionTyper to hold original drlConstraintString) but if you are fine with the `currentConstraintDescr` approach, I hope I can keep it for this PR. Thanks.

lucamolteni: yes thanks for checking out @tkobayas let's keep the original variable

hellowdan (approved):

PR #3247, opened 2020-11-18
[DROOLS-5730] optimize executable model's constraint evaluator

mariofusco:


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/ConstraintEvaluator.java
```diff
@@ -226,4 +214,109 @@ public boolean isTemporal() {
...
+    static abstract class InnerEvaluator {
+
+        private final Declaration patternDeclaration;
+
+        protected InnerEvaluator( Declaration patternDeclaration ) {
+            this.patternDeclaration = patternDeclaration;
+        }
+
+        public abstract boolean evaluate( InternalFactHandle handle, InternalWorkingMemory workingMemory ) throws Exception;
+        public abstract boolean evaluate(InternalFactHandle handle, Tuple tuple, InternalWorkingMemory workingMemory) throws Exception;
```
triceo, line 231: @mariofusco These two lines (among others) are a great example that maybe Drools would benefit from a consistent code style.
In OptaPlanner, we've employed this plugin, which enforces a consistent (and configurable) code style during the build, just like Quarkus does:
https://code.revelc.net/formatter-maven-plugin/

Maybe you could consider that too? (Just a suggestion.)

mariofusco: @triceo this is a good suggestion, indeed. I just don't want that a formatter plugin will make the whole build to fail only for of a formatting inconsistency like this, also because this will increase the barrier for eventual external contributors. Is this the case? Or maybe is the plugin configurable?

triceo: I believe you can configure the plugin to not fail when it has to reformat code, but at the same time, I believe it defeats the purpose, as unformatted code can still hit the repo.

See more on Quarkus Github:
https://github.com/quarkusio/quarkus/blob/master/CONTRIBUTING.md#ide-config-and-code-style

PR #3249, opened 2020-11-19
[JBPM-9437] Updated data object is not take into account when the pro…

elguardian:
jira : https://issues.redhat.com/browse/JBPM-9437


elguardian:
Jenkins retest this

PR #3259, opened 2020-11-23
[DROOLS-5849] Fix jdk11 jaxb configuration for kie-dmn-pmml-tests-trusty

danielezonca:
Fix enforcer error when building module `kie-dmn-pmml-tests-trusty` with jdk-11

NOTE: It is an integration test so I can change the scope of `org.glassfish.jaxb:jaxb-runtime` to `test` if needed

**JIRA**: https://issues.redhat.com/browse/DROOLS-5849

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
+++ b/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
@@ -70,7 +70,7 @@
 
   <profiles>
     <profile>
-      <id>java11-pmml</id>
+      <id>jdk11</id>
       <activation>
         <jdk>[11,)</jdk>
       </activation>
--- a/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
+++ b/kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
@@ -55,28 +55,6 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
-
-    <!-- prepare for JDK11 -->
-    <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-xjc</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-core</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-impl</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.jboss.spec.javax.xml.bind</groupId>
-      <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
 
   <dependencyManagement>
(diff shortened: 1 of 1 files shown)


tarilabs:
```
Test Result (1 failure / +1)
org.drools.mvel.integrationtests.concurrency.BasicConcurrentInsertionsTest.testConcurrentInsertionsNewSessionEachThreadUpdateFacts
```

🤷‍♂️ 🎱 

Jenkins retest this

kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
```diff
@@ -89,4 +67,19 @@
...
+  <profiles>
+    <profile>
+      <id>java11-pmml</id>
+      <activation>
+        <jdk>[11,)</jdk>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>org.glassfish.jaxb</groupId>
+          <artifactId>jaxb-runtime</artifactId>
```
tarilabs, line 80: I'm not sure I understand this fix, shouldn't the usual 4 dependencies be included if JDK detected is a >=11 ?

danielezonca: The problem is that both `com.sun.xml.bind:jaxb-impl` and `org.glassfish.jaxb:jaxb-runtime` provide jaxb interfaces and impl needed in jdk11 so the enforcer fails because of duplicated classes. `kie-pmml-trusty` already brings in the glassfish one so the possible fixes are:
- exclude transitive dependency from `kie-pmml-trusty` and keep here `jaxb-impl` etc: not sure it if works because as far as I remember `kie-pmml-trusty` works only with glassfish impl (probably because of `jpmml-model` but not sure)
- remove everything from this module and leverage the transitive dependency from `kie-pmml-trusty`: I prefer to declare external transitive dependencies
- change glassfish dependency from `kie-pmml-trusty` to be not transitive (i.e. `provided`): this will require many more changes because ~all `kie-pmml-trusty` test modules need to be changed so I prefer to postpone a similar change to a future ticket
- declare here the same `glassfish` dependency to avoid the duplication that is what I did

I also make this dependency active only if jdk>=11

danielezonca:
jenkins retest this please

kie-dmn/kie-dmn-pmml-tests-parent/kie-dmn-pmml-tests-trusty/pom.xml
```diff
@@ -89,4 +67,19 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
+
+  <profiles>
+    <profile>
+      <id>java11-pmml</id>
```
mareknovotny: i think it would be good to have the same profile name/id for jdk11 instead of java11_pmml

danielezonca: Done

mareknovotny:
jenkins do cdb

danielezonca:
jenkins do cdb

danielezonca:
jenkins do cdb

danielezonca:
@mareknovotny 
CDB is breaking on `appformer-js` builds. I think it is safe to proceed
`[2020-12-02T13:19:12.676Z] Caused by: com.github.eirslett.maven.plugins.frontend.lib.TaskRunnerException: 'yarn run init --registry=https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/' failed.`

@ederign @tiagobento fyi ^

tarilabs (approved):
Approving with noted limitation from https://github.com/kiegroup/drools/pull/3259#discussion_r529292470

danielezonca:
jenkins do cdb

PR #3261, opened 2020-11-24
[DROOLS-5700] Implement range index in Alpha Network Compiler

tkobayas:
**JIRA**: 

https://issues.redhat.com/browse/DROOLS-5700

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-alphanetwork-compiler/src/main/java/org/drools/ancompiler/DeclarationsHandler.java
+++ b/drools-alphanetwork-compiler/src/main/java/org/drools/ancompiler/DeclarationsHandler.java
@@ -16,18 +16,22 @@
 
 package org.drools.ancompiler;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
 import org.drools.core.base.ValueType;
 import org.drools.core.reteoo.AlphaNode;
 import org.drools.core.reteoo.BetaNode;
 import org.drools.core.reteoo.LeftInputAdapterNode;
 import org.drools.core.reteoo.Sink;
 import org.drools.core.reteoo.WindowNode;
-import org.drools.core.reteoo.CompositeObjectSinkAdapter.FieldIndex;
 import org.drools.core.rule.ContextEntry;
 import org.drools.core.rule.IndexableConstraint;
 import org.drools.core.util.index.AlphaRangeIndex;
-
-import java.util.*;
 
 /**
  * This handler is used to create the member declarations section of a generated subclass of a {@link CompiledNetwork}.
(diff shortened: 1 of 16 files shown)


tkobayas:
Jenkins retest this

tkobayas:
Please review @lucamolteni 

SonarCloud tells:

> 9.0% Duplication

Indeed, we have some duplication between AssertHandler and ModifyHandler. How about changing ModifyHandler to be a subclass of AssertHandler? Does it make sense?

tkobayas:
Benchmark result with https://github.com/tkobayas/kie-benchmarks/blob/DROOLS-5700-bench/drools-benchmarks/src/main/java/org/drools/benchmarks/operators/AlphaNodeRangeIndexingBenchmark.java
```
Benchmark                             (alphanetworkCompilerEnabled)  (rangeIndexingEnabled)  (rulesAndFactsNumber)  Mode    Cnt    Score   Error  Units
AlphaNodeRangeIndexingBenchmark.test                          false                   false                     12    ss  20000   47.952 ± 0.342  us/op
AlphaNodeRangeIndexingBenchmark.test                          false                   false                     24    ss  20000  133.599 ± 1.727  us/op
AlphaNodeRangeIndexingBenchmark.test                          false                   false                     48    ss  20000  477.013 ± 2.687  us/op
AlphaNodeRangeIndexingBenchmark.test                          false                    true                     12    ss  20000   42.489 ± 1.624  us/op
AlphaNodeRangeIndexingBenchmark.test                          false                    true                     24    ss  20000  126.070 ± 0.855  us/op
AlphaNodeRangeIndexingBenchmark.test                          false                    true                     48    ss  20000  391.990 ± 2.298  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                   false                     12    ss  20000   40.054 ± 0.372  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                   false                     24    ss  20000  112.956 ± 1.297  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                   false                     48    ss  20000  374.519 ± 1.768  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                    true                     12    ss  20000   39.906 ± 0.215  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                    true                     24    ss  20000  124.887 ± 0.501  us/op
AlphaNodeRangeIndexingBenchmark.test                           true                    true                     48    ss  20000  347.360 ± 1.804  us/op
```
seems to be expected results. Better with ANC enabled. Better with RangeIndex enabled.

lucamolteni:
> Please review @lucamolteni
> 
> SonarCloud tells:
> 
> > 9.0% Duplication
> 
> Indeed, we have some duplication between AssertHandler and ModifyHandler. How about changing ModifyHandler to be a subclass of AssertHandler? Does it make sense?

I'm not sure - I haven't touched the original implementation as it's mostly done concatenating strings and I personally don't like it. The only place where I've replaced it with JavaParser is when I needed to partition the evaluation in order to avoid a "Code too large" error. 

I'd personally keep it as it is, and consider migrating it to JP when we have a good reason to do so. WDYT?

Thank you very much!

lucamolteni:
> seems to be expected results. Better with ANC enabled. Better with RangeIndex enabled.

This is good news! 

If we manage to demonstrate a consistent superiority of the ANC we could promote it to default eventually...

drools-alphanetwork-compiler/src/main/java/org/drools/ancompiler/DeclarationsHandler.java
```diff
@@ -140,4 +148,33 @@ public void startHashedAlphaNodes(IndexableConstraint indexableConstraint) {
...
+        return PRIVATE_MODIFIER + " " + AlphaRangeIndex.class.getName() + " " + variableName + "; // including " + comment + " etc.";
+    }
+
+    public Map<String, AlphaRangeIndex> getRangeIndexDeclarationMap() {
+        return rangeIndexDeclarationMap;
+    }
+
+    @Override
+    public void startRangeIndexedAlphaNode(AlphaNode alphaNode) {
+        builder.append(getAlphaNodeVariableDeclaration(alphaNode)).append(NEWLINE); // range index uses AlphaNode rather than MVELConstraint
```
lucamolteni: I thought one of the optimisation of the ANC was to unwrap the alpha node and use the constraint directly, effectively inlining the call. Can you please double check?

lucamolteni (approved):

tkobayas:
> I'd personally keep it as it is, and consider migrating it to JP when we have a good reason to do so. WDYT?

Agreed with the plan. Thanks!

drools-alphanetwork-compiler/src/main/java/org/drools/ancompiler/DeclarationsHandler.java
tkobayas: Yeah, I noticed that I don't need these AlphaNode variables! Actually I don't even need the internal constraints because we can get matching alphaNode's id from the range index and go to the next sink. I removed useless codes so now it's much simplified. Thanks!

tkobayas:
Jenkins retest this

tkobayas:
@lucamolteni FYI), Here is the generated class with Range Index enabled : https://gist.github.com/tkobayas/1a2ddf11968a5e7644d3fa91c2a93d20

tkobayas:
@hellowdan SonarCloud failed because of code Duplication but Luca and I agreed to keep it as-is for now. See the above discussion.

hellowdan (approved):

hellowdan:
> @hellowdan SonarCloud failed because of code Duplication but Luca and I agreed to keep it as-is for now. See the above discussion.

agreed, thanks!!

mariofusco (approved):

PR #3274, opened 2020-12-03
[DROOLS-5878] Fix "ApplyPmmlModelCommand.isMining()" implementation

gitgabrio:
@danielezonca @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-5878

### DIFF
--- a/drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
+++ b/drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
@@ -6,12 +6,15 @@ import org.drools.core.impl.KnowledgeBaseFactory;
 import org.junit.Test;
 import org.kie.api.KieBase;
 import org.kie.api.pmml.PMML4Result;
+import org.kie.api.pmml.PMMLConstants;
 import org.kie.api.pmml.PMMLRequestData;
 import org.kie.internal.command.RegistryContext;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
-
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 public class ApplyPmmlModelCommandTest {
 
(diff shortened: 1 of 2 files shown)


danielezonca:
Jenkins run fdb

danielezonca:
jenkins do fdb

danielezonca (approved):

drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
```diff
@@ -61,5 +64,16 @@ public void testHasRequestData() {
         assertEquals("ERROR-1", resultCode);
     }
 
+    @Test
```
jiripetrlik, line 67: Can we use a test prefix for this method. Something like testIsMining.

drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
```diff
@@ -61,5 +64,16 @@ public void testHasRequestData() {
...
+    public void isMining() {
+        ApplyPmmlModelCommand cmd = new ApplyPmmlModelCommand();
+        assertNull(cmd.getHasMining());
+        assertFalse(cmd.isMining());
+        cmd.setHasMining(false);
+        assertFalse(cmd.isMining());
+        cmd.setHasMining(true);
+        assertTrue(cmd.isMining());
```
jiripetrlik, line 75: Maybe we should test more options. Maybe I'm doing something wrong, but I've changed the logic of isMining method backwards and the test passed anyway.

drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
gitgabrio: @jiripetrlik 
Done

drools-ruleunit/src/test/java/org/drools/ruleunit/command/pmml/ApplyPmmlModelCommandTest.java
gitgabrio: @jiripetrlik 
Hi. Can't follow exactly what you mean.

jiripetrlik: I reverted "ApplyPmmlModelCommand.java" and run the same test. But it did not fail. So, it seems to me that it does not test your changes. Or maybe I'm doing something wrong.

jiripetrlik: @gitgabrio 
Thank for your explanation about this offline. As logic did not change I think it is ok.

jiripetrlik (approved):

PR #3275, opened 2020-12-03
DROOLS-5879 Optional newline handling in MVEL parser

evacchi:
https://issues.redhat.com/browse/DROOLS-5762

MVEL parser explicitly handles newlines so that statements can be terminated automatically without semicolons. However, in some situations this is not necessary. For instance, 
- while parsing constraints, newlines may be treated as non-significant whitespace. 
- the consequence, unless the dialect is MVEL, should treat code as regular Java code. 

We add support to a new lexer state, that can be switched at run-time (parser instantiation time), to allow for whitespace to be fully ignored, newlines included. 

This feature will be also used in the DRLX research experiment.

The change is actually quite minimal

1. we add a new lexical state called NEWLINES_SKIPPED
2. we explicitly add all tokens to the DEFAULT lexical state (which, unless specified, is indeed the default state)
3. we add newlines to SPECIAL_TOKENS only in the NEWLINES_SKIPPED lexical state
4. we add newlines to TOKENS only in the DEFAULT lexical state
5. we add a method `setDefaultState(state)` to the TokenManager (the lexer) so that it is possible to set the default state of the lexer (to NEWLINES_SKIPPED or DEFAULT).  

there is then a bit of boilerplate to make sure that when a lexical state transition really is necessary (`IN_MULTI_LINE_COMMENT`) we return to the new default lexical state, instead of just DEFAULT

Finally, a flag is exposed in `MvelParser` via a boolean field. We may consider forking `ParserConfiguration` and add it there (currently we are reusing the same class in JavaParser's library, so that can't be customized).

enabling the special behavior is as easy as setting the flag with `mvelParser.setIgnoreNewlines(true)`; e.g:

```
         String expr = "1 \n + 1";

        Expression r1 = parseExpression(parser, expr).getExpr();
        assertEquals("Parsing should stop at newline", "1", printConstraint(r1));

        MvelParser mvelParser = new MvelParser();
        mvelParser.setIgnoreNewlines(true);
        Expression r2 = mvelParser.parse(GeneratedMvelParser::Expression, new StringProvider(expr)).getResult().get();
        assertEquals("Parsing must ignore newlines","1 + 1", printConstraint(r2)); 
```

we may consider flipping the default (i.e. default to nonsignificant whitespace) but I wanted to keep the change minimal for now.

### DIFF
--- a/drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
+++ b/drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
@@ -30,6 +30,7 @@ import java.util.HashSet;
 import java.util.concurrent.TimeUnit;
 
 import com.github.javaparser.ParseProblemException;
+import com.github.javaparser.ParseResult;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.expr.BinaryExpr;
 import com.github.javaparser.ast.expr.BinaryExpr.Operator;
@@ -52,10 +53,12 @@ import org.junit.Ignore;
 import org.junit.Test;
 
 import static org.drools.mvel.parser.DrlxParser.parseExpression;
+import static org.drools.mvel.parser.Providers.provider;
 import static org.drools.mvel.parser.printer.PrintUtil.printConstraint;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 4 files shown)


drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
```diff
@@ -903,6 +906,18 @@ public void testArrayCreation2() {
...
     }
 
+    @Test
+    public void testSpecialNewlineHandling() {
+        String expr = "1 \n + 1";
+
+        Expression r1 = parseExpression(parser, expr).getExpr();
+        assertEquals("Parsing should stop at newline", "1", printConstraint(r1));
+
+        MvelParser mvelParser = new MvelParser();
```
lucamolteni: Maybe add an overload constructor? There should be template for the generated class

lucamolteni (approved):
Can you please also enable the failing test I showed to you? Thanks

mariofusco (approved):

drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
evacchi: memo for myself: this is the failing test https://github.com/kiegroup/drools/blob/6408b5e1caa1e5c5fe94e45f82e1bedfbf55e049/drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java#L125

PR #3276, opened 2020-12-03
DROOLS-5876: Display actual test results instead of a generic message

yesamer:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5876  and https://issues.redhat.com/browse/DROOLS-5402

Client-side PR contains minor changes only on UI (popover is larger and error message shorter).
Server-side PR introduced a new DTO, `ExpressionEvaluatorResult` which holds detailed information related to the result of an Expression Evaluation. In particular, it contains the original boolean which represents the computation result (true = successful, false = failure), and adds the wrong value and the path to reach it. 
Please note, currently is not possible to provide a correct value to the user, because a List evaluation doesn't consider the list order. 

Examples:
- DMN List
Screenshot from 2021-02-03 10-03-02

- Rule Map
Screenshot from 2021-02-03 10-26-14

- Rule List<String> (Simple list type)
Screenshot from 2021-02-03 10-28-08

- Rule List<Author> 
Screenshot from 2021-02-03 10-30-35

**referenced Pull Requests**: 

* https://github.com/kiegroup/drools-wb/pull/1447
* https://github.com/kiegroup/drools/pull/3276
<details>
<summary>
</summary>

</details>


drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/expression/ExpressionEvaluatorResult.java
```diff
@@ -0,0 +1,93 @@
...
+    public void addFieldItemStepToPath(String fieldName) {
+        pathToWrongValue.add(0, "Field \"" + fieldName+ "\"");
+    }
+
+    public void addListItemStepToPath(int elementNumber) {
+        pathToWrongValue.add(0, "List's item: " + elementNumber);
+    }
+
+    public void addMapItemStepToPath(String key) {
+        pathToWrongValue.add(0, "Map item with \"" + key + "\" key");
```
Rikkola: Will it be a problem for i18n that these are only in English?

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationBackendMessages.java
```diff
@@ -0,0 +1,47 @@
...
+        return "Value <strong>\"" + wrongValue + "\"</strong>" +
+                " is wrong inside:\n" +
+                "<em>" + String.join("\n", pathToWrongValue) + "</em>";
+    }
+
+    private static String getCollectionWithoutWrongValueHTMLErrorMessage(List<String> pathToWrongValue) {
+        if (pathToWrongValue == null || pathToWrongValue.isEmpty()) {
+            return getGenericCollectionErrorMessage();
+        }
+        return "Following path is wrong:\n" +
```
Rikkola: Same thing in this class. i18n might be an issue.

jstastny-cz:
@yesamer I have two comments on the presented screenshots:
1. there's no mention of that was the actual - no help to the user what the actual value was.
2. The details about exact position in List or Map are not straight forward, it's kinda complicated.
  ```
  Value "pop" is wrong inside:
  List's item: 3
  Field "author"
  Field "name"
  ```
  ```
  Value "asda" is wrong inside:
  Map Item with "one" key
  Field "name"
  ```
  I am wondering wouldn't be a structured output be better understandable?
  ```
  Error reason:
    Fact Type: List
    Item \#: 3
    Field: "author" -> "name"
    Value: "pop"
  ```
  ```
  Error reason:
    Fact Type: Map
    Item key: "one"
    Field: "name"
    Value: "asda"
  ```

yesamer:
@jstastny-cz 
1. As I wrote in the first comment this is not currently possible: `Please note, currently is not possible to provide a correct value to the user, because a List evaluation doesn't consider the list order.`
Currently, a list is equal to another list if they have the same elements in ANY order.
Eg. This case will return a success
GIVEN LIST:   'a' 'b' 'c' 'd'
EXPECTED LIST 'd' 'b' 'a' 'c' 
Eg. This case will return a failure
GIVEN LIST:   'a' 'b' 'c' 'd'
EXPECTED LIST 'f' 'g' 'e' 'h' 
But I can't tell to the user which is the correct value to put in the first place, because any of 'a' 'b' 'c' 'd' are acceptable.
What we can do is to provide the user a set of possible values, but I would consider this as a future enhancement of the feature. 
We are informing the user where the error is, and I think this is a big step. Consider that we show a very generic message `"Impossible to find elements in the collection to satisfy the conditions."`

2. I can try to find a way to improve it, consider there are several cases to manage, so I would prefer to use a simple way to write the message. I'll let you know.

jstastny-cz:
1) is clear now, thanks.
2) I would suggest to look into how to get some structure into the message, ideally a list or tabular form, so that it is well-testable in UI tests.

yesamer:
@jstastny-cz I partially applied your proposed format.
New version:

Assertion failed in:
  _Item #: 3
  Field: "author"
  Field: "name"
  Value: "**pop**"_

Assertion failed in:
  _Item key: "one"
  Field: "name"
  Value: "**asada**"_


I think is not necessary to specify if the collection is a Map or a List because this info is already present in the grid.
Regarding using a single line for fields, I prefer to use multiple lines in the case of multiple nested types.
Reason is that in future we could tackle the case of very complex nested type (eg. "author" -> "book" -> Item: 1 -> "name" )...
In addition, I think the user common case is to have only a level of the field.
WDYT?

Rikkola (approved):

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationBackendMessages.java
yesamer: @Rikkola Thank you for spotting, you're right. Usually, I can easily manage i18n on the front end side, but how I should manage it backend side?

karreiro: @yesamer @Rikkola My suggestion would to return keys/maps from the backend, and consume I18n API in the front end :-)

karreiro (commented):
Thank you for this PR, @yesamer!

I've noticed that we're not showing the buttons "Keep" and "Apply" on these results (but we are on more primitive values):
Screenshot 2021-02-12 at 13 57 29

Is that expected?

Thanks again! :-)

yesamer:
@karreiro Yes, it is. In this case, we are just informing the user where the problem is. We are not telling the actual value, so the user can't apply anything.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationBackendMessages.java
yesamer: @Rikkola @karreiro Can you please point me an example? :)

jstastny-cz:
@yesamer 

For me it was not easy to understand what sort (if any) of hierarchy is there between the two lines:
```
Field: "author"
Field: "name"
```

Secondly I would like to ask for some sort of DOM structure for presenting these data, not just strings (even though formatted).

Would sth like this be possible?
```
<ul>
<li xy="item-identifier">
    <span data-field="title">Item #</span>:<span data-field="value">3</span>
</li>
.
.
.
```
In such case the Field structure can be achieved by a nested list, so that we have indenting.
<ul>
  <li>
    <span data-field="field-name">author</span>
    <ul>
        <li>
          <span data-field="field-name">name</span>
        </li>
      </ul>
  </li>
</ul>

```
<ul>
  <li>
    <span data-field="field-name">author</span>
    <ul>
        <li>
          <span data-field="field-name">name</span>
        </li>
     </ul>
  </li>
</ul>
```

So overall look could be:
<ul>
<li xy="item-identifier">
    <span data-field="title">Item #</span>: <span data-field="value">3</span>
</li>
<li xy="field-structure">
<span data-field="title">Field</span>:
<ul>
  <li>
    <span data-field="field-name">author</span>
    <ul>
        <li>
          <span data-field="field-name">name</span>
        </li>
      </ul>
  </li>
</ul>
</li>
<li xy="value">
    <span data-field="title">Value</span>: <span data-field="value">"pop"</span>
</li>
</ul>

yesamer:
@jstastny-cz It's an excellent proposal, and I had a very similar idea during the analysis of the task.
The problem I see to implement that is the popover: in that widget, we have a very limited space to implement a tree as you proposed. I can try to figurate out how it could be.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/ScenarioSimulationBackendMessages.java
Rikkola: @yesamer Yes what @karreiro said. I do this with Verification. For the errors/warnings/info messages I have an ID for each type that is just an enum. Then in the client side I have a Report Generator that gets a message based on the ID. If the message template needs variable, I have them in the message object. 

In the example I produce HTML, but might as well just use the same for more simple i18n.
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-verifier/kie-wb-common-verifier-reporting-client/src/main/java/org/kie/workbench/common/services/verifier/reporting/client/reporting/ExplanationProvider.java

yesamer: @Rikkola Thank you, I moved all messages on the client-side and managed them with i18n.

yesamer:
@karreiro @Rikkola @jstastny-cz Ready to be reviewed.
Applied changes:
- All messages are now handled in client-side
- UI Changes suggested by @jstastny-cz applied (preparing screenshots)

yesamer:
@jstastny-cz @Rikkola @karreiro 
Screenshot from 2021-02-17 16-08-14
Screenshot from 2021-02-17 16-06-35
Screenshot from 2021-02-17 16-05-18
Snapshots of the final version. Can you please finalize the review? Thank you!

karreiro (approved):
Thank you, @yesamer LGTM!

jstastny-cz (approved):
LGTM.

yesamer:
Failing Tests: 
`org.drools.mvel.integrationtests.concurrency.SharedSessionParallelTest.testNoExceptions[Enforced jitting=false, Serialize KieBase=false]
org.drools.mvel.integrationtests.waltz.ReteOOWaltzTest.testWaltz`

yesamer:
jenkins retest this please

yesamer:
@karreiro Can you please merge it?

PR #3277, opened 2020-12-03
[DROOLS-5875] phreak groupby node

mariofusco:


mariofusco:
replaced by https://github.com/kiegroup/drools/pull/3298

PR #3280, opened 2020-12-04
DROOLS-5879 Optional newline handling in MVEL parser

evacchi:
https://issues.redhat.com/browse/DROOLS-5762

This is an alternative IMO cleaner version of https://github.com/kiegroup/drools/pull/3275. 

Instead of a new lexer state, we take inspiration for [this antlr JavaScript grammar](https://github.com/antlr/grammars-v4/blob/master/javascript/javascript/JavaScriptParser.g4#L506-L510)

The change is even more minimal that the other PR, plus it makes IMO cleaner the grammar, by getting rid of `EOL()`.

- it makes all whitespace, including new lines, SPECIAL_TOKENs -- so they are processed by the lexer, but not emitted directly to the parser
- it introduces a new rule in the parser EOS() -- End Of Statement
- EOS() is either a semicolon or EOF or a _semantic lookahead_ (i.e. a predicate)
- in the semantic lookahead -- if semicolons are optional -- we look for either:
   (1) a newline on the "special" channel before the next token;
   (2) a closing brace '}';
   (3) an identifier (this is probably a special case)
   (4) if it's a `with` or `modify` statement, a comma may also terminate the statement

(4) uses a minor hack which assumes that `modify` and `with` cannot nest; i.e. I am using a **boolean** when we descend into a `with` or `modify` statement (I think the grammar may still allow for nesting but I think that probably we can safely disallow it). 

Finally, a flag is exposed in `MvelParser` via a boolean field. We may consider forking `ParserConfiguration` and add it there (currently we are reusing the same class in JavaParser's library, so that can't be customized).

enabling the special behavior is as easy as setting the flag with `mvelParser.setOptionalSemicolon(false)` (see the test)

we may consider flipping the default (i.e. default to nonsignificant whitespace) but I wanted to keep the change minimal for now.

### DIFF
--- a/drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/MvelParser.java
+++ b/drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/MvelParser.java
@@ -64,6 +64,7 @@ import static org.drools.mvel.parser.Providers.resourceProvider;
  */
 public final class MvelParser {
     private final ParserConfiguration configuration;
+    private boolean ignoreNewlines = false;
 
     private GeneratedMvelParser astParser = null;
     private static ParserConfiguration staticConfiguration = new ParserConfiguration();
@@ -118,6 +119,10 @@ public final class MvelParser {
         return this.configuration;
     }
 
+    public void setIgnoreNewlines(boolean ignoreNewlines) {
+        this.ignoreNewlines = ignoreNewlines;
+    }
+
     private GeneratedMvelParser getParserForProvider(Provider provider) {
         if (astParser == null) {
             astParser = new GeneratedMvelParser(provider);
@@ -126,6 +131,10 @@ public final class MvelParser {
         }
         astParser.setTabSize(configuration.getTabSize());
         astParser.setStoreTokens(configuration.isStoreTokens());
+        if (ignoreNewlines) {
+            astParser.getTokenSource()
+                    .setDefaultState(GeneratedMvelParserTokenManager.NEWLINES_SKIPPED);
+        }
         return astParser;
     }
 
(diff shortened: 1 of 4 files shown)


drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
```diff
@@ -716,14 +722,14 @@ public void testWithConstructor() {
...
+                        "a()" + newLine() +
+                        "b()" + newLine() +
                         "}";
 
         BlockStmt expression = MvelParser.parseBlock(expr);
         assertEquals("{" + newLine() +
-                             "    a;" + newLine() +
-                             "    b;" + newLine() +
+                             "    a();" + newLine() +
+                             "    b();" + newLine() +
```
evacchi, line 732: I updated this test because I don't think this parses in a regular java grammar. Field/variable accesses are not statements and this was failing IIRC

drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
```diff
@@ -815,8 +821,6 @@ public void commentsWithEmptyStatements() {
 
         BlockStmt expression = MvelParser.parseBlock(expr);
         assertEquals("{" + newLine() +
-                             "    ;" + newLine() +
-                             "    ;" + newLine() +
```
evacchi, line 819: I removed these newlines because they are now treated as whitespace and not empty statements. I think this a good behavior.

drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/MvelParser.java
```diff
@@ -91,6 +92,13 @@ public MvelParser() {
     public MvelParser(ParserConfiguration configuration) {
         this.configuration = configuration;
         configuration.getPostProcessors().clear();
+        this.optionalSemicolon = true;
```
mariofusco: Nitpick: you could make this constructor to simply call the other, passing `true` as 2nd argument.

mariofusco (approved):

lucamolteni (approved):

evacchi:
is this ok to merge? @mariofusco @lucamolteni

PR #3299, opened 2020-12-22
[DROOLS-5915] avoid spurious firing in self join forall when no fact …

mariofusco:
…matching the forall is present

**JIRA**: https://issues.redhat.com/browse/DROOLS-5915


lucamolteni (approved):

mariofusco:
jenkins retest

PR #3301, opened 2020-12-23
[DROOLS-5918] fix nested accumulates in executable model

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5918


lucamolteni (changes requested):
Rather than adding the same parameter to each method, can we promote it to a field? These are all non-static functions

lucamolteni (approved):

PR #3304, opened 2021-01-04
[DROOLS-5910] Enable range index for JoinNode

tkobayas:
- WIP

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-5910

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
@@ -72,7 +72,7 @@ public class LambdaConstraint extends AbstractConstraint {
                     indexingDeclaration = indexingDeclaration.clone();
                     org.drools.model.Index.ConstraintType constraintType = index.getConstraintType();
                     Class<?> accessorFieldType;
-                    if (((BetaIndex) index).getRightReturnType() != null) {
+                    if (constraintType.isComparison() && ((BetaIndex) index).getRightReturnType() != null) {
                         accessorFieldType = ((BetaIndex) index).getRightReturnType();
                     } else {
                         accessorFieldType = index.getIndexedClass();
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
@@ -70,7 +70,14 @@ public class LambdaConstraint extends AbstractConstraint {
                 indexingDeclaration = evaluator.getRequiredDeclarations()[0];
                 if ( indexingDeclaration.getExtractor() instanceof PatternExtractor ) {
                     indexingDeclaration = indexingDeclaration.clone();
-                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), index.getIndexedClass(), (( BetaIndex ) index).getRightOperandExtractor() ) );
+                    org.drools.model.Index.ConstraintType constraintType = index.getConstraintType();
+                    Class<?> accessorFieldType;
+                    if (((BetaIndex) index).getRightReturnType() != null) {
+                        accessorFieldType = ((BetaIndex) index).getRightReturnType();
+                    } else {
+                        accessorFieldType = index.getIndexedClass();
+                    }
+                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), accessorFieldType, (( BetaIndex ) index).getRightOperandExtractor() ) );
                 }
             }
         }
@@ -172,7 +179,7 @@ public class LambdaConstraint extends AbstractConstraint {
 
     @Override
     public boolean isIndexable( short nodeType ) {
-        return getConstraintType().isIndexableForNode(nodeType);
+        return getConstraintType().isIndexableForNode(nodeType, this);
     }
 
     @Override
@@ -212,6 +219,11 @@ public class LambdaConstraint extends AbstractConstraint {
     @Override
     public InternalReadAccessor getFieldExtractor() {
         return readAccessor;
+    }
+
+    @Override
+    public Declaration getIndexingDeclaration() {
+        return indexingDeclaration;
     }
 
     public static class LambdaContextEntry implements ContextEntry {
--- a/drools-core/src/main/java/org/drools/core/base/CoercionUtil.java
+++ b/drools-core/src/main/java/org/drools/core/base/CoercionUtil.java
@@ -163,4 +163,30 @@ public class CoercionUtil {
         }
         return ret;
     }
+
+    public static Number coerceToNumber(Number value, Class<?> toClass) {
+        Number ret = null;
+        if (value != null) {
+            if (toClass.equals(BigDecimal.class)) {
+                ret = MathUtils.getBigDecimal(value);
+            } else if (toClass.equals(BigInteger.class)) {
+                ret = MathUtils.getBigInteger(value);
+            } else if (toClass.equals(Double.class)) {
+                ret = value.doubleValue();
+            } else if (toClass.equals(Float.class)) {
+                ret = value.floatValue();
+            } else if (toClass.equals(Long.class)) {
+                ret = value.longValue();
+            } else if (toClass.equals(Integer.class)) {
+                ret = value.intValue();
+            } else if (toClass.equals(Short.class)) {
+                ret = value.shortValue();
+            } else if (toClass.equals(Byte.class)) {
+                ret = value.byteValue();
+            } else {
+                throw new ClassCastException("Not possible to coerce [" + value + "] from class " + value.getClass() + " to class " + toClass);
+            }
+        }
+        return ret;
+    }
 }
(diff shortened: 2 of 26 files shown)


tkobayas:
Jenkins retest this

tkobayas:
Jenkins retest this

tkobayas:
Jenkins retest this

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
```diff
@@ -70,7 +71,14 @@ private void initIndexes() {
...
                     indexingDeclaration = indexingDeclaration.clone();
-                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), index.getIndexedClass(), (( BetaIndex ) index).getRightOperandExtractor() ) );
+                    org.drools.model.Index.ConstraintType constraintType = index.getConstraintType();
+                    Class<?> accessorFieldType;
+                    if (constraintType.isComparison() && indexingDeclaration.getValueType() != null) {
+                        accessorFieldType = indexingDeclaration.getValueType().getClassType(); // For comparison index, it requires the actual right hand type. However, this is always "Object"
+                    } else {
+                        accessorFieldType = index.getIndexedClass();
+                    }
+                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), accessorFieldType, (( BetaIndex ) index).getRightOperandExtractor() ) );
```
tkobayas: Hi @mariofusco @lucamolteni   I want your advise regarding a tricky issue with exec-model `LambdaConstraint` index. 

While enabling range index for JoinNode, I see there are some type coercion problems. So I added a method to check constraint operands (`IndexUtil.areRangeIndexCompatibleOperands()`) so we use range index only when Number comparison or same Comparable class comparison. I think the criteria is fine.

Now the problem is that it's difficult to convey the type information of the right hand operand of `LambdaConstraint` (involving `PatternExtractor` : e.g. `Person( age > $pet.age )`). Currently `indexingDeclaration.readAccessor.fieldType` is populated with `index.getIndexedClass()` so I cannot see the actual return type of the right operand. (`MvelConstraint.indexingDeclaration` has the actual return type on the other hand)

`LambdaConstraint.initIndexes()`:
```
                ...
                if ( indexingDeclaration.getExtractor() instanceof PatternExtractor ) {
                    indexingDeclaration = indexingDeclaration.clone();
                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), index.getIndexedClass(), (( BetaIndex ) index).getRightOperandExtractor() ) );
                }
```

So I tried to use the indexingDeclaration's fieldType but it's always "`Object`". Probably it's not a right approach.

```
                if ( indexingDeclaration.getExtractor() instanceof PatternExtractor ) {
                    indexingDeclaration = indexingDeclaration.clone();
                    org.drools.model.Index.ConstraintType constraintType = index.getConstraintType();
                    Class<?> accessorFieldType;
                    if (constraintType.isComparison() && indexingDeclaration.getValueType() != null) {
                        accessorFieldType = indexingDeclaration.getValueType().getClassType(); // For comparison index, it requires the actual right hand type. However, this is always "Object"
                    } else {
                        accessorFieldType = index.getIndexedClass();
                    }
                    indexingDeclaration.setReadAccessor( new LambdaReadAccessor( index.getIndexId(), accessorFieldType, (( BetaIndex ) index).getRightOperandExtractor() ) );
                }
```

Btw, `BetaIndexImpl.rightOperandExtractor` is a `LambdaExtractor` so I cannot get the actual return type as well.

A) Is there a way to get the return type of the LambdaConstraint's right operand?

B) Should I enhance `BetaIndexImpl` so I can populate `LambdaConstraint.indexingDeclaration.readAccessor.fieldType` with the actual return type? (It may be over-killing?)

C) Or can I simply remove `D.betaIndexedBy() `during the PATTERN DSL generation when it's a comparison operator (>, <, ...) and doesn't meet the criteria (Number or same Comparable class) because it will not be used anyway?

Feel free to let me know if the explanation is not enough or clear. Thanks!

mariofusco: Hi @tkobayas I don't know if you saw [this fix](https://github.com/kiegroup/drools/commit/e33abb6536b1b7cfe83df79bd1109a4fbcf40b4f) that I made last week, but, if I'm not mistaken, I believe it is related with your problem and hopefully part of the solution. To implement this fix I needed to change the signature of the [betaIndexedBy](https://github.com/kiegroup/drools/commit/e33abb6536b1b7cfe83df79bd1109a4fbcf40b4f#diff-e901257553ca0a67b065a65b472fc94b666ce8826840b18f3d3eaf9bfa833e6fL107) method and in particular relax the constraint of the function return the right value to be indexed that now can be of any arbitrary type and doesn't have to be of the same type of the left value.

What happens at runtime is that the `indexedClass` is used to determine the `ValueType` (that is now an [enum](https://github.com/kiegroup/drools/commit/e33abb6536b1b7cfe83df79bd1109a4fbcf40b4f#diff-e7bfee3d4873f23d06adbdf9ce9af7f3e244a7a411d09d6e1fbfff0be54f1d29R31)) of the index. Then the indexing data structure, (my fix was for equality index, so the `TupleIndexHashTable` in my case) [coerces the right value](https://github.com/kiegroup/drools/commit/e33abb6536b1b7cfe83df79bd1109a4fbcf40b4f#diff-d8f10aae640557cdafd70e600877d5ca8aff7d588094c86ff5e9364e282e2cdaR341) to be indexed to the `ValueType` of the index.

As anticipated I'm probably missing some details of your issue, since I haven't tried to give a run at your pull request yet, but I'm hoping that you could leverage the same mechanism and reuse it in the range indexing also invoking the ValueType coercion from there. Am I making sense? Is there something preventing you to use this strategy?

tkobayas: Thank you @mariofusco  Yeah, your fix is related but I'm still stuck at the point "What is the return type of the right operand?".

For example, when I run the following rule (JoinNodeRangeIndexingTest.testRangeIndexForJoin) for exec-mode,

```
when
   $pet : Pet()
   Person( age > $pet.age )
then
   ...
```

1) During PATTERN DSL generation, at PatternExpressionBuilder.buildIndexedBy(), we know that the return type of "$pet.getAge()" is "int" from `TypedExpression right`.

2) When PATTERN DSL is written, we no longer have the return type of "$pet.getAge()".

```
D.betaIndexedBy(
  int.class,
  org.drools.model.Index.ConstraintType.GREATER_THAN,
  DomainClassesMetadataC03F4BCD5D1100AD260234F5B01B3A51.org_drools_testcoverage_common_model_Person_Metadata_INSTANCE.getPropertyIndex("age"),
  defaultpkg.PA0.LambdaExtractorA004CC49B66BA2D3FA53699059788395.INSTANCE,
  defaultpkg.PD6.LambdaExtractorD67BF1C9955A873441E78F0C494C8236.INSTANCE
)
```
```
public enum LambdaExtractorD67BF1C9955A873441E78F0C494C8236 implements org.drools.model.functions.Function1<org.drools.testcoverage.common.model.Pet, java.lang.Object>, org.drools.model.functions.HashedExpression {

    INSTANCE;

    public static final String EXPRESSION_HASH = "7D1018A5F1910E6120D92E3BEE901EC7";

    public java.lang.String getExpressionHash() {
        return EXPRESSION_HASH;
    }

    @Override()
    public java.lang.Object apply(org.drools.testcoverage.common.model.Pet $pet) {
        return $pet.getAge();
    }
}
```

3) When LambdaConstraint is created (= kbase build), LambdaConstraint knows only ConstraintEvaluator and PredicateInformation.

... Oh, can I use PredicateInformation to bring the return type information? Let me try...

tkobayas:
I can convey the right return type by using PredicateInformation. @lucamolteni  suggested me that in case of non-externalized lambda, we cannot use PredicateInformation. In this case, I disable range index for the constraint. Luca told me that lambda vs externalised behaviour shouldn't diverge much. But range index is an internal optimization and the rule outcome would be the same so I think it's justified...

Now unit tests passed and I started working on benchmark. I'll share the results later...

mariofusco:
@tkobayas what I'm not understanding is why the equality index doesn't need to know the right type (it's enough to coerce the right object extracted by the function to the left type, whatever it is, through the `ValueType` enumeration) while this is not enough for the range indexing. Can you please clarify this?

> I can convey the right return type by using PredicateInformation. 

In case you need to know the right type, I don't like this solution (that seems more a hack) that, as probably @lucamolteni already pointed out, only works when the executable model is generated from the drl. We want to fully support direct use of executable model, like for instance optaplanner is already doing these days. In essence if the beta indexes eventually need also that right type we need a way to explicitly provide that information. For this reason my suggestion is to add a `betaIndexedBy` overload (please leave the existing method as it is) also taking the class of the right type as argument.

tkobayas:
@mariofusco  Thank you very much for the suggestions.

>  why the equality index doesn't need to know the right type

Oops, I just focused to work on inequality cases. Sure, I can set the right type in the equality case as well.

> my suggestion is to add a betaIndexedBy overload (please leave the existing method as it is) also taking the class of the right type as argument.

Thanks, sure, I will implement it instead of using PredicateInformation.

tkobayas:
Jenkins retest this

tkobayas:
Superseded by https://github.com/kiegroup/drools/pull/3343
Closing.

tkobayas:
Superseded by https://github.com/kiegroup/drools/pull/3343
Closing.

PR #3305, opened 2021-01-04
[DROOLS-5924] String vs Number Coercion behavior difference between s…

tkobayas:
…tandard-drl and exec-model

- Test case

**JIRA**:

https://issues.redhat.com/browse/DROOLS-5924

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/base/CoercionUtil.java
+++ b/drools-core/src/main/java/org/drools/core/base/CoercionUtil.java
@@ -137,4 +137,30 @@ public class CoercionUtil {
         }
         throw new RuntimeException("Unable to coerce " + value + " into an Integer");
     }
+
+    public static Number coerceToNumber(String value, Class<?> toClass) {
+        Number ret = null;
+        if (value != null) {
+            if (toClass.equals(BigDecimal.class)) {
+                ret = MathUtils.getBigDecimal(value);
+            } else if (toClass.equals(BigInteger.class)) {
+                ret = MathUtils.getBigInteger(value);
+            } else if (toClass.equals(Double.class)) {
+                ret = Double.valueOf(value);
+            } else if (toClass.equals(Float.class)) {
+                ret = Float.valueOf(value);
+            } else if (toClass.equals(Long.class)) {
+                ret = Long.valueOf(value);
+            } else if (toClass.equals(Integer.class)) {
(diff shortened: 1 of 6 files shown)


tkobayas:
Jenkins retest this

tkobayas:
Jenkins retest this

lucamolteni (approved):

hellowdan (approved):

mariofusco (changes requested):
@tkobayas This PR looks formally correct to me, but I'm a bit worried about performances. In practice, correct me if I'm wrong, but we're penalizing each and every comparison with that coercion code, just to cover something that it's quite an extreme edge case. I think that it could be possible to determine these situations at executable model generation time and invoke specific methods performing that coercion only in those cases. Let me know what you think about it and if my suggestion makes sense and is feasible.

tkobayas:
@mariofusco 

The methods which are added in this PR are actually for the edge cases where one of arguments is Object.
```
public static boolean greaterThan(Object o1, Comparable c2) {
   ...

public static boolean lessThan(Object o1, Comparable c2) {
   ...

public static boolean greaterThan(Comparable c1, Object o2) {
   ...

public static boolean lessThan(Comparable c1, Object o2) {
   ...
```

In normal cases, exec-model uses existing methods where arguments are both Comparable or Number so doesn't do type coercion.
```
    public static boolean greaterThan(Comparable c1, Comparable c2) {
        return c1 != null && c2 != null && c1.compareTo( c2 ) > 0;
    }
   ...

    public static boolean greaterThanNumbers(Number n1, Number n2) {
   ...
```

I confirmed that only test cases which explicitly use Object.class go though the EvaluationUtil.coerceAndEvaluate() (e.g. CompilerTest.testBigDecimalIntCoercion()).

Do you think it's fine? If you think it's better to use explicit method names for the cases, e.g. greaterThanWithCoercion() instead of overloading, I would implement that. WDYT?

mariofusco:
> Do you think it's fine? If you think it's better to use explicit method names for the cases, e.g. greaterThanWithCoercion() instead of overloading, I would implement that. WDYT?

@tkobayas sorry I misread your PR and didn't notice that you added new methods, leaving unchanged and still calling the old ones when possible. That's ok for me. Merging.

mariofusco (approved):

PR #3309, opened 2021-01-05
[DROOLS-5926] fix interpretation of numeric strings with leading 0 in executable model

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5926


lucamolteni (changes requested):
Why not use 
```Integer.valueOf("0008800")```
this will support any number of trailing zeroes

lucamolteni (approved):

PR #3310, opened 2021-01-05
[DROOLS-5930] allow an accumulate function to state if it supports re…

mariofusco:
…verse for the specific data to be reversed

**JIRA**: https://issues.redhat.com/browse/DROOLS-5930

### DIFF
--- a/drools-core/src/main/java/org/drools/core/spi/Accumulator.java
+++ b/drools-core/src/main/java/org/drools/core/spi/Accumulator.java
@@ -103,14 +103,14 @@ public interface Accumulator
      * @param workingMemory
      * @throws Exception
      */
-    void reverse(Object workingMemoryContext,
-                 Object context,
-                 Tuple leftTuple,
-                 InternalFactHandle handle,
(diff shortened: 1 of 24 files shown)


drools-core/src/main/java/org/drools/core/spi/Accumulator.java
```diff
@@ -103,14 +103,14 @@ Object accumulate(Object workingMemoryContext,
...
-                 Declaration[] innerDeclarations,
-                 WorkingMemory workingMemory) throws Exception;
+    boolean tryReverse(Object workingMemoryContext,
+                       Object context,
+                       Tuple leftTuple,
+                       InternalFactHandle handle,
+                       Object value,
+                       Declaration[] declarations,
+                       Declaration[] innerDeclarations,
+                       WorkingMemory workingMemory) throws Exception;
```
triceo, line 113: Note to self: have to do another OptaPlanner PR to adapt to this.

lucamolteni (approved):

PR #3313, opened 2021-01-08
[DROOLS-5938] fix resolution of accumulating variable in executable m…

mariofusco:
…odel

**JIRA**: https://issues.redhat.com/browse/DROOLS-5938

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
@@ -24,18 +24,22 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.commons.math3.util.Pair;
 import org.assertj.core.api.Assertions;
 import org.drools.core.ClockType;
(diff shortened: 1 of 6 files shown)


lucamolteni (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
```diff
@@ -907,4 +911,38 @@ public void testAfterWithAnd() throws Exception {
 
         assertEquals( 0, ksession.fireAllRules() );
     }
+
+    @Test
+    public void testTwoAccumulatesWithVarBindingModel() {
```
triceo, line 916: IMO we don't need to add this test. It's my original reproducer, which has since been converted to smaller (and much nicer) tests by Mario. 

I may be wrong, but I think this test is just bloat now.

mariofusco:
jenkins retest this

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
mariofusco: @triceo this test is not exactly the same thing of the drl based ones that I also added, as demonstrated by the fact that it was not failing with 7.48.0.Final while mine did. If you don't mind I'll keep this.

PR #3314, opened 2021-01-08
[RHDM-1550] add test case for new line in mvel consequence

mariofusco:
**JIRA**: https://issues.redhat.com/browse/RHDM-1550


mariofusco:
@lucamolteni I only added the test case which is already successful on master. It seems to be fixed by @evacchi's work on mvel parser.

mariofusco:
jenkins retest this

lucamolteni (approved):

PR #3324, opened 2021-01-13
[DROOLS-5945] fix type coercion for a string bound variable in execut…

mariofusco:
…able model

**JIRA**: https://issues.redhat.com/browse/DROOLS-5945

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/ConstraintParser.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/ConstraintParser.java
@@ -464,10 +464,10 @@ public class ConstraintParser {
         Expression rightExpr = right.uncastExpression();
 
         if (isLeftNumber) {
-            if ( right.getBoxedType().map( String.class::isAssignableFrom ).orElse( false ) ) {
+            if ( isString( right ) ) {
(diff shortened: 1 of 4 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/ConstraintParser.java
```diff
@@ -449,18 +449,32 @@ private static String getExpressionSymbol(Expression expr) {
...
+
+        String equalsMethod = isLeftNumber && isRightNumber ?
                 "org.drools.modelcompiler.util.EvaluationUtil.areNumbersNullSafeEquals" :
                 "org.drools.modelcompiler.util.EvaluationUtil.areNullSafeEquals";
 
+        Expression leftExpr = left.uncastExpression();
+        Expression rightExpr = right.uncastExpression();
+
+        if (isLeftNumber) {
+            if ( right.getBoxedType().map( String.class::isAssignableFrom ).orElse( false ) ) {
```
lucamolteni: I think you can extract a few methods in here

lucamolteni (changes requested):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/ConstraintParser.java
mariofusco: @lucamolteni good point, done.

lucamolteni (approved):

PR #3326, opened 2021-01-18
DROOLS-5951 DroolsMvelParserTest fails due to line endings \r\n on Windows

evacchi:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5951

the issue is that two strings are being compared. 
The fix is as simple as changing the assertion. 

In fact, the "expected" string uses `\n`, but the `printConstraint()` string returns a platform-dependent newline encoding; thus it uses `\r\n` on Windows. – in fact the number of semicolons in both strings is 2; so that should mean that the `expr` string has been parsed correctly


hellowdan (approved):
I tested the fix locally and it worked on both Windows and Linux.

evacchi:
jenkins run cdb

mariofusco (approved):

PR #3327, opened 2021-01-18
DROOLS-5997 Initial DRLX support

evacchi:
Adds changes to the MVEL.jj grammar -- disconnected from the rest of the grammar (basically a new type of CompilationUnit); handles those changes in a Visitor that translate the DrlxCompilationUnit into DRL Descrs. 

Connects it to the build via an assembler.

**JIRA**:
- https://issues.redhat.com/browse/DROOLS-5997

**referenced Pull Requests**:

- https://github.com/kiegroup/droolsjbpm-knowledge/pull/495


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/drlx/DrlxVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+        for (TypeDeclaration<?> typeDeclaration : u.getTypes()) {
+            RuleDeclaration rd = (RuleDeclaration) typeDeclaration;
+            this.visit(rd, null);
+        }
+        PackageDescr descr = builder.getDescr();
+        return descr;
+    }
+
+    @Override
+    public ImportDescr visit(ImportDeclaration decl, Void v) {
```
lucamolteni: I'm not sure how I understood how this works though, do we need to return the Descr as the return type? It seems like the return type of the visit is ignored at :37

lucamolteni (approved):
Good job!

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/drlx/DrlxVisitor.java
mariofusco: If I'm understanding correctly this visit is doing an AST (CompilationUnit) to AST (PackageDescr) transformation which is a good idea and perfectly fine in this context. I agree with @lucamolteni that having a visit method that return a value and consistently ignoring it is at least confusing.

mariofusco (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/drlx/DrlxVisitor.java
evacchi: IIRC the reason why this is ignoring the return type most of the time is that *builders automatically append their children*; so using the return type in general is only useful to the root transformation because the caller is "outside". This in general only happens for PackageDescr, but it would also allow to "enter" the visit at arbitrary nodes (e.g. only apply to consequence or an expression).

An alternative may be to use a `DrlVoidVisitor` and then a `getPackageDescr()` method; but then if you ever had to enter the visit at arbitrary places, you could never get intermediate results. This could be useful e.g. for finer-grained testing. Or, in general, some other transformation may need that return value.

Maybe I can at least add a few lines of comment, with what I have described above.

lucamolteni: I'd rather use the return type by explicitly adding the transformed node after calling visit or either change the return type to void and put the root return value as a field in the `DrlxVisitor` itself

evacchi: I think that by adding the node explicitly you would end up with the same node being appended twice (the first by the builder, the second when you add it). I will use the field.

PR #3335, opened 2021-01-21
DROOLS-5950 Support returning PackageDescr from assemblers

evacchi:
We add a hook for assemblers that runs _before_ rules, to make sure that PackageDescrs are picked up at the right time by the runtime. This allows for improved processing of PMML and (near future) experimental DRLX -- a follow-up PR

This PR includes a tiny test case for Assemblers. The test case has been added in the exec model compiler module just to make sure that PackageDescrs are not evaluated "dynamically" using MVEL. The exec model compiler module is one of the simplest module without the MVEL run-time dependency.

**JIRA**: https://issues.redhat.com/browse/DROOLS-5950

**referenced Pull Requests**: 

- https://github.com/kiegroup/drools/pull/3335
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/497


evacchi:
jenkins run cdb

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
```diff
@@ -186,9 +189,34 @@ private static ResourceWithConfiguration descrToResourceWithConfiguration(Resour
         buildResource(packages, ResourceType.TEMPLATE, ResourceToPkgDescrMapper.TEMPLATE_TO_PKG_DESCR);
         buildResource(packages, ResourceType.GDST, ResourceToPkgDescrMapper.GUIDED_DTABLE_TO_PKG_DESCR);
         this.resourcesByType.remove(ResourceType.DRT); // drt is a template for dtables but doesn't have to be built on its own
+
+        buildResourcesFromAssemblers();
```
gitgabrio: @evacchi 
I would prefer to have this method with a different name (e.g. `buildBeforeRules`) invoked directly inside the `build` one, i.e.

`      ....
       kBuilder.registerBuildResources(getResources());
        buildResources();
        buildBeforeRules();
        if(buildRules) {
        ....
`
So that:
1) the `build` method is a sort of "orchestrator", clearly defining different "life-cycle" invocations
2) the invoked name clearly express its scope/semantic
3) the `buildResources` method keep a consistent (even if by chance) semantic, i.e. it deal with no-assembler resources

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
```diff
@@ -186,9 +189,34 @@ private static ResourceWithConfiguration descrToResourceWithConfiguration(Resour
...
+
         return packages.values();
     }
 
+    private void buildResourcesFromAssemblers() {
```
gitgabrio: @evacchi 
Please see above comment and https://github.com/kiegroup/droolsjbpm-knowledge/pull/497#discussion_r564353874 about name inconsistence: here we have a `buildResourcesFromAssemblers` (here it is not a "what" nor a "when", but sort of "how") that invokes a `addResourceAsPackageDescr`. Could you please unify all that naming with same semantic ?

gitgabrio (changes requested):
@evacchi 
Many thanks for those two PRs: can't wait to have them merged (it would allow me to remove at least some/much of unwanted code).
I would just kindly ask you to review the naming used all over.

lucamolteni (approved):

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
evacchi: `buildResourcesFromAssemblers` <--what about `buildAssemblerResources` ?

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
evacchi: ah yes there was a reason for it being there -- it was originally `buildResourcesFromAssemblers(packages)` -- but it's no longer the case, so it can definitely be moved outside

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
evacchi: I propose:

- `buildResourcesFromAssemblers` --> `buildAssemblerResourcesBeforeRules` 
- `buildOthers` --> `buildAssemblerResourcesAfterRules`

gitgabrio: > I propose:
> 
> * `buildResourcesFromAssemblers` --> `buildAssemblerResourcesBeforeRules`
> * `buildOthers` --> `buildAssemblerResourcesAfterRules`

That would be great, thanks!!!

hellowdan (approved):

mariofusco (approved):

evacchi:
jenkins run cdb

evacchi:
jenkins run cdb

gitgabrio (approved):
@evacchi 
LGTM - thx!

PR #3344, opened 2021-01-26
[DROOLS-5964] Implement PMML-Trusty/Scesim backend integration

gitgabrio:
@danielezonca @yesamer  @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-5964

Please let me know if `RuntNotifier` `RunListener` usage inside `PMMLRunnerTest` could be improved.

Needed by https://github.com/kiegroup/drools-wb/pull/1458


drools-scenario-simulation/drools-scenario-simulation-api/src/main/java/org/drools/scenariosimulation/api/model/ScenarioSimulationModel.java
```diff
@@ -25,11 +25,12 @@
...
     public enum Type {
         RULE,
-        DMN
+        DMN,
+        PMML
     }
 
     @XStreamAsAttribute()
-    private String version = "1.8";
+    private String version = "1.9";
```
danielezonca, line 33: @yesamer 
Do you think this is a problem for client side marshaller or in general do you expect any impact on Kogito/BC side?

yesamer:
Hi @gitgabrio thank you for your PR.
Before review it, I think we need to discuss how to manage the impact of this PR on Kogito version of the editor.
Please consider that Kogito version of the editor currently supports version **1.8** only. This means that, after merging this PR, all scesim files managed in BC will be updated to 1.9 and therefore no more compatible with Kogito. This is a scenario we should avoid. My suggestion is to include a PR which enables the kogito scesim marshaller to manage at least version 1.8 and 1.9.

gitgabrio:
> Hi @gitgabrio thank you for your PR.
> Before review it, I think we need to discuss how to manage the impact of this PR on Kogito version of the editor.
> Please consider that Kogito version of the editor currently supports version **1.8** only. This means that, after merging this PR, all scesim files managed in BC will be updated to 1.9 and therefore no more compatible with Kogito. This is a scenario we should avoid. My suggestion is to include a PR which enables the kogito scesim marshaller to manage at least version 1.8 and 1.9.

Thanks @yesamer !

gitgabrio:
@danielezonca @yesamer 
Switched to draft because it also need modifications on drools-wb.
Please let me know if you prefer to keep as draft until someone may complete the client-side part.

drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/fluent/PMMLScenarioExecutableBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+import org.kie.pmml.evaluator.core.PMMLContextImpl;
+import org.kie.pmml.evaluator.core.utils.PMMLRequestDataBuilder;
+
+public class PMMLScenarioExecutableBuilder {
+
+    public static final String DEFAULT_APPLICATION = "defaultApplication";
+    public static final String PMML_RESULT = "pmmlResult";
+    public static final String PMML_MODEL = "pmmlModel";
+
+    private static final String SEPARATOR = FileSystems.getDefault().getSeparator();
```
danielezonca, line 41: I think it is better to normalise the path to use `/` instead of handling platform specific separator: with this code I expect that if the scesim file is created on Linux and executed then on a Windows machine it will fail

drools-scenario-simulation/drools-scenario-simulation-backend/src/test/java/org/drools/scenariosimulation/backend/PMMLRunnerTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private final String PMML_FILE = "CategoricalVariablesRegression.pmml";
+    private final String SCESIM_FILE = "scesim-pmml.scesim";
+
+    @Test
+    public void runSuccessScenario() throws Exception {
+        KieContainer kieContainer = getKieContainer();
+        ScenarioRunnerDTO scenarioRunnerDTO = getScenarioRunnerDTO();
+        PMMLScenarioRunner runner = new PMMLScenarioRunner(kieContainer, scenarioRunnerDTO);
+        runner.run(getRunNotifier());
```
danielezonca, line 48: What about adding an assertion on the result also when it succeed?

gitgabrio:
Suspended - waiting for better indications

PR #3360, opened 2021-02-05
[DROOLS-5995] beta indexes for constraints depending from more than o…

mariofusco:
…ne declaration of the left tuple

**JIRA**: https://issues.redhat.com/browse/DROOLS-5995

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
@@ -948,15 +948,15 @@ public class PatternDSLTest {
 
     @Test
     public void testBetaIndexOn2ValuesOnLeftTuple() {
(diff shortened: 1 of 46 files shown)


lucamolteni (approved):
LGTM, can we add also an example of usage of such index in PatternDSLTest?

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PatternDSLTest.java
```diff
@@ -945,4 +945,38 @@ public void testTwoAccumulatesWithVarBindingModel() {
         session.insert(new Person("Lukas", 35));
         session.fireAllRules();
     }
+
+    @Test
+    public void testBetaIndexOn2ValuesOnLeftTuple() {
+        final Variable<Integer> var_GENERATED_$pattern_Integer$3$ = D.declarationOf(Integer.class, "GENERATED_$pattern_Integer$3$");
```
lucamolteni: Rename machine-generated variables to human-readable plase

lucamolteni (approved):

PR #3369, opened 2021-02-09
[DROOLS-5995] rename BetaIndex interface to keep executable model bac…

mariofusco:
…kward compatible + add BetIndex with 4-arity


triceo (approved):
I've seen the referenced downstream PR pass locally.

mariofusco:
The 2 test failures are unrelated and should be fixed by this #3370

PR #3387, opened 2021-02-16
[DROOLS-6031] fix filtering on group key

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6031


triceo (approved):
Works with downstream OptaPlanner master.
Integrating my own changes will wait until the other bugs are fixed.

PR #3392, opened 2021-02-18
[DROOLS-6031] fix pattern binding of an accumulate result

mariofusco:


triceo (approved):
Solves the problem.

triceo:
Tested the latest changes and all the reported issues are fixed.
We will handle nested groupBy another time.
Good to merge.

PR #3402, opened 2021-02-23
[DROOLS-6041] "RHS doesn't have a type" error when a semicolon is mis…

tkobayas:
…sing with exec-model and mvel dialect

- WIP : just to spot the issue

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-6041

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
@@ -925,4 +925,28 @@ public class MvelDialectTest extends BaseModelTest {
 
         assertTrue(result.contains("R"));
     }
+
+    @Test
+    public void testLineBreakAtTheEndOfStatementWithoutSemicolon() {
+        final String str =
+                "import " + Person.class.getCanonicalName() + ";\n" +
+                           "\n" +
+                           "rule R\n" +
+                           "dialect \"mvel\"\n" +
+                           "when\n" +
+                           "  Person(name == \"Mario\")\n" +
+                           "then\n" +
+                           "  Person p2 = new Person(\"John\");\n" +
+                           "  p2.age = 30\n" + // a line break at the end of the statement without a semicolon
+                           "  insert(p2);\n" +
+                           "end";
+
+        KieSession ksession = getKieSession(str);
+
+        Person p = new Person("Mario", 40);
+        ksession.insert(p);
+        int fired = ksession.fireAllRules();
+
+        assertEquals(1, fired);
+    }
 }
--- a/drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
+++ b/drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
@@ -91,6 +91,7 @@ import com.github.javaparser.JavaToken;
 @Generated("JavaCC")
 public final class GeneratedMvelParser extends GeneratedMvelParserBase {
     private boolean optionalSemicolon = true;
+    private int nestedBlocks = 0;
     private boolean isWithOrModify = false;
 
     public void setOptionalSemicolon(boolean optionalSemicolon) {
@@ -110,14 +111,16 @@ public final class GeneratedMvelParser extends GeneratedMvelParserBase {
 
     }
 
-    public boolean doesCrossLineBreak(Token token) {
-        if (token.specialToken == null) {
-            return false;
-        }
-        if (token.specialToken.kind == UNIX_EOL || token.specialToken.kind == WINDOWS_EOL) {
-            return true;
-        }
-        return doesCrossLineBreak(token.specialToken);
+    public boolean inBlock() {
+        return nestedBlocks > 0;
+    }
+
+    public void enterBlock() {
+        this.nestedBlocks++;
+    }
+
+    public void leaveBlock() {
+        this.nestedBlocks--;
     }
 
     /* Returns the JavaParser specific token type of the last matched token */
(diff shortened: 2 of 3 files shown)


tkobayas:
@evacchi  @mariofusco 

DROOLS-5879 (https://github.com/kiegroup/drools/commit/e37605f2b77b0714ea3018684eef85934ba2e5de) changed the newline handling so we can parse an expression across a newline. It solved the issue in RHDM-1550, which has RHS like this. (`MvelDialectTest#testSetOnMvel()`)

```
        $fact.method2("message1", "message2"
);
```

On the other hand, in DROOLS-6041, RHS like this (see that the second line doesn't have a semicolon) fails in exec-model because the parser treats "30 insert p2" as a `PointFreeExpr`. standard-drl passes the test because it treats the newline as a delimiter. (`MvelDialectTest#testLineBreakAtTheEndOfStatementWithoutSemicolon`). `MvelParser.optionalSemicolon` doesn't affect this behaviour.

```
  Person p2 = new Person("John");
  p2.age = 30
  insert(p2);
```

This draft PR is just to spot the issue and start a discussion. I added `doesCrossLineBreak()` to pass the test but it's a very patchy approach and there should be a better way.

I think the key point is that whether we should have EOL as TOKEN or SPECIAL_TOKEN.

If EOL is TOKEN, we consider the baseline is "a newline is a delimiter". I think it's more closer to the standard-drl mvel. Then we would fix issues like RHDM-1550 on case-by-case basis. 

If EOL is SPECIAL_TOKEN, we consider the baseline is "a newline is not a delimiter". I think it's more closer to languages like java. Then we would fix issues like DROOLS-6041 on case-by-case basis. 

I may have not yet understood the parser behaviour well. Feel free to share your thoughts on this. Thanks!

evacchi:
EOL as a token was the previous approach, but it polluted quite a bit the code of the parser. With this approach, similar to what happens in [some JS grammars](https://github.com/antlr/grammars-v4/blob/master/javascript/ecmascript/JavaScript/ECMAScript.g4#L727),  we are using a conditional lookahead; if a newline is found then it should automatically "terminate" the line. 

see https://github.com/kiegroup/drools/blob/8f2dce45d88695c36037a4185ec6c326ffa8323a/drools-model/drools-mvel-parser/src/main/javacc/mvel.jj#L3497

```
void EOS():
{
}
{
   <SEMICOLON> | <EOF> | LOOKAHEAD( { this.shouldTerminateStatement() } ) {}
}
```

unfortunately it looks like it breaks in this case 🤔 

EDIT: here is a minimal test case:

```
        @Test
        public void testLineBreakAtTheEndOfStatementWithoutSemicolon() {
            String expr =
                    "{  Person p2 = new Person(\"John\");\n" +
                    "  p2.age = 30\n" + // a line break at the end of the statement without a semicolon
                    "insert(p2);\n }";

            MvelParser mvelParser = new MvelParser(new ParserConfiguration(), true);
            ParseResult<BlockStmt> r = mvelParser.parse(GeneratedMvelParser::BlockParseStart, new StringProvider(expr));
            BlockStmt blockStmt = r.getResult().get();
            assertEquals("Should parse 3 statements", 3, blockStmt.getStatements().size());

        }
```

if we look closer, we'll see that the third statement is erroneously merged with the second:

```
{
    Person p2 = new Person("John");
    p2.age = 30 insert p2;
}
```

and in fact, as you noticed, the insert statement is parsed as the PointFreeExpr `30 insert p2` which is most probably _not_ what we want; IIRC PointFreeExpr make sense only on the LHS;

It kind of suggest that a fix may be to add boolean check if the parser is parsing the LHS or the RHS 🤔 

EDIT2: the latter proposed change (a boolean check) would be kind of disruptive for the usages of the grammar -- but IMO it may be still worth revisiting in the future. Assuming it does not break other tests, the check you added can be substituted for the already present `shouldTerminateStatement()`; i.e.:

```
            left = PrimaryExpression()
            {
                if (shouldTerminateStatement()) {
                    return left;
                }
```
 
~~which does not feel that hacky after all IMO~~ unfortunately I don't think this works, but breaks a lot of other stuff


EDIT3:

TBF I don't think point-free expressions should be handled at all in the consequence; now, because a consequence is always a BlockStmt, and because we may assume that a pointFreeExpr is never correct within a BlockStmt; then we can *short-circuit* and "cut" the PointFreeExpr branch -- as you @tkobayas did -- but only when the parsing occurs within a BlockStmt. 

- we add a boolean(*) for checking if we are in a BlockStmt; 
- if we are in a BlockStmt, then we short-circuit because it can't be a pointFree expr
- the boolean because is flipped when we leave the block stmt -> everything should work

(*) except it can't be a boolean, because BlockStmt can nest; so we add a counter instead; same effect


fix here https://github.com/evacchi/drools/commit/53f007038161d9b1d8ec0bd160d8b1fbd3f57f31

tkobayas:
Thank you very much, @evacchi ! I agree that short-circuiting PointFreeExpr in BlockStmt makes sense.

tkobayas:
Failed with 3 tests. I believe they are not related to this PR. But just in case, one more time.
```
org.drools.compiler.integrationtests.TimerAndCalendarTest.testIntervalRuleInsertion[KieBase type=STREAM_IDENTITY]
org.drools.compiler.integrationtests.TimerAndCalendarTest.testTimerWithNot[KieBase type=STREAM_IDENTITY]
org.drools.mvel.integrationtests.KieContainerTest.testIncrementalCompilationSynchronization
```

tkobayas:
Jenkins retest this

evacchi (approved):

mariofusco (approved):

hellowdan (approved):

PR #3405, opened 2021-02-23
[DROOLS-6048] Enforcing before rules/after rules semantic

gitgabrio:

See https://issues.redhat.com/browse/DROOLS-6048

Scope of this PR is to enforce/clarify the "before rules/after rules" semantic introduced by DROOLS-5950.
In detail:

1. renamed some methods for consistency
2. uniformed methods
3. replaced addResource/addResources with addResourceAfterRules / addResourcesAfterRules

@tarilabs @evacchi FYI

Depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/503

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
@@ -111,7 +111,7 @@ public class CompositeKnowledgeBuilderImpl implements CompositeKnowledgeBuilder 
             kBuilder.buildPackagesWithoutRules(packages);
         }
         buildProcesses();
-        buildAssemblersAfterRules();
+        buildAssemblerResourcesAfterRules();
         kBuilder.postBuild();
         resourcesByType.clear();
         if (buildException != null) {
@@ -153,19 +153,6 @@ public class CompositeKnowledgeBuilderImpl implements CompositeKnowledgeBuilder 
                     kBuilder.setAssetFilter(null);
                 }
             }
-        }
-    }
-
-    private void buildAssemblersAfterRules() {
-        try {
-            for (Map.Entry<ResourceType, List<ResourceDescr>> entry : resourcesByType.entrySet()) {
-                List<ResourceWithConfiguration> rds = entry.getValue().stream().map(CompositeKnowledgeBuilderImpl::descrToResourceWithConfiguration).collect(Collectors.toList());
-                kBuilder.addPackageForExternalType(entry.getKey(), rds);
-            }
-        } catch (RuntimeException e) {
-            throw e;
-        } catch (Exception e) {
-            throw new RuntimeException( e );
         }
     }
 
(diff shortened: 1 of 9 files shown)


tarilabs:
Thank you for the FYI.
I didn't understand if the test which you mentioned previously error-ed on your work-in-progress, now are again green after the full set of code change?

gitgabrio:
Jenkins run cdb

gitgabrio:
> Thank you for the FYI.
> I didn't understand if the test which you mentioned previously error-ed on your work-in-progress, now are again green after the full set of code change?

Looking at the code I had I could not find a reason for that error. I tried to debug a little, but since it was taking too much time,  I've restarted such renaming from scratch - with a slightly different approach - and the error is gone.

tarilabs:
> I've restarted such renaming from scratch - with a slightly different approach - and the error is gone.

👍  glad to hear you achieved the desired result, possibly it was just the IDE missed some update during the 1st round of refactoring.

danielezonca (approved):

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
```diff
@@ -215,6 +202,19 @@ private void buildAssemblerResourcesBeforeRules() {
...
     }
 
+    private void buildAssemblerResourcesAfterRules() {
+        KieAssemblers assemblers = ServiceRegistry.getService(KieAssemblers.class);
+        try {
+            for (Map.Entry<ResourceType, List<ResourceDescr>> entry : resourcesByType.entrySet()) {
+                List<ResourceWithConfiguration> rds = entry.getValue().stream().map(CompositeKnowledgeBuilderImpl::descrToResourceWithConfiguration).collect(Collectors.toList());
+                assemblers.addResourcesAfterRules(kBuilder, rds, entry.getKey());
+            }
+        } catch (RuntimeException e) {
```
jiripetrlik, line 212: This looks quite unusual to me. Can we for example unify both catch blocks and always throw Runtime exception with reasonable description and original exception inside? Description might be something like "Building assembler resources failed".

jiripetrlik (commented):
Looks good. I have one suggestion.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/CompositeKnowledgeBuilderImpl.java
gitgabrio: @jiripetrlik 
Hi! Scope of this PR is just plain renaming for sake of improved readability.
Behind those two different exception management (buildAssemblerResourcesBeforeRules and buildAssemblerResourcesAfterRules) there is the way how the whole flow is managed and the exception are used (pls look inside CompositeKnowledgeBuilderImpl.build(boolean buildRules))
I can't exactly follow why they are so different, but it is a little bit out of scope from this PR.

gitgabrio:
@danielezonca Hi! Could you please merge this, squashing commits? Thanks!

PR #3411, opened 2021-02-25
[DROOLS-6053] Reduce unnecessary classloading by parent classloader

tkobayas:
**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6053

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core-reflective/src/main/java/org/drools/reflective/classloader/ProjectClassLoader.java
+++ b/drools-core-reflective/src/main/java/org/drools/reflective/classloader/ProjectClassLoader.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


drools-core-reflective/src/main/java/org/drools/reflective/classloader/ProjectClassLoader.java
```diff
@@ -145,6 +148,10 @@ public static ProjectClassLoader createProjectClassLoader(ClassLoader parent, Ma
         try {
             return super.loadClass(name, resolve);
         } catch (ClassNotFoundException e) {
+            if (ownedClassNameSet.contains(name)) {
+                // No need to check parent classloader
+                throw e;
+            }
             return Class.forName(name, resolve, getParent());
```
mariofusco, line 155: @tkobayas I see what you did, but I'm not particularly happy of adding another Set of class names to the ProjectClassLoader. What's missing here is that when this `Class.forName` on the parent fails it doesn't add the name of the class to the `nonExistingClasses` Set. Wouldn't this be enough to solve or at least hugely mitigate the problem?

tkobayas: @mariofusco  It would help only when the class is repeatedly loaded, right? If there are tons of generated classes, those classes will still be examined by the parent classloader at least once.

Anyway I can take the approach and check with the customer how much it mitigates the problem. Thanks!

tkobayas:
superseded by https://github.com/kiegroup/drools/pull/3415

PR #3428, opened 2021-03-08
[DROOLS-6052] avoid to build assembler based resources when creating …

mariofusco:
…a KieBase from the executable model

**JIRA**: https://issues.redhat.com/browse/DROOLS-6052

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
@@ -77,7 +77,7 @@ public class ModelBuilderImpl<T extends PackageSources> extends KnowledgeBuilder
     }
 
     @Override
-    protected void processPackages(Collection<CompositePackageDescr> packages) { }
+    protected void doFirstBuildStep( Collection<CompositePackageDescr> packages) { }
 
     @Override
     public void addPackage(final PackageDescr packageDescr) {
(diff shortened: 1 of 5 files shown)


mariofusco:
jenkins execute fdb

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
```diff
@@ -78,7 +78,7 @@ public ModelBuilderImpl(Function<PackageModel, T> sourcesGenerator, KnowledgeBui
     }
 
     @Override
-    public void buildPackages(Collection<CompositePackageDescr> packages) {
+    public void processPackages(Collection<CompositePackageDescr> packages) {
```
gitgabrio: As reminder from our private conversation, I think it would be clearer to follo java naming convention for this method that actually is a setter and, IMO, it should not override the parent one (of course I do not know all the implications of that)

gitgabrio (commented):
As per private conversation, the main issue I see is that client code (e.g. PMML) has no knowledge of "KnowledgeBuilderImpl" children, so a public method is needed that regardless of actual instance (KnowledgeBuilderImpl vs any of its children) actully execute the build of the packages.

mariofusco:
jenkins execute fdb

mariofusco:
jenkins execute fdb

gitgabrio (commented):
@mariofusco 
As reminder for our private conversation, the only little bit to change, IMO, it is the naming of those two methods "processPackages" and "postBuild": it would be useful to have names  clearly explaining that the two are two differents "build", invoked on one or the other implementation. A very small javadoc would also help.

mariofusco:
jenkins execute fdb

gitgabrio (approved):
LGTM, thanks!

PR #3436, opened 2021-03-11
[JBPM-9656] Provide configuration parameters for OptimisticLockRetryInterceptor

elguardian:
https://issues.redhat.com/browse/JBPM-9656


elguardian:
Jenkins retest this

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/OptimisticLockRetryInterceptor.java
```diff
@@ -45,9 +45,9 @@
 
     private static final Logger logger = LoggerFactory.getLogger(OptimisticLockRetryInterceptor.class);
 
-    private int retries = 3;
-    private long delay = 50;
-    private long delayFactor = 4;
+    private int retries = Integer.getInteger("org.kie.optlock.retries", 5);
```
afalhambra: why are we increasing the number of retries? probably there won't be any side effect out of this (thinking about backward compatibility mainly), but just curious why we don't keep the same number of retries?

afalhambra (approved):

PR #3439, opened 2021-03-11
[DROOLS-6079] fix memory leak in groupBy

mariofusco:


triceo (approved):
Verified that the vaccination quickstart leak is fixed.
Verified that no other OptaPlanner test is failing.
Did not run any performance benchmarks.

PR #3446, opened 2021-03-15
[DROOLS-6086] allow direct firing from rule terminal node

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6086

### DIFF
--- a/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
+++ b/drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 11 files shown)


drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
```diff
@@ -231,9 +231,6 @@
 
     private Environment environment;
 
-    // this is a counter of concurrent operations happening. When this counter is zero,
-    // the engine is idle.
-    private AtomicLong opCounter;
```
triceo, line 236: So does `DirectFiring.YES` also enable this non-thread-safe session?
Because I don't see any other switch to enable it separately.

mariofusco: @triceo no, that's opCounter is totally unrelated with both non-thread safe session and direct firing. We introduced it ages ago as, but in practice we never used it and it was also harmful for perfs, so I agreed with @mdproctor to just no-op the session's start and end operation methods.

triceo (approved):

drools-core/src/main/java/org/drools/core/impl/StatefulKnowledgeSessionImpl.java
mariofusco: @triceo I was wrong and that `opCounter` was actually used in some cep and jpa use cases. I restored it and will eventually make it optional with the non-thread safe ksession work.

lucamolteni (approved):

PR #3450, opened 2021-03-16
[DROOLS-6087] add KieSession option to flag it as non-thread-safe

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6087


lucamolteni (approved):

triceo (approved):

PR #3453, opened 2021-03-17
[DROOLS-6096] disable property reactivity for updates with unknown pr…

mariofusco:
…operties in mvel

**JIRA**: https://issues.redhat.com/browse/DROOLS-6096


lucamolteni (approved):

lucamolteni:
was there a regression?

mariofusco:
> was there a regression?

The behaviour appears to be a regression for that specific use case. In reality we introduced a more fine grained property reactivity analysis but in mvel dialect forgot to disable property reactivity when there is an unknown property as we did in java one.

PR #3456, opened 2021-03-17
[DROOLS-6094] Fix package name declaration from files containing dash in the name 

gitgabrio:

see https://issues.redhat.com/browse/DROOLS-6094 https://issues.redhat.com/browse/RHPAM-3521


jiripetrlik (approved):
@gitgabrio 
Thank you for quick fix. Looks good to me!

kie-pmml-trusty/kie-pmml-commons/src/main/java/org/kie/pmml/commons/utils/KiePMMLModelUtils.java
```diff
@@ -27,7 +27,7 @@ private KiePMMLModelUtils() {
      * @return
      */
     public static String getSanitizedPackageName(String modelName) {
-        return modelName.replace(" ", "").replace("_", "").toLowerCase();
+        return modelName.replace(" ", "").replace("-", "").replace("_", "").toLowerCase();
```
danielezonca, line 30: What about "clone" the same logic of this method?
https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CodegenStringUtil.java#L31-L50

danielezonca (approved):
Thanks for the PR, approving with minor suggestion

kie-pmml-trusty/kie-pmml-commons/src/main/java/org/kie/pmml/commons/utils/KiePMMLModelUtils.java
gitgabrio: I prefer to not have "_" or similar in packagename/classname. And the logic of that method is different of what is needed here

PR #3464, opened 2021-03-19
[DROOLS-6179] optimize LambdaConsequence

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6179


triceo (approved):

lucamolteni (approved):

tkobayas (approved):

hellowdan (approved):

triceo:
I ran my benchmark on this new code and I'm seeing some interesting results.

Before this, on Thursday, I got CS-D 18 % behind CS-B.
With this PR, I am now consistently getting CS-D only 10 % behind.
Good stuff!

Additional observation: CH seems to be slower at first (20 % behind) and then LS picks up, speeds up and brings us to only 10 % behind. Therefore if we're looking for further improvements, I'd say CH is the place to look.

PR #3465, opened 2021-03-21
[DROOLS-6064] Inconsistent behavior when accumulate function returns …

tkobayas:
…null

- WIP
- Not yet worked on exec-model side

https://issues.redhat.com/browse/DROOLS-6064

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/AgendaItem.java
+++ b/drools-core/src/main/java/org/drools/core/common/AgendaItem.java
@@ -16,6 +16,7 @@
 package org.drools.core.common;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
@@ -25,6 +26,7 @@ import org.drools.core.beliefsystem.simple.SimpleMode;
 import org.drools.core.phreak.RuleAgendaItem;
 import org.drools.core.reteoo.LeftTuple;
 import org.drools.core.reteoo.TerminalNode;
+import org.drools.core.rule.SingleAccumulate;
 import org.drools.core.spi.Activation;
 import org.drools.core.spi.PropagationContext;
 import org.drools.core.spi.Tuple;
@@ -75,7 +77,12 @@ public interface AgendaItem<T extends ModedAssertion<T>> extends Activation<T> {
             if ( entry.getFactHandle() != null ) {
                 Object o = ((InternalFactHandle) entry.getFactHandle()).getObject();
                 if (!(o instanceof QueryElementFactHandle || o instanceof InitialFact)) {
-                    list.add(o);
+                    if (SingleAccumulate.ENABLE_D6064 && o instanceof Object[]) {
+                        List<Object> tmpList = Arrays.asList((Object[])o);
+                        list.addAll(tmpList);
+                    } else {
+                        list.add(o);
+                    }
                     list.addAll( entry.getAccumulatedObjects() );
                 }
             }
--- a/drools-core/src/main/java/org/drools/core/rule/SingleAccumulate.java
+++ b/drools-core/src/main/java/org/drools/core/rule/SingleAccumulate.java
@@ -36,6 +36,9 @@ import org.kie.internal.security.KiePolicyHelper;
 
 public class SingleAccumulate extends Accumulate {
     private Accumulator accumulator;
+    private boolean resultAsObjectArray;
+
+    public static boolean ENABLE_D6064 = true;
 
     public SingleAccumulate() { }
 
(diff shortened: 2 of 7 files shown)


drools-core/src/main/java/org/drools/core/rule/SingleAccumulate.java
```diff
@@ -36,6 +36,9 @@
 
 public class SingleAccumulate extends Accumulate {
     private Accumulator accumulator;
+    private boolean resultAsObjectArray;
+
+    public static boolean ENABLE_D6064 = true;
```
tkobayas, line 41: This is a temporal flag for development use

drools-mvel/src/main/java/org/drools/mvel/java/JavaAccumulateBuilder.java
```diff
@@ -182,12 +182,27 @@ private Accumulate buildExternalFunctionCall( RuleBuildContext context,
...
+            if (SingleAccumulate.ENABLE_D6064) {
+                if (pattern.getObjectType().getClassType().equals(Object.class) && !fc.isUnification()) {
+                    // implicit Pattern by accumulate is 'Object'. In this case, we return Object[] (same as MultiAccumulate) so it can contain null
+                    InternalReadAccessor reader = new SelfReferenceClassFieldReader( Object[].class );
+                    bindReaderToDeclaration(context, accumDescr, pattern, fc, new ArrayElementReader(reader, 0, function.getResultType()), function.getResultType(), -1);
+                    resultAsObjectArray = true;
+                } else {
+                    // from accumulate: Returns specified Type
+                    // unification: Requires original readAccessor
+                    bindReaderToDeclaration(context, accumDescr, pattern, fc, new SelfReferenceClassFieldReader( function.getResultType() ), function.getResultType(), -1);
```
tkobayas, line 195: In case of "accumulate", the implicit Pattern is `Object`. So we can return `Object[]` just like `MultiAccumulate` does.
Exceptional case:
- "from accumulate" has its specific Pattern (e.g. `Number`). In this case, we don't propagate 'null'. Right? (Maybe another inconsistency though)
- in case of unification, we need to use the original `readAccesor`. So we cannot use `Object[]` approach. I haven't yet investigated deeper if it causes a problematic scenario.

drools-core/src/main/java/org/drools/core/common/AgendaItem.java
```diff
@@ -75,7 +77,12 @@
             if ( entry.getFactHandle() != null ) {
                 Object o = ((InternalFactHandle) entry.getFactHandle()).getObject();
                 if (!(o instanceof QueryElementFactHandle || o instanceof InitialFact)) {
-                    list.add(o);
+                    if (SingleAccumulate.ENABLE_D6064 && o instanceof Object[]) {
+                        List<Object> tmpList = Arrays.asList((Object[])o);
+                        list.addAll(tmpList);
+                    } else {
+                        list.add(o);
+                    }
```
tkobayas, line 85: Here we explode `Object[]` for `getObjectsDeep()`. But it doesn't guarantee that the fact `o` is the result of accumulate. So it's not precise logic. Should we introduce a specific Class (e.g. AccumulateResultFact) instead of `Object[]` ?

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3790,8 +3791,10 @@ public void testAverageWithNoFacts() throws Exception {
         list.clear();
 
         kieSession.delete( fh );
-        assertEquals(0, kieSession.fireAllRules() );
-        assertEquals(0, list.size() );
+        // changed by DROOLS-6064
+        assertEquals(1, kieSession.fireAllRules() );
+        assertEquals(1, list.size() );
+        assertNull(list.get(0));
```
tkobayas, line 3797: This test was introduced by DROOLS-2595. At that time, "not fire" was intentional. But now we change the behavior.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/PropertySpecificTest.java
```diff
@@ -2704,7 +2704,8 @@ public void testAccBigDecimal() {
                 "import " + BigDecimalFact.class.getCanonicalName() + "\n" +
                 "rule R when\n" +
                 "        accumulate( BigDecimalFact( $bdVal: bdVal), $minVal : min($bdVal))\n" +
-                "        accumulate( BigDecimalFact( $bdVal2: bdVal, $bdVal2 > $minVal), $minVal2 : min($bdVal2))\n" +
+                "        accumulate( BigDecimalFact( $bdVal2: bdVal, $bdVal2 > $minVal), $minVal2 : min($bdVal2); $minVal2 != null)\n" + // guard for DROOLS-6064
```
tkobayas, line 2707: If this change breaks existing rules, we will need to add a null check like this.

tkobayas:
Superseded by https://github.com/kiegroup/drools/pull/3479

PR #3466, opened 2021-03-22
[DROOLS-6175] Change CanonicalKieModule metadata path from a dot cont…

tkobayas:
…aining directory to standard package structure

**JIRA**:
https://issues.redhat.com/browse/DROOLS-6175

**referenced Pull Requests**: 
* https://github.com/kiegroup/droolsjbpm-integration/pull/2432

<details>
<summary>
</summary>

</details>


tkobayas:
Jenkins run fdb

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModuleProvider.java
```diff
@@ -50,7 +50,7 @@ private InternalKieModule createCanonicalKieModule( InternalKieModule internalKi
             return new CanonicalKieModule(internalKieModule);
         } else {
             if (log.isInfoEnabled()) {
-                log.info( "No executable model found for artifact " + internalKieModule.getReleaseId() + ". Falling back to resources parsing." );
+                log.info( "No executable model found for artifact " + internalKieModule.getReleaseId() + ". Falling back to resources parsing. If you built a executable model kjar with older drools version, please build it again with the same drools version which you load with." );
```
tkobayas: Note that because of this change, "drools-model" file in kjars built by older drools versions will not be found . This log message is good enough?

lucamolteni: I'm not sure we should break backward compatibility here, let me add @mariofusco  in the loop

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModuleProvider.java
gitgabrio: @tkobayas 
Could you add a "backward" method to retrieve file generated in old ways if not found in "new" way ?
I mean: look for file without dot, if they are not found, look for file with dot.

tkobayas:
Jenkins run fdb

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModuleProvider.java
tkobayas: @gitgabrio Yeah, I was a little wondering if I should add it. You encouraged me :)

tkobayas: So now we don't break backward compatibility

hellowdan (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/CanonicalKieModule.java
```diff
@@ -454,7 +464,19 @@ private void initModels(CanonicalKieModuleModel kmodel) throws ClassNotFoundExce
 
     private Collection<String> findRuleClassesNames() {
         ReleaseId releaseId = internalKieModule.getReleaseId();
-        String modelFiles = readExistingResourceWithName(getModelFileWithGAV(releaseId));
+        String modelFiles;
+
+        String modelFilePath = getModelFileWithGAV(releaseId);
```
gitgabrio, line 469: @tkobayas 
Forgive me, be patient! Would it be possible to create a getModelFileWIthGav method that transparently return one or 
the other (something similar to
`resourceFileExists(getModelFileWithGAV(releaseId)) || resourceFileExists(getModelFileWithGAV(releaseId))`
)
and then use that string ? 
It would remove the needs of that `if` and improve readability and - I think - testability

mariofusco (commented):
I'm sorry but I see no valid reason for this change. In other words I see only a big con that isn't compensated by any evident pro.

The big con is that with this PR we will have to deal with 2 different folder formats for backward compatibility reasons and I think this will be very error-prone in future. Already now that `getModelFileWithGAV` method seems to be used in a [couple of places ](https://github.com/kiegroup/drools/search?q=getModelFileWithGAV) but this PR seems to address only one of the two.

Conversely the [supposed pros](https://github.com/kiegroup/droolsjbpm-integration/pull/2430#discussion_r597501868) of this change seems to be quite weak. Yes it's true that "splits on dot when managing directory/file names is a pretty common practice" but indeed it is just a practice, there isn't anything that mandates this. Moreover how this file position looks inside an IDE is totally irrelevant since this is just a text file that is written inside the generated jar. 

In essence if we were developing this thing now for the first time, I'd agree of avoiding dots in the folder name for the convention mentioned by @gitgabrio, but introducing it now is a totally different story because it solves a problem that is (at most) theoretical by introducing another problem that conversely is very practical.

Sorry again, but I'm STRONGLY against this change.

tkobayas:
Hi @gitgabrio  , Do you agree with @mariofusco ? TBH, I agree that the con is larger and I'm fine that this PR will not be merged (Sorry about my indecisiveness!)

gitgabrio:
> Hi @gitgabrio , Do you agree with @mariofusco ? TBH, I agree that the con is larger and I'm fine that this PR will not be merged (Sorry about my indecisiveness!)

@mariofusco @tkobayas 
Nope, I disagree with both of you, for the following reasons
1) the "con" that Mario points out depends on the fact that the management of two different folders should have been hidden inside one single method, so that it would be transparent for the rest of the code (this was the reason of my suggestion)
2) we can't always accumulate small or big issues hoping/waiting for something that allows us to make breaking change
3) we should not keep working pretending there is no-one else out there that is using our code, and that eventually does exactly what we expect: it does not work that way
4) we are plenty of such technical debt, and we should start to tackle them, maybe beginning with small one (like that) and then proceeding with bigger ones

The issue I discovered on the maven-plugin (related of the javadoc generation and due to the "naive" management of the generated-source directory ) is an example of that, surely not the first nor the only I found.

Having said that, at the end of the day, it all resolve around personal mindset, experience, and idea about what there should be, and what not, in a project like that.

mariofusco:
@gitgabrio I agree with all you wrote, except with the fact that having a dot in the name of a folder is perfectly normal and NOT to be considered a bug.

PR #3468, opened 2021-03-22
[DROOLS-6180] use default java compiler (ecj one when available) also…

mariofusco:
… in executable model

**JIRA**: https://issues.redhat.com/browse/DROOLS-6180

### DIFF
--- a/drools-compiler/pom.xml
+++ b/drools-compiler/pom.xml
@@ -41,6 +41,10 @@
       <artifactId>kie-memory-compiler</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-ecj</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-api</artifactId>
     </dependency>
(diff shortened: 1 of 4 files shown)


gitgabrio (commented):
@mariofusco 
Thanks for the PR. At first sight it looks good, but I would need to test it in my modified code and right now I can't do: could you please wait that (hopefully today or tomorrow) ? Thanks!

drools-compiler/pom.xml
```diff
@@ -40,10 +40,6 @@
       <groupId>org.kie</groupId>
       <artifactId>kie-memory-compiler</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-ecj</artifactId>
-    </dependency>
```
danielezonca: Why this change? As far as I know ecj is the default and should be added by default

gitgabrio (approved):
@mariofusco 
It works as expected, LGTM. Thanks!

danielezonca (approved):

mariofusco:
jenkins execute fdb

mariofusco:
The only test failure is in `org.kie.server.springboot.samples.KeycloakKieServerTest.org.kie.server.springboot.samples.KeycloakKieServerTest` which is clearly unrelated with this change.

PR #3467, opened 2021-03-22
[DROOLS-6183] Refactor DATA_TYPE.getActualValue. Move ConverterTypeUtil

gitgabrio:
@danielezonca @mariofusco @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-6183

Basically, this PR is just a small refactoring of DATA_TYPE method that need moving of ConverterTypeUtil.

@kostola  FYI

PR #3469, opened 2021-03-22
[DROOLS-6176] Fix missing node score in rule generation for tree model

gitgabrio:

See https://issues.redhat.com/browse/DROOLS-6176

Some of the changed files are actually coming from https://github.com/kiegroup/drools/pull/3467

@kostola @tteofili FYI


kie-pmml-trusty/kie-pmml-models/kie-pmml-models-mining/kie-pmml-models-mining-tests/pom.xml
```diff
@@ -76,6 +76,10 @@
       <scope>test</scope>
     </dependency>
     <!-- KIE -->
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-ecj</artifactId>
```
danielezonca, line 81: Why do you need this dependency explicitly? Isn't already available as transitive dependency?

kie-pmml-trusty/kie-pmml-evaluator/kie-pmml-evaluator-core/src/main/java/org/kie/pmml/evaluator/core/service/PMMLRuntimeInternalImpl.java
```diff
@@ -159,6 +163,23 @@ protected void executeTransformations(final KiePMMLModel model, final PMMLContex
...
+    /**
+     * Verify that the returned value has the required type as defined inside <code>DataDictionary/MiningSchema</code>
+     * @param toUpdate
+     */
+    void updateTargetValueType(final KiePMMLModel model, final PMML4Result toUpdate) {
+        DATA_TYPE dataType = model.getMiningFields().stream()
+                .filter(miningField -> model.getTargetField().equals(miningField.getName()))
+                .map(MiningField::getDataType)
+                .findFirst()
+                .orElseThrow(() -> new KiePMMLException("Failed to find DATA_TYPE for " + model.getTargetField()));
```
danielezonca, line 175: Can this happen? Because at this point the execution succeeds so I don't know if it makes sense to fail

danielezonca (approved):
Thanks for the PR, approving 👍

kie-pmml-trusty/kie-pmml-evaluator/kie-pmml-evaluator-core/src/main/java/org/kie/pmml/evaluator/core/service/PMMLRuntimeInternalImpl.java
gitgabrio: Yes, currently this could happen, and if it does not it is only "by chance". I have to implement a common validation step: with that in place, this check won't be needed anymore

kie-pmml-trusty/kie-pmml-models/kie-pmml-models-mining/kie-pmml-models-mining-tests/pom.xml
gitgabrio: @danielezonca before modification made Mario, I had to add this to be sure that the ecj compiler was picked up.

jiripetrlik (approved):
LGTM

PR #3486, opened 2021-03-30
DROOLS-6203 rule builder should now cope with subsequant accumulate r…

mdproctor:
…ewritting. Added more test foverage for executed rule model output.


mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/3488

PR #3490, opened 2021-03-31
DROOLS-6236 More tweaks to make rewrites and nested rewrites more robust

mdproctor:


mariofusco:
The tests you added don't work in the executable model simply because binding with a function with more than one argument it is  also not allowed in drl and indeed you don't have a way in plain drl to trigger this problem. In other words, there's no way that you could generate the executable model you wrote programmatically from a drl.

I see 3 possibilities here:

1. making this work in the executable model (and if we are able to do so, maybe also making it possible from drl) but at the moment I don't see how
2. throwing an exception at runtime in the executable model when it meets such binding
3. changing our dsl in order to disallow that binding at compile time

I'd go with option 3. (if possible I need to check).

mariofusco (approved):

PR #3501, opened 2021-04-01
[DROOLS-5971] ailow to modify justified facts

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5971


lucamolteni (approved):

mariofusco:
This PR always assumes that a justified object will become stated after an update. Also it does this only for the primary fact supported by a belief set not taking any action for all other objects eventually supported by the same set. We decided that this assumptions are too strong and don't provide enough flexibility for users, that conversely should be allowed to have more control on this. We need to review the requirements and the syntax of the update/modify statement in order to provide such flexibility.

PR #3504, opened 2021-04-02
[DROOLS-6256] remove flow dsl from executable model

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6256


mariofusco:
jenkins retest this

lucamolteni (changes requested):
This class was created to have `FlowExpressionBuilder` and `PatternExpressionBuilder`, when removing Flow it has to be deleted

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
```diff
@@ -777,11 +775,7 @@ private MethodCallExpr buildRulesField( ClassOrInterfaceDeclaration rulesClass )
 
     private void manageImportForCompilationUnit(CompilationUnit cu) {
         // fixed part
-        if(isPattern) {
-            cu.addImport("org.drools.modelcompiler.dsl.pattern.D");
-        } else {
-            cu.addImport("org.drools.modelcompiler.dsl.flow.D");
-        }
+        cu.addImport("org.drools.modelcompiler.dsl.pattern.D");
```
lucamolteni, line 778: is the DSL package name still `D` then?

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
```diff
@@ -240,7 +232,7 @@ protected boolean isStringToDateExpression(Expression expression) {
     }
 
     public static AbstractExpressionBuilder getExpressionBuilder(RuleContext context) {
-        return context.isPatternDSL() ? new PatternExpressionBuilder( context ) : new FlowExpressionBuilder( context );
+        return new PatternExpressionBuilder( context );
```
lucamolteni, line 235: This class was created to have `PatternExpressionBuilder` and `FlowExpressionBuilder` we remove FlowDSL  but we introduced `EvalExpressionBuidler` can you clarify?

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
mariofusco: Part of the original `FlowExpressionBuilder` is still in use. In particular it is now only used by `EvalVisitor` that used it to generate an excutable model `expr` outside of a pattern that is indeed the way how `eval`s are defined in the executable model. To make this clearer I also renamed the class `EvalExpressionBuidler`.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
mariofusco: For now yes. We could change later it if we decide to, but I'd prefer to avoid this even because this will break backward incompatibility for all old generated executable models.

mariofusco:
jenkins execute fdb

mariofusco:
jenkins execute cdb

mariofusco:
jenkins execute cdb

mariofusco:
jenkins execute cdb

tkobayas (approved):

hellowdan (approved):

mariofusco:
jenkins execute cdb

mariofusco:
jenkins retest this

PR #3556, opened 2021-04-21
[DROOLS-6291] avoid unnecessary object hashing in fact handle

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6291


triceo (approved):
Fixes the issue.

PR #3558, opened 2021-04-22
[DROOLS-6291] avoid unnecessary object hashing in fact handle (#3556)

mariofusco:


triceo (approved):
Please do not merge until we have approval for the product.

mareknovotny:
now the question is why it was merged without saying as it is not blocker looking at jira.

@mariofusco any reason you hurry up with this merging?

PR #3564, opened 2021-04-23
[RHDM-1713] fix parsing of mvel expressions with nulls

mariofusco:
**JIRA**: https://issues.redhat.com/browse/RHDM-1713


drools-model/drools-mvel-compiler/src/main/java/org/drools/mvelcompiler/ast/NullLiteralExpressionT.java
```diff
@@ -0,0 +1,34 @@
+package org.drools.mvelcompiler.ast;
+
+import java.lang.reflect.Type;
+import java.util.Optional;
```
lucamolteni: This is `UnalteredTypedExpression` no need for a custom type if it doesn't do much

lucamolteni (changes requested):

drools-model/drools-mvel-compiler/src/main/java/org/drools/mvelcompiler/ast/NullLiteralExpressionT.java
mariofusco: @lucamolteni fixed, thanks

lucamolteni (approved):

PR #3581, opened 2021-05-03
[DROOLS-6202] fix nested accumulates in executable model

mariofusco:
**JIRA**: _https://issues.redhat.com/browse/DROOLS-6202


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -73,29 +76,28 @@
...
+    public AccumulateVisitor(ModelGeneratorVisitor modelGeneratorVisitor, RuleContext context, PackageModel packageModel) {
         this.context = context;
         this.modelGeneratorVisitor = modelGeneratorVisitor;
         this.packageModel = packageModel;
+        this.expressionBuilder = new PatternExpressionBuilder(context);
     }
 
-    protected BaseDescr input;
-
-    Optional<NewBinding> optNewBinding;
```
mariofusco, line 97: @lucamolteni This is the most relevant change. Visitors cannot have a mutable state otherwise they won't work when called recursively.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -516,18 +525,99 @@ Expression buildConstraintExpression(Expression expr, Collection<String> usedDec
...
+            pattern.setParentNode( newBindingExpression );
+        });
+    }
+
+    private MethodCallExpr replaceBindingWithPatternBinding(MethodCallExpr bindExpression, MethodCallExpr lastPattern) {
+        final Expression bindingId = lastPattern.getArgument(0);
+
+        bindExpression.findFirst(NameExpr.class, e -> e.equals(bindingId)).ifPresent( name -> {
+            bindExpression.remove(name);
+            LambdaExpr lambda = (LambdaExpr)bindExpression.getArgument( bindExpression.getArguments().size()-1 );
```
lucamolteni, line 606: can you clarify what it does here? Maybe add an example with a unit test

lucamolteni (approved):

PR #3595, opened 2021-05-07
DROOLS-6193 : Subnetwork Optimization - DO NOT MERGE YET

mdproctor:
All tests pass. Although it should add no changes in behaviour, I added some additional coverage in areas I was concerned about. I need to review changes, for cleanups, remove commented code etc. We also need to benchmark this, to make sure there are no actual regressions. It is still searching for the LT, we could optimise this further to record it in the constructor at the start - it won't matter much for small subnetworks, but would do if there was a bigger subnetwork.

When benchmark not only put lots of data inside of the subnetwork, make a lot of LTs prior to it, which we want to trigger updates on - which is also common in OP, with nesting of groupby's. These updates are the ones that typically cost the most, due to double update either side. Updates inside should be less problematic, if there is no chaining of groupbys. Need to benchmark not/exists/foreach and accumulate.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
@@ -24,6 +24,7 @@ import org.drools.core.reteoo.PathMemory;
 import org.drools.core.reteoo.SegmentMemory;
 
 import static org.drools.core.phreak.AddRemoveRule.forceFlushLeftTuple;
+import static org.drools.core.phreak.AddRemoveRule.forceFlushWhenRiaNode;
 
 public class SegmentPropagator {
 
@@ -39,8 +40,25 @@ public class SegmentPropagator {
         }
                 
         processPeers(sourceSegment, leftTuples, wm);
-    }    
-    
+
+        // @TODO Could a boolean be cached on sourceSegment, to avoid this forloop (mdp)?
+        for (SegmentMemory smem = sourceSegment.getFirst(); smem != null; smem = smem.getNext() ) {
+            if (smem.hasDataDrivenPathMemories()) {
+                for (PathMemory dataDrivenPmem : smem.getDataDrivenPathMemories()) {
+                    if (smem.getStagedLeftTuples().getDeleteFirst() == null &&
+                        smem.getStagedLeftTuples().getUpdateFirst() == null &&
+                        !dataDrivenPmem.isRuleLinked()) {
+                        // skip flushing segments that have only inserts staged and the path is not linked
+                        continue;
+                    }
+                    forceFlushLeftTuple(dataDrivenPmem, smem, wm, smem.getStagedLeftTuples());
+                    forceFlushWhenRiaNode(wm, dataDrivenPmem);
+                }
+            }
+        }
+
+    }
+
     private static void processPeers(SegmentMemory sourceSegment, TupleSets<LeftTuple> leftTuples, InternalWorkingMemory wm) {
         SegmentMemory firstSmem = sourceSegment.getFirst();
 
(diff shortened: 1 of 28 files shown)


drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
```diff
@@ -39,8 +40,25 @@ public static void propagate(SegmentMemory sourceSegment, TupleSets<LeftTuple> l
...
+            if (smem.hasDataDrivenPathMemories()) {
+                for (PathMemory dataDrivenPmem : smem.getDataDrivenPathMemories()) {
+                    if (smem.getStagedLeftTuples().getDeleteFirst() == null &&
+                        smem.getStagedLeftTuples().getUpdateFirst() == null &&
+                        !dataDrivenPmem.isRuleLinked()) {
+                        // skip flushing segments that have only inserts staged and the path is not linked
+                        continue;
+                    }
+                    forceFlushLeftTuple(dataDrivenPmem, smem, wm, smem.getStagedLeftTuples());
+                    forceFlushWhenRiaNode(wm, dataDrivenPmem);
```
mariofusco, line 55: Is forcing this flush necessary? I have the feeling that this could have a performance impact and defeat the biggest part of phreak advantages over rete ( as it already happens for CEP :( )

drools-core/src/main/java/org/drools/core/reteoo/BetaNode.java
```diff
@@ -426,7 +426,7 @@ public FastIterator getRightIterator(TupleMemory memory, RightTuple rightTuple)
         }
     }
 
-    public FastIterator getLeftIterator(TupleMemory memory) {
+    public FastIterator getLeftIterator1(TupleMemory memory) {
```
mariofusco, line 429: Not sure if we could find a better name for these methods, but if possible please make them private so we won't use them mistakenly.

drools-core/src/main/java/org/drools/core/spi/Tuple.java
```diff
@@ -61,6 +61,13 @@
      */
     InternalFactHandle get(int pattern);
 
+    /**
+     * Returns the tuple at the given index
+     * @param index
+     * @return
+     */
+    Tuple getTuple(int index);
```
mariofusco, line 69: This seems to be used only to retrieve the root tuple. If possible I'd leave this method out and keep only the one for the root tuple, even because it would be unnecessarily time consuming if used mistakenly.

drools-core/src/main/java/org/drools/core/phreak/SegmentPropagator.java
mdproctor: Inside of that method is an if statement, it's avoided in non data driven situations. I believe this was the same as before:
if (pmem.isDataDriven() && pmem.getNodeType() == NodeTypeEnums.RightInputAdaterNode) {

drools-core/src/main/java/org/drools/core/reteoo/BetaNode.java
mdproctor: good point. No the "1" was temporary, I just didn't come back to it, will correct that now.

mariofusco:
jenkins retest this

mariofusco:
I ran [this benchmark](https://github.com/kiegroup/kie-benchmarks/pull/117) with and without this commit and got the following results

WITHOUT subnetwork optimization 
```
Benchmark                      (joinCount)  (nestedAccumulates)  (objectsPerSegment)  (perSegmentUpdate)  (segmentCount)  Mode  Cnt    Score   Error  Units
SubNetworkBenchmark.benchmark            2                 true                    5                true               3    ss  200    0.941 ± 0.067  ms/op
SubNetworkBenchmark.benchmark            2                 true                    5               false               3    ss  200    0.954 ± 0.054  ms/op
SubNetworkBenchmark.benchmark            2                 true                   15                true               3    ss  200    1.539 ± 0.165  ms/op
SubNetworkBenchmark.benchmark            2                 true                   15               false               3    ss  200    1.561 ± 0.089  ms/op
SubNetworkBenchmark.benchmark            2                false                    5                true               3    ss  200    1.830 ± 0.101  ms/op
SubNetworkBenchmark.benchmark            2                false                    5               false               3    ss  200    1.697 ± 0.091  ms/op
SubNetworkBenchmark.benchmark            2                false                   15                true               3    ss  200  137.876 ± 2.528  ms/op
SubNetworkBenchmark.benchmark            2                false                   15               false               3    ss  200   82.678 ± 1.578  ms/op
```

WITH subnetwork optimization 
```
Benchmark                      (joinCount)  (nestedAccumulates)  (objectsPerSegment)  (perSegmentUpdate)  (segmentCount)  Mode  Cnt   Score   Error  Units
SubNetworkBenchmark.benchmark            2                 true                    5                true               3    ss  200   1.050 ± 0.092  ms/op
SubNetworkBenchmark.benchmark            2                 true                    5               false               3    ss  200   0.994 ± 0.099  ms/op
SubNetworkBenchmark.benchmark            2                 true                   15                true               3    ss  200   2.147 ± 0.145  ms/op
SubNetworkBenchmark.benchmark            2                 true                   15               false               3    ss  200   2.043 ± 0.108  ms/op
SubNetworkBenchmark.benchmark            2                false                    5                true               3    ss  200   2.446 ± 0.108  ms/op
SubNetworkBenchmark.benchmark            2                false                    5               false               3    ss  200   1.973 ± 0.121  ms/op
SubNetworkBenchmark.benchmark            2                false                   15                true               3    ss  200  31.513 ± 0.952  ms/op
SubNetworkBenchmark.benchmark            2                false                   15               false               3    ss  200  22.349 ± 0.418  ms/op
```

These results have been obtained after one single fork with 100 warmup iterations and 200 benchmark iterations, so probably this would deserve a longer run before drawing any definitive conclusion. From this preliminary results, it is evident that this optimization has a huge positive impact on the longest running scenarios (sequential accumulates, 15 objects inserted per segment), while it seems to have some slightly negative consequences on the other shorter use cases.

mariofusco:
jenkins retest this

PR #3630, opened 2021-05-21
[RHDM-1735] optimize runtime performances and replace legacy data str…

mariofusco:
…uctures

**JIRA**: https://issues.redhat.com/browse/RHDM-1735

For now I profiled `org.drools.benchmarks.turtle.runtime.RetractFactsFromWmExpertBenchmark` which is the most penalized one by the migration from Java 8 to 11 according to @hellowdan's results.

With a very quick run on my machine, before this change I got

```
Benchmark                                                           Mode  Cnt     Score    Error  Units
RetractFactsFromWmExpertBenchmark.timeFactsInsertionAndRulesFiring    ss   20  1676.355 ± 87.505  ms/op
```

while using it I obtained

```
Benchmark                                                           Mode  Cnt     Score     Error  Units
RetractFactsFromWmExpertBenchmark.timeFactsInsertionAndRulesFiring    ss   20  1380.140 ± 107.072  ms/op
```


lucamolteni (approved):

tkobayas (approved):

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/compiler/beliefsystem/defeasible/DefeasibilityTest.java
```diff
@@ -362,7 +363,7 @@ public void testRemoveDefeasibleJustifier() {
 
 
 
-    @Test(timeout = 10000 )
+    @Test(timeout = 10000 ) @Ignore
```
hellowdan, line 366: Is it possible to add a comment here explaining the purpose of skipping this test?

drools-traits/src/test/java/org/drools/traits/compiler/factmodel/traits/TraitTest.java
```diff
@@ -5389,7 +5388,7 @@ public void testHierarchyEncodeOnPackageMerge() {
 
 
 
-    @Test
+    @Test @Ignore
```
hellowdan, line 5391: Is it possible to add a comment here explaining the purpose of skipping this test?

hellowdan (approved):

PR #3637, opened 2021-05-26
[DROOLS-6371] decouple agenda-groups management from the agenda and p…

mariofusco:
…rovide a simplified implementation not supporting them

**JIRA**: https://issues.redhat.com/browse/DROOLS-6371

### DIFF
--- a/drools-core/src/main/java/org/drools/core/impl/InternalKnowledgeBase.java
+++ b/drools-core/src/main/java/org/drools/core/impl/InternalKnowledgeBase.java
@@ -150,4 +150,6 @@ public interface InternalKnowledgeBase extends KieBase {
 
     List<AsyncReceiveNode> getReceiveNodes();
     void addReceiveNode(AsyncReceiveNode node);
+
+    boolean hasMultipleAgendaGroups();
 }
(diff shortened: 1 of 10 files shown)


mariofusco:
jenkins execute fdb

drools-core/src/main/java/org/drools/core/impl/InternalKnowledgeBase.java
```diff
@@ -150,4 +150,6 @@
 
     List<AsyncReceiveNode> getReceiveNodes();
     void addReceiveNode(AsyncReceiveNode node);
+
+    boolean hasMultipleAgendaGroups();
```
evacchi, line 154: wondering if this boolean could be instead with some `instanceof` (if it makes sense)

mariofusco: Not understanding what you're suggesting: that boolean is calculated [here](https://github.com/kiegroup/drools/pull/3637/files#diff-55968e0506f23499453664e5c0e35cc1e6737f9e110798316ba3f7b127ea2d7bR1529) checking if a rule belongs to group different than the main one when it's added to the KieBase. There aren't different implementations to discriminate with an `instanceof`.

mariofusco:
jenkins execute fdb

lucamolteni (approved):

PR #3644, opened 2021-05-27
DROOLS-6375 DMN Listener BKM invocation params and result

tarilabs:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6375

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeInvokeBKMEvent.java
+++ b/kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeInvokeBKMEvent.java
@@ -16,11 +16,17 @@
 
 package org.kie.dmn.api.core.event;
 
+import java.util.List;
+
 import org.kie.dmn.api.core.ast.BusinessKnowledgeModelNode;
 
 public interface BeforeInvokeBKMEvent
         extends DMNEvent {
 
     BusinessKnowledgeModelNode getBusinessKnowledgeModel();
+
+    default List<Object> getInvocationParameters() {
+        return null;
+    }
 
 }
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
@@ -17,6 +17,7 @@
 package org.kie.dmn.core.ast;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -145,9 +146,10 @@ public class DMNFunctionDefinitionEvaluator
             // context, but for now, cloning the original context
             DMNContextFEELCtxWrapper dmnContext = new DMNContextFEELCtxWrapper(ctx, resultContext.getContext().getMetadata().asMap());
             dmnContext.enterFrame();
+            Object invocationResult = null;
             try {
                 if (originatorNode instanceof BusinessKnowledgeModelNode) {
-                    DMNRuntimeEventManagerUtils.fireBeforeInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext);
+                    DMNRuntimeEventManagerUtils.fireBeforeInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext, Arrays.asList(params));
                 }
                 if( evaluator != null ) {
                     closureContext.getAll().forEach(dmnContext::set);
@@ -172,7 +174,8 @@ public class DMNFunctionDefinitionEvaluator
                     resultContext.setContext( dmnContext );
                     EvaluatorResult result = evaluator.evaluate( eventManager, resultContext );
                     if( result.getResultType() == ResultType.SUCCESS ) {
-                        return result.getResult();
+                        invocationResult = result.getResult();
+                        return invocationResult;
                     }
                     return null;
                 } else {
@@ -199,7 +202,7 @@ public class DMNFunctionDefinitionEvaluator
                 return null;
             } finally {
                 if (originatorNode instanceof BusinessKnowledgeModelNode) {
-                    DMNRuntimeEventManagerUtils.fireAfterInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext);
+                    DMNRuntimeEventManagerUtils.fireAfterInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext, invocationResult);
                 }
                 resultContext.setContext( previousContext );
                 dmnContext.exitFrame();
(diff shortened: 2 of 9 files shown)


mswiderski:
@tarilabs this one introduces BKM inputs/results to be accessible from the before/after events. Is there other work to be done to have same for decision service?

I will test this out either later today or tomorrow. thanks a lot @tarilabs

tarilabs:
Yes as the title and JIRA says, this tackle BKM first. We can discuss DS as a further step

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeInvokeBKMEvent.java
```diff
@@ -16,11 +16,17 @@
...
+import java.util.List;
+
 import org.kie.dmn.api.core.ast.BusinessKnowledgeModelNode;
 
 public interface BeforeInvokeBKMEvent
         extends DMNEvent {
 
     BusinessKnowledgeModelNode getBusinessKnowledgeModel();
 
+    default List<Object> getInvocationParameters() {
```
mswiderski, line 28: can the result type of this method be a `Map` instead of list as getting only raw values it's impossible to know what value corresponds to which variable/param?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
```diff
@@ -145,9 +146,10 @@ public Object invoke(EvaluationContext ctx, Object[] params) {
             // context, but for now, cloning the original context
             DMNContextFEELCtxWrapper dmnContext = new DMNContextFEELCtxWrapper(ctx, resultContext.getContext().getMetadata().asMap());
             dmnContext.enterFrame();
+            Object invocationResult = null;
             try {
                 if (originatorNode instanceof BusinessKnowledgeModelNode) {
-                    DMNRuntimeEventManagerUtils.fireBeforeInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext);
+                    DMNRuntimeEventManagerUtils.fireBeforeInvokeBKM(eventManager, (BusinessKnowledgeModelNode) originatorNode, resultContext, Arrays.asList(params));
```
mswiderski, line 152: params here could be replaced after they are matched with names that is done few lines below when populating the context. So maybe we could event consider `context.getAllValues()` as the params instead?

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeInvokeBKMEvent.java
tarilabs: That would be performance penalty, I am proposing an alternative solution with 695bb79fc7 which does extract the same relation "offline"

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
tarilabs: unless I am mistaken, this is not what is happening.
During function evaluation, global context entries are overridden with closure context entries, which are overridden with parameters entries; per the specification.
So I don't think getting context values are necessary for the scope of determining the parameter of a BKM invocation?

mswiderski: what I mean by context was the context that is populated here and not the global one. As context here is based on parameters array matched with names.

kie-dmn/kie-dmn-api/src/main/java/org/kie/dmn/api/core/event/BeforeInvokeBKMEvent.java
mswiderski: so this utility class should be used on top of the event to extract the values. Sounds good to me.

Thanks @tarilabs

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
tarilabs: for the use-case of BKM it should suffice

tarilabs:
Jenkins retest this

mswiderski (approved):
@tarilabs with the latest changes it does exactly what I needed. 

From my point of view it is good to go. Thanks a lot.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNEventUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.kie.dmn.api.core.event.BeforeInvokeBKMEvent;
+import org.kie.dmn.core.ast.BusinessKnowledgeModelNodeImpl;
+import org.kie.dmn.core.ast.DMNFunctionDefinitionEvaluator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DMNEventUtils {
```
danielezonca, line 29: Is this class intended to be used by end user? If so, what about move it somewhere else outside `dmn-core` and `org.kie.dmn.core.impl` package?
It should be nice to have it in `api` but it is not probably possible considering the cast to internal impl 🤔

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
```diff
@@ -172,7 +174,8 @@ public Object invoke(EvaluationContext ctx, Object[] params) {
                     resultContext.setContext( dmnContext );
                     EvaluatorResult result = evaluator.evaluate( eventManager, resultContext );
                     if( result.getResultType() == ResultType.SUCCESS ) {
-                        return result.getResult();
+                        invocationResult = result.getResult();
+                        return invocationResult;
```
danielezonca, line 178: Do we have a test that breaks if this assignment to `invocationResult` is skipped? If not what about at least to add a comment to highlight that it is used in the `finally` block?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNEventUtils.java
tarilabs: not end-user, but developers integrating specific use-cases.
The reason is not in API is that indeed this can only be done in specific use-cases.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNFunctionDefinitionEvaluator.java
tarilabs: in the case of an exception, the error is already logged as part of the DMN Message mechanism. So it is expected for this to be `null` per previous line, and the semantic for DMN remains unchanged where you would check presence of Error messages per standard error reporting mechanism?

danielezonca (approved):

hellowdan (approved):

PR #3650, opened 2021-06-03
[DROOLS-6386] BitMask is not generated for modify in named consequenc…

tkobayas:
…e with exec-model

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6386

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
@@ -75,7 +75,6 @@ import static org.drools.modelcompiler.builder.PackageModel.DOMAIN_CLASSESS_META
 import static org.drools.modelcompiler.builder.PackageModel.DOMAIN_CLASS_METADATA_INSTANCE;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.addCurlyBracesToBlock;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.findAllChildrenRecursive;
-import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.getClassFromType;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.isNameExprWithName;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.parseBlock;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.toClassOrInterfaceType;
@@ -359,7 +358,10 @@ public class Consequence {
                     if ( !initializedBitmaskFields.contains( updatedVar ) ) {
                         Set<String> modifiedProps = findModifiedProperties( methodCallExprs, updateExpr, updatedVar, updatedClass );
                         MethodCallExpr bitMaskCreation = createBitMaskInitialization( updatedClass, modifiedProps );
-                        ruleBlock.addStatement( createBitMaskField( updatedVar, bitMaskCreation ) );
+                        AssignExpr bitMaskAssign = createBitMaskField(updatedVar, bitMaskCreation);
+                        if (!DrlxParseUtil.hasDuplicateExpr(ruleBlock, bitMaskAssign)) {
+                            ruleBlock.addStatement(bitMaskAssign);
+                        }
                     }
 
                     updateExpr.addArgument( "mask_" + updatedVar );
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
@@ -920,6 +920,10 @@ public class DrlxParseUtil {
         return exprType == Boolean.class || exprType == boolean.class;
     }
 
+    public static boolean hasDuplicateExpr(BlockStmt ruleBlock, Expression expr) {
+        return ruleBlock.findFirst(expr.getClass(), expr::equals).isPresent();
+    }
+
     private DrlxParseUtil() {
         // It is not allowed to create instances of util classes.
     }
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/NamedConsequenceVisitor.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/NamedConsequenceVisitor.java
@@ -158,7 +158,7 @@ public class NamedConsequenceVisitor {
             context.addCompilationError(new InvalidExpressionErrorResult("Unknown consequence name: " + namedConsequence.getName()));
             return null;
         }
-        BlockStmt ruleVariablesBlock = new BlockStmt();
+        BlockStmt ruleVariablesBlock = context.getRuleVariablesBlock();
         createVariables(ruleVariablesBlock, packageModel, context);
         return new Consequence(context).createCall(null, namedConsequenceString, ruleVariablesBlock, namedConsequence.isBreaking() );
     }
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
@@ -523,20 +523,20 @@ public class NamedConsequencesTest extends BaseModelTest {
     }
 
     @Test
-    public void testModifyInNamedConsequence2() {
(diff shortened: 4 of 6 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java
```diff
@@ -359,7 +358,10 @@ private boolean rewriteRHS(BlockStmt ruleBlock, BlockStmt rhs) {
                     if ( !initializedBitmaskFields.contains( updatedVar ) ) {
                         Set<String> modifiedProps = findModifiedProperties( methodCallExprs, updateExpr, updatedVar, updatedClass );
                         MethodCallExpr bitMaskCreation = createBitMaskInitialization( updatedClass, modifiedProps );
-                        ruleBlock.addStatement( createBitMaskField( updatedVar, bitMaskCreation ) );
+                        AssignExpr bitMaskAssign = createBitMaskField(updatedVar, bitMaskCreation);
+                        if (!DrlxParseUtil.hasDuplicateExpr(ruleBlock, bitMaskAssign)) {
+                            ruleBlock.addStatement(bitMaskAssign);
+                        }
```
tkobayas, line 364: Just checking duplication here. But we would need to create different BitMasks if named consequences modify different properties. I'll split it to another JIRA. (Filed https://issues.redhat.com/browse/DROOLS-6388)

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/NamedConsequenceVisitor.java
```diff
@@ -154,7 +154,7 @@ private MethodCallExpr onDSL(NamedConsequenceDescr namedConsequence) {
             context.addCompilationError(new InvalidExpressionErrorResult("Unknown consequence name: " + namedConsequence.getName()));
             return null;
         }
-        BlockStmt ruleVariablesBlock = new BlockStmt();
+        BlockStmt ruleVariablesBlock = context.getRuleVariablesBlock();
```
tkobayas, line 161: We need to assemble `ruleVariablesBlock` that is modified for named consequences and a default consequence. So rather than creating a new `ruleVariablesBlock`, bring it around using `RuleContext`.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
```diff
@@ -920,6 +920,14 @@ public static boolean isBooleanBoxedUnboxed(java.lang.reflect.Type exprType) {
         return exprType == Boolean.class || exprType == boolean.class;
     }
 
+    public static boolean hasDuplicateExpr(BlockStmt ruleBlock, Expression expr) {
```
lucamolteni, line 923: I think you can use `findFirst` of BlockStmt passing a Expression class here

lucamolteni (approved):

hellowdan (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
tkobayas: Updated. Thanks!

mariofusco (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
```diff
@@ -489,4 +489,67 @@ public void testMultipleIfElseInARowWithJoin2() {
...
+
+        ksession.fireAllRules();
+
+        assertEquals(1, results.size());
+        assertTrue(results.contains("stilton"));
+    }
+
+    @Test
+    public void testModifyInNamedConsequence2() {
+        String str = "import " + Cheese.class.getCanonicalName() + ";\n " +
```
pibizza, line 527: Possibly finding a better name for the test? It seems like this is another variant of the previous test, but it is unclear from the name which is the difference

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
```diff
@@ -489,4 +489,67 @@ public void testMultipleIfElseInARowWithJoin2() {
...
+        ksession.setGlobal("results", results);
+
+        Cheese stilton = new Cheese("stilton", 5);
+        Cheese cheddar = new Cheese("cheddar", 7);
+
+        ksession.insert(stilton);
+        ksession.insert(cheddar);
+
+        ksession.fireAllRules(10);
+
```
pibizza, line 553: Why we have here a limit of 10? How this is relevant to this specific fix and this specific test?

pibizza:
Tests added are clearly integration level tests - they create a rule, pass it to drools together with data and check that the correct results are present. 

However they expose lots of low level details in that, and that makes the relevant aspect of the test less clear.

Since it seems there are lots of tests like that I would suggest creating a simple infrastructure that can abstract all these low level details, and provide a simple high level test with as inputs a) the drool rule, b) the initial data and c) the final results.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
mariofusco: The test is exercising property reactivity and for this specific use case Toshiya is setting a limit of 10 firings otherwise it would fall in an infinite loop, which is expected. 

@tkobayas maybe you can just add an assertion to check that it actually fired 10 times?

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
tkobayas: Thanks, changed the name to `test2ModifyBlocksInNamedConsequences`.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NamedConsequencesTest.java
tkobayas: Sorry, 10 was just a leftover of writing a test back-and-forth (and yes, it's property reactivity) and now it's just fine with `fireAllRules()` and the rule is fired 2 times. Fixed.

tkobayas:
Hi @pibizza ,

> Tests added are clearly integration level tests - they create a rule, pass it to drools together with data and check that the correct results are present.

Yes.

> However they expose lots of low level details in that, and that makes the relevant aspect of the test less clear.

Could let me know which lines you mention as "low level details"? Thanks!

pibizza:
> Hi @pibizza ,

> Could let me know which lines you mention as "low level details"? Thanks!

Well lots of them:

- creating a session is a detail
- inserting facts via the insert in the session is a detail
- putting result as a global and using it to check the results is a detail
- using fireAllRules is a detail
- and so on.

My point is that our test is not really related to any of these details, and should not be affected by them. 
My feeling is that we can hide most of them and try to go for a more declarative form of the test.

Something like

given(Rule1)
andFacts(Fact1,Fact2,Fact3)
afterRulesFireAgendaIs(AgendaItem1, AgendaItem2, AgendaItem3)

In a test like that if we change ksession.insert to ksession.add our test is not affected. If we change fireAllRules to executeAllRules our test is not affected and so on.

My two cents, happy to discuss about it and find out that I missed lots of important stuff.

tkobayas:
Thanks, @pibizza I got your point.

I agree that a good amount of unit tests in drools-model-compiler and drools-test-coverage have the similar fashion:

- Create a ksession from a String DRL
- Set a List global (or no global) 
- Insert a few facts
- fireAllRules()
- Assert the number of fired rules
- Assert the global's elements or fact's properties

We can create declarative test methods to hide the drools APIs. It would provide some consistency and robustness for the test cases that meet the fashion. 

Rewriting existing tests would be a huge effort but we may be able to start with newly created tests.

@mariofusco @lucamolteni @hellowdan  Do you have any thoughts on this?

lucamolteni:
> Rewriting existing tests would be a huge effort but we may be able to start with newly created tests.
> 
> @mariofusco @lucamolteni @hellowdan Do you have any thoughts on this?


In my opinion the current kie-api (and the one we use in test) is a good tradeoff between high level and composability, it gives us the flexibility to change the test however we want and it's quite compact.

Maybe we could try and experiment a new API and decide whether we can apply it

pibizza (approved):

PR #3657, opened 2021-06-07
[DROOLS-6392] restoring a fireUntilHalt after an incremental compilat…

mariofusco:
…ion either on the same thread where the fireUntilHalt was running or on a new thread

**JIRA**: https://issues.redhat.com/browse/DROOLS-6392


tkobayas (approved):
Thanks!

lucamolteni (approved):

PR #3659, opened 2021-06-08
Rule name logic Included in Rule.java in drools-templates

MvkZ:
### String Replacement for the function drools.getRule().getName()


MvkZ:
Can any of the admin review this pull request?

PR #3667, opened 2021-06-10
[DROOLS-6401] First cut for DSL for testing drools rules

pibizza:

**JIRA**: https://issues.redhat.com/browse/DROOLS-6401

Overall approach

* create a composable set of methods to create the tests
* hide all the details that are irrelevant 
* make visible only the data relevant for the tests. 

Created so far
* simple initial API for testing
* one test file fully migrated

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
@@ -16,10 +16,13 @@
 
 package org.drools.modelcompiler;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import org.assertj.core.api.Assertions;
 import org.drools.compiler.kie.builder.impl.DrlProject;
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
 import org.drools.core.impl.InternalKnowledgeBase;
@@ -44,6 +47,7 @@ import static java.util.Arrays.asList;
 import static org.drools.modelcompiler.BaseModelTest.RUN_TYPE.PATTERN_DSL;
 import static org.drools.modelcompiler.BaseModelTest.RUN_TYPE.PATTERN_WITH_ALPHA_NETWORK;
 import static org.drools.modelcompiler.BaseModelTest.RUN_TYPE.STANDARD_WITH_ALPHA_NETWORK;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
 @RunWith(Parameterized.class)
@@ -94,6 +98,14 @@ public abstract class BaseModelTest {
     }
 
     protected final CompilerTest.RUN_TYPE testRunType;
+
+	private String str;
+
+	private KieSession ksession;
+
+	protected List<Object> list;
+
+	protected int firedRules;
 
     public BaseModelTest( CompilerTest.RUN_TYPE testRunType ) {
         this.testRunType = testRunType;
@@ -225,4 +237,45 @@ public abstract class BaseModelTest {
         }
         return null;
     }
+
+	protected void whenWeFireAllRules() {
+	    firedRules = ksession.fireAllRules();
+	}
+
+	protected void withFacts(Object... facts) {
+		for (Object fact: facts) {
+			ksession.insert(fact);
+		}
+	}
+
+	protected void withSessionAndGlobals() {
+	    withStandardSession();
+	
+	    list = new ArrayList<>();
+	    ksession.setGlobal("list", list);	
+	
+	}
+
+	protected void withStandardSession() {
+		ksession = getKieSession(str);
+	}
+
+	protected void withRule(String... rules) {
+		this.str = String.join("\n", rules);
+	}
+	
+    protected void firedRulesAre(int i) {
+    	assertEquals( i, firedRules );
+		
+	}
+
+	protected void resultContainsInAnyOrder(Object... values) {
+	    Assertions.assertThat(list).containsExactlyInAnyOrder(values);
+	}
+
+	protected void sessionContainsExactly(String... values) {
+		Collection<String> results = getObjectsIntoList(ksession, String.class);
+		Assertions.assertThat(results).containsExactlyInAnyOrder(values);
+	}
+
 }
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
@@ -29,7 +29,6 @@ import org.assertj.core.api.Assertions;
 import org.drools.modelcompiler.domain.Address;
 import org.drools.modelcompiler.domain.InternationalAddress;
 import org.drools.modelcompiler.domain.Person;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.kie.api.builder.Message;
 import org.kie.api.builder.Results;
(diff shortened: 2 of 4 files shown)


tkobayas (approved):
Very nice! Thanks!

mariofusco:
@pibizza as a first iteration this looks nice. At the beginning I was skeptical about not writing our tests using our public API, but there's no reason why we shouldn't do this and in fact now I see as an improvement to have our tests abstracted from it. This however reminds me of the [`KieHelper`](https://github.com/kiegroup/droolsjbpm-knowledge/blob/master/kie-internal/src/main/java/org/kie/internal/utils/KieHelper.java) that I introduced years ago. When I created it my intention was just to have a small utility class, only for internal use, to write tests in a more concise way. Nevertheless what happened next is that many users saw it used in our tests, found it nice to use and started employing both within their own tests and also in production code. I foresee a similar pattern for this testing DSL and this time I'd like to learn the `KieHelper`'s lesson and plan in advance to this purpose.

In essence I'm in favor to create a testing DSL, but I'd like to have something that could be used not only by the Drools team, but also by end-users as a Drools testing suite. To allow this we need to make this DSL a little richer and more flexible. We can also review is style and maybe doing something similar to what [cucumber](https://cucumber.io/) does (or decide to use cucumber directly).

lucamolteni (approved):
Fine by me, try adapt this API to some different testing to see how flexible it is please

pibizza:
> Fine by me, try adapt this API to some different testing to see how flexible it is please

@lucamolteni  Good point. Any specific suggestion for possible candidates?

pibizza:
> In essence I'm in favor to create a testing DSL, but I'd like to have something that could be used not only by the Drools team, but also by end-users as a Drools testing suite. To allow this we need to make this DSL a little richer and more flexible. We can also review is style and maybe doing something similar to what [cucumber](https://cucumber.io/) does (or decide to use cucumber directly).

@mariofusco I agree with you. Honestly this was the original idea, but I wanted to first define the API by checking the required cases, and then make it a separate product to the client.

I am very interested in using Cucumber or JBehave - I clearly got inspiration from them - and this would be a very nice step towards transforming our integration tests into executable documentation. In theory we should be able to match the existing Drools rule docs with various tests.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
```diff
@@ -95,6 +99,14 @@ public boolean isExecutableModel() {
 
     protected final CompilerTest.RUN_TYPE testRunType;
 
+	private String str;
+
+	private KieSession ksession;
+
+	protected List<Object> list;
+
+	protected int firedRules;
```
danielezonca, line 108: What about wrap the logic in a builder/delegate class that is local to test instead of a shared field? These fields need to be cleaned up after every test to be on the safe side and in general this makes the test impossible to run in parallel.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
```diff
@@ -225,4 +237,45 @@ protected ObjectTypeNode getObjectTypeNodeForClass( KieSession ksession, Class<?
...
+	}
+	
+    protected void firedRulesAre(int i) {
+    	assertEquals( i, firedRules );
+		
+	}
+
+	protected void resultContainsInAnyOrder(Object... values) {
+	    Assertions.assertThat(list).containsExactlyInAnyOrder(values);
+	}
```
danielezonca, line 274: Related to previous comment, these methods can be moved to the other class and become fluent to allow composition

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
```diff
@@ -130,44 +129,38 @@ public void testMVELmodify() {
...
+                "  retract($s)",
+                "end");
+        withStandardSession();
+        withFacts(47);
+        whenWeFireAllRules();
+        sessionContainsExactly("Modified person age to 1 for: Matteo");
     }
 
-    @Test
+	@Test
```
danielezonca, line 163: Minor comment related to the whole PR, have a look to code formatting rules because it looks like the formatting is broken most of the classes

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
```diff
@@ -225,4 +237,45 @@ protected ObjectTypeNode getObjectTypeNodeForClass( KieSession ksession, Class<?
...
+
+	protected void withSessionAndGlobals() {
+	    withStandardSession();
+	
+	    list = new ArrayList<>();
+	    ksession.setGlobal("list", list);	
+	
+	}
+
+	protected void withStandardSession() {
```
danielezonca, line 259: What do you mean with "standard"? Do you mean like kieSession with default configuration?

danielezonca:
> In essence I'm in favor to create a testing DSL, but I'd like to have something that could be used not only by the Drools team, but also by end-users as a Drools testing suite. To allow this we need to make this DSL a little richer and more flexible. We can also review is style and maybe doing something similar to what [cucumber](https://cucumber.io/) does (or decide to use cucumber directly).

I agree, if one of the goal is to create a "reusable" API, I would prefer a delegate to an independent class/DSL that we can use initially to structure our tests but then also let user in the future to reuse it.
In addition to cucumber we use a lot RestAssured api in Kogito to [test REST](https://github.com/kiegroup/kogito-examples/blob/stable/dmn-quarkus-example/src/test/java/org/kie/dmn/kogito/quarkus/example/TrafficViolationTest.java#L31-L47) that has the same given/when/then structure of cucumber. Of course it is very specific to REST:
```java
        given()
                .body("{\n" +
                        "    \"Driver\": {\n" +
                        "        \"Points\": 2\n" +
                        "    },\n" +
                        "    \"Violation\": {\n" +
                        "        \"Type\": \"speed\",\n" +
                        "        \"Actual Speed\": 120,\n" +
                        "        \"Speed Limit\": 100\n" +
                        "    }\n" +
                        "}")
                .contentType(ContentType.JSON)
                .when()
                .post("/Traffic Violation")
                .then()
                .statusCode(200)
                .body("'Should the driver be suspended?'", is("No"));
```

We can consider to preserve the "structure": 
```java
   given(DroolsTest.class)
      .rules(....)
      .globals(...)
      .insert(...)
      .customization(testContext -> ....)  // testContext is a "map" with kieSession and other
   .when(testContext -> ....)              // action
   .then()
      .assertion(testContext -> ...)
      .assertion(testContext -> ...)
```

Btw let me know if I'm in the completely wrong direction :)
In this case please ignore my comment :)

lucamolteni:
> > Fine by me, try adapt this API to some different testing to see how flexible it is please
> 
> @lucamolteni Good point. Any specific suggestion for possible candidates?

Not really, maybe starts taking a look at the  `org.drools.testcoverage.functional;` package

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/BaseModelTest.java
pibizza: The suggestions has its own merits, expecially if we want to make it a separate library. 

However in Junit istance variables are per test - each test is run in a separate instance, so using istance variables in tests is a standard pattern.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/MvelDialectTest.java
pibizza: Thanks, I didn't notice that.  I will review the patch for formatting issues.

pibizza:
> > In essence I'm in favor to create a testing DSL, but I'd like to have something that could be used not only by the Drools team, but also by end-users as a Drools testing suite. To allow this we need to make this DSL a little richer and more flexible. We can also review is style and maybe doing something similar to what [cucumber](https://cucumber.io/) does (or decide to use cucumber directly).
> 
> I agree, if one of the goal is to create a "reusable" API, I would prefer a delegate to an independent class/DSL that we can use initially to structure our tests but then also let user in the future to reuse it.

baldimir:
I like the idea of having a public testing API, however we need to be careful, because often the tests serve as examples for how to use Drools. If we stop writing Drools tests completely using a public Drools API, we will remove this context and hide the actual Drools usage examples.

mdproctor:
I would avoid anything "public", as we don't want to get locked into these apis. But it's  to explore things that are re-used by ourselves, and those submitting back to the project.

pibizza:
I am going to close this pull request since we have decided to move into a different direction for the moment.

PR #3678, opened 2021-06-15
DROOLS-6416 DMN FEEL GWT Showcase only on explicit GWT properties

tarilabs:
# nothing specified, eg. `mvn clean install`

image

# specifying profile, eg: `... -PfullProfile`

image

# specify with property `... -Dfull`

image

---

**JIRA**: https://issues.redhat.com/browse/DROOLS-6416

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
+++ b/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
@@ -1,0 +1,8 @@
+To compile and run this project in the main Maven build, ensure the necessary profile property is set.
+
+Eg.
+```
+mvn clean install -Pgwt-showcase
+ or
+mvn clean install -DgwtShowcase
+```
--- a/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
+++ b/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
@@ -2,7 +2,7 @@ To compile and run this project in the main Maven build, ensure the necessary pr
 
 Eg.
 ```
-mvn clean install -Pgwt-showcase
+mvn clean install -PfullProfile
  or
-mvn clean install -DgwtShowcase
+mvn clean install -Dfull
 ```
--- a/kie-dmn/pom.xml
+++ b/kie-dmn/pom.xml
@@ -22,7 +22,7 @@
     <module>kie-dmn-feel</module>
     <module>kie-dmn-feel-gwt-functions</module>
     <module>kie-dmn-feel-gwt</module>
-    <module>kie-dmn-feel-gwt-showcase</module>
+    <!-- <module>kie-dmn-feel-gwt-showcase</module> -->
     <module>kie-dmn-backend</module>
     <module>kie-dmn-core</module>
     <module>kie-dmn-validation-bootstrap</module>
@@ -48,5 +48,20 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
+  
+  <profiles>
+    <profile>
+      <id>gwt-showcase</id>
+      <activation>
+        <activeByDefault>false</activeByDefault>
+        <property>
+          <name>gwt.compiler.skip</name>
+        </property>
+      </activation>
+      <modules>
+        <module>kie-dmn-feel-gwt-showcase</module>
+      </modules>
+    </profile>
+  </profiles>
 
 </project>
--- a/kie-dmn/pom.xml
+++ b/kie-dmn/pom.xml
@@ -50,11 +50,10 @@
   
   <profiles>
     <profile>
-      <id>gwt-showcase</id>
+      <id>fullProfile</id>
       <activation>
-        <activeByDefault>false</activeByDefault>
         <property>
-          <name>gwtShowcase</name>
+          <name>full</name>
         </property>
       </activation>
       <modules>
--- a/kie-dmn/pom.xml
+++ b/kie-dmn/pom.xml
@@ -22,7 +22,6 @@
     <module>kie-dmn-feel</module>
     <module>kie-dmn-feel-gwt-functions</module>
     <module>kie-dmn-feel-gwt</module>
-    <!-- <module>kie-dmn-feel-gwt-showcase</module> -->
     <module>kie-dmn-backend</module>
     <module>kie-dmn-core</module>
     <module>kie-dmn-validation-bootstrap</module>
@@ -55,7 +54,7 @@
       <activation>
         <activeByDefault>false</activeByDefault>
         <property>
-          <name>gwt.compiler.skip</name>
+          <name>gwtShowcase</name>
         </property>
       </activation>
       <modules>


kie-dmn/pom.xml
```diff
@@ -48,5 +48,20 @@
...
     </dependencies>
   </dependencyManagement>
+  
+  <profiles>
+    <profile>
+      <id>gwt-showcase</id>
+      <activation>
+        <activeByDefault>false</activeByDefault>
+        <property>
+          <name>gwt.compiler.skip</name>
```
danielezonca: I would prefer to use a different property (`gwtShowcase`) as property so that we don't overload the existing `gwt.compiler.skip` property with a different meaning

kie-dmn/pom.xml
```diff
@@ -22,7 +22,7 @@
     <module>kie-dmn-feel</module>
     <module>kie-dmn-feel-gwt-functions</module>
     <module>kie-dmn-feel-gwt</module>
-    <module>kie-dmn-feel-gwt-showcase</module>
+    <!-- <module>kie-dmn-feel-gwt-showcase</module> -->
```
danielezonca: Please remove the commented line

danielezonca (commented):
Thanks for the PR, can you add a README in [kie-dmn-feel-gwt-showcase](https://github.com/kiegroup/drools/tree/master/kie-dmn/kie-dmn-feel-gwt-showcase) folder?

kie-dmn/pom.xml
tarilabs: done.

kie-dmn/pom.xml
tarilabs: done.

tarilabs:
https://github.com/kiegroup/drools/pull/3678#pullrequestreview-683726725 done with 
bb6d003

danielezonca (approved):
Thanks for the changes 👍

Rikkola (approved):
Another trick is to use the `-Dfull` to include the GWT module in the regular `Linux - Pull Request`. The same practice is used in other GWT projects to either skip the GWT build or run a lighter version of it, while enabling faster failure and more complete build test coverage on our PR process. Just proposing this, even without it the build will then fail in `kie-wb-common` with the FDB.

karreiro (commented):
Thank you for this PR, @tarilabs.

However, I've noticed it removes the execution of `org/kie/dmn/feel/client/showcase/FEELShowcaseIT.java`.

This test represents the beginning of a single FEEL implementation running on the client and the server-side. Still, it already ensures that FEEL is being executed in the browser successfully.

 If we remove this test from execution, we may get a green build that actually breaks some FEEL use cases on the client-side and, consequently, the FEEL IntelliSense suggestions in the DMN editor.

So, maybe we may have a fast profile for excluding this check. But, it seems valuable to keep it in the project build.

tarilabs:
https://github.com/kiegroup/drools/pull/3678#pullrequestreview-684847836
why does these tests need to go through a GUI, @karreiro ?

All I see are expression evaluations and autocompletion assertions. I assume these functionalities exists beyond the actual GUI. Can you migrate the test "upstream" and without the need of a GUI ?

Rikkola:
Ok, so I feel like the question will end up being if this should move up stream or not. We can move the tests upstream, but then each DMN PR needs to run FDB, however they might already have to do that.

tarilabs:
I don't think is acceptable to impose arbitrary requirements such as:

> each DMN PR needs to run FDB

what I've been asking is that FEEL expression evaluation once transpiled/compiled/translated-into JavaScript can happen without the need to compile a full GUI; and just operate on the JS code which is the FEEL engine. Makes sense?

Rikkola:
Makes sense and that is on the road map. Status on that: I wanted to do that with J2CL and ran into some trouble, Dmitrii stepped in and found out some of our GWT/J2CL dependencies were not yet J2CL compatible. So there are two paraller efforts now: one to get DMN to JS with J2CL and another where I'm working on FEEL functions. The end goal is to not use this showcase for automated testing, but it is used for now before we get the J2CL trouble figured out.

However even with J2CL transpiling will of course be slow. It might make sense to also skip that in a similar manner as here.

tarilabs:
Closing because

> The end goal is to not use this showcase for automated testing, but it is used for now before we get the J2CL trouble figured out

tarilabs:
> However even with J2CL transpiling will of course be slow. It might make sense to also skip that in a similar manner as here.

Just no, please.
Of course please make sure to not have in the transpilation scope things which are not needed for FEEL engine evaluations, i.e. out of scopes, e.g. a full featured GUI.
But please do **not** skip on having tests that operates FEEL expression evaluation once transpiled/compiled/translated-into JavaScript, which just operate on the JS code of the FEEL engine.

Rikkola:
@tarilabs We can still use have the build behind  `-Dfull` if that makes life easier. I think the GWT devs are used to using that handle and I checked that also the `drools` repository build with "Jenkins retest this" activates that profile. So by default the GWT based modules are skipped, but you can't get a PR in without running those tests.

tarilabs:
https://github.com/kiegroup/drools/pull/3678#issuecomment-862177357
thank you for the clarification.

I am pending confirmation that GHA will always use `-Dfull`

tarilabs:
reopening on confirmation that both Jenkins AND GitHub Actions do make use of `-Dfull`.
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/.ci/pull-request-config.yaml#L7

kie-dmn/kie-dmn-feel-gwt-showcase/README.md
```diff
@@ -0,0 +1,8 @@
+To compile and run this project in the main Maven build, ensure the necessary profile property is set.
+
+Eg.
+```
+mvn clean install -Pgwt-showcase
+ or
+mvn clean install -DgwtShowcase
```
danielezonca: Please update instruction accordingly

tarilabs: done. thanks I forgot

hellowdan (approved):

karreiro (approved):
Thank you for the changes and clarification, @tarilabs,@Rikkola, and @danielezonca.

tarilabs:
```
[2021-06-16T13:33:45.864Z] Body did not finish within grace period; terminating with extreme prejudice
[Pipeline] }
[Pipeline] // script
Error when executing always post condition:
org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
	at org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution.cancel(TimeoutStepExecution.java:165)
	at org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution.access$100(TimeoutStepExecution.java:38)
	at org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution$1.run(TimeoutStepExecution.java:139)
	at jenkins.security.ImpersonatingScheduledExecutorService$1.run(ImpersonatingScheduledExecutorService.java:67)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

unrelated, Jenkins retest this

tarilabs:
So to recap,
this solution as of now would activate the `fullProfile` Maven profile only via Jenkins or GHA. The developer can manually check locally this by using `-PfullProfile` or `-Dfull` per usual convention, in any case the tests are run and integrated both on Jenkins and GHA. 

This solution can be applied analogously when transpilation to JS will be migrated from GWT to J2CL per this https://github.com/kiegroup/drools/pull/3678#issuecomment-862168817 and this https://github.com/kiegroup/drools/pull/3678#issuecomment-862177357 confirms.

This way, we keep local development time smooth, but we keep confidence in test coverage (executing FEEL expression evaluation by means of the FEEL engine _after_ being transpiled to JS, so really using the JS resulting engine) being continously integrated both Jenkins and GHA.

tarilabs:
https://eng-jenkins-csb-business-automation.apps.ocp4.prod.psi.redhat.com/job/KIE/job/master/job/pullrequest/job/drools-master.pr/308 is a 404.

🤷‍♂️ 
Jenkins retest this.

PR #3680, opened 2021-06-16
[DROOLS-6328] Prevent same key for Alpha Node range index

tkobayas:
**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6328

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
@@ -482,7 +482,7 @@ public class CompositeObjectSinkAdapter implements ObjectSinkPropagator {
 
     /**
      * Returns a FieldIndex which Keeps a count on how many times a particular field is used with a range check
-     * in the sinks. Stores the FieldIndex for asc and desc separately
+     * in the sinks.
      */
     private FieldIndex registerFieldIndexForRange(final int index,
                                                   final InternalReadAccessor fieldExtractor) {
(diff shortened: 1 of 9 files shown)


tkobayas:
If multiple alpha nodes have the same field value but node sharing is not working for those alpha nodes (e.g. DROOLS-6313), only one of them can be added to the index hence rule evaluation would go wrong (typically, miss to fire). Ideally, we should fix every node sharing issue but as we cannot forecast all possible cases, a guard should be introduced for safety.

@mariofusco Could you check if this approach makes sense? There are different approaches (e.g. put the alpha node of the duplicate key into `otherSinks` instead of disabling range index) but they could be more difficult because we have a little complex logic about `rangeIndexableSinks` maintenance. Disabling range index is a little radical but I assume this is a rare case. If okay, I would add some more tests in `CompositeObjectSinkAdapterTest`

tkobayas:
Talked with Mario. I'll update to throw an Exception rather than automatic disabling range index.

mariofusco (approved):

pibizza (approved):
Two points, non blockers for the patch.

Minor point. 
I suggest being consistent and putting the various commands of the drools rule on different lines, corresponding to the new lines.

Major point
I do not like the new integration test - it is too complicated, contains lots of unnecessary details and does too many things together. 

However this is a problem of all the tests in the file, not just the specific test. 

I will try to make a patch to show what I have in mind in terms of test style.

Approved.

hellowdan (approved):

tkobayas:
Hi @pibizza 

>  I suggest being consistent and putting the various commands of the drools rule on different lines, corresponding to the new lines.

Yeah, you meant lines like this, right? Sorry, I did it because if I wrote the rules with all different lines, it would be long vertically and getting hard to read... but you are right, if it's too long, I should use a separated DRL file.
```
  "then\n end\n" +
  "rule test2\n when\n" +
```

>  I do not like the new integration test - it is too complicated, contains lots of unnecessary details and does too many things together.
>  I will try to make a patch to show what I have in mind in terms of test style.

You mean CompositeObjectSinkAdapterTest.java, right? It is a unit test rather than an integration test (contrary to the package name). So they have details but I look forward to your style suggestion.

Thanks!

PR #3703, opened 2021-06-25
[DROOLS-6423] avoid storing executable model generated classes into t…

mariofusco:
…he static class loader


mariofusco:
jenkins retest this

danielezonca (approved):

PR #3704, opened 2021-06-25
[DROOLS-6284] Rules not fired when RUN_TYPE==STANDARD_FROM_DRL and wh…

tkobayas:
…en the constraint is jitted

Fixed for these 3 tests:
org.drools.modelcompiler.TypeObjectCoercionTest.testJoinStringToObject2[STANDARD_FROM_DRL]
org.drools.modelcompiler.TypeObjectCoercionTest.testJoinObjectToString2[STANDARD_FROM_DRL]
org.drools.modelcompiler.TypeObjectCoercionTest.testJoinStringToObjectNonComparable[STANDARD_FROM_DRL]

Also tested drools-model-compiler and drools-test-coverage with `mvn clean test -Ddrools.jittingThreshold=0` locally.

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-6284
https://issues.redhat.com/browse/RHDM-1595

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/TypeObjectCoercionTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/TypeObjectCoercionTest.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.Map;
 
 import org.assertj.core.api.Assertions;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.kie.api.runtime.KieSession;
 
@@ -211,18 +210,6 @@ public class TypeObjectCoercionTest extends BaseModelTest {
         ksession.insert(new ObjectHolder("DEF"));
         assertEquals(1, ksession.fireAllRules()); // standard-drl : "DEF" < "ABC" (String comparison)
         ksession.dispose();
-    }
-
-    @Ignore("in case of standard-drl, MathProcessor.doOperationNonNumeric() returns false when the left operand is not Comparable. But it doesn't make sense to keep compatibility for this behaviour")
-    @Test
-    public void testJoinStringToObjectNonComparable() {
-
-        KieSession ksession = getKieSessionForJoinStringToObject();
-
-        // Object > String "10"
-        ksession.insert(new StringHolder("10"));
-        ksession.insert(new ObjectHolder(new Object())); // not Comparable
-        assertEquals(0, ksession.fireAllRules()); // in case of standard-drl, MathProcessor.doOperationNonNumeric() returns false when the left operand is not Comparable
     }
 
     private KieSession getKieSessionForJoinIntegerToObject() {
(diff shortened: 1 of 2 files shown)


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/TypeObjectCoercionTest.java
```diff
@@ -212,6 +213,7 @@ public void testJoinStringToObject5() {
         ksession.dispose();
     }
 
+    @Ignore("in case of standard-drl, MathProcessor.doOperationNonNumeric() returns false when the left operand is not Comparable. But it doesn't make sense to keep compatibility for this behaviour")
```
hellowdan: Hi Toshiya, thanks for the PR! Would it be possible to add a test (or change this one) that somehow expects NumberFormatException for this scenario instead of ignoring it?

hellowdan (approved):

mariofusco (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/TypeObjectCoercionTest.java
mariofusco: @tkobayas I agree with @hellowdan and in general I think that it doesn't make sense to keep ignored tests in our codebase if we know in advance that they are structurally broken or anyway we can't or won't to fix them. Please change this test according to Daniel's suggestion or if not possible just remove it.

tkobayas: Hi @mariofusco @hellowdan , It means we will change Mvel to throw an Exception (probably RuntimeException or IllegalArgumentException rather than NumberFormatException because it's not only for number) when a left operand is not Comparable? It's already an irregular constraint but the change would have a wide effect.

https://github.com/mvel/mvel/blob/master/src/main/java/org/mvel2/math/MathProcessor.java#L269-L279

And then other engines (Jitted standard-drl and exec-model) throw the same Exception for compatibility.  Am I understanding correctly? I think it's not impossible but we may not want to do it for such a minor use case so shall I remove the test? Please confirm, thanks!

mariofusco: @tkobayas that's too much effort and there are too many corner cases imo. If my understanding is correct and this test is referring to a wrong or misleading behaviour that we don't want to support, please just totally remove the test.

hellowdan: I agree

tkobayas: Okay, removed.

PR #3708, opened 2021-06-30
[DROOLS-6462] exec-model : compilation error when bind var produced b…

tkobayas:
…y accumulate is used in the condition of subsequent accumulate

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-6462

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
@@ -356,6 +356,7 @@ public class RuleContext {
         }
         this.scopedDeclarations.put(d.getBindingId(), d);
         this.allDeclarations.put(d.getBindingId(), d);
+        definedVars.put(bindingId, bindingId);
     }
 
     public void addOOPathDeclaration(DeclarationSpec d) {
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
@@ -46,6 +46,7 @@ import org.drools.modelcompiler.builder.errors.InvalidExpressionErrorResult;
 import org.drools.modelcompiler.builder.generator.TypedExpression;
 import org.drools.modelcompiler.builder.generator.UnificationTypedExpression;
 
+import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.toClassOrInterfaceType;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.toJavaParserType;
 import static org.drools.modelcompiler.util.ClassUtil.toNonPrimitiveType;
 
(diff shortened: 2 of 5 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
```diff
@@ -356,6 +356,7 @@ public void addDeclarationReplacing(DeclarationSpec d) {
         }
         this.scopedDeclarations.put(d.getBindingId(), d);
         this.allDeclarations.put(d.getBindingId(), d);
+        definedVars.put(bindingId, bindingId);
```
tkobayas, line 359: The issue was that `RuleContext.getVar()` was not able to find the bind variable `$maxName`  in `RuleContext.definedVars` so it renamed `$maxName` to `var_$maxName_2_sCoPe` which doesn't exist.

So my fix is to simply put the bind variable name to `RuleContext.definedVars`. I did it only in `RuleContext.addDeclarationReplacing()` to solve the test cases.

triceo:
@tkobayas Coincidentally, I filed an issue today that looks awfully similar:
https://issues.redhat.com/browse/DROOLS-6466

Would you mind double-checking if it's the same thing? The reproducer is attached.

tkobayas:
@triceo  Yes, DROOLS-6466 is solved by the same fix. I'm adding your unit test. Thanks!

tkobayas:
@mariofusco @hellowdan @pibizza  Please review, thanks!

mariofusco (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
pibizza: Two small doubts

-  should we remove the bindingId from the definedVars when we call the removeDeclarationById?
- should we add the bindingId to the definedVars when we call addDeclaration?
-

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
```diff
@@ -180,6 +181,8 @@ private static TypedExpression coerceToString(TypedExpression typedExpression) {
             coercedExpression = typedExpression.cloneWithNewExpression(new MethodCallExpr(new NameExpr("String"), "valueOf", NodeList.nodeList(expression)));
         } else if (typedExpression.getType() == Object.class) {
             coercedExpression = typedExpression.cloneWithNewExpression(new MethodCallExpr(expression, "toString"));
+        } else if (expression instanceof NameExpr) {
+                coercedExpression = typedExpression.cloneWithNewExpression(new CastExpr(toClassOrInterfaceType(String.class), expression));
         } else {
```
pibizza, line 186: Since the CoercedExpressionTest already exists, we could add this new case as a test case, unless it is too complicated.

tkobayas:
Thanks @pibizza , I filed a JIRA to revisit your points: https://issues.redhat.com/browse/DROOLS-6474

PR #3723, opened 2021-07-07
[DROOLS-6402] make ChainedProperties in RuleBaseConfiguration transient

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6402

Under a security point of view I agree with the issue reporter that the `System.properties` shouldn't be part of the serialized `KieBase`. I found that this can be easily achieved by completely avoiding the serialization of the `ChainedProperties` inside the `RuleBaseConfiguration`. In fact by doing so the `ChainedProperties` is automatically recreated from scratch in the context, and then using the `System.properties`, of the JVM where the `KieBase` gets deserialized. This could be more correct in some situation and less in other, I don't have a strong opinion on this, but anyway I will document this change in case we will decide to merge this PR. The other issue that I see is that this change will break all existing serialized `KieBase`s, making them incompatible with the next release. We never guaranteed seriialized `KieBase` compatibility between different versions (which is also the reason why we want to discourage the use of seriialized `KieBase` regardless of this issue), so I believe that this also won't be a issue.

@tkobayas what do you think?

### DIFF
--- a/drools-core/src/main/java/org/drools/core/RuleBaseConfiguration.java
+++ b/drools-core/src/main/java/org/drools/core/RuleBaseConfiguration.java
@@ -127,7 +127,7 @@ public class RuleBaseConfiguration
 
     protected static final transient Logger logger = LoggerFactory.getLogger(RuleBaseConfiguration.class);
 
-    private transient ChainedProperties chainedProperties;
+    private ChainedProperties chainedProperties;
 
     private boolean immutable;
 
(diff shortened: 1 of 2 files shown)


tkobayas (approved):
@mariofusco I agree with the change. Also we have already stated that KieBase serialization is not supported. So compatibility is not an issue.

https://docs.jboss.org/drools/release/latest/drools-docs/html_single/#_thekmodulexmlfile
```
KieBase and KiePackage don’t support serialization since Drools 6. You need to build KieBase through KieContainer. On the other hand, KieSession can be marshalled/unmarshalled by KieMashaller. See Marshalling.
```

Btw, I don't have a permission to access DROOLS-6402. It's okay if it doesn't have further information about this topic.

PR #3732, opened 2021-07-15
[DROOLS-6492] 2nd incremental update causes misfiring of a rule with …

tkobayas:
…'from'

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6492

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
+++ b/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.compiler.integrationtests.incrementalcompilation;
+
+import java.io.StringReader;
+import java.util.Collection;
+
+import org.drools.compiler.kie.builder.impl.DrlProject;
+import org.drools.core.impl.InternalKieContainer;
+import org.drools.modelcompiler.ExecutableModelProject;
+import org.drools.testcoverage.common.util.KieBaseTestConfiguration;
+import org.drools.testcoverage.common.util.TestParametersUtil;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.kie.api.KieServices;
+import org.kie.api.builder.KieBuilder;
+import org.kie.api.builder.KieFileSystem;
+import org.kie.api.builder.ReleaseId;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieContainer;
+import org.kie.api.runtime.KieSession;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+@RunWith(Parameterized.class)
+public class MultipleIncrementalCompilationTest {
+
+    private final KieBaseTestConfiguration kieBaseTestConfiguration;
+
+    public MultipleIncrementalCompilationTest(final KieBaseTestConfiguration kieBaseTestConfiguration) {
+        this.kieBaseTestConfiguration = kieBaseTestConfiguration;
+    }
+
+    @Parameterized.Parameters(name = "KieBase type={0}")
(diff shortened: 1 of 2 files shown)


tkobayas:
Jenkins run fdb

tkobayas:
On the 3rd run (= 2nd incremental update), FromNode.resultClass's ClassLoader is old so the list element objects don't match the resultClass.

Why this happens on the 3rd run? Because

- The 1st run : KnowledgePackageImpl doesn't have typeResolver.
- The 2nd run : KieBaseUpdaterImpl lets KnowledgePackageImpl create a ClassTypeResolver which caches classes in cachedImports.
- The 3rd run : KieBaseUpdaterImpl reuses the KnowledgeBaseImpl, KnowledgePackageImpl and the ClassTypeResolver.

TypeResolver.cachedImports may return a Class with an old ClassLoader but KieBaseUpdaterImpl calls kbuilder.rewireAllClassObjectTypes() to solve the classloader mismatch. The problem is that FromNode holds resultClass as a bare Class object which is not reached by wiring.

mariofusco (approved):

hellowdan (approved):

tkobayas:
FDB failure is not related
```
org.kie.karaf.itest.InstallFeaturesKarafIntegrationTest.testInstallFeature[13: jbpm-human-task]
```

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
```diff
@@ -0,0 +1,173 @@
...
+        final String personListSrc3 = PERSON_LIST_COMMON_SRC +
+                                      "    public void someMethod1() { }\n" +
+                                      "    public void someMethod2() { }\n" +
+                                      "}";
+
+        final KieServices ks = KieServices.Factory.get();
+        final ReleaseId releaseId = ks.newReleaseId("org.test", "myTest", "1.0.0-SNAPSHOT");
+
+        // 1st run
+        buildKjar(releaseId, personListSrc1);
```
pibizza, line 93: I like this test a lot. There is a clear separation of the issue - different classloading - and the implementation details on how we arrive there.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
```diff
@@ -0,0 +1,173 @@
...
+        kfs.write(ks.getResources()
+                    .newReaderResource(new StringReader(drl))
+                    .setResourceType(ResourceType.DRL)
+                    .setSourcePath("org/test/rules.drl"));
+
+        if (kieBaseTestConfiguration.isExecutableModel()) {
+            kieBuilder.buildAll(ExecutableModelProject.class);
+        } else {
+            kieBuilder.buildAll(DrlProject.class);
+        }
```
pibizza, line 162: A minor point. Would it be possible to remove the if and use a parameter that is passed during the test? This would make the test cleaner. 

However this is a minor point.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+        assertFalse(kieBuilder.getResults().hasMessages(org.kie.api.builder.Message.Level.ERROR));
+    }
+
+    private void runRules(KieContainer kieContainer) {
+        final KieSession ksession = ((InternalKieContainer) kieContainer).getKieSession(); // use the same ksession
```
pibizza, line 168: Other minor point - since the real issue here is to reuse  the same ksession, could you just put it in a field of the test and reuse it in the method? This would remove the need to pass the kieContainer. 

If you instead want to be explicit that you are reusing the same ksession, I suggest putting the kSession in a local variable of the test and passing the kSession to the method runRules.

pibizza (changes requested):
This is clearly an integration test - we have to play with classloading and the effects require a network of objects in place to perform the test.

I have added two comments, but they are minor points.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
tkobayas: @pibizza Thanks, I think I will be able to add such a method in KieUtil.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/incrementalcompilation/MultipleIncrementalCompilationTest.java
tkobayas: @pibizza Yes, it's a possible approach. But in this test, I'm simulating a kie-server use case (which was originally reported in https://issues.redhat.com/browse/RHDM-1781). Eventually it's the same ksession but kieContainer.getKieSession() is also a part of the integration test scenario.

PR #3734, opened 2021-07-16
[DROOLS-6498] synchronize evalation of mvel constraint till the AST i…

mariofusco:
…s not fully materialized

**JIRA**: https://issues.redhat.com/browse/DROOLS-6498


lucamolteni (approved):

drools-mvel/src/main/java/org/drools/mvel/MVELConditionEvaluator.java
```diff
@@ -160,42 +159,54 @@ private void evaluateIfNecessary( InternalFactHandle handle, InternalWorkingMemo
...
+        return isEvaluated( getRootNode( executableStatement ) );
     }
 
-    private boolean isEvaluated(ASTNode node) {
+    private static boolean isEvaluated(ASTNode node) {
         node = unwrapSubstatement(node);
         if (node instanceof Contains) {
             return ((Contains)node).getFirstStatement().getAccessor() != null;
         }
         return node instanceof BinaryOperation ? ((BooleanNode) node).getLeft().getAccessor() != null : node.getAccessor() != null;
```
tkobayas: for example, in case of ConstraintWithAndOrConcurrencyTest's expression : `TOSTRING(karaoke.dvd[$title].artist) == "BBB" || TOSTRING(karaoke.dvd[$title].artist) >= "01" && TOSTRING(karaoke.dvd[$title].artist) <= "39" `, its root node is `Or`. `Or`'s accessor is always null even if the constraint is optimized. (e.g. by the 1st optimization, accessor of `Or`'s `left`'s `left` is populated). It means `SynchronizedTillEvaluated.evaluate()` is always synchronized for constraints which contain `&&` or `||`. Is it intentional?

drools-mvel/src/main/java/org/drools/mvel/expr/MvelEvaluator.java
```diff
@@ -26,23 +27,46 @@
...
 import org.slf4j.LoggerFactory;
 
+import static org.drools.mvel.MVELConditionEvaluator.isFullyEvaluated;
+
 public class MvelEvaluator<T> {
-    private static final boolean THREAD_SAFE = true;
+
+    private static final boolean THREAD_UNSAFE = Boolean.valueOf( System.getProperty("drools.mvel.thread.unsafe") );
+
+    private static final EvaluatorType DEFAULT_EVALUATOR_TYPE = EvaluatorType.THREAD_SAFE_ON_FIRST_EVAL;
```
tkobayas: It's great that we can choose from 4 types. But why don't we set `DEFAULT_EVALUATOR_TYPE` to `SYNCHRONIZED_TILL_EVALUATED` which is safer? I filed a consequence test PR which fails with `THREAD_SAFE_ON_FIRST_EVAL`.

https://github.com/mariofusco/drools/pull/17

If you choose `THREAD_SAFE_ON_FIRST_EVAL` as default for performance reasons, we can make `DEFAULT_EVALUATOR_TYPE` configurable for users. WDYT?

drools-mvel/src/main/java/org/drools/mvel/MVELConstraint.java
```diff
@@ -267,38 +265,10 @@ protected boolean evaluate(InternalFactHandle handle, InternalWorkingMemory work
...
-                            } finally {
-                                mvelOptimized.countDown();
-                            }
-                            if (invocationCounter.getAndIncrement() == jittingThreshold) {
-                                jitEvaluator(handle, workingMemory, tuple);
-                            }
-                            return result;
-                        }
-                    }
+                    conditionEvaluator = createMvelConditionEvaluator(workingMemory);
```
tkobayas, line 271: Now we don't synchronize `createMvelConditionEvaluator()`. It means that `MVELConstraint.conditionEvaluator` can be swapped under concurrent requests. I confirmed that it doesn't return wrong values in `ConstraintConcurrencyTest` but it causes multiple `compileExpression` calls and optimizations for the same `MVELConstraint`. Isn't better to synchronize it?  (Removing `mvelOptimized` is completely fine)

mariofusco: I believe that the creation of multiple instances for that `conditionEvaluator` doesn't cause any a problem other than being wasteful. Moreover this happens only in the very specific use case that we're trying to solve which is not common at all. I prefer to have it not synchronized and then as fast as possible in 99.9% and pay the cost of (harmless) multiple creations for this scenario.

tkobayas: Sure, understood.

drools-mvel/src/main/java/org/drools/mvel/MVELConditionEvaluator.java
mariofusco: This should be fixed.

drools-mvel/src/main/java/org/drools/mvel/expr/MvelEvaluator.java
mariofusco: Checking if an mvel constraint has been fully evaluated is also a time consuming task. At the moment that check works only for LHS constraints and making it to work for any valid mvel block of code will make it even more expensive. I was hoping of doing this only for constraints, but I see your point that this could break mvel consequences. If we also want this case and play safe, ma suggestion is to just give user the possibility of having a fully synchronized evaluation.

tkobayas: Yes, it makes sense to be configurable.

tkobayas (approved):

mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/3756

PR #3737, opened 2021-07-19
[DROOLS-6493] hardwired kie conf

mariofusco:
https://issues.redhat.com/browse/DROOLS-6493


mariofusco:
jenkins execute fdb

mariofusco:
We decided to solve the problem moving kie.conf files in different folders.

PR #3743, opened 2021-07-22
DROOLS-6511: Use ConcurrentHashMap for FactHandleMap

jomarko:

[DROOLS-6511](https://issues.redhat.com/browse/DROOLS-6511)

* https://github.com/kiegroup/drools-wb/pull/1505

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/ClassAwareObjectStore.java
+++ b/drools-core/src/main/java/org/drools/core/common/ClassAwareObjectStore.java
@@ -26,7 +26,6 @@ import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.Lock;
 
@@ -609,7 +608,7 @@ public class ClassAwareObjectStore implements Externalizable, ObjectStore {
         public FactHandleMap() { }
 
         public FactHandleMap(boolean identity) {
-            facts = new ConcurrentHashMap<>(identity ? new IdentityHashMap<>() : new HashMap<>());
+            facts = identity ? new IdentityHashMap<>() : new HashMap<>();
         }
 
         @Override
--- a/drools-core/src/main/java/org/drools/core/common/ClassAwareObjectStore.java
+++ b/drools-core/src/main/java/org/drools/core/common/ClassAwareObjectStore.java
@@ -26,6 +26,7 @@ import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


mariofusco:
@jomarko Good catch. I see what you did and why and I believe that your fix is correct, but we'd need a test in Drools to reproduce this problem. I think you could reproduce it by trying to insert facts concurrently from multiple threads into the same KieSession instance. Can you please give a try to write such a test or do you want me to attempt this?

jomarko:
@mariofusco I would like to ask you for help with this as you are the drools expert. My changes fixes the test scenario case, see new drools-wb test passed [1], however it seems it breaks a lot of drools tests [2].



[1]
https://eng-jenkins-csb-business-automation.apps.ocp4.prod.psi.redhat.com/job/KIE/job/master/job/fdb/job/drools-wb-master.fdb/12/testReport/org.drools.workbench.screens.testscenario.backend.server/ScenarioRunnerServiceTest/testMemberOfFormula/

[2]
https://eng-jenkins-csb-business-automation.apps.ocp4.prod.psi.redhat.com/job/KIE/job/master/job/pullrequest/job/drools-master.pr/421/

mariofusco:
@jomarko Digging into our codebase I found a whole [set of tests](https://github.com/kiegroup/drools/blob/master/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/concurrency/BasicConcurrentInsertionsTest.java) checking exactly what I suggested in my former comment and indeed I would have been surprised if we didn't have any. I believe (but this is just wild guess) that the workbench is bypassing the synchronizations in the session and entry-point and directly accessing to the object store in some way. If this is true we either need to prevent wb from doing this or at least recreate the same situation with a test in drools. If it isn't we however need to recreate the problem in drools in some way otherwise it will impossible for me to validate this change and I don't want to introduce a concurrent map in the core engine if not strictly necessary.

jomarko:
@mariofusco thank you for response. I understand we should minimize the risk and try the keep the scope of changes as small as posisble. I reverted the `ClassAwareObjectStore` changes and did basically the same however just for workbench legacy test scenarios. Is such a solution acceptable? Or it is rather a masking real issue?

Rikkola (approved):

mariofusco:
@jomarko ok, now I see your second commit and better understand why you got that `ConcurrentModificationException`: you're deleting facts from the session while iterating the list of the same facts provided by the session. Your fix is ok for me, but I bet that the following could also work, can you please give it a try?
```
List<FactHandle> facts = new ArrayList<>(ksession.getFactHandles());
for (FactHandle factHandle : facts) {
    ksession.delete(factHandle);
}
```

jomarko:
@mariofusco thank you, I pushed the simplified version of the fix

mariofusco (approved):

jomarko:
Jenkins do cdb

PR #3745, opened 2021-07-22
[DROOLS-6508] Implement Aggregation

gitgabrio:
@danielezonca @kostola @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-6508

To be viewed/merged after #3740

PR #3751, opened 2021-07-23
[DROOLS-6512] Split BUILTIN_FUNCTIONS in specific types

gitgabrio:

See https://issues.redhat.com/browse/DROOLS-6512

Some notes
1) I wanted to do a conservative split and not modify "public API" of BUILTIN_FUNCTIONS
2)  it would be possible to use a sort of map <String, Function> to avoid creation of "sub-type" enums, but this way seems more readable IMO

To be viewed/merged after https://github.com/kiegroup/drools/pull/3745


gitgabrio:
Included in #3772

PR #3756, opened 2021-07-26
[DROOLS-6498] synchronize evalation of mvel constraint till the AST is not fully materialized

mariofusco:
JIRA: https://issues.redhat.com/browse/DROOLS-6498


tkobayas (approved):
Thanks!

PR #3767, opened 2021-07-28
[DROOLS-6480] emit a proper compile time error when the setters in a …

mariofusco:
…modify block are not comma separated using the executable model

**JIRA**: https://issues.redhat.com/browse/DROOLS-6480

### DIFF
--- a/drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
+++ b/drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
@@ -817,7 +817,7 @@ public class DroolsMvelParserTest {
 
     @Test
     public void testModifyMultiple() {
-        String expr = "{ modify($p) { setAge(1)," + newLine() + " setAge(2);setAge(3)" + newLine() + "setAge(4); }; }";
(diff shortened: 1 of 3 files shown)


drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
```diff
@@ -817,7 +817,7 @@ public void testModifySemiColon() {
 
     @Test
     public void testModifyMultiple() {
-        String expr = "{ modify($p) { setAge(1)," + newLine() + " setAge(2);setAge(3)" + newLine() + "setAge(4); }; }";
+        String expr = "{ modify($p) { setAge(1)," + newLine() + " setAge(2), setAge(3)," + newLine() + "setAge(4); }; }";
```
lucamolteni, line 820: Shouldn't we put a unit test for semi colon as well?

lucamolteni (changes requested):

drools-model/drools-mvel-parser/src/test/java/org/drools/mvel/parser/DroolsMvelParserTest.java
mariofusco: The semicolon is not allowed there (according to drl specs) and the test that I added in the executable model demonstrates the compilation failure it causes.

lucamolteni (approved):

hellowdan (approved):

PR #3773, opened 2021-08-01
[DROOLS-6531] fix OR of 2 accumulates in executable model

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6531

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/KiePackagesBuilder.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/KiePackagesBuilder.java
@@ -153,6 +153,7 @@ import static java.util.stream.Collectors.toList;
 
 import static org.drools.compiler.rule.builder.RuleBuilder.buildTimer;
 import static org.drools.core.rule.GroupElement.AND;
+import static org.drools.core.rule.GroupElement.OR;
 import static org.drools.core.rule.Pattern.getReadAcessor;
 import static org.drools.model.DSL.declarationOf;
 import static org.drools.model.DSL.entryPoint;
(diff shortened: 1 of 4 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/KiePackagesBuilder.java
```diff
@@ -690,12 +691,18 @@ private ConditionalBranch buildConditionalConsequence(RuleContext ctx, Condition
...
         for (int i = 0; i < subconditions.size(); i++) {
             RuleConditionElement element = conditionToElement( ctx, ge, subconditions.get(i) );
             if (element != null) {
                 ge.addChild( element );
             }
         }
+        if (ge.getType() == OR) {
+            ctx.endOrCondition();
+        }
         if (ge.getType() == AND && ge.getChildren().size() == 1) {
```
pibizza, line 706: Is it possible to have "or" conditions at multiple levels? I am wondering if the end can be called when we have a complex condition.

pibizza (commented):
As a general comment we miss unit tests. I understand that many of these classes have no unit test, and they require lots of initial effort to be put in a unit test fixture. On the other side, once the fixture is in place adding another test is typically quite simple.

I suggest to identify the classes that we know we are going to modify a lot, and add fixtures and initial unit tests for them.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/KiePackagesBuilder.java
mariofusco: @pibizza it is possible to have nested ORs and then I added in the `RuleContext` class a stack of List of Variables to take count of this situation.

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/KiePackagesBuilder.java
pibizza: Ok, of course I didn't notice that. Thanks. Looks good to me.

pibizza (approved):

hellowdan (approved):

PR #3781, opened 2021-08-04
[JBPM-9841] Start process with correlation key rest api should return…

abhijithumbe:
… 400 http status if correlation key already exist

**JIRA**: _(https://issues.redhat.com/browse/JBPM-9841)_ 

**referenced Pull Requests**: _(https://github.com/kiegroup/droolsjbpm-integration/pull/2564)_

### DIFF
--- a/drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
+++ b/drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
@@ -30,6 +30,7 @@ import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
 
 import org.drools.core.command.IdentifiableResult;
 import org.drools.core.runtime.impl.ExecutionResultImpl;
+import org.drools.core.util.process.StartCorrelatedProcess;
 import org.drools.core.xml.jaxb.util.JaxbMapAdapter;
 import org.kie.api.command.ExecutableCommand;
 import org.kie.api.runtime.Context;
@@ -38,7 +39,6 @@ import org.kie.api.runtime.process.ProcessInstance;
 import org.kie.internal.command.CorrelationKeyCommand;
 import org.kie.internal.command.RegistryContext;
 import org.kie.internal.jaxb.CorrelationKeyXmlAdapter;
-import org.kie.internal.process.CorrelationAwareProcessRuntime;
 import org.kie.internal.process.CorrelationKey;
 
 @XmlRootElement
@@ -143,7 +143,7 @@ public class StartCorrelatedProcessCommand implements ExecutableCommand<ProcessI
                 ksession.insert(o);
             }
         }
-        ProcessInstance processInstance = ((CorrelationAwareProcessRuntime)ksession).startProcess(processId, correlationKey, parameters);
+        ProcessInstance processInstance = StartCorrelatedProcess.startProcess(ksession, processId, correlationKey, parameters);
         if ( this.outIdentifier != null ) {
             ((RegistryContext) context).lookup( ExecutionResultImpl.class ).setResult(this.outIdentifier, processInstance.getId());
         }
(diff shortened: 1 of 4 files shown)


drools-core/src/main/java/org/drools/core/command/runtime/process/StartProcessFromNodeIdsCommand.java
```diff
@@ -101,7 +102,19 @@ public ProcessInstance execute(Context context) {
         if (correlationKey == null) {
             processInstance = ksession.startProcessFromNodeIds(getProcessId(), getParameters(), ids);
         } else {
-            processInstance = ((CorrelationAwareProcessRuntime) ksession).startProcessFromNodeIds(getProcessId(), correlationKey, getParameters(), ids);
+            try {
+                processInstance = ((CorrelationAwareProcessRuntime) ksession).startProcessFromNodeIds(getProcessId(), correlationKey, getParameters(), ids);
+            } catch (RuntimeException e) {
+                Throwable t = e.getCause();
+                while ((t != null) && !(t instanceof ConstraintViolationException)) {
```
fjtirado: unneeded parentheses

drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
```diff
@@ -139,11 +140,25 @@ public ProcessInstance execute(Context context) {
...
-            for (Object o: data) {
+            for (Object o : data) {
                 ksession.insert(o);
             }
         }
-        ProcessInstance processInstance = ((CorrelationAwareProcessRuntime)ksession).startProcess(processId, correlationKey, parameters);
+        ProcessInstance processInstance = null;
+        try {
+            processInstance = ((CorrelationAwareProcessRuntime) ksession).startProcess(processId, correlationKey, parameters);
+        } catch (RuntimeException e) {
```
fjtirado: Put this block inside an utility method that is shared by both commands

drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
```diff
@@ -139,11 +140,25 @@ public ProcessInstance execute(Context context) {
         KieSession ksession = ((RegistryContext) context).lookup( KieSession.class );
 
         if (data != null) {
-            for (Object o: data) {
+            for (Object o : data) {
```
fjtirado: unneeded?

fjtirado (commented):
Besides minor comments, a general one, I think rather than including hibernate dependency, you can search for a SQLException as cause exception (it will be wrapperd inside hibernate one) and check getSQLState againts this table https://en.wikipedia.org/wiki/SQLSTATE
You probably needs to check which one is used when an object already exist. 
With this approach we avoid hibernate dependency (it is safer to assume DB is relational than we are using hibernate, which is kind of hack)

drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
abhijithumbe: new class file added to avoid duplicate code.

drools-core/src/main/java/org/drools/core/command/runtime/process/StartCorrelatedProcessCommand.java
abhijithumbe: This was done during formatting of new code. Reverted this change.

abhijithumbe:
> Besides minor comments, a general one, I think rather than including hibernate dependency, you can search for a SQLException as cause exception (it will be wrapperd inside hibernate one) and check getSQLState againts this table https://en.wikipedia.org/wiki/SQLSTATE
> You probably needs to check which one is used when an object already exist.
> With this approach we avoid hibernate dependency (it is safer to assume DB is relational than we are using hibernate, which is kind of hack)

As suggested used SQLException and SQLState to check exception.

elguardian:
@abhijithumbe  exceptions treatment should be transformed properly in the rest endpoint not here.

please do the transformation in here
https://github.com/kiegroup/droolsjbpm-integration/pull/2564

elguardian:
closing this one. as it is superseeded by the integration PR

PR #3784, opened 2021-08-10
[DROOLS-6351] PredicateInformation contains only one rule name even w…

tkobayas:
…hen shared

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6351

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
@@ -35,12 +35,12 @@ public class MaterializedLambdaPredicateTest {
 
         //language=JAVA
         String expectedResult = "" +
-                "package org.drools.modelcompiler.util.lambdareplace.PEE;\n" +
+                "package org.drools.modelcompiler.util.lambdareplace.PDB;\n" +
                 "import static rulename.*; " +
                 "import org.drools.modelcompiler.dsl.pattern.D; " +
                 "" +
                 "@org.drools.compiler.kie.builder.MaterializedLambda() " +
-                "public enum LambdaPredicateEE863708F0E52E1DD7FBE6537EB76024 implements org.drools.model.functions.Predicate1<org.drools.modelcompiler.domain.Person>, org.drools.model.functions.HashedExpression {\n" +
+                "public enum LambdaPredicateDB79C39D8DCF73551C2CE63B976491B8 implements org.drools.model.functions.Predicate1<org.drools.modelcompiler.domain.Person>, org.drools.model.functions.HashedExpression {\n" +
                 " INSTANCE; \n" +
                 "public static final String EXPRESSION_HASH = \"4DEB93975D9859892B1A5FD4B38E2155\";" +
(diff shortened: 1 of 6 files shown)


tkobayas:
Hi @lucamolteni 

I implemented DROOLS-6351 for `PredicateInformation` so now the error message can print like this when a constraint is shared by multiple rules.
```
Error evaluating constraint 'money < salary * 20' in [Rule "R1", "R2" in r0.drl]
```

In this PR, this is available only for exec-model. So non-exec-model still prints only one rule name.

```
"Error evaluating constraint 'money < salary * 20' in [Rule "R1" in r0.drl]",
```

I wrote another PR which covers non-exec-model.

https://github.com/kiegroup/drools/pull/3785

Concerns of that PR are:
- now `drools-canonical-model` depends on `drools-core` because I want to share some util logic
- We don't want to implement non critical improvement to non-exec-model at this stage?

Which one do you prefer? (Or we may merge exec-model first and consider non-exec-model later)  I'd like to hear your thoughts on this.

Thanks!

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
```diff
@@ -52,8 +52,11 @@ public void createClassWithOneParameter() {
                 "        }\n" +
                 "        @Override()\n" +
                 "        public org.drools.model.functions.PredicateInformation predicateInformation() {\n" +
-                "            return new org.drools.model.functions.PredicateInformation(\"p.age > 35\", \"rule1\", \"rulefilename.drl\");" +
+                "            org.drools.model.functions.PredicateInformation info = new org.drools.model.functions.PredicateInformation(\"p.age > 35\");\n" +
```
lucamolteni, line 55: it it possible to change this test so that is shows there will be n `info.addRuleName` instead of one?

lucamolteni (approved):
Great job! Minor comment aside

lucamolteni:
Great job! 

As for which one PR to merge, I think that having the same behaviour in both engines is better, especially because you've already done and it's ready to merge.

Having said that, I think that it was made on purpose to make the `drools-canonical-model` not to depend on `drools-core`.  I think the idea was to have a model of an evaluation engine not bounded to a specific implementation. It's probably over-engineering, but I'd keep it this way as we're talking about an helper method.

One way to break this dependency could be to expose an interface and use the service loader (`drools-core`'s `kie.conf`) to inject the implementation of `MessageUtils`. I've done it in the past for example with the `drools-traits` and it worked pretty good. 

Let me know what you think

tkobayas:
@lucamolteni  Thanks! I'll go forward with https://github.com/kiegroup/drools/pull/3785 applying your suggestions.

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
tkobayas: Sure, will do.

tkobayas:
Superseded by https://github.com/kiegroup/drools/pull/3785

PR #3785, opened 2021-08-11
[DROOLS-6351] PredicateInformation contains only one rule name even when shared

tkobayas:
**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6351

**referenced Pull Requests**: 

This is another version of https://github.com/kiegroup/drools/pull/3784
This time, I implement the *better error message* (print all associated rule names and rule file names) for non-exec-model as well.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/util/MessageUtils.java
+++ b/drools-core/src/main/java/org/drools/core/util/MessageUtils.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.core.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class MessageUtils {
+
+    private MessageUtils() {}
+
+    public static String defaultToEmptyString(String str) {
+        return Optional.ofNullable(str).orElse("");
+    }
+
+    public static String formatConstraintErrorMessage(String expression, Map<String, Set<String>> ruleNameMap) {
+        return String.format("Error evaluating constraint '%s' in %s", expression, formatRuleNameMap(ruleNameMap));
+    }
+
+    public static String formatRuleNameMap(Map<String, Set<String>> ruleNameMap) {
+        if (ruleNameMap == null || ruleNameMap.isEmpty()) {
+            return "";
+        }
+        final StringBuilder sb = new StringBuilder();
+        List<String> ruleFileNameList = ruleNameMap.keySet().stream().sorted().collect(Collectors.toList());
+        ruleFileNameList.forEach(ruleFileName -> {
+            sb.append("[Rule ");
+            ruleNameMap.get(ruleFileName).stream().sorted().forEach(ruleName -> {
+                sb.append("\"" + ruleName + "\", ");
+            });
+            if (!ruleFileNameList.isEmpty()) {
+                sb.delete(sb.length() - 2, sb.length());
+            }
+            sb.append(" in " + ruleFileName + "] ");
+        });
+        sb.deleteCharAt(sb.length() - 1);
+        return sb.toString();
+    }
+}
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
@@ -389,14 +389,15 @@ public abstract class AbstractExpressionBuilder {
     protected String createExprId(SingleDrlxParseSuccess drlxParseResult) {
         String exprId = drlxParseResult.getExprId(context.getPackageModel().getExprIdGenerator());
 
-        context.getPackageModel().indexConstraint(exprId, new PredicateInformation(
-                drlxParseResult.getOriginalDrlConstraint(),
-                context.getRuleName(),
-                Optional.ofNullable(context.getRuleDescr())
-                    .map(RuleDescr::getResource)
-                    .map(Resource::getSourcePath)
-                    .orElse("")
-        ));
+        String stringConstraint = drlxParseResult.getOriginalDrlConstraint();
+        String ruleName = context.getRuleName();
+        String ruleFileName = Optional.ofNullable(context.getRuleDescr())
+                                      .map(RuleDescr::getResource)
+                                      .map(Resource::getSourcePath)
+                                      .orElse("");
+
+        context.getPackageModel().indexConstraint(exprId, stringConstraint, ruleName, ruleFileName);
+
         return exprId;
     }
 
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/ConstraintEvaluationException.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/ConstraintEvaluationException.java
@@ -1,28 +1,31 @@
 /*
- * Copyright (c) 2020. Red Hat, Inc. and/or its affiliates.
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- *       http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
+ *
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
 
-package org.drools.mvel;
+package org.drools.modelcompiler.constraints;
 
-import org.drools.mvel.MVELConstraint.EvaluationContext;
+import org.drools.model.functions.PredicateInformation;
 
 import static org.drools.core.util.MessageUtils.formatConstraintErrorMessage;
 
 public class ConstraintEvaluationException extends RuntimeException {
 
-    private static final long serialVersionUID = -3413225194510143529L;
+    private static final long serialVersionUID = 7880877148568087603L;
 
-    public ConstraintEvaluationException(String expression, EvaluationContext evaluationContext, Throwable cause) {
-        super(formatConstraintErrorMessage(expression, evaluationContext.getRuleNameMap()), cause);
+    public ConstraintEvaluationException(PredicateInformation predicateInformation, Throwable cause) {
+        super(formatConstraintErrorMessage(predicateInformation.getStringConstraint(), predicateInformation.getRuleNameMap()), cause);
     }
 }
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
@@ -175,20 +175,28 @@ public class LambdaConstraint extends AbstractConstraint {
         try {
             return evaluator.evaluate(handle, workingMemory);
         } catch (RuntimeException e) {
-            throw predicateInformation.betterErrorMessage(e);
+            throw new ConstraintEvaluationException(predicateInformation, e);
         }
     }
 
     @Override
     public boolean isAllowedCachedLeft(ContextEntry context, InternalFactHandle handle) {
         LambdaContextEntry lambdaContext = ((LambdaContextEntry) context);
-        return evaluator.evaluate(handle, lambdaContext.getTuple(), lambdaContext.getWorkingMemory());
+        try {
+            return evaluator.evaluate(handle, lambdaContext.getTuple(), lambdaContext.getWorkingMemory());
+        } catch (RuntimeException e) {
+            throw new ConstraintEvaluationException(predicateInformation, e);
+        }
     }
 
     @Override
     public boolean isAllowedCachedRight(Tuple tuple, ContextEntry context) {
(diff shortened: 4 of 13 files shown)


drools-core/src/main/java/org/drools/core/util/MessageUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.drools.core.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class MessageUtils {
```
lucamolteni, line 25: See https://github.com/kiegroup/drools/pull/3784#issuecomment-896651405

drools-model/drools-canonical-model/pom.xml
```diff
@@ -18,6 +18,10 @@
   </properties>
 
   <dependencies>
+    <dependency>
```
lucamolteni: See https://github.com/kiegroup/drools/pull/3784#issuecomment-896651405

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
```diff
@@ -182,13 +182,21 @@ public boolean isAllowed(InternalFactHandle handle, InternalWorkingMemory workin
     @Override
     public boolean isAllowedCachedLeft(ContextEntry context, InternalFactHandle handle) {
         LambdaContextEntry lambdaContext = ((LambdaContextEntry) context);
-        return evaluator.evaluate(handle, lambdaContext.getTuple(), lambdaContext.getWorkingMemory());
+        try {
+            return evaluator.evaluate(handle, lambdaContext.getTuple(), lambdaContext.getWorkingMemory());
+        } catch (RuntimeException e) {
```
lucamolteni, line 187: why wasn't this needed in https://github.com/kiegroup/drools/pull/3784/files ?

lucamolteni (changes requested):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/LambdaConstraint.java
tkobayas: I noticed it when I added more test cases :)

tkobayas:
Hi @lucamolteni   , thank you for the suggestion about `kie.conf` solution. However, I feel it's a little too much change just for util class. Instead, I moved the "better error message" call from `PredicateInformation` to `org.drools.modelcompiler.constraints.ConstraintEvaluationException` which exists in `drools-model-compiler` so it can call MessageUtils. I think it's reasonable because Exception class is responsible for "error message". Now `drools-canonical-model` no longer depends on `drools-core`. Do you think it's good?

lucamolteni:
> Hi @lucamolteni , thank you for the suggestion about `kie.conf` solution. However, I feel it's a little too much change just for util class. Instead, I moved the "better error message" call from `PredicateInformation` to `org.drools.modelcompiler.constraints.ConstraintEvaluationException` which exists in `drools-model-compiler` so it can call MessageUtils. I think it's reasonable because Exception class is responsible for "error message". Now `drools-canonical-model` no longer depends on `drools-core`. Do you think it's good?

Perfect - thank you!

hellowdan (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
```diff
@@ -28,19 +28,22 @@
...
         //language=JAVA
         String expectedResult = "" +
-                "package org.drools.modelcompiler.util.lambdareplace.PEE;\n" +
+                "package org.drools.modelcompiler.util.lambdareplace.PA1;\n" +
                 "import static rulename.*; " +
                 "import org.drools.modelcompiler.dsl.pattern.D; " +
                 "" +
                 "@org.drools.compiler.kie.builder.MaterializedLambda() " +
-                "public enum LambdaPredicateEE863708F0E52E1DD7FBE6537EB76024 implements org.drools.model.functions.Predicate1<org.drools.modelcompiler.domain.Person>, org.drools.model.functions.HashedExpression {\n" +
+                "public enum LambdaPredicateA1BF1CB9AB1BC4D6F67E08AE32E64041 implements org.drools.model.functions.Predicate1<org.drools.modelcompiler.domain.Person>, org.drools.model.functions.HashedExpression {\n" +
```
lucamolteni: Just FYI we might use the same approach used here 

https://github.com/kiegroup/drools/blob/cab7fcf290ca6111b688d66aacd7233b14aa2105/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaConsequenceTest.java#L40

to avoid having to change the class name each time

lucamolteni (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/util/lambdareplace/MaterializedLambdaPredicateTest.java
tkobayas: That's nice! I updated MaterializedLambdaPredicateTest with the same approach.

tkobayas:
Jenkins retest this

tkobayas:
Jenkins run fdb

tkobayas:
FBD failures are not related to this PR (found in daily builds).
```
org.jbpm.executor.impl.jms.JmsAvaiableJobExecutorTest.testAsyncAuditErrorProducer

java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.jbpm.executor.impl.jms.JmsAvaiableJobExecutorTest.testAsyncAuditErrorProducer(JmsAvaiableJobExecutorTest.java:153)


org.kie.workbench.common.services.datamodel.backend.server.ModuleDataModelConcurrencyTest.testConcurrentResourceUpdates

WELD-000049: Unable to invoke private void org.guvnor.common.services.builder.ResourceChangeIncrementalBuilder.setup() on org.kie.workbench.common.services.backend.builder.core.ResourceChangeIncrementalBuilder@7e98077c

java.lang.AssertionError: WELD-000049: Unable to invoke private void org.guvnor.common.services.builder.ResourceChangeIncrementalBuilder.setup() on org.kie.workbench.common.services.backend.builder.core.ResourceChangeIncrementalBuilder@7e98077c
	at org.junit.Assert.fail(Assert.java:89)
	at org.kie.workbench.common.services.datamodel.backend.server.ModuleDataModelConcurrencyTest.testConcurrentResourceUpdates(ModuleDataModelConcurrencyTest.java:168)
```

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/ConstraintEvaluationException.java
```diff
@@ -0,0 +1,14 @@
+package org.drools.modelcompiler.constraints;
```
danielezonca, line 18: Missing license

drools-mvel/src/main/java/org/drools/mvel/MVELConstraint.java
```diff
@@ -980,27 +984,54 @@ public void registerEvaluationContext(BuildContext buildContext) {
...
         public void writeExternal(ObjectOutput out) throws IOException {
-            out.writeObject(evaluatedRules);
+            out.writeObject(ruleNameMap);
         }
 
         @Override
         public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
-            evaluatedRules = (Collection<String>) in.readObject();
+            ruleNameMap = (Map<String, Set<String>>) in.readObject();
         }
```
danielezonca, line 1030: This change probably breaks backward compatibility of existing serialised `KieBase`s, I don't know if this is a problem or in general if we consider this aspect 🤔

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/constraints/ConstraintEvaluationExceptionTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public ConstraintEvaluationExceptionTest(RUN_TYPE testRunType) {
+        super(testRunType);
+    }
+
+    @Test
+    public void testMultipleRules() {
+        constraint("age > 20", "R1", "sample.drl");
+        addRuleName("R2", "sample.drl");
+        addRuleName("R3", "sample.drl");
```
danielezonca: I find method names quite misleading: I was expecting that the `addRuleName` method to just add a new rule that is not sharing a constraint with the previous one while looking at the implementation all the methods work on the same field. This also implies that `addRuleName` can only be invoked after `constraint` to avoid a NPE.

I suggest to improve it with one of the following:
1) **Rename `constraint` as `initConstraintTestField` and `addRuleName` as `addRuleToConstraintTestField`**
In this way at least it is a bit more clear that it is performing a side effect on a shared single instance

2) **Making `constraint` returning the instance of the constraint as `Object` (to work with both exec-model and non) and making `addRuleName` accepting this value**
This is still not ideal but at least it will be explicit that `addRuleName` is linked to the previous defined instance

3) **Create a local `RuleIdentifier` class that is a pair of RuleName + RuleFileName and update `constraint` to accept varargs**
The code will become something like
```java
constraint("age > 20", new RuleIdentifier("R1", "sample.drl"), 
      new RuleIdentifier("R2", "sample.drl"), 
      new RuleIdentifier("R3", "sample.drl"));
```

4) **Apply a builder pattern so that `addRuleName` is a method of the builder so it is clear that is working on the same constraint**
This will probably require quite a lot of additional code just to have 
```java
 buildConstraint("age > 20", "R1", "sample.drl")
        .addRuleName("R2", "sample.drl")
        .addRuleName("R3", "sample.drl")
        .build();
```

Any other similar idea :)
I think option 3 could be a good tradeoff but feel free to follow any similar approach. My main point is to make clear that all the tests are referring to the same field and creating a single shared constraint

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/CompilerTest.java
```diff
@@ -2748,6 +2748,106 @@ public void testNPEOnConstraint() {
         ksession.fireAllRules();
     }
 
+    @Test
+    public void testSharedPredicateInformation() {
```
danielezonca, line 2837: Can you add a test where there is at least one rule that has no shared constraint to verify that this rule is not included in the error?

Something like
```java
        exceptionRule.expect(RuntimeException.class);
        exceptionRule.expectMessage(equalTo("Error evaluating constraint 'money < salary * 20' in [Rule \"R1\", \"R3\" in r0.drl]"));

        String str =
                "import " + Person.class.getCanonicalName() + ";" +
                     "rule R1 when\n" +
                     "  $p : Person(money < salary * 20 )\n" +
                     "then\n" +
                     "end\n" +
                     "rule R2 when\n" +
                     "  $p : Person()\n" +
                     "then\n" +
                     "end\n" +
                     "rule R3 when\n" +
                     "  $p : Person(money < salary * 20 )\n" +
                     "then\n" +
                     "end";

```

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
```diff
@@ -389,14 +389,21 @@ protected boolean containsThis(TypedExpression typedExpression) {
...
+                                      .map(Resource::getSourcePath)
+                                      .orElse("");
+
+        Optional<PredicateInformation> optInfo = context.getPackageModel().findConstraintWithExprId(exprId);
+        if (optInfo.isPresent()) {
+            PredicateInformation info = optInfo.get();
+            info.addRuleName(ruleName, ruleFileName);
+        } else {
+            context.getPackageModel().indexConstraint(exprId, new PredicateInformation(stringConstraint, ruleName, ruleFileName));
+        }
```
danielezonca: Quite unrelated to this PR and I don't know how often this "pattern" is used because if this is the only usage probably it doesn't make sense but what about implement a `computeIfAbsent`/`compute` like methods to `PackageModel` to simplify this `getOrCreate` logic.
Wdyt? 
If you think it makes sense feel free to create a ticket otherwise just ignore this comment :)

danielezonca (commented):
Overall looks good to me, just some minor comments

pibizza (approved):
Seems good to me - sorry for taking so much time for the review.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/constraints/ConstraintEvaluationException.java
tkobayas: Thank you. Added.

drools-mvel/src/main/java/org/drools/mvel/MVELConstraint.java
tkobayas: @danielezonca  We have already stated that KieBase serialization is not supported.
https://docs.jboss.org/drools/release/latest/drools-docs/html_single/#_thekmodulexmlfile
```
KieBase and KiePackage don’t support serialization since Drools 6. You need to build KieBase through KieContainer. On the other hand, KieSession can be marshalled/unmarshalled by KieMashaller. See Marshalling.
```
Even though, some users may use KieBase serialization at their own risk, we don't need to worry about its backward compatibility. WDYT, @mariofusco  ?

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/constraints/ConstraintEvaluationExceptionTest.java
tkobayas: @danielezonca  Thanks for suggestions! I chose option 1 :)

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/CompilerTest.java
tkobayas: Sure, added. Thanks.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expression/AbstractExpressionBuilder.java
tkobayas: It's true that this is the only usage but I agree `compute` based implementation is cleaner. So I refactored. Thanks!

tkobayas:
Jenkins run fdb

drools-mvel/src/main/java/org/drools/mvel/MVELConstraint.java
mariofusco: @danielezonca, what @tkobayas wrote is correct. We already discussed `KieBase`s serialization more in general and agreed that we don't consider it a issue.

mariofusco (approved):

PR #3786, opened 2021-08-11
[DROOLS-6545] HalfBinary constraint with || And && causes StackOverfl…

tkobayas:
…owError

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6545

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
@@ -110,7 +110,7 @@ import static java.util.Optional.of;
 import static java.util.stream.Collectors.toList;
 import static org.drools.core.util.MethodUtils.findMethod;
 import static org.drools.modelcompiler.builder.generator.DslMethodNames.PATTERN_CALL;
-import static org.drools.modelcompiler.builder.generator.expressiontyper.ExpressionTyper.findLeftLeafOfNameExpr;
+import static org.drools.modelcompiler.builder.generator.expressiontyper.ExpressionTyper.findLeftLeafOfNameExprTraversingParent;
 import static org.drools.modelcompiler.util.ClassUtil.toRawClass;
 import static org.drools.mvelcompiler.util.TypeUtils.toJPType;
 
@@ -362,7 +362,7 @@ public class DrlxParseUtil {
 
             HalfBinaryExpr halfBinaryExpr = (HalfBinaryExpr) drlxExpr;
 
-            Expression parentLeft = findLeftLeafOfNameExpr( parent.get() );
+            Expression parentLeft = findLeftLeafOfNameExprTraversingParent( parent.get() );
             Operator operator = toBinaryExprOperator(halfBinaryExpr.getOperator());
             return new BinaryExpr(parentLeft, halfBinaryExpr.getRight(), operator);
         }
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
@@ -359,10 +359,8 @@ public class DrlxParseUtil {
     public static Expression trasformHalfBinaryToBinary(Expression drlxExpr) {
         final Optional<Node> parent = drlxExpr.getParentNode();
         if(drlxExpr instanceof HalfBinaryExpr && parent.isPresent()) {
-
             HalfBinaryExpr halfBinaryExpr = (HalfBinaryExpr) drlxExpr;
-
-            Expression parentLeft = findLeftLeafOfNameExprTraversingParent( parent.get() );
+            Expression parentLeft = findLeftLeafOfNameExprTraversingParent( halfBinaryExpr );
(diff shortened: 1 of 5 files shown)


tkobayas:
The github action test failures are not related to this PR:
```
kiegroup_drools/drools-persistence/drools-persistence-jpa/src/test/java/org/drools/persistence/timer/integrationtests/TimerAndCalendarTest.java:1553 | expected:<2> but was:<1>
```

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expressiontyper/ExpressionTyper.java
```diff
@@ -1156,6 +1156,18 @@ private static String getFieldName( Expression drlxExpr, SimpleName fieldName )
         return fieldName.getIdentifier();
     }
 
+    public static Expression findLeftLeafOfNameExprTraversingParent(Node expression) {
```
lucamolteni, line 1163: Check if you can use `DrlxParseUtil.findRootNodeViaParent`

or instead `node.findAncestor()` which is the JavaParser method we could have used instead (but I didn't know of its existence at the time)

lucamolteni (changes requested):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/expressiontyper/ExpressionTyper.java
tkobayas: Thank you for the suggestion, @lucamolteni 

`node.findAncestor()` seems to be a good fit so I rewrite the code with the method. Thanks! (`DrlxParseUtil.findRootNodeViaParent` doesn't fit because we want to traverse from closer parents, not jumping to the root parent. See `HalfBinaryTest.testNestedHalfBinaryOrAndAmpersand`)

hellowdan (approved):

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/DrlxParseUtil.java
```diff
@@ -359,10 +359,8 @@ public static Node replaceAllHalfBinaryChildren(Node parent) {
     public static Expression trasformHalfBinaryToBinary(Expression drlxExpr) {
         final Optional<Node> parent = drlxExpr.getParentNode();
         if(drlxExpr instanceof HalfBinaryExpr && parent.isPresent()) {
-
             HalfBinaryExpr halfBinaryExpr = (HalfBinaryExpr) drlxExpr;
-
-            Expression parentLeft = findLeftLeafOfNameExpr( parent.get() );
+            Expression parentLeft = findLeftLeafOfNameExprTraversingParent( halfBinaryExpr );
```
danielezonca, line 363: @tkobayas 
I don't fully understand the fix: as far as I can see in the [Jira ticket](https://issues.redhat.com/browse/DROOLS-6545), the method `ExpressionTyper.toTypedExpressionRec` can produce a StackOverflow recursion in certain conditions (based on the input).
So I was expecting to see some change in that method to prevent (even throwing an IllegalArgumentException if necessary) it but as far as I can understand this fix introduces a new path that does a "preliminary" navigation/search to avoid to trigger this SO.

Said differently: is it possible to trigger the same SO invoking the existing `findLeftLeafOfNameExpr` function with a wrong param? If so I think we should fix this too.

tkobayas: @danielezonca  Thank you for reviewing. Before the fix, `HalfBinaryTest.testHalfBinaryOrAndAmpersand` was executed like:
1. Parse binaryExpr `> 20  && < 30`
2. Call `toTypedExpression` for its `left`, which is `> 20`
3. Call `toTypedExpressionRec` for `> 20`
4. Call `trasformHalfBinaryToBinary` for `> 20`
5. Call `findLeftLeafOfNameExpr` for the parent `> 20 && < 30`
6. `findLeftLeafOfNameExpr` finds `> 20` as the left leaf
7. `trasformHalfBinaryToBinary` combines the left leaf `> 20` with the original HalfBinaryExpr into a BinaryExpr. It results in `> 20 > 20`. (Bad!)
8. Call `toTypedExpressionRec` for `> 20 > 20`
9. Call `toTypedExpressionRec` for its `left`, which is `> 20`
 -> loops from step 3 to step 9 (This time, the parent is `> 20 > 20`)

The fix is to traverse parents until it finds an Expr (which is not a HalfBinaryExpr) so that it can construct an expected BinaryExpr.

> So I was expecting to see some change in that method to prevent (even throwing an IllegalArgumentException if necessary) it but as far as I can understand this fix introduces a new path that does a "preliminary" navigation/search to avoid to trigger this SO.
> Said differently: is it possible to trigger the same SO invoking the existing findLeftLeafOfNameExpr function with a wrong param? If so I think we should fix this too.

`findLeftLeafOfNameExpr` alone doesn't trigger the SO. But as suggested, I added checks to prevent from StackOverflow in `toTypedExpressionRec`. Also expanded the test coverage to HalfPointFreeExpr cases.

Thanks!

danielezonca (approved):
Thanks for the PR 👍

PR #3790, opened 2021-08-19
[DROOLS-6547] NullSafeDereferencing misses null check in LambdaExtrac…

tkobayas:
…tor for index with exec-model

- Make sure to place null-check constraint at the beginning of the constraint if possible. So it will be an AlphaNode so index of the main constraint will not be used when the field is null (same as non-exec-model).
-  If a constraint contains `||`, combine null-check constraint with `&&` in the expected order. In this case (= complex constraint), index is not created.
- `instanceof` check will be placed before null-check because `instanceof` may ensure that the property is accessible.

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6547

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/SingleDrlxParseSuccess.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/SingleDrlxParseSuccess.java
@@ -34,6 +34,7 @@ import org.drools.model.Index;
 import org.drools.modelcompiler.builder.generator.DRLIdGenerator;
 import org.drools.modelcompiler.builder.generator.TypedExpression;
 import org.drools.modelcompiler.builder.generator.UnificationTypedExpression;
+import org.drools.modelcompiler.util.StreamUtils;
 
 import static java.util.Optional.ofNullable;
 import static org.drools.modelcompiler.builder.generator.DrlxParseUtil.toClassOrInterfaceType;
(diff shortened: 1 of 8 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/SingleDrlxParseSuccess.java
```diff
@@ -410,9 +423,18 @@ public DrlxParseResult combineWith( DrlxParseResult other, BinaryExpr.Operator o
                 .setRight(new TypedExpression(otherDrlx.expr, right != null ? right.getType() : boolean.class))
                 .setBoundExpr(left)
                 .setIsPredicate(this.isPredicate && otherDrlx.isPredicate)
+                .setNullSafeExpressions(newNullSafeExpressions)
                 .setExprBinding(this.exprBinding); // only left exprBinding
     }
 
+    private Expression combineExpressions(Expression mainExpr, List<Expression> prefixExpressions) {
```
lucamolteni: maybe 
```
combinePredicatesWithAnd(Expression firstPredicate, List<Expression> otherPredicates)
```

tkobayas:
Jenkins run fdb

tkobayas:
Removing review request. Sorry, there are still test failures. Will check.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -152,20 +153,23 @@ private void generatePatternIdentifierIfMissing() {
...
-        }
 
-        addImplicitCastExpr(constraintParser, pattern.getIdentifier(), patternConstraintParseResults);
-        addNullSafeExpr(constraintParser, pattern.getIdentifier(), patternConstraintParseResults);
+            // Cast-check should be placed earlier than Null-check (calling the add method later means pushing the constraint earlier)
+            addNullSafeExpr(constraintParser, pattern.getIdentifier(), patternConstraintParseResultsPerConstraintDescr);
+            addImplicitCastExpr(constraintParser, pattern.getIdentifier(), patternConstraintParseResultsPerConstraintDescr);
+
+            patternConstraintParseResults.addAll(patternConstraintParseResultsPerConstraintDescr);
+        }
```
tkobayas, line 172: I moved `addImplicitCastExpr` and `addNullSafeExpr` inside "per constraint" loop so that we can place those instanceof-check and null-check at the beginning of the processed constraint (not at the beginning of the all constraint in the Pattern). `NullSafeDereferencingTest.testNullSafeDereferncingWithInstanceof` is the test case which explains why this is required.

lucamolteni (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/SingleDrlxParseSuccess.java
tkobayas: Thank you for the suggestion. I changed to:
```
combinePredicatesWithAnd(Expression mainPredicate, List<Expression> prefixPredicates)
```
These arg names explain the purpose better.

tkobayas:
Jenkins run fdb

hellowdan (approved):

pibizza (approved):

mariofusco (approved):

PR #3802, opened 2021-08-26
[DROOLS-6486] avoid creating a FactHandle for an object already prese…

mariofusco:
…nt in the working memory while evaluating a FromNode

**JIRA**: https://issues.redhat.com/browse/DROOLS-6486

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/FromNode.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/FromNode.java
@@ -47,6 +47,7 @@ import org.drools.core.util.bitmask.AllSetBitMask;
 import org.drools.core.util.bitmask.BitMask;
 import org.drools.core.util.index.TupleList;
 
+import static org.drools.core.base.DefaultKnowledgeHelper.getFactHandleFromWM;
 import static org.drools.core.reteoo.PropertySpecificUtil.calculateNegativeMask;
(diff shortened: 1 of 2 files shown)


mariofusco:
jenkins execute fdb

lucamolteni (approved):

lucamolteni:
@mariofusco I'm not sure I got the fix, is this something that wasn't needed in DRL?

mariofusco:
> @mariofusco I'm not sure I got the fix, is this something that wasn't needed in DRL?

@lucamolteni this is both an improvement for the whole engine and a fix for the executable model. It's an improvement because in all cases it avoids to create a duplicated FactHandle for a fact that is already present in the working memory. 

As for the fix, the drl in its generated consequence has a resolution mechanism that eventually checks for the existence of the duplicated FactHandle and retrieves the "right one" to be used with a modify. This mechanism doesn't exist for the executable model consequence and I didn't want to add it also there (also for performance reasons), so avoiding to create the duplicated FactHandle at all indirectly also solves the problem for the executable model.

mariofusco:
jenkins retest this

mariofusco:
The only failing test is `org.kie.wb.test.rest.functional.ProjectIntegrationTest.testGetBranches` which is clearly unrelated to this.

hellowdan (approved):

pibizza (approved):
Change is good, but we are still missing proper unit testing.

PR #3804, opened 2021-08-26
[DROOLS-6560] use ! operator to perform negation of a simple expressi…

mariofusco:
…on constraint

**JIRA**: https://issues.redhat.com/browse/DROOLS-6560

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/lang/descr/ExprConstraintDescr.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/lang/descr/ExprConstraintDescr.java
@@ -22,7 +22,6 @@ package org.drools.compiler.lang.descr;
 public class ExprConstraintDescr extends BaseDescr implements ExpressionDescr {
 
     private static final long serialVersionUID = 520l;
-    private boolean negated = false;
 
     public static enum Type {
         NAMED, POSITIONAL;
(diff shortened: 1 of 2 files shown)


lucamolteni (approved):

hellowdan (approved):

drools-compiler/src/main/java/org/drools/compiler/lang/descr/ExprConstraintDescr.java
```diff
@@ -80,12 +79,7 @@ public String toString() {
 
     @Override
     public BaseDescr negate() {
-        negated = true;
+        setText("!(" + getText() + ")");
         return this;
     }
```
pibizza, line 84: From a logical point of view I would expect that if I negate something, then when I ask isNegated that should return true (and if I negate it again, it should return false).

I think this is because I do not fully understand the semantic of negate.

mariofusco: @pibizza you're right, let me explain. In this case when the `negate()` method is called **either** I can take an immediate action to negate the expression without the need of any further work (as I'm doing with this fix) **or** I have to flag the expression as negated meaning that the expression has to be negated, but I'm not able to perform the negation right now, so some other action need to be taken in the remaining part of the parsing process to complete it.

PR #3809, opened 2021-08-31
[DROOLS-6584] Expose Drools metrics via Micrometer

triceo:

[DROOLS-6584](https://issues.redhat.com/browse/DROOLS-6584)

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1745
* https://github.com/kiegroup/kie-docs/pull/3860

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-metric/pom.xml
+++ b/drools-metric/pom.xml
@@ -17,6 +17,7 @@
 
   <properties>
     <java.module.name>org.drools.metric</java.module.name>
+    <version.io.micrometer>1.7.3</version.io.micrometer>
   </properties>
 
   <dependencies>
@@ -24,6 +25,16 @@
     <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-jmx</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.drools</groupId>
@@ -53,5 +64,20 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <dependencyManagement> <!-- TODO: move to parent -->
+    <dependencies>
+      <dependency>
+        <groupId>io.micrometer</groupId>
+        <artifactId>micrometer-core</artifactId>
+        <version>${version.io.micrometer}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.micrometer</groupId>
+        <artifactId>micrometer-registry-jmx</artifactId>
+        <version>${version.io.micrometer}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
 </project>
--- a/drools-metric/pom.xml
+++ b/drools-metric/pom.xml
@@ -17,7 +17,6 @@
 
   <properties>
     <java.module.name>org.drools.metric</java.module.name>
-    <version.io.micrometer>1.7.3</version.io.micrometer>
   </properties>
 
   <dependencies>
--- a/drools-metric/pom.xml
+++ b/drools-metric/pom.xml
@@ -32,11 +32,6 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>io.micrometer</groupId>
-      <artifactId>micrometer-registry-jmx</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-compiler</artifactId>
       <scope>test</scope>
@@ -68,20 +63,5 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
-  <dependencyManagement> <!-- TODO: move to parent -->
-    <dependencies>
-      <dependency>
-        <groupId>io.micrometer</groupId>
-        <artifactId>micrometer-core</artifactId>
-        <version>${version.io.micrometer}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.micrometer</groupId>
-        <artifactId>micrometer-registry-jmx</artifactId>
-        <version>${version.io.micrometer}</version>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
 
 </project>
--- a/drools-metric/pom.xml
+++ b/drools-metric/pom.xml
@@ -52,7 +52,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
     <!-- Logging -->
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
+++ b/drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
@@ -26,7 +26,6 @@ import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.search.Search;
 import org.drools.core.common.BaseNode;
-import org.drools.core.common.InternalWorkingMemory;
 import org.kie.api.definition.rule.Rule;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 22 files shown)


drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
```diff
@@ -65,22 +86,54 @@ public void incrementEvalCount() {
...
+                                        long elapsedTimeInNanos) {
+        // TODO This takes a long time; cache this somehow.
+        Tag nodeIdTag = Tag.of("node.id", Long.toString(node.getId()));
+        Stream<Tag> allTags = Stream.of(nodeIdTag);
+        for (Rule rule : node.getAssociatedRules()) {
+            String ruleName = rule.getPackageName() + "." + rule.getName(); // TODO sanitize rule names (whitespace etc.)
+            Tag ruleTag = Tag.of("rule", ruleName);
+            allTags = Stream.concat(allTags, Stream.of(ruleTag));
+        }
+        Iterable<Tag> tagsIterable = allTags.collect(Collectors.toSet()); // TODO Somehow identify the session/kiebase too.
```
triceo: Is there an unique session/kiebase identifier that I could use?

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
```diff
@@ -65,22 +86,54 @@ public void incrementEvalCount() {
...
+        Timer registeredTimer = Search.in(meterRegistry)
+                .tags(tagsIterable)
+                .timer();
+        if (registeredTimer == null) { // If timer does not exist, create one.
+            registeredTimer = Timer.builder("org.drools.metric.time.elapsed")
+                    .tags(tagsIterable)
+                    .register(meterRegistry);
+        }
+        // Now record the average elapsed time.
+        registeredTimer.record(Duration.ofNanos(elapsedTimeInNanos / evalCount));
```
triceo: I don't like this "bulk processing". It'd be better for this purpose if drools-metric allowed me to measure individual operations directly.

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
```diff
@@ -34,6 +46,15 @@
 
     private static final MetricLogUtils INSTANCE = new MetricLogUtils();
 
+    private static MeterRegistry getMicrometerRegistryIfEnabled() {
```
triceo: This needs changing. I need to make Micrometer actually optional.

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
tkobayas:
```
long ksessionId = internalWorkingMemory.getIdentifier();
String kbaseId = internalWorkingMemory.getKnowledgeBase().getId();

KieContainer kieContainer = ((KnowledgeBaseImpl)internalWorkingMemory.getKnowledgeBase()).getKieContainer();
if (kieContainer != null) {
   String releaseId = kieContainer.getReleaseId().toString();
}
```
If you use KieContainer, releaseId would be the meaningful unique id. kbaseId and ksessionId will be under the hierarchy. 

Btw, do you mean the use case is that multiple kbases and multiple ksessions are running concurrently?

triceo: Yes, this should work for multiple sessions running at the same time.

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
tkobayas: I may misunderstand you but from `startMetrics` to `logAndEndMetrics` is not "bulk". It's just one execution of `PhreakXXXNode.doNode()`. During the one execution, constraints could be evaluated `evalCount` times.

In many bottle-neck cases (e.g. described in https://blog.kie.org/2021/07/how-to-find-a-bottle-neck-in-your-rules-for-performance-analysis.html ) , a large number of `evalCount` could be a sign of a problem. So `elapsedTimeInNanos / evalCount` may not help many cases. I would rather want to see both `elapsedTimeInNanos` and `evalCount` separately.

triceo: Good point. Still, I want to see the averaging statistics - but we can probably have both.

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
tkobayas: Note that the drools-metric module is not for production environment use because CI doesn't test drools test suites with drools-metric enabled. drools-metric was developed to find a bottle-neck in rules so it's supposed to be used in test environment. I think Micrometer enhancement is still useful in test environment (e.g. observe performance in pre-prod tests) but if you think of usage in production environment, we would need some discussion.

https://docs.jboss.org/drools/release/latest/drools-docs/html_single/#performance-tuning-drl-ref_drl-rules

triceo: Good point. This is definitely not for production use, and therefore we likely do not need to track multiple sessions.
That helps, since these tags are a performance burden - so the fewer there are, the better.

triceo:
Please squash when merging, @tkobayas.
Also, I wasn't able to find the source code for docs - if you could point me to it, I will update the docs to mention this enhancement.

drools-metric/src/main/java/org/drools/metric/util/MicrometerUtils.java
```diff
@@ -0,0 +1,75 @@
...
+                        .tags(tags)
+                        .register(meterRegistry),
+                timer -> timer.record(Duration.ofNanos(elapsedTimeInNanos)),
+                node);
+        MicrometerUtils.triggerMicrometer(evaluationCountCache,
+                tags -> Counter.builder("org.drools.metric.evaluation.count")
+                        .tags(tags)
+                        .register(meterRegistry),
+                counter -> counter.increment(evalCount),
+                node);
```
triceo, line 55: I wasn't sure here.

For the `elapsed.time.per.evaluation counter`, it is clear that it has to be a timer. We record average times per evaluation.
For the `evaluation.count`, that has to be a counter. An ever increasing sum of evaluations.
For the `elapsed.time`, should that be a timer or a counter? In other words, should it treat `Duration`s individually and provide statistics, or should it rather just sum those durations and provide that as a simple number?

I'll leave the decision up to you.

tkobayas: I think `elapsed.time` is fine to be a timer. It would be used to find abnormally large elapsed time (= a bottle-neck).

drools-metric/src/test/java/org/drools/metric/MetricLogUtilsTest.java
```diff
@@ -16,27 +16,45 @@
...
     public void setup() {
         System.setProperty(MetricLogUtils.METRIC_LOGGER_ENABLED, "true");
         System.setProperty(MetricLogUtils.METRIC_LOGGER_THRESHOLD, "-1");
+        this.meterRegistry = new SimpleMeterRegistry();
+        Metrics.globalRegistry.add(this.meterRegistry);
+    }
+
+    @After
+    public void after() {
+        Metrics.globalRegistry.remove(this.meterRegistry);
```
tkobayas: Here, I think we need also:
```
        Metrics.globalRegistry.clear();
```
Without `clear`,  `meterRegistry.meterMap` is populated with timers (of a previous test) by `globalRegistry` on every setup so the assertions of the number of timers are incorrect. For example, expected timers should be 2 in `testJoin()`. Please confirm. Thanks!

drools-metric/pom.xml
```diff
@@ -17,6 +17,7 @@
 
   <properties>
     <java.module.name>org.drools.metric</java.module.name>
+    <version.io.micrometer>1.7.3</version.io.micrometer>
```
mariofusco: This doesn't seem to be used and anyway the micrometer version is defined in our bom.

drools-metric/src/main/java/org/drools/metric/util/MetricLogUtils.java
```diff
@@ -26,6 +26,7 @@
 
     public static final String METRIC_LOGGER_ENABLED = "drools.metric.logger.enabled";
     private boolean enabled = Boolean.parseBoolean(System.getProperty(METRIC_LOGGER_ENABLED, "false"));
+    private boolean micrometerAvailable = isMicrometerAvailable();
```
mariofusco, line 29: Couldn't this be a `static final` constant?

triceo: I can do it, but the entire class is a singleton, so I'd argue it doesn't matter.

tkobayas:
> Please squash when merging, @tkobayas.
> Also, I wasn't able to find the source code for docs - if you could point me to it, I will update the docs to mention this enhancement.

Doc source is here:
https://github.com/kiegroup/kie-docs/blob/main/doc-content/drools-docs/src/main/asciidoc/AuthoringAssets/performance-tuning-drl-ref.adoc

Btw, I don't have a permission to merge so @mariofusco will :)

tkobayas (approved):
Great job. Thanks!

triceo:
@mariofusco When you're happy with this, please merge with squash.

mariofusco (approved):

PR #3812, opened 2021-08-31
[DROOLS-6580] make java native compiler to work with JPMS

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6580

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NativeCompilerTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NativeCompilerTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2005 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.modelcompiler;
+
+import org.drools.compiler.compiler.JavaDialectConfiguration;
+import org.drools.modelcompiler.domain.Person;
+import org.junit.Test;
+import org.kie.api.runtime.KieSession;
+import org.kie.memorycompiler.JavaConfiguration;
+
+import static org.junit.Assert.assertEquals;
+
+public class NativeCompilerTest extends BaseModelTest {
+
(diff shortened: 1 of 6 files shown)


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NativeCompilerTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class NativeCompilerTest extends BaseModelTest {
+
+    public NativeCompilerTest(RUN_TYPE testRunType ) {
+        super( testRunType );
+    }
+
+    @Test(timeout = 5000)
```
gitgabrio, line 33: @mariofusco 
Out of curiosity: why this timeout ?

kie-memory-compiler/src/main/java/org/kie/memorycompiler/KieMemoryCompiler.java
```diff
@@ -123,7 +125,7 @@ private KieMemoryCompiler() { }
         }
         JavaConfiguration javaConfiguration = new JavaConfiguration();
         javaConfiguration.setCompiler(compilerType);
-        javaConfiguration.setJavaLanguageLevel("1.8");
+        javaConfiguration.setJavaLanguageLevel(findJavaVersion());
```
gitgabrio, line 128: @mariofusco 
Thx! :smile:

gitgabrio (approved):
HI @mariofusco. Thx for the PR. LGTM!

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/NativeCompilerTest.java
mariofusco: @gitgabrio I took a random executable model test and tried it with the native compiler. In particular this test checks if property reactivity is properly working for that rule and in case there's a problem and it doesn't the rule will fall in an endless loop. In this case we want the test to fail with a timeout instead of looping forever.

gitgabrio: @mariofusco 
:+1:

lucamolteni (approved):

kie-memory-compiler/src/main/java/org/kie/memorycompiler/KieMemoryCompiler.java
hellowdan: Is there any possibility of findJavaVersion() returns an empty value?

hellowdan (approved):
Approving with minor comment. If JAVA_LANG_LEVEL_PROPERTY is somehow not defined, what's the expected behaviour for findJavaVersion() method?

mariofusco:
> Approving with minor comment. If JAVA_LANG_LEVEL_PROPERTY is somehow not defined, what's the expected behaviour for findJavaVersion() method?

That's the most common case actually and in this case it defaults to `System.getProperty("java.version")` that must be there.

PR #3818, opened 2021-09-02
[DROOLS-6542] fix split statement when a String contains a single quote

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6542

### DIFF
--- a/drools-core/src/test/java/org/drools/core/util/StringUtilsTest.java
+++ b/drools-core/src/test/java/org/drools/core/util/StringUtilsTest.java
@@ -15,6 +15,8 @@
 
 package org.drools.core.util;
 
+import java.util.List;
+
 import org.junit.Assert;
 import org.junit.Test;
 import org.kie.api.builder.ReleaseId;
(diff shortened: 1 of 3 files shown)


tkobayas (approved):

hellowdan (approved):

drools-core/src/test/java/org/drools/core/util/StringUtilsTest.java
```diff
@@ -320,6 +323,18 @@ public String toExternalForm() {
...
+        String text =
+                "System.out.println(\"'\");" +
+                "$visaApplication.setValidation( Validation.FAILED );" +
+                "drools.update($visaApplication);";
+        List<String> statements = splitStatements(text);
+        assertEquals(3, statements.size());
+        assertEquals("System.out.println(\"'\")", statements.get(0));
+        assertEquals("$visaApplication.setValidation( Validation.FAILED )", statements.get(1));
+        assertEquals("drools.update($visaApplication)", statements.get(2));
     }
```
pibizza, line 339: I like a lot that we have at least a new unit test for a public method. 

I would suggest adding more unit tests to explore all corner cases, and possibly write an helper method to simplify checks and reduce duplicated code. Something like "stringsAre(String... expected) that checks that the number of splitted strings are correct.

I would not add any further integration test. Once we are sure that we are performing the split, verifying how the split works is for the unit test level.

pibizza (approved):
Comment re: adding more unit tests. This will help covering us for other possible issues.

PR #3820, opened 2021-09-03
[DROOLS-6541] fix oopath following a not pattern

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6541

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/ReactiveFromNodeLeftTuple.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/ReactiveFromNodeLeftTuple.java
@@ -45,9 +45,10 @@ public class ReactiveFromNodeLeftTuple extends FromNodeLeftTuple {
                                      final boolean leftTupleMemoryEnabled) {
         super(leftTuple, rightTuple, currentLeftChild, currentRightChild, sink, leftTupleMemoryEnabled);
 
-        objects = new Object[leftTuple.getIndex() + 2];
-        System.arraycopy( leftTuple.toObjects(), 0, objects, 0, leftTuple.getIndex()+1 );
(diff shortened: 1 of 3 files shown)


pibizza (approved):
Seems good to me

hellowdan (approved):

drools-core/src/main/java/org/drools/core/reteoo/ReactiveFromNodeLeftTuple.java
```diff
@@ -45,9 +45,10 @@ public ReactiveFromNodeLeftTuple(final LeftTuple leftTuple,
                                      final boolean leftTupleMemoryEnabled) {
         super(leftTuple, rightTuple, currentLeftChild, currentRightChild, sink, leftTupleMemoryEnabled);
 
-        objects = new Object[leftTuple.getIndex() + 2];
-        System.arraycopy( leftTuple.toObjects(), 0, objects, 0, leftTuple.getIndex()+1 );
-        objects[leftTuple.getIndex()+1] = rightTuple.getFactHandle().getObject();
+        Object[] leftObjects = leftTuple.toObjects();
```
lucamolteni: why a +2 became a +1? Please explain these magic numbers by putting a comment

lucamolteni (changes requested):

drools-core/src/main/java/org/drools/core/reteoo/ReactiveFromNodeLeftTuple.java
mariofusco: @lucamolteni I found a further issue (another constructor wasn't applying the same fix), so I refactored and also added a comment to clarify it. Please review again.

lucamolteni (approved):
Ok thanks

PR #3822, opened 2021-09-07
[DROOLS-6530] sanitize paths

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6530

The goal of this pull request is to always have a `Path` as an internal representation of a file path instead of a `String` as we had so far. This not only has the advantage of improving Windows compatibility, but also allows to get rid of tons of ugly and error-prone string manipulations when dealing with paths. Unfortunately we cannot avoid the use of Strings to represents paths everywhere also because in many cases they're exposed in our public API. 

This pull request is bigger and more invasive than I expected and very likely I'm still missing some parts where we could eliminate the use of Strings. However I consider it a good starting point.

This is still a draft and I expect that it is breaking a few projects downstream, including Kogito, I will send other PRs to fix them accordingly. I wouldn't be surprised even if it will break a few currently working tests under Windows. I ran the build only under Linux, I'm trusting the Windows CI, but I appreciate it if you could run this on your Windows machine if you have any. As I said I couldn't remove all String manipulations and there are other around that I'm probably missing and the partial use of `Path` could have broken them under Windows.


mariofusco:
jenkins execute fdb

mariofusco:
jenkins execute fdb

mariofusco:
jenkins execute cdb

drools-core/src/main/java/org/drools/core/util/ClassUtils.java
```diff
@@ -138,8 +143,7 @@ public static String convertClassToResourcePath(final Class cls) {
     }
 
     public static String convertClassToResourcePath(final String pName) {
-        return pName.replace( '.',
-                              '/' ) + ".class";
+        return pName.replace( '.', File.separatorChar ) + ".class";
     }
```
triceo, line 147: This doesn't seem to be a public API, so why not return `Path` too?

drools-compiler/src/main/java/org/drools/compiler/kproject/models/KieModuleModelImpl.java
```diff
@@ -30,10 +32,11 @@
...
     public static final String KMODULE_FILE_NAME = "kmodule.xml";
-    public static final String KMODULE_JAR_PATH = "META-INF/" + KMODULE_FILE_NAME;
-    public static final String KMODULE_INFO_JAR_PATH = "META-INF/kmodule.info";
-    public static final String KMODULE_SRC_PATH = "src/main/resources/" + KMODULE_JAR_PATH;
-    public static final String KMODULE_SPRING_JAR_PATH = "META-INF/kmodule-spring.xml";
+    public static final String KMODULE_INFO_FILE_NAME = "kmodule.info";
+    public static final Path KMODULE_JAR_PATH = Paths.get("META-INF", KMODULE_FILE_NAME);
+    public static final Path KMODULE_INFO_JAR_PATH = Paths.get("META-INF", KMODULE_INFO_FILE_NAME);
+    public static final Path KMODULE_SRC_PATH = Paths.get("src", "main", "resources", "META-INF", KMODULE_FILE_NAME);;
+    public static final Path KMODULE_SPRING_JAR_PATH = Paths.get("META-INF", "kmodule-spring.xml");;
```
triceo:
```suggestion
    public static final Path KMODULE_SRC_PATH = Paths.get("src", "main", "resources", "META-INF", KMODULE_FILE_NAME);
    public static final Path KMODULE_SPRING_JAR_PATH = Paths.get("META-INF", "kmodule-spring.xml");
```

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/DeclaredTypeWriter.java
```diff
@@ -39,8 +41,8 @@ public String getSource() {
                 generatedPojo);
     }
 
-    public String getName() {
-        return pkgModel.getPathName() + "/" + name + ".java";
+    public Path getPath() {
+        return pkgModel.getPath().resolve( name + ".java" );
     }
```
triceo, line 46: Duplication with `AccumulateClassWriter` - perhaps there is more that these classes could share?
(Also, I've seen the "java" extension being added in other parts of the code.)

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
```diff
@@ -209,7 +212,11 @@ public String getName() {
     }
 
     public String getPathName() {
-        return name.replace('.', '/');
+        return name.replace('.', File.separatorChar);
+    }
+
+    public Path getPath() {
+        return Paths.get(getPathName());
     }
```
triceo: Why not invert the logic? Keep the `Path`, get the `pathName` from it.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/RuleWriter.java
```diff
@@ -119,8 +120,8 @@ private RuleFileSource(String name, CompilationUnit source) {
             this.source = source;
         }
 
-        public String getName() {
-            return pkgModel.getPathName() + "/" + name + ".java";
+        public Path getPath() {
+            return pkgModel.getPath().resolve( name + ".java" );
```
triceo, line 124: Again, the same code duplication.

triceo (commented):
Went through the code, made a couple of observations.
I'll run the code w/ OptaPlanner on Windows.

mariofusco:
I consider this an unsuccessful experiment, for a few reasons:

- the normalization of paths as a String coming from external API is error-prone.
- in most cases we deal with files that are only in memory and are never written on the disk, so identifying them with a path that is dependent on the hosting operating system is a mere and useless "styling exercise".
- in some situations, like for files extracted from a jar or a zip, the path separator is always '/' regardless of the operating system and using the Windows separator there is a mistake.
- in other situations, like retrieving a resource from the class loader, it is not possible to refer to it with a Path, but only through a String thus forcing the need of the opposite conversion from Path to String. Also in this case the conversion works only with the Unix separator.
- Changing the signature from String to Path in many internal methods breaks many use cases of the downstream projects.

Considering that also Windows is able to deal with paths using the Unix separator, provided that in a given path you don't mix the 2 types of separators, I decided to use a different strategy: 

- always use the Unix-style separator as internal representation
- normalize any path coming from public API to that style
- get rid of any usage of `File.separator` and `File.separatorChar` in our code

PR #3823, opened 2021-09-08
[DROOLS-6442] duplicate case label error in DomainClassesMetadata

tkobayas:
**JIRA**:

https://issues.redhat.com/browse/DROOLS-6442

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
@@ -222,6 +222,15 @@ public class RuleContext {
         }
     }
 
+    public void addCompilationWarning( KnowledgeBuilderResult warn ) {
+        if ( warn instanceof BaseKnowledgeBuilderResultImpl ) {
+            (( BaseKnowledgeBuilderResultImpl ) warn).setResource( ruleDescr.getResource() );
+        }
+        synchronized (kbuilder) {
+            kbuilder.addBuilderResult(warn);
+        }
+    }
+
     public boolean hasCompilationError() {
         return hasCompilationError;
     }
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GetterOverloadingTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GetterOverloadingTest.java
@@ -1,0 +1,452 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ *
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.modelcompiler;
+
+import java.util.List;
+
+import org.junit.Ignore;
+import org.junit.Test;
+import org.kie.api.KieServices;
+import org.kie.api.builder.KieBuilder;
+import org.kie.api.builder.Message.Level;
+import org.kie.api.runtime.KieContainer;
+import org.kie.api.runtime.KieSession;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+
+public class GetterOverloadingTest extends BaseModelTest {
+
+    public GetterOverloadingTest(RUN_TYPE testRunType) {
+        super(testRunType);
+    }
+
+    // NOTE: Drools/Mvel accepts 5 kinds of getters
+    // e.g. for "resource" property,
+    //
+    // isResource()   (only for boolean)
+    // isresource()   (only for boolean)
+    // getResource()
+    // getresource()
+    // resource()
+    //
+    // If a class has multiple getters of those 5, one getter has to be chosen based on the above priority order
+
+    @Test
+    public void testDuplicateDifferentPropertyInClassHierarchy() {
+        // ClassA.resource is boolean : isResource()
+        // ClassB.resource is String  : getResource()
+        // Mvel picks a method from getResource or isResoure depending on the order of Class.getMethods() -> unreliable. So let's make this ERROR
+        final String str =
+                "import " + ClassB.class.getCanonicalName() + ";" +
+                           "rule r1\n" +
+                           "when\n" +
+                           "    ClassB( resource == \"ABC\" )\n" +
+                           "then\n" +
+                           "end\n";
+
+        KieBuilder kieBuilder = createKieBuilder(str);
+        List<org.kie.api.builder.Message> messages = kieBuilder.getResults().getMessages(Level.ERROR);
+        assertThat(messages.get(0).getText()).contains("Imcompatible Getter overloading detected");
+    }
+
+    public static class ClassA {
+
+        private boolean resource;
+
+        public boolean isResource() {
+            return resource;
+        }
+
+        public void setResource(boolean resource) {
+            this.resource = resource;
+        }
+    }
+
+    public static class ClassB extends ClassA {
+
+        private String resource;
+
+        public String getResource() {
+            return resource;
+        }
+
+        public void setResource(String resource) {
+            this.resource = resource;
+        }
+    }
+
+    @Ignore("This test randomly fails with STANDARD_DRL until Mvel is updated")
+    @Test
+    public void testBooleanAccessorOverload() {
+        // ClassC implements both isResource() and getResource() for boolean (This is acceptable according to Javabeans spec)
+        // isResource() has to be prioritized per Javabeans spec.
+        // No Warning
+        final String str =
+                "import " + ClassC.class.getCanonicalName() + ";" +
+                           "rule r1\n" +
+                           "when\n" +
+                           "    ClassC( resource == true )\n" +
+                           "then\n" +
+                           "end\n";
+
+        KieBuilder kieBuilder = createKieBuilder(str);
+        List<org.kie.api.builder.Message> messages = kieBuilder.getResults().getMessages();
+        assertThat(messages).isEmpty();
+
+        KieContainer kieContainer = KieServices.get().newKieContainer(kieBuilder.getKieModule().getReleaseId());
+        KieSession ksession = kieContainer.newKieSession();
+
+        ClassC c = new ClassC();
+        c.setResource(true);
+
+        ksession.insert(c);
+        int fired = ksession.fireAllRules();
+        assertEquals(1, fired);
+    }
+
+    public static class ClassC {
+
+        private boolean resource;
+
+        public boolean isResource() {
+            return resource;
+        }
+
+        public boolean getResource() {
+            return !resource; // this getter should not be called!
+        }
+
+        public void setResource(boolean resource) {
+            this.resource = resource;
+        }
+    }
+
+    @Test
+    public void testAccessorInMultipleInterfaces() {
+        // 2 super interfaces have the same abstract method
+        // Valid overriding. No warning.
+        final String str =
+                "import " + InterfaceC.class.getCanonicalName() + ";" +
+                           "rule r1\n" +
+                           "when\n" +
+                           "    InterfaceC( name == \"ABC\" )\n" +
+                           "then\n" +
+                           "end\n";
+
+        KieBuilder kieBuilder = createKieBuilder(str);
+        List<org.kie.api.builder.Message> messages = kieBuilder.getResults().getMessages();
+        assertThat(messages).isEmpty();
+
+        KieContainer kieContainer = KieServices.get().newKieContainer(kieBuilder.getKieModule().getReleaseId());
+        KieSession ksession = kieContainer.newKieSession();
+
+        ClassD d = new ClassD("ABC");
+
+        ksession.insert(d);
+        int fired = ksession.fireAllRules();
+        assertEquals(1, fired);
+    }
+
+    public static interface InterfaceA {
+
+        String getName();
+    }
+
+    public static interface InterfaceB {
+
+        String getName();
+    }
+
+    public static interface InterfaceC extends InterfaceA, InterfaceB {
+
+    }
+
+    public static class ClassD implements InterfaceC {
+
+        private String name;
+
+        public ClassD(String name) {
+            this.name = name;
+        }
+
+        @Override
+        public String getName() {
+            return name;
+        }
+    }
+
+    @Test
+    public void testAccessorInSuperClassAndInterface() {
+        // Valid overriding from super class and interface. No Warning
+        final String str =
+                "import " + ClassF.class.getCanonicalName() + ";" +
+                           "rule r1\n" +
+                           "when\n" +
+                           "    ClassF( name == \"ABC\" )\n" +
+                           "then\n" +
+                           "end\n";
+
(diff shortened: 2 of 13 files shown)


tkobayas:
@mariofusco @lucamolteni 

As you see in the test case `testDuplicatePropertyInClassHierarchy`, this is not a recommended class design. But I wonder whether we should raise an Exception or just a warning (and execute the rule).

Currently, non-exec-model raises a warning. The message is:

```
Getter overloading detected in class org.drools.modelcompiler.CompilerTest$ClassB : getResource (class java.lang.String) vs isResource (boolean)
```

and executes the rule. The chosen accessor depends on the order of result of clazz.getMethods()... It doesn't sound good.

https://github.com/mvel/mvel/blob/mvel2-2.4.12.Final/src/main/java/org/mvel2/util/PropertyTools.java#L108-L117

In this PR, I raise an Exception in case of exec-model. So it's not aligned with non-exec-model. Which is better do you think? (I like explicit Exception/Error for both but it might break existing rule when a user updates the drools version... Hmm, introduce a system property switch?)

lucamolteni (approved):

lucamolteni:
I'd either warn in both cases or throw an exception in both cases. I'd avoid the system property.
As for breaking Users' code, I don't think it's a big deal if the error message is good enough. Thanks!

drools-mvel/src/main/java/org/drools/mvel/asm/ClassFieldInspectorImpl.java
```diff
@@ -336,11 +337,23 @@ private void storeGetterSetter( Method method,
...
+                    if (method.getReturnType().isAssignableFrom(existingMethod.getReturnType())) {
+                        // a more specialized getter (covariant overload) has been already indexed, so skip this one
+                        return;
+                    } else if (existingMethod.getReturnType().isAssignableFrom(method.getReturnType())) {
+                        // this method is a more specialized getter. Overwrite with this one
+                    } else {
+                        // returnType is different so it would likely produce a wrong result
+                        addResult( fieldName, new IncompatibleGetterOverloadError( classUnderInspection,
+                                                                         this.getterMethods.get( fieldName ).getName(), this.fieldTypes.get( fieldName ),
+                                                                         method.getName(), method.getReturnType() ) );
```
tkobayas, line 352: This is the major change. If a fact class has ambiguous getters (including inherited methods) with different return types (e.g. `boolean isResource()` and `String getResource()`), it results in a compile Error rather than a Warning.

It was already mentioned by Geoffrey https://issues.redhat.com/browse/DROOLS-18?focusedCommentId=12753871&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12753871  but somehow @mariofusco implemented it with Warning. Is it okay to change? Do you still prefer Warning?

tkobayas:
Thanks @lucamolteni . Now I added more test cases which:
- We want to raise a build error
- We want to raise a build warning. We can execute the rule
- No warning. We can execute the rule

while exec-model and non-exec-model give the same results.

Please review @lucamolteni @mariofusco @hellowdan @pibizza  Thanks!

tkobayas:
Jenkins run fdb

lucamolteni (approved):

tkobayas:
PR test failure:

Not related to this PR. Hmm, is this a flaky test? If I will see the failure again, I would consider an adjustment.
```
[2021-09-14T11:12:37.716Z] [ERROR] testHaltAfterSomeTimeThenRestartButNoLongerHolding[KieBase type=STREAM_IDENTITY_MODEL_PATTERN](org.drools.compiler.integrationtests.TimerAndCalendarTest)  Time elapsed: 0.826 s  <<< FAILURE!
[2021-09-14T11:12:37.716Z] java.lang.AssertionError: expected:<2> but was:<3>
[2021-09-14T11:12:37.716Z]      at org.junit.Assert.fail(Assert.java:89)
[2021-09-14T11:12:37.716Z]      at org.junit.Assert.failNotEquals(Assert.java:835)
[2021-09-14T11:12:37.716Z]      at org.junit.Assert.assertEquals(Assert.java:647)
[2021-09-14T11:12:37.716Z]      at org.junit.Assert.assertEquals(Assert.java:633)
[2021-09-14T11:12:37.716Z]      at org.drools.compiler.integrationtests.TimerAndCalendarTest.testHaltAfterSomeTimeThenRestartButNoLongerHolding(TimerAndCalendarTest.java:1560)
```

FDB test failures:

Not related to this PR.
```
[2021-09-14T13:17:28.628Z] [ERROR] Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 8.646 s <<< FAILURE! - in org.jbpm.executor.impl.jms.JmsAvaiableJobExecutorTest
[2021-09-14T13:17:28.628Z] [ERROR] testAsyncAuditErrorProducer(org.jbpm.executor.impl.jms.JmsAvaiableJobExecutorTest)  Time elapsed: 2.397 s  <<< FAILURE!
[2021-09-14T13:17:28.628Z] java.lang.AssertionError
[2021-09-14T13:17:28.628Z]      at org.junit.Assert.fail(Assert.java:87)
[2021-09-14T13:17:28.628Z]      at org.junit.Assert.assertTrue(Assert.java:42)
[2021-09-14T13:17:28.628Z]      at org.junit.Assert.assertTrue(Assert.java:53)
[2021-09-14T13:17:28.628Z]      at org.jbpm.executor.impl.jms.JmsAvaiableJobExecutorTest.testAsyncAuditErrorProducer(JmsAvaiableJobExecutorTest.java:153)
```

Not related to this PR. Hopefully, DROOLS-6600 will address this.
```
[2021-09-14T18:38:09.644Z] [ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 57.371 s <<< FAILURE! - in org.kie.karaf.itest.blueprint.KieBlueprintjBPMPersistenceKarafIntegrationTest
[2021-09-14T18:38:09.644Z] [ERROR] testStartProcessEnv(org.kie.karaf.itest.blueprint.KieBlueprintjBPMPersistenceKarafIntegrationTest)  Time elapsed: 10.3 s  <<< ERROR!
[2021-09-14T18:38:09.644Z] org.ops4j.pax.swissbox.tracker.ServiceLookupException: gave up waiting for service org.kie.api.runtime.KieSession
```

tkobayas:
Jenkins run fdb

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GetterOverloadingTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+        public void setResource(String resource) {
+            this.resource = resource;
+        }
+    }
+
+    @Test
+    public void testBooleanAccessorOverload() {
+        // ClassC implements both isResource() and getResource() for boolean
+        // This is acceptable overloading assuming they return the same result.. Just Warning.
```
mariofusco: I disagree that this is acceptable and I think we should emit a compilation error even here. If the 2 methods were implemented as it follows, which of the 2 accessors would you call? What would be the correct result?

```
        public boolean isResource() {
            return resource;
        }

        public boolean getResource() {
            return !resource;
        }
```

tkobayas: Thanks for raising the point, @mariofusco 

JavaBeans spec says:

```
8.3.2
Boolean properties
In addition, for boolean properties, we allow a getter method to match the pattern:
public boolean is<PropertyName>();
This “is<PropertyName>” method may be provided instead of a “get<PropertyName>” meth-
od, or it may be provided in addition to a “get<PropertyName>” method.
In either case, if the “is<PropertyName>” method is present for a boolean property then we will
use the “is<PropertyName>” method to read the property value.
```
So I assume that some users may implement both `is<PropertyName>` and `get<PropertyName>`. And I thought that rather than surprising them with a build error (when updating drools version), it would be better to execute the rule (if they implement such inconsistent accessors, they will get a wrong result). However, I'm not very opinionated. If you still want to raise an error, I would follow. Please let me know.

Btw, 

> which of the 2 accessors would you call?

Even though the javabeans spec prioritizes `is<PropertyName>`, Mvel picks an accessor depending on the order of the result of Class.getMethods() so it's non-deterministic. (yeah, that's bad)

mariofusco: @tkobayas ok, your point is totally valid. The problem is either we are able to completely follow the spec and then we guarantee that we always use the `is<PropertyName>` method when available (and in this case we could also get rid of the warning) or, if this is not possible, it's safer to raise compile time error instead of randomly picking one of the 2 accessors. 

My suggestion is to change the `getResource` method as I suggested (returning the negate of `resource`) and check if the rules fires as expected (it correctly uses `isResource` as accessor) or not, both with and without the executable model. If this is the case we are fine and we can also remove the warning since, as you pointed out, having both accessors doesn't violate the JavaBeans spec. If not we have a bug and either we are to fix it, for both implementations, or in my opinion it's safer to raise the compilation error that I was suggesting.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
```diff
@@ -222,6 +222,15 @@ public void addCompilationError( KnowledgeBuilderResult error ) {
...
     }
 
+    public void addCompilationWarning( KnowledgeBuilderResult warn ) {
+        if ( warn instanceof BaseKnowledgeBuilderResultImpl ) {
+            (( BaseKnowledgeBuilderResultImpl ) warn).setResource( ruleDescr.getResource() );
+        }
+        synchronized (kbuilder) {
+            kbuilder.addBuilderResult(warn);
+        }
+    }
```
pibizza, line 232: I think these synchronization belong to the addBuilderResult method. I would move the synch in the signature of the method itself. That said, it seems strange that only these calls are synchronized.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/Misc2Test.java
```diff
@@ -3377,12 +3377,9 @@ public void testIsGetClash() {
...
         List<org.kie.api.builder.Message> errors = results.getMessages(org.kie.api.builder.Message.Level.ERROR);
-        assertTrue(errors.toString(), errors.isEmpty());
-
-        List<org.kie.api.builder.Message> warnings = results.getMessages(org.kie.api.builder.Message.Level.WARNING);
-        assertEquals( 2, warnings.size() );
-        for (org.kie.api.builder.Message warning : warnings) {
-            System.out.println( warning );
+        assertEquals( 2, errors.size() );
+        for (org.kie.api.builder.Message error : errors) {
+            System.out.println( error );
```
pibizza: I think it is not a good idea to have tests that print stuff on the console - expecially error messages. It can get confusing.

pibizza (approved):
Mostly minor comments. Code looks good to me.

hellowdan (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GetterOverloadingTest.java
tkobayas: @mariofusco 

> My suggestion is to change the getResource method as I suggested (returning the negate of resource) and check if the rules fires as expected (it correctly uses isResource as accessor) or not, both with and without the executable model. 

### non-exec-model

It randomly fails because the picked method depends on the order of the result of `Class.getMethods()`.

https://github.com/mvel/mvel/blob/mvel2-2.4.12.Final/src/main/java/org/mvel2/util/PropertyTools.java#L108-L117

We can fix Mvel like this:

https://github.com/mvel/mvel/pull/274/files

### exec-model

`isResource` is always chosen because `ClassUtils.accessorNameCandidates` has this order. Later method can overwrite the bestCandidate.

https://github.com/tkobayas/drools/blob/DROOLS-6442/drools-core/src/main/java/org/drools/core/util/ClassUtils.java#L480-L486

Now the next question arises. Should we difine the priority order of all possible getters? (like I did in the above Mvel fix example)

Drools ClassUtils and Mvel PropertyTools both accept 5 getters. Below is my candidate priority order.

- isResource()
- isresource()
- getResource()
- getresource()
- resource()

So approaches would be:

A) Implement the fix and priority order in both Drools and Mvel. I think it's possible but it will require mvel upgrade.
B) Raise a build error if a class has multiple getter overload (among the above 5 getters). 

I like A) better because less users will be surprised by drools upgrade. Please share your thoughts. Thanks!

mariofusco: @tkobayas thanks a lot for your great work on this, I also prefer strategy A)
I suggest to proceed as it follows:

1. Merge this pull request ignoring some of the test you added since I'm afraid that they will cause some random failures when running with drl, correct?
2. Also merge your fix to mvel. I gave a look at it and it seems ok to me, any reason why you flagged it with "DO NOT MERGE"? Is there something still missing?
3. Deploy a new release of mvel containing your fix.
4. Bump drools to the newer mvel version and unignore the tests.

Is that ok for you?

mariofusco (approved):

drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/GetterOverloadingTest.java
tkobayas: Thanks, @mariofusco 

>     1. Merge this pull request ignoring some of the test you added since I'm afraid that they will cause some random failures when running with drl, correct?
> 
>     2. Also merge your fix to mvel. I gave a look at it and it seems ok to me, any reason why you flagged it with "DO NOT MERGE"? Is there something still missing?
> 
>     3. Deploy a new release of mvel containing your fix.
> 
>     4. Bump drools to the newer mvel version and unignore the tests.

Yes, I'll go ahead with the plan. For 1. and 2. , I will add some more test coverage. Thanks!

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/RuleContext.java
tkobayas: I imitated the syncronization in `RuleContext.addCompilationError` which was added by https://github.com/kiegroup/drools/commit/2d28395dde177d2a61e8759761fb4e77eccc92dc

This is to protect `kbuilder` only in exec-model (which is potentially called in parallel) so we don't syncronize the `addBuilderResult` method itself.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/Misc2Test.java
tkobayas: Thanks, removed.

tkobayas:
Jenkins run fdb

tkobayas:
Updated. But even if every test is successful, probably it's better to wait to merge until the pom version is bumped to 7.61.0-SNAPSHOT?

tkobayas:
Also updated https://github.com/mvel/mvel/pull/274/

tkobayas:
FDB failure is not related to this PR:
```
org.uberfire.java.nio.fs.jgit.JGitFileSystemImplProviderTest.testNewFileSystemCloneAndRescan
```

tkobayas:
Hi @mariofusco 

As we will keep using 7.60.0-SNAPSHOT this week, we can go ahead with this plan, right?

>     1. Merge this pull request ignoring some of the test you added since I'm afraid that they will cause some random failures when running with drl, correct?
> 
>     2. Also merge your fix to mvel. I gave a look at it and it seems ok to me, any reason why you flagged it with "DO NOT MERGE"? Is there something still missing?
> 
>     3. Deploy a new release of mvel containing your fix.
> 
>     4. Bump drools to the newer mvel version and unignore the tests.

Please merge this PR and https://github.com/mvel/mvel/pull/274 (So step 1 and 2). Thanks!

PR #3829, opened 2021-09-14
[DROOLS-6382] Switch scorecard/tree default implementations

gitgabrio:

See https://issues.redhat.com/browse/DROOLS-6382

Needed by https://github.com/kiegroup/droolsjbpm-integration/pull/2604


jiripetrlik (approved):
I would suggest to remove old implementations completely and not only to switch.

gitgabrio:
> I would suggest to remove old implementations completely and not only to switch.

HI @jiripetrlik 
It was my first attempt, but after discussion with @danielezonca we agreed on keeping it, without updating with new features.

danielezonca (approved):

PR #3830, opened 2021-09-14
[DROOLS-6497] when updating a parent rule during the incremental comp…

mariofusco:
…ilation also recompile all its children

**JIRA**: https://issues.redhat.com/browse/DROOLS-6497


mariofusco:
Jankins failure of `org.drools.compiler.integrationtests.TimerAndCalendarTest.testHaltAfterSomeTimeThenRestartButNoLongerHolding` is unrelated

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1258,12 +1258,38 @@ private void preProcessRules(PackageDescr packageDescr, PackageRegistry pkgRegis
...
     }
 
+    private Collection<RuleImpl> findChildrenRulesToBeRemoved(PackageDescr packageDescr, Collection<RuleImpl> rulesToBeRemoved) {
+        Collection<String> childrenRuleNamesToBeRemoved = new HashSet<>();
+        Collection<RuleImpl> childrenRulesToBeRemoved = new HashSet<>();
+        for (RuleImpl rule : rulesToBeRemoved) {
+            if (rule.hasChildren()) {
+                for (RuleImpl child : rule.getChildren()) {
+                    if (!rulesToBeRemoved.contains(child)) {
+                        // if a rule has child not already contained among the ones to be removed ...
```
lucamolteni: If a rule has a child rule not marked to be removed

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -1258,12 +1258,38 @@ private void preProcessRules(PackageDescr packageDescr, PackageRegistry pkgRegis
...
+                        childrenRulesToBeRemoved.add(child);
+                        childrenRuleNamesToBeRemoved.add(child.getName());
+                        // ... remove the child rule but also add it back to the PackageDescr in order to readd it when also the parent rule will be ...
```
lucamolteni: Will be what? :D

lucamolteni (changes requested):

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
mariofusco: I meant "readd it when also the parent rule will be [readded]". Is the sentence unclear?

hellowdan (approved):

pibizza (approved):
Sounds good for me.

lucamolteni (approved):

PR #3838, opened 2021-09-17
Performance Fix by caching DMN Profile

anandkshitiz:
Hi Team,

As you might have noticed from my earlier posts that I am part of evaluation of Drools for DMN execution in our project.

While doing the performance benchmarking of Drools we saw a huge bottleneck and we were able to find the root cause of this bottleneck in the Drools codebase for which we did an internal fix.
I am sharing the use case which we tried , bottleneck which we faced and the code fix which we did to resolve it.
Now, we want guidance on how can we push this patch to Drools codebase.

Use Case :
Execute 5000 rules(1 rule = 1 DMN Decision Table) spread across 3800 DMN files(nested DMN files using Decision Service) in Drools.

Observation :
While executing these 5000 rules in a single call to dmnRuntime.evaluateByName(rootModel,dmnContext,"<RootDecisionName>") it took around 6 seconds.

Fix:
We did a lot of analysis for this performance issue by giving better computation power to run the SpringBoot app where Drools engine was running, sharding the DMNs and executing DMNs using multiple DMNRuntime. Ultimately, we looked into the codebase and found the following issue :

Currently while evaluation of every decision, the DMNRuntimeKBWrappingIKB** checks for update to profiles by iterating over each profile. Time taken to execute increases with the no. of profiles(dmns) in memory.

Our fix is to avoid this check for every decision evaluation, So, the time of update to Kbase is maintained in pkgsLastUpdatedAt and checked before iterating over all the models, this significantly reduces the time taken to complete each decision(DT).

// DMNRuntimeKBWrappingIKB.java
public List<DMNProfile> getProfiles() {
...
for (DMNProfile p : dmnPkg.getProfiles()) {
if (!profiles.contains(p)) {
profiles.add(p);
}
....
}
...
}

Observation after this fix: With the above fix, we are able to execute 5000 rules spread across 3800 DMN files in 508ms.

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeKBWrappingIKB.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeKBWrappingIKB.java
@@ -16,6 +16,7 @@
 
 package org.kie.dmn.core.impl;
 
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -34,6 +35,7 @@ import org.drools.core.definitions.ResourceTypePackageRegistry;
 import org.drools.core.impl.InternalKnowledgeBase;
 import org.drools.core.impl.KnowledgeBaseImpl;
 import org.kie.api.KieBase;
+import org.kie.api.definition.KiePackage;
 import org.kie.api.io.ResourceType;
 import org.kie.api.runtime.KieContainer;
 import org.kie.api.runtime.KieRuntimeFactory;
@@ -53,6 +55,8 @@ public class DMNRuntimeKBWrappingIKB implements DMNRuntimeKB {
     private static final Logger logger = LoggerFactory.getLogger( DMNRuntimeKBWrappingIKB.class );
 
     private final InternalKnowledgeBase        knowledgeBase;
+    private List<DMNProfile> dmnProfileList;
+    private Instant pkgsLastUpdatedAt;
 
     public DMNRuntimeKBWrappingIKB(InternalKnowledgeBase knowledgeBase) {
         this.knowledgeBase = knowledgeBase;
@@ -151,19 +155,29 @@ public class DMNRuntimeKBWrappingIKB implements DMNRuntimeKB {
 
     @Override
     public List<DMNProfile> getProfiles() {
-        // need list to preserve ordering
-        List<DMNProfile> profiles = new ArrayList<>();
-        knowledgeBase.getKiePackages().forEach(kpkg -> {
-            DMNPackageImpl dmnPkg = (DMNPackageImpl) ((InternalKnowledgePackage) kpkg).getResourceTypePackages().get(ResourceType.DMN);
-            if (dmnPkg != null) {
-                for (DMNProfile p : dmnPkg.getProfiles()) {
-                    if (!profiles.contains(p)) {
-                        profiles.add(p);
+        if(isProfileCacheInvalid()) {
+            // need list to preserve ordering
+            List<DMNProfile> profiles = new ArrayList<>();
+            knowledgeBase.getKiePackages().forEach(kpkg -> {
+                DMNPackageImpl dmnPkg = (DMNPackageImpl) ((InternalKnowledgePackage) kpkg)
+                    .getResourceTypePackages().get(ResourceType.DMN);
+                if (dmnPkg != null) {
+                    for (DMNProfile p : dmnPkg.getProfiles()) {
+                        if (!profiles.contains(p)) {
+                            profiles.add(p);
+                        }
                     }
                 }
-            }
-        });
-        return profiles;
+            });
+            pkgsLastUpdatedAt = knowledgeBase.getPkgsLastUpdatedAt();
+            dmnProfileList = profiles;
+        }
+        return dmnProfileList;
+    }
+
+    private boolean isProfileCacheInvalid() {
+        return dmnProfileList == null || dmnProfileList.isEmpty() || pkgsLastUpdatedAt == null
+            || !pkgsLastUpdatedAt.equals(knowledgeBase.getPkgsLastUpdatedAt());
     }
 
     @Override
--- a/drools-core/src/main/java/org/drools/core/impl/InternalKnowledgeBase.java
+++ b/drools-core/src/main/java/org/drools/core/impl/InternalKnowledgeBase.java
@@ -16,6 +16,7 @@
 package org.drools.core.impl;
 
 import java.io.IOException;
+import java.time.Instant;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -153,6 +154,7 @@ public interface InternalKnowledgeBase extends KieBase {
 
     List<AsyncReceiveNode> getReceiveNodes();
     void addReceiveNode(AsyncReceiveNode node);
+    Instant getPkgsLastUpdatedAt();
 
     boolean hasMultipleAgendaGroups();
 }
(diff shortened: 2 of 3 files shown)


anandkshitiz:
@tarilabs  @lucamolteni Could you please have a look into this PR as per our discussion on Zulip chat(https://kie.zulipchat.com/#narrow/stream/232677-drools/topic/DMN.20Execution.20Performance/near/242022179)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeKBWrappingIKB.java
```diff
@@ -151,19 +155,29 @@ public DMNModel getModelById(String namespace, String modelId) {
...
-                        profiles.add(p);
+        if(isProfileCacheInvalid()) {
+            // need list to preserve ordering
+            List<DMNProfile> profiles = new ArrayList<>();
+            knowledgeBase.getKiePackages().forEach(kpkg -> {
+                DMNPackageImpl dmnPkg = (DMNPackageImpl) ((InternalKnowledgePackage) kpkg)
+                    .getResourceTypePackages().get(ResourceType.DMN);
+                if (dmnPkg != null) {
+                    for (DMNProfile p : dmnPkg.getProfiles()) {
+                        if (!profiles.contains(p)) {
```
lucamolteni, line 166: I'm not sure about the `pkgsLastUpdatedAt` logic, but to remove the quadratic behaviour you could try changing the profiles collection from an `ArrayList` to a `SortedSet`

lucamolteni (changes requested):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeKBWrappingIKB.java
anandkshitiz: Thanks @lucamolteni for your comment.

Please note that it is not the quadratic behavior which is causing the performance issue. The fact that ```getProfiles``` method is getting called for every recursive call to executeDMN and goes through all the packages is causing the performance issue when a DMN has a very deep structure like the 3800 DMN example which i had given in my initial comment.

By keeping the ```List<DMNProfile>``` as class variable we are kind of caching it and ```pkgsLastUpdatedAt``` is nothing but for the invalidation of the cache.

tarilabs (changes requested):
Hi @anandkshitiz ,
thank you for raising this PR.

We really appreciate community contributions, and we evaluate seriously each one.

Unfortunately some of the premises/claims in your original report, are not demonstrated by either some tests or some benchmark, making difficult to really asses if this solution is generalised enough.

My recommended next-action would be for you to properly track the problem you are facing, in order to have a throughout investigation.

To properly track, kindly either:
- create a Red Hat Customer Portal issue (https://access.redhat.com/support/cases/#) if you have a Red Hat subscription
- create a Drools JIRA (https://issues.redhat.com/projects/DROOLS) if you don't have a Red Hat subscription

When you will raise the ticket, don't hesitate to link this github thread.

Specifically, I've tested to verify your claims using `DMNDecisionTableRuntimeTest#testDecisionTableWithCalculatedResult_parallel` which executes about 10'000 decision tables.
There is never a call to `DMNRuntimeKBWrappingIKB#getProfiles`, as I would expect, as the models are fully "compiled" in the DMNRuntime, hence it does not need parsing of the KB/Profiles.

I even tried to change the test using method `dmnRuntime.evaluateByName(...)` here:
https://github.com/kiegroup/drools/blob/d6a5c3fe6cccf1a7f74148b02cce72979ef43f18/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNDecisionTableRuntimeTest.java#L92
to check specifically if something not immediately obvious, but again there is no call to `DMNRuntimeKBWrappingIKB#getProfiles`, as I would expect for the reasons explained above.

This makes me believe there is something which might be impacting in the way you are interacting with the Kie `Drools DMN engine` API, or specifically about your models etc.
We would need a reproducer to check what's going wrong.
We can understand you might not be in the position to share the full DMN models, but we would need at least an understanding of the Kie API calls, and the overall structure of some simplified/generic example DMN models which exhibit the issue you are claiming.
Especially if you have a Red Hat subscription, opening a Red Hat Customer Portal case would guide you in this way.

Kindly track the issue, and let's work together on the technical solution once there is a valid reproducer.
Thank you in advance!

anandkshitiz:
@tarilabs Acknowledging your comment.
Will follow the next action items and will also attach the evidence in terms of Java Profiler which clearly explains that ```getProfiles``` method is getting called for each recursive call to ```executeDMN```

ubaveenther:
@tarilabs I'm also part of @anandkshitiz team. 

- We've run the application with Java profiler and performed small amount of performance test by using JMeter with the configuration of 3 users for 5 mins. For this example, we've taken 500 DMN files and found that **DMNRuntimeKBWrappingIKB#getProfiles** is getting called for every recursive call which is taking significant amount of time for execution. In this case, this method is taking around **7.9% of CPU time**. 

- Also we've performed the same test **with the performance cache fix jar** version and this time this method is not getting called for every recursive call because of cache fix and it is taking **less than 1% of CPU time**.

- For your reference I've attached the dump of java profiler, kindly have a look into it.
[java_profiler_dump.zip](https://github.com/kiegroup/drools/files/7606021/java_profiler_dump.zip)

tarilabs:
Hi @ubaveenther thank you for your message, but as explained in https://github.com/kiegroup/drools/pull/3838#pullrequestreview-758867424 I would need the reference Customer Portal ticket ID# to follow-up on this request.

In the message https://github.com/kiegroup/drools/pull/3838#issuecomment-923701682 I understood @anandkshitiz was going to create a ticket.

I'm happy to investigate further your point, but I need
1. the ticket reference
2. the DMN model you have been using to collect these metrics in order to reproduce your problem (as explained in https://github.com/kiegroup/drools/pull/3838#pullrequestreview-758867424 your problem does not reproduce on a benchmark)

With the ticket and the DMN model, I'm more than happy to investigate in more details your claim.
Please let me know!

ubaveenther:
@tarilabs Thanks for your quick response. 

- I've created JIRA ticket for this issue https://issues.redhat.com/browse/DROOLS-6722
- Also in that same JIRA ticket I've attached the DMN models which I've used for my experiment.

tarilabs:
Hi @ubaveenther , @anandkshitiz 

thank you for having attached the DMN models and the profile measurement data.
With that, I could investigate better your claims.

Earlier, without the actual DMN model and the measurement data, your claims could not be verified at the time of this https://github.com/kiegroup/drools/pull/3838#pullrequestreview-758867424 and now I can see why, this use case relates specifically to making use of DMN Literal Expressions.

With the DMN models and the measurement data,
I digged further into the problem which you are seeing:

image

but the root cause of the issue is not really caused by the `DMNProfile`,
the root cause problem, is that the DMN Literal expression node was not internalizing the FEEL evaluator.

I have created a specific PR to correct that:
 - https://github.com/kiegroup/drools/pull/4048

the provisional benchmark data confirms the improvement, but I will ask my colleagues to run more extended benchmarks.

Fixing the actual root cause will make this PR superseeded, and I will eventually close this PR.

Thank you for contributing in making the Drools DMN Engine better!
Don't hesitate to share DMN model + code to demonstrate use-case in the future, if you spot other areas of improvements.

tarilabs:
As mentioned in https://github.com/kiegroup/drools/pull/3838#issuecomment-981062402
This PR is superseeded by a fix addressing the root-cause of the performance here: 
 - https://github.com/kiegroup/drools/pull/4048#pullrequestreview-824008049

Thank you for contributing in making the Drools DMN Engine better!
Don't hesitate to share DMN model + code to demonstrate use-case in the future, if you spot other areas of improvements.

PR #3844, opened 2021-09-21
[DROOLS-5826] fix PropertySpecificTest with executable model

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5826

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
@@ -55,6 +55,7 @@ import org.drools.modelcompiler.builder.generator.drlxparse.ParseResultVoidVisit
 import org.drools.modelcompiler.builder.generator.drlxparse.SingleDrlxParseSuccess;
 import org.drools.modelcompiler.builder.generator.visitor.DSLNode;
 import org.drools.modelcompiler.builder.generator.visitor.FromVisitor;
+import org.kie.api.definition.rule.Watch;
 
 import static org.drools.model.impl.NamesGenerator.generateName;
 import static org.drools.model.impl.VariableImpl.GENERATED_VARIABLE_PREFIX;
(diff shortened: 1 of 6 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -320,19 +321,30 @@ public void buildPattern() {
...
         return settableWatchedProps;
     }
 
-    private void populateSettableWatchedProps(String prop, Collection<String> settableProps, Set<String> settableWatchedProps, boolean raiseErrorForNonSettableProp) {
-        String actualProp = prop.startsWith("!") ? prop.substring(1) : prop;
+    private void populateSettableWatchedProps(String prop, Collection<String> settableProps, Set<String> settableWatchedProps, boolean raiseErrorForNonSettableOrDuplicatedProp) {
+        String actualProp = prop;
+        if (prop.startsWith("!")) {
+            actualProp = prop.substring(1).trim();
+            prop = "!" + actualProp;
```
lucamolteni: substring returns a new string, `prop` hasn't be changed so no need to reappend this

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -320,19 +321,30 @@ public void buildPattern() {
...
+        }
         if (actualProp.equals("*") || settableProps.contains(actualProp)) {
+            if (raiseErrorForNonSettableOrDuplicatedProp && (settableWatchedProps.contains(actualProp) || settableWatchedProps.contains("!" + actualProp))) {
```
lucamolteni: isn't `! + actualProp` `prop`?

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
```diff
@@ -320,19 +321,30 @@ public void buildPattern() {
...
+            lookAheadProps.stream().forEach(prop -> populateSettableWatchedProps(prop, settableProps, settableWatchedProps, false)); // okay to have non-settable prop in lookAhead
+        }
+        
         return settableWatchedProps;
     }
 
-    private void populateSettableWatchedProps(String prop, Collection<String> settableProps, Set<String> settableWatchedProps, boolean raiseErrorForNonSettableProp) {
-        String actualProp = prop.startsWith("!") ? prop.substring(1) : prop;
+    private void populateSettableWatchedProps(String prop, Collection<String> settableProps, Set<String> settableWatchedProps, boolean raiseErrorForNonSettableOrDuplicatedProp) {
+        String actualProp = prop;
```
lucamolteni: find better name for prop

lucamolteni (changes requested):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
mariofusco: unfortunately it is necessary because of the trim(), to eventually turn `! prop` into `!prop`

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
mariofusco: no, it is only if `prop` was negated. If it wasn't `prop` and `actualProp` are the same but I want to check if the set didn't already contain `!prop` while inserting `prop`.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/visitor/pattern/PatternDSL.java
lucamolteni: Please rename it to `trimmedProperty` then

mariofusco:
@lucamolteni I refactored the variables names, please review again, thanks.

lucamolteni (approved):

pibizza (approved):
Minor comment on a TODO. Good for me otherwise.

PR #3851, opened 2021-09-27
DROOLS-6624 DMN DRGAnalysis utils

tarilabs:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6624

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/internal/utils/DRGAnalysisUtils.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/internal/utils/DRGAnalysisUtils.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.dmn.core.internal.utils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.kie.dmn.api.core.DMNModel;
+import org.kie.dmn.api.core.ast.DMNNode;
+import org.kie.dmn.api.core.ast.InputDataNode;
+import org.kie.dmn.core.ast.BusinessKnowledgeModelNodeImpl;
+import org.kie.dmn.core.ast.DMNBaseNode;
+import org.kie.dmn.core.ast.DecisionNodeImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/internal/utils/DRGAnalysisUtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        Assertions.assertThatThrownBy(() -> DRGAnalysisUtils.inputDataOfDecision(dmnModel, "Routing rules")).isInstanceOf(IllegalArgumentException.class);
+        
+        Collection<String> adjudicationIDs = DRGAnalysisUtils.inputDataOfDecision(dmnModel, "Adjudication");
+        assertThat(adjudicationIDs).containsExactlyInAnyOrder("Supporting documents", "Bureau data", "Requested product", "Applicant data");
+        
+        Collection<String> routingIDs = DRGAnalysisUtils.inputDataOfDecision(dmnModel, "Routing");
+        assertThat(routingIDs).containsExactlyInAnyOrder("Bureau data", "Requested product", "Applicant data");
+        
+        Collection<String> strategyIDs = DRGAnalysisUtils.inputDataOfDecision(dmnModel, "Strategy");
+        assertThat(strategyIDs).containsExactlyInAnyOrder("Requested product", "Applicant data");
```
tarilabs, line 72: @tteofili
from the perspective of TrustyAI I believe you are mainly interested to know for a given Decision node (square) which InputData nodes are impacting (round corner boxes)

This test case make the example using the DMN spec CH11 DMN-example-model:
image

let me know what do you think?
/cc @danielezonca

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/internal/utils/DRGAnalysisUtils.java
```diff
@@ -0,0 +1,125 @@
...
+import org.kie.dmn.core.ast.DMNBaseNode;
+import org.kie.dmn.core.ast.DecisionNodeImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DRGAnalysisUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DRGAnalysisUtils.class);
+    
+    public static Collection<DRGDependency> dependencies(DMNModel model, String nodeName) {
```
danielezonca, line 39: Is it possible to compile a "partial" `DMNModel`?
What I mean is that we already export the xml as plain string and this can be probably used to recreate the `DMNModel` but it doesn't have the imported models (if any) but as far as I see this type of analysis should work also in this case (for example we don't check imported types).
If not I think we need to revisit the way to export models to include all related models too (PMML too probably)

tarilabs: Not, it's not possible.
A `DMNModel` is a "compiled" representation, so it needs all imported resources.

What we could do, is that I implement a similar capability, but directly on the (unmarshalled) xml, but the results of the DRG graph analysis might not be 100% in all cases, as the semantic "basis" is naturally different.

Wdyt, shall we go ahead with this one as-is (it's helpful in general) and I implement a variant of analogous DRG Analysis as a separate PR?

danielezonca: Yes I agree, this feature is a useful "insight API" for DMN models in general so I'm fine to merge and then see if this can fit TrustyAI needs and how :)

danielezonca (approved):

hellowdan (approved):

PR #3864, opened 2021-09-30
[DROOLS-6625] Validate input data

gitgabrio:

See https://issues.redhat.com/browse/DROOLS-6625

This PR
1) define three "validation" steps inside PreProcess: `convertInputData`, `verifyFixInvalidValues`, `verifyAddMissingValues`
2) modify one integration test for each model to verify correct application
3) modify MiningModel evaluator to propagate results from one segment as input to the next one

### DIFF
--- a/kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-commons/src/test/java/org/kie/pmml/compiler/commons/builders/KiePMMLModelCodegenUtilsTest.java
+++ b/kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-commons/src/test/java/org/kie/pmml/compiler/commons/builders/KiePMMLModelCodegenUtilsTest.java
@@ -16,55 +16,139 @@
 
 package org.kie.pmml.compiler.commons.builders;
 
-import java.io.IOException;
 import java.util.List;
-import java.util.Map;
 
-import javax.xml.bind.JAXBException;
-
-import com.github.javaparser.utils.Pair;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.ConstructorDeclaration;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.Statement;
+import org.dmg.pmml.Field;
+import org.dmg.pmml.Model;
 import org.dmg.pmml.PMML;
+import org.dmg.pmml.TransformationDictionary;
+import org.junit.BeforeClass;
 import org.junit.Test;
-import org.kie.pmml.api.enums.DATA_TYPE;
+import org.kie.pmml.api.exceptions.KiePMMLException;
+import org.kie.pmml.compiler.commons.utils.JavaParserUtils;
 import org.kie.pmml.compiler.commons.utils.KiePMMLUtil;
-import org.xml.sax.SAXException;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.kie.pmml.compiler.commons.CommonTestingUtils.getFieldsFromDataDictionary;
-import static org.kie.pmml.compiler.commons.builders.KiePMMLModelCodegenUtils.getMissingValueReplacementsMap;
-import static org.kie.pmml.compiler.commons.builders.KiePMMLModelCodegenUtils.getRequiredFieldsList;
+import static org.kie.pmml.compiler.commons.mocks.TestingModelImplementationProvider.KIE_PMML_TEST_MODEL_TEMPLATE;
+import static org.kie.pmml.compiler.commons.mocks.TestingModelImplementationProvider.KIE_PMML_TEST_MODEL_TEMPLATE_JAVA;
+import static org.kie.pmml.compiler.commons.utils.JavaParserUtils.MAIN_CLASS_NOT_FOUND;
+import static org.kie.pmml.compiler.commons.utils.ModelUtils.getFieldsFromDataDictionaryTransformationDictionaryAndModel;
 import static org.kie.test.util.filesystem.FileUtils.getFileInputStream;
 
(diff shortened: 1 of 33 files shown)


kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-commons/src/test/java/org/kie/pmml/compiler/commons/builders/KiePMMLModelCodegenUtilsTest.java
```diff
@@ -16,41 +16,139 @@
...
+        modelTemplate = cloneCU.getClassByName("className")
+                .orElseThrow(() -> new KiePMMLException(MAIN_CLASS_NOT_FOUND + ": " + "className"));
     }
 
+    @Test
+    public void init() {
+        ConstructorDeclaration constructorDeclaration = modelTemplate.getDefaultConstructor().get();
+        KiePMMLModelCodegenUtils.init(modelTemplate, fields, transformationDictionary, model);
+        BlockStmt body = constructorDeclaration.getBody();
+        Statement expected = JavaParserUtils.parseConstructorBlock("{\n" +
```
jiripetrlik, line 72: I would suggest to move these long strings with code to files in test-resources. It will be much more readable.

jiripetrlik (approved):
I have one minor comment. My one concern is what will happen if it throws an exception inside of DMN evaluation and what is the correct behavior here.

kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-commons/src/test/java/org/kie/pmml/compiler/commons/builders/KiePMMLModelCodegenUtilsTest.java
gitgabrio: > I have one minor comment. My one concern is what will happen if it throws an exception inside of DMN evaluation and what is the correct behavior here.

DMN  engine should verify/manage the situation: there is already a "catch" somewhere, anyway...

gitgabrio: > I would suggest to move these long strings with code to files in test-resources. It will be much more readable.

@jiripetrlik 
You are perfectly right. I would prefer to do that refactor globally, once and for all, soon after this PR: see https://issues.redhat.com/browse/DROOLS-6635. WDYT ?

jiripetrlik: Thank you for your info and :+1: for DROOLS-6635.

kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-core/src/main/java/org/kie/pmml/compiler/executor/PMMLCompilerImpl.java
```diff
@@ -144,47 +149,49 @@
...
+//            localTransformations.getDerivedFields().stream().map(Field.class::cast)
+//                    .forEach(toReturn::add);
+//        }
+//        Output output =  model.getOutput();
+//        if (output != null && output.hasOutputFields()) {
+//            output.getOutputFields().stream().map(Field.class::cast)
+//                    .forEach(toReturn::add);
+//        }
+//        return toReturn;
+//    }
```
danielezonca: To be removed

kie-pmml-trusty/kie-pmml-evaluator/kie-pmml-evaluator-core/src/main/java/org/kie/pmml/evaluator/core/utils/PreProcess.java
```diff
@@ -106,9 +190,111 @@ static void executeTransformations(final ProcessingDTO processingDTO,
...
+    static void manageInvalidValues(final MiningField miningField, final ParameterInfo parameterInfo,
+                                    final List<ParameterInfo> toRemove) {
+        INVALID_VALUE_TREATMENT_METHOD invalidValueTreatmentMethod =
+                miningField.getInvalidValueTreatmentMethod() != null ?
+                        miningField.getInvalidValueTreatmentMethod()
+                        : INVALID_VALUE_TREATMENT_METHOD.RETURN_INVALID;
+        Object originalValue = parameterInfo.getValue();
+        switch (invalidValueTreatmentMethod) {
+            case RETURN_INVALID:
+                throw new KiePMMLException("Invalid value " + originalValue + " for " + miningField.getName());
```
danielezonca: I think `KiePMMLException ` is a bit too generic, what about create a subclass like `KiePMMLValidationException` or similar so that user can catch differently validation and runtime errors?

kie-pmml-trusty/kie-pmml-compiler/kie-pmml-compiler-core/src/main/java/org/kie/pmml/compiler/executor/PMMLCompilerImpl.java
gitgabrio: @danielezonca 

done

kie-pmml-trusty/kie-pmml-evaluator/kie-pmml-evaluator-core/src/main/java/org/kie/pmml/evaluator/core/utils/PreProcess.java
gitgabrio: @danielezonca 
I created KiePMMLInputDataException (KiePMMLValidationException already exists and it is dedicated validation errors of the pmml model itself).
Please note that in a couple of places I kept KiePMMLException - but those exception should really never happen

danielezonca (approved):

gitgabrio:
Jenkins retest this

PR #3900, opened 2021-10-13
DROOLS-6647 DMN Signavio MID with multiple decision

tarilabs:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6647

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/MultiInstanceDecisionLogic.java
+++ b/kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/MultiInstanceDecisionLogic.java
@@ -20,7 +20,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import org.kie.dmn.api.core.DMNContext;
@@ -39,6 +41,8 @@ import org.kie.dmn.core.compiler.DecisionCompiler;
 import org.kie.dmn.core.impl.DMNContextImpl;
 import org.kie.dmn.core.impl.DMNModelImpl;
 import org.kie.dmn.core.impl.DMNResultImpl;
+import org.kie.dmn.core.internal.utils.DRGAnalysisUtils;
+import org.kie.dmn.core.internal.utils.DRGAnalysisUtils.DRGDependency;
 import org.kie.dmn.feel.FEEL;
 import org.kie.dmn.feel.runtime.functions.AllFunction;
 import org.kie.dmn.feel.runtime.functions.AnyFunction;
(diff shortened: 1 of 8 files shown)


tarilabs:
@j-beyer please let me know if your team concurs this fixes the originally reported problem. If you have additional test cases don't hesitate to share, so to incorporate them directly.

kie-dmn/kie-dmn-signavio/src/main/java/org/kie/dmn/signavio/MultiInstanceDecisionLogic.java
```diff
@@ -117,13 +121,30 @@ public void compileEvaluator(DMNNode node, DMNCompilerImpl compiler, DMNCompiler
...
+                // DROOLS-6647 multi-instance decisions with multiple levels of decisions inside
+                List<String> reqDecisionKeys = new ArrayList<>();
+                for (Entry<String, DMNNode> candidate : di.getDependencies().entrySet()) { 
+                    if (candidate.getValue() instanceof DecisionNodeImpl && allWrappedDeps.contains(candidate.getValue())) {
+                        DecisionNodeImpl reqDecision = (DecisionNodeImpl) candidate.getValue();
+                        recurseNodeToRemoveItAndDepsFromModelIndex(reqDecision, cModel);
+                        reqDecisionKeys.add(candidate.getKey());
+                        miEvaluator.addReqDecision(reqDecision);
+                    }
+                }
```
j-beyer: Given that we don't want to rely on the MID's dependencies, could we also simply iterate over the topLevelDecisions dependencies like this?
```suggestion
                List<String> reqDecisionKeys = new ArrayList<>();
                for (DMNNode candidate : allWrappedDeps) { 
                    if (candidate.getValue() instanceof DecisionNodeImpl) {
                        DecisionNodeImpl reqDecision = (DecisionNodeImpl) candidate.getValue();
                        recurseNodeToRemoveItAndDepsFromModelIndex(reqDecision, cModel);
                        reqDecisionKeys.add(candidate.getKey());
                        miEvaluator.addReqDecision(reqDecision);
                    }
                }
```

j-beyer: Would the test still be green with this change, with the existing XML and also if you remove the `dmn:requiredDecision` tags from the multi instance decision?

j-beyer (commented):
Looks good!
I've added a [comment on the ticket](https://issues.redhat.com/browse/DROOLS-6647?focusedCommentId=19151329&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19151329) that adds a test case, and also describes that we would rather not rely on the MID dependencies. I've added comments here in the PR suggesting what could be changed.

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/SignavioTest.java
```diff
@@ -293,4 +293,35 @@ public void testSignavioIterateMultiinstanceWithComplexInputs() {
...
+        DMNResult evaluateAll = runtime.evaluateAll(model0, context);
+        LOG.info("{}", evaluateAll);
+    
+        assertThat(evaluateAll.getDecisionResultByName("overallage").getResult(), is(new BigDecimal("18")));
+    }
+    
+    @Test
+    public void testSignavioIterateMultiinstanceMultipleDecisionsOutside() {
+        DMNRuntime runtime = createRuntime("MID with outside requirement.dmn");
+
```
pibizza, line 315: I have a doubt that using names with underscores can be an annoyance, but I do not have strong opinions on that. However I suggest to keep the naming of the files consistent - right now some include the Signavio name, some use underscores, some use the word Test. Any of them is good as long as it is consistently applied.

j-beyer:
Looks great, thank you! We'd run some tests tomorrow again and will let you know of the results 🙂

kie-dmn/kie-dmn-signavio/src/test/java/org/kie/dmn/signavio/SignavioTest.java
tarilabs: Fair comment, but this PR addresses a bug and I didn't want to include additional scope (refactor existing unaffected code).

- all test method indeed starts with `test` already
- I have removed the one using `_` with 84fa087
- _some_ used Signavio in the name to signify in the contribution that it works sensibly different than original DMN spec (MID extension) I believe --while on the contrary it is _relatively_ expected different Implementors might adopt their own set of extended built-in FEEL functions

pibizza: Ok, my fault. I did not explain myself. I was referring to the names of the resources used in the tests, "MID with outside requirement.dmn" and how they are all different in the resources dir. However I think your comment about this being a fix is fair. We can address this in a separate commit, if we feel it is necessary.

pibizza (approved):

j-beyer:
With the recent major version update in drools, we're struggling to build the kie-dmn-signavio jar.
Could you provide us with a JAR with your changes that we can use for testing?

tarilabs:
@j-beyer it could be likely as this refers to `main` which corresponds to Drools 8 series.

You might use https://github.com/kiegroup/drools/pull/3912 to try do that instead.
Usually I backport once done on a default branch, but in this case I will keep the 2 PR aligned if anything else comes up.

Let me know if using the other PR targeting `7.x` branch doesn't help.

tarilabs:
@j-beyer any update on this one, following my https://github.com/kiegroup/drools/pull/3900#issuecomment-943495568 did that helped for your team to test? Thx

j-beyer:
Hi, sorry for the delayed response!
Yes, we were able to test using the branch based on version 7. Thanks for the hint!

During testing, we've discovered a small issue with the implementation.
If a decision inside of an MID depends on a decision outside of an MID, we will remove the outside decision and its dependencies from the model index via `recurseNodeToRemoveItAndDepsFromModelIndex`. As a result, it's not possible to evaluate the outside decision separately anymore, you can only evaluate the MID.

I.e. take the example of `test/resources/org/kie/dmn/signavio/MID with outside requirement.dmn` - here it's not possible anymore to evaluate the decision `determineModifier` on its own.

We believe that removing the dependencies from the model index only makes sense for the nodes inside of the MID, but not for anything on the outside that is independent of the MID.
Currently, my colleague is investigating if we can determine which dependencies are inside vs. outside from the requirements tree, to see if that would be a possible fix.
If you can imagine another solution approach, we'd also be glad to hear that.

Again, sorry for the slow asynchronous communication 😕

TommyNeubert:
@tarilabs  the changes look good and cover our idea

PR #3929, opened 2021-10-18
[DROOLS-6623] remove KieBase serialization

mariofusco:

**NOTE!:** Double check the target branch for this PR.
The default is `main` so it will target Drools 8 / Kogito / Optaplanner 8.
If this PR is not strictly related to drools and kogito project in `drools.git`, it should probably target `7.x`as a branch

**JIRA**: https://issues.redhat.com/browse/DROOLS-6623


mdproctor:
I wouldn't rush to delete kbase serialisation. It doesn't get in the way of maintenance or tests. we can make it internal only. But it may come in useful in the future, and having the touch points is useful. if it got in the way, i'd say delete it - but it's mostly harmless. For example, we may need ultra fast ways to get a drools session up and running from cold, in edge based computing. Re-adding all this from scratch again, would be cumbersome.
 
I don't think it should be a end user feature, but I could imagine embedded situations in edge computing where we own the full life cycle of the engine and would want this faster restart.

mariofusco:
@mdproctor I think we don't need kbase serialization at the moment. If in future we will find that we want or need it we could better reimplement it with a visitor patter or with an external mechanism like protobuf, rather than using plain java serialization. Finally in case we will need this serialization urgently e won't have time to implement a better serialization mechanism we could always revert this pull request.

fbraun4358:
@mdproctor 
From an end user point of view, having the `KieBase` implement `Serializable` and not actually be serializable if you use any of the more advanced features is both extremely confusing and cumbersome to deal with.

We were using serialization in Drools 6 and also doing things which should not actually work when serialized. Switching to 7 has been a bit of a nightmare and we have files like 7 or 8 bugs for the executable model to get edge cases working.

See [DROOLS-2552](https://issues.redhat.com/browse/DROOLS-2552) and [DROOLS-6446](https://issues.redhat.com/browse/DROOLS-6446) for examples of it being both confusing/underdocumented to use and just not working in 7 respectively.

mariofusco:
We decided to keep KieBase serialization as it is for now.

radtriste:
jenkins rerun kogito-apps tests

radtriste:
jenkins rerun kogito-examples tests

PR #3566, opened 2021-04-27
KOGITO-4647 : [DMN] FEEL client-side - Remove mirror classes

Rikkola:
**JIRA**:  

[KOGITO-4647](https://issues.redhat.com/browse/KOGITO-4647)

**referenced Pull Requests**: 

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1632
* https://github.com/kiegroup/drools/pull/3566
* https://github.com/kiegroup/kie-wb-common/pull/3712

**Notable changes**:
- `ClassUtil`and `StringUtil` wrap GWT problematic methods. Might be better to have these wrappers around. If another problematic method arises then the same approach can be taken without the need for core DMN engineer to update the JS version. The problem is that the GWT versions of `Class` and `String` do not have these methods, I can make a Maven plugin that appends them to the GWT source, but that might be harder to manage, even if it touches less of the code in `kie-dmn-feel`.
- Less GwtIngore, due to extended super classes
- Java reflection has been super classed and this will be used to [provide the functions in JS runtime](https://issues.redhat.com/browse/KOGITO-4646)

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #3963, opened 2021-10-29
[7.x] [DROOLS-6688] -"10" results in -10 in FEEL

tkobayas:
**Ports**

Forward-port of this PR is https://github.com/kiegroup/drools/pull/3992

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-6688

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
+++ b/kie-dmn/kie-dmn-feel-gwt-showcase/README.md
@@ -1,8 +1,27 @@
 To compile and run this project in the main Maven build, ensure the necessary profile property is set.
 
 Eg.
+
 ```
 mvn clean install -PfullProfile
  or
 mvn clean install -Dfull
 ```
+
+### Trouble-shooting
+If you hit a strange error by gwt-maven-plugin after changing a class in `kie-dmn-feel` complaining that the class doesn't seem to be changed, check if the same FQCN class exists in `kie-dmn-feel-gwt`. You would need to change the class in the same way.
+
+for example)
+I added `CANNOT_BE_SIGNED` to `org.kie.dmn.feel.util.Msg` in `kie-dmn-feel`.
+
+```
+[2021-11-02T07:18:51.749Z] [INFO] --- gwt-maven-plugin:2.9.0:compile (default) @ kie-dmn-feel-gwt-showcase ---
+[2021-11-02T07:18:51.749Z] [INFO] auto discovered modules [org.kie.dmn.feel.FEELShowcaseWebapp]
+[2021-11-02T07:19:14.267Z] [INFO] Compiling module org.kie.dmn.feel.FEELShowcaseWebapp
+[2021-11-02T07:19:14.267Z] [INFO]    Tracing compile failure path for type 'org.kie.dmn.feel.lang.ast.SignedUnaryNode'
+[2021-11-02T07:19:14.267Z] [INFO]       [ERROR] Errors in 'jar:file:/home/jenkins/workspace/KIE/7.x/pullrequest/drools-7.x.pr/bc/kiegroup_drools/kie-dmn/kie-dmn-feel/target/kie-dmn-feel-7.62.0-SNAPSHOT-sources.jar!/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java'
+[2021-11-02T07:19:14.267Z] [INFO]          [ERROR] Line 66: CANNOT_BE_SIGNED cannot be resolved or is not a field
+[2021-11-02T07:19:14.267Z] [INFO]    [ERROR] Aborting compile due to errors in some input files
+```
+
+In this case, I have to change `org.kie.dmn.feel.util.Msg` in `kie-dmn-feel-gwt` as well.
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
@@ -458,11 +458,6 @@ public class CompiledFEELSupport {
         return null;
     }
 
-    public static Object notifyCompilationWarn(EvaluationContext feelExprCtx, String message) {
-        feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.WARN, message, null));
-        return null;
-    }
-
     public static Object coerceNumber(Object value) {
         return EvalHelper.coerceNumber(value);
     }
@@ -514,15 +509,6 @@ public class CompiledFEELSupport {
         return new MethodCallExpr(
                 new NameExpr("CompiledFEELSupport"),
                 "notifyCompilationError",
-                new NodeList<>(
-                        new NameExpr("feelExprCtx"),
-                        Expressions.stringLiteral(msg)));
-    }
-
-    public static MethodCallExpr compiledWarnExpression(String msg) {
-        return new MethodCallExpr(
-                new NameExpr("CompiledFEELSupport"),
-                "notifyCompilationWarn",
                 new NodeList<>(
                         new NameExpr("feelExprCtx"),
                         Expressions.stringLiteral(msg)));
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
@@ -531,4 +531,12 @@ public class CompiledFEELSupport {
     public static BigDecimal pow(BigDecimal l, BigDecimal r) {
         return BigDecimalMath.pow( l, r, MathContext.DECIMAL128 );
     }
+
+    public static BigDecimal negate(EvaluationContext feelExprCtx, Object value) {
+        if (value instanceof String) {
+            feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.ERROR, Msg.createMessage(Msg.CANNOT_NEGATE), null));
+            return null;
+        }
+        return ((BigDecimal)value).negate();
+    }
 }
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
@@ -533,10 +533,27 @@ public class CompiledFEELSupport {
     }
 
     public static BigDecimal negate(EvaluationContext feelExprCtx, Object value) {
-        if (value instanceof String) {
-            feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.ERROR, Msg.createMessage(Msg.CANNOT_NEGATE), null));
+        if (isValidSignedType(feelExprCtx, value)) {
+            return ((BigDecimal) value).negate();
+        } else {
             return null;
         }
-        return ((BigDecimal)value).negate();
+    }
+
+    public static BigDecimal positive(EvaluationContext feelExprCtx, Object value) {
+        if (isValidSignedType(feelExprCtx, value)) {
+            return (BigDecimal) value;
+        } else {
+            return null;
+        }
+    }
+
+    private static boolean isValidSignedType(EvaluationContext feelExprCtx, Object value) {
+        if (value instanceof String) {
+            feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.ERROR, Msg.createMessage(Msg.CANNOT_BE_SIGNED), null));
+            return false;
+        } else {
+            return true;
+        }
     }
 }
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
@@ -458,6 +458,11 @@ public class CompiledFEELSupport {
         return null;
     }
 
+    public static Object notifyCompilationWarn(EvaluationContext feelExprCtx, String message) {
+        feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.WARN, message, null));
+        return null;
+    }
+
     public static Object coerceNumber(Object value) {
         return EvalHelper.coerceNumber(value);
     }
@@ -509,6 +514,15 @@ public class CompiledFEELSupport {
         return new MethodCallExpr(
                 new NameExpr("CompiledFEELSupport"),
                 "notifyCompilationError",
+                new NodeList<>(
+                        new NameExpr("feelExprCtx"),
+                        Expressions.stringLiteral(msg)));
+    }
+
+    public static MethodCallExpr compiledWarnExpression(String msg) {
+        return new MethodCallExpr(
+                new NameExpr("CompiledFEELSupport"),
+                "notifyCompilationWarn",
                 new NodeList<>(
                         new NameExpr("feelExprCtx"),
                         Expressions.stringLiteral(msg)));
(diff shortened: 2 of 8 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
```diff
@@ -458,6 +458,11 @@ public static Object notifyCompilationError(EvaluationContext feelExprCtx, Strin
         return null;
     }
 
+    public static Object notifyCompilationWarn(EvaluationContext feelExprCtx, String message) {
+        feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.WARN, message, null));
+        return null;
+    }
+
```
tkobayas: In order to raise a WARN for JAVA_TRANSLATED test, I added this method.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELValuesConstantsTest.java
```diff
@@ -48,7 +48,8 @@
                 { "\"šomeÚnicodeŠtriňg\"", "šomeÚnicodeŠtriňg" , null},
                 { "\"横綱\"", "横綱" , null},
                 { "\"thisIsSomeLongStringThatMustBeProcessedSoHopefullyThisTestPassWithItAndIMustWriteSomethingMoreSoItIsLongerAndLongerAndLongerAndLongerAndLongerTillItIsReallyLong\"", "thisIsSomeLongStringThatMustBeProcessedSoHopefullyThisTestPassWithItAndIMustWriteSomethingMoreSoItIsLongerAndLongerAndLongerAndLongerAndLongerTillItIsReallyLong" , null},
-                { "\"\"", "" , null}
+                { "\"\"", "" , null},
+                { "-\"10\"", null , FEELEvent.Severity.WARN}
```
tarilabs: here you could add something like 
```java
{ "-string(\"10\")", null , FEELEvent.Severity.ERROR}
```

to prove the solution is general enough

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java
```diff
@@ -66,7 +66,12 @@ public Object evaluate(EvaluationContext ctx) {
             ctx.notifyEvt( astEvent(Severity.WARN, Msg.createMessage(Msg.NEGATING_A_NULL)));
             return null;
         } else if ( Sign.NEGATIVE == sign ) {
-            return BigDecimal.valueOf( -1 ).multiply( result );
+            if (expression instanceof StringNode) {
+                ctx.notifyEvt( astEvent(Severity.WARN, Msg.createMessage(Msg.NEGATING_A_NULL)));
+                return null;
+            } else {
+                return BigDecimal.valueOf( -1 ).multiply( result );
```
tarilabs: I think you are getting closer to a solution!

Unfortunately this solution mixes run-time with compile-time (when calling `evaluate(EvaluationContext ctx)` we will be at run-time) so I believe we can do something that only consider run-time.
In other words, I'm not yet entirely convinced `if (expression instanceof StringNode) {` is a good idea here..

For Interpreted FEEL, I would do the following:
1. extract `expression.evaluate( ctx )` from line 64 into a expressionResult variable
2. check if expressionResult instanceof String, then emit as Error a new type of Msg, something like ~"cannot negate a value which is not a number"

Then the FEEL->Java translation should align with this behaviour.
I believe we need to produce some code that if it's attempting to cast to BigDecimal here:
https://github.com/kiegroup/drools/blob/d1adc0588c0acd2e0756662817d4876cd4fd5945/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/Expressions.java#L106

it will emit (at run-time) the same error mentioned above.

I don't believe we need to modify ASTCompilerVisitor.java nor CompiledFEELSupport.java 

I hope I was able to explain the reason behind the different approach needed?
I hope the additional test case example (`-string("10")`) I suggest as part of the review helps to better support this.

What do you think?

tkobayas:
@tarilabs  Thank you very much for the valuable advice!

I changed the `AST_INTERPRETED` code accordingly.

However, I couldn't yet implement it for `JAVA_TRANSLATED`.

> I believe we need to produce some code that if it's attempting to cast to BigDecimal here:
> it will emit (at run-time) the same error mentioned above.

So I'd like to to produce code like
```
    public Object apply(EvaluationContext feelExprCtx) {
        Object evaluatedResult = XXX;
        if (evaluatedResult instanceof String) {
            feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.ERROR, Msg.createMessage(Msg.CANNOT_NEGATE), null));
            return null;
        }
        return ((java.math.BigDecimal) evaluatedResult).negate();
    }
```

but DirectCompilerResult can have one Expression (not Statement) so it's not possible? Should I make use of CompiledFEELSupport? Sorry if I'm misunderstanding JavaParser capability.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
```diff
@@ -531,4 +531,12 @@ public void onEvent(FEELEvent evt) {
...
         return BigDecimalMath.pow( l, r, MathContext.DECIMAL128 );
     }
+
+    public static BigDecimal negate(EvaluationContext feelExprCtx, Object value) {
+        if (value instanceof String) {
+            feelExprCtx.notifyEvt(() -> new ASTEventBase(Severity.ERROR, Msg.createMessage(Msg.CANNOT_NEGATE), null));
+            return null;
+        }
+        return ((BigDecimal)value).negate();
+    }
```
tkobayas: Implemented for JAVA_TRANSLATED as `CompiledFEELSupport` method. Is it okay?

tkobayas:
Linux - Pull Request:

```
[2021-11-02T07:18:51.749Z] [INFO] --- gwt-maven-plugin:2.9.0:compile (default) @ kie-dmn-feel-gwt-showcase ---
[2021-11-02T07:18:51.749Z] [INFO] auto discovered modules [org.kie.dmn.feel.FEELShowcaseWebapp]
[2021-11-02T07:19:14.267Z] [INFO] Compiling module org.kie.dmn.feel.FEELShowcaseWebapp
[2021-11-02T07:19:14.267Z] [INFO]    Tracing compile failure path for type 'org.kie.dmn.feel.lang.ast.SignedUnaryNode'
[2021-11-02T07:19:14.267Z] [INFO]       [ERROR] Errors in 'jar:file:/home/jenkins/workspace/KIE/7.x/pullrequest/drools-7.x.pr/bc/kiegroup_drools/kie-dmn/kie-dmn-feel/target/kie-dmn-feel-7.62.0-SNAPSHOT-sources.jar!/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java'
[2021-11-02T07:19:14.267Z] [INFO]          [ERROR] Line 66: CANNOT_NEGATE cannot be resolved or is not a field
[2021-11-02T07:19:14.267Z] [INFO]    [ERROR] Aborting compile due to errors in some input files
```
I can reproduce the error on my end just by building kie-dmn-feel-gwt-showcase. Hmm, CANNOT_NEGATE is already added to org.kie.dmn.feel.util.Msg which exists in the same kie-dmn-feel.jar. @tarilabs  Have you experienced such an error?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
tarilabs: yep looks good to me now, as it is consistent with the interpreted version, good work !

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELValuesConstantsTest.java
```diff
@@ -48,7 +48,9 @@
                 { "\"šomeÚnicodeŠtriňg\"", "šomeÚnicodeŠtriňg" , null},
                 { "\"横綱\"", "横綱" , null},
                 { "\"thisIsSomeLongStringThatMustBeProcessedSoHopefullyThisTestPassWithItAndIMustWriteSomethingMoreSoItIsLongerAndLongerAndLongerAndLongerAndLongerTillItIsReallyLong\"", "thisIsSomeLongStringThatMustBeProcessedSoHopefullyThisTestPassWithItAndIMustWriteSomethingMoreSoItIsLongerAndLongerAndLongerAndLongerAndLongerTillItIsReallyLong" , null},
-                { "\"\"", "" , null}
+                { "\"\"", "" , null},
+                { "-\"10\"", null , FEELEvent.Severity.ERROR},
+                { "-string(\"10\")", null , FEELEvent.Severity.ERROR}
```
tarilabs: Just to make sure, I would add a few tests to increase coverage, like:
```
+10   // I would expect return BigDecimal of 10
+"10"   //null, and an error
```

tarilabs (commented):
Great work! 👍  thank you, just minor comments.

For the gwt error, I think we unfortunately have to deal with the code duplication.
There is an equivalent `Msg` class in one of the gwt module, it should be sufficient to apply the same new field constant also there.
You can reproduce locally by running maven with `-Dfull` or `-PfullProfile`.

Sorry for the inconvenience, this is a known issue with the GWT implementation
/cc @danielezonca

tkobayas:
> There is an equivalent `Msg` class in one of the gwt module, it should be sufficient to apply the same new field constant also there.

Thanks! I added the tips to `README.md` in `kie-dmn-feel-gwt-showcase`

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELValuesConstantsTest.java
tkobayas: > `+10   // I would expect return BigDecimal of 10`

This test already has `{ "+50", new BigDecimal( "50" ) , null},` so we don't need to add?

> `+"10"   //null, and an error`

Sure, Added. I added another method `positive` for the case.

Btw, this test doesn't have `+string("10")` yet. I noticed that `+string("10")` works differently from `-string("10")`. `expression.evaluate( ctx )` returns `StringFunction`, not `String`. Probably I need to look into the parser phase. May I split it to another JIRA? Or shall I work on it in the same JIRA/PR? Let me know your thoughts. Thanks!

tarilabs: > This test already has { "+50", new BigDecimal( "50" ) , null}, so we don't need to add?

I agree 👍 

> Btw, this test doesn't have +string("10") yet

I believe this might not be a problem actually, as I'm not sure it's even a valid FEEL expression tbh. If you want to track it as a JIRA, is worth to be tracked and eventually investigated, but indeed as a lower priority

tkobayas: > I believe this might not be a problem actually, as I'm not sure it's even a valid FEEL expression tbh. If you want to track it as a JIRA, is worth to be tracked and eventually investigated, but indeed as a lower priority

Thanks. Then I leave it without filing a JIRA at the moment.

tkobayas:
Now it's ready. Please review, @tarilabs @hellowdan @pibizza . Thanks!

kie-dmn/kie-dmn-feel-gwt-showcase/README.md
```diff
@@ -1,8 +1,27 @@
...
+
 ```
 mvn clean install -PfullProfile
  or
 mvn clean install -Dfull
-```
\ No newline at end of file
+```
+
+### Trouble-shooting
```
tarilabs, line 11: Thank you for taking the extra step of documenting this 🙏

tarilabs (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
```diff
@@ -531,4 +531,29 @@ public void onEvent(FEELEvent evt) {
...
+    public static BigDecimal positive(EvaluationContext feelExprCtx, Object value) {
+        if (isValidSignedType(feelExprCtx, value)) {
+            return (BigDecimal) value;
+        } else {
+            return null;
+        }
+    }
+
+    private static boolean isValidSignedType(EvaluationContext feelExprCtx, Object value) {
+        if (value instanceof String) {
```
danielezonca, line 552: Minor note: are you sure the only options are `Sting` or number? This code assumes that it can be a string or must be a number. Isn't possible to have other options (i.e. I remember `LocalDate` class but I don't know if here can happen, or even a `List`).
I might be wrong and this cannot happen because in this codepath we are sure that this cannot happen, my main concern is that it looks inconsistent: the error message is specific for "everything except numbers" but the condition is verified only for string so I'm guessing is there are other (not covered) options 🤔 
For example we can consider to change the error message as "sign cannot be applied to a string" 

@tarilabs wdyt?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java
```diff
@@ -61,7 +61,12 @@ public BaseNode getExpression() {
     @Override
     public Object evaluate(EvaluationContext ctx) {
         if (expression == null) return null;
-        BigDecimal result = EvalHelper.getBigDecimalOrNull( expression.evaluate( ctx ) );
+        Object expressionResult = expression.evaluate( ctx );
+        if (expressionResult instanceof String) {
```
danielezonca, line 65: Similar comment here.

kie-dmn/kie-dmn-feel-gwt-showcase/README.md
```diff
@@ -1,8 +1,27 @@
...
+[2021-11-02T07:18:51.749Z] [INFO] --- gwt-maven-plugin:2.9.0:compile (default) @ kie-dmn-feel-gwt-showcase ---
+[2021-11-02T07:18:51.749Z] [INFO] auto discovered modules [org.kie.dmn.feel.FEELShowcaseWebapp]
+[2021-11-02T07:19:14.267Z] [INFO] Compiling module org.kie.dmn.feel.FEELShowcaseWebapp
+[2021-11-02T07:19:14.267Z] [INFO]    Tracing compile failure path for type 'org.kie.dmn.feel.lang.ast.SignedUnaryNode'
+[2021-11-02T07:19:14.267Z] [INFO]       [ERROR] Errors in 'jar:file:/home/jenkins/workspace/KIE/7.x/pullrequest/drools-7.x.pr/bc/kiegroup_drools/kie-dmn/kie-dmn-feel/target/kie-dmn-feel-7.62.0-SNAPSHOT-sources.jar!/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java'
+[2021-11-02T07:19:14.267Z] [INFO]          [ERROR] Line 66: CANNOT_BE_SIGNED cannot be resolved or is not a field
+[2021-11-02T07:19:14.267Z] [INFO]    [ERROR] Aborting compile due to errors in some input files
+```
+
+In this case, I have to change `org.kie.dmn.feel.util.Msg` in `kie-dmn-feel-gwt` as well.
```
danielezonca, line 27: Side comment this duplication issue should be reduced by this PR https://github.com/kiegroup/drools/pull/3566 🤞

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/codegen/feel11/CompiledFEELSupport.java
tarilabs: @danielezonca it's nice you want to voluntarily check into this detail; this fixes DROOLS-6688 per title, while a generalized solution is tracked under DROOLS-6685 separately.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/SignedUnaryNode.java
tarilabs: as above

kie-dmn/kie-dmn-feel-gwt-showcase/README.md
tarilabs: one can always wish/dream, but I prefer to keep realistic.

pibizza (commented):
Looks good to me.

PR #3981, opened 2021-11-03
[KOGITO-2391] first lightweight session implementation

mariofusco:

**NOTE!:** Double check the target branch for this PR.
The default is `main` so it will target Drools 8 / Kogito / Optaplanner 8.
If this PR is not strictly related to drools and kogito project in `drools.git`, it should probably target `7.x`as a branch

**Ports** If a forward-port or a backport is needed, paste the forward port PR here

[link](https://www.example.com)

**JIRA**: https://issues.redhat.com/browse/KOGITO-2391

[link](https://www.example.com)

* https://github.com/kiegroup/drools/pull/3981
* https://github.com/kiegroup/kogito-runtimes/pull/1693

<details>
<summary>
</summary>

</details>


drools-core/src/main/java/org/drools/core/base/MapGlobalResolver.java
```diff
@@ -98,8 +98,11 @@ public void set(String identifier, Object value) {
     }
 
     public void setGlobal(String identifier, Object value) {
-        this.map.put( identifier,
-                      value );
+        if (value == null) {
+            this.map.remove(identifier);
+        } else {
+            this.map.put(identifier, value);
+        }
```
danielezonca, line 105: Is this an unrelated change?
Why `setGlobal` has a different behaviour now? `removeGlobal` method already exists

drools-core/src/main/java/org/drools/core/common/ReteEvaluator.java
```diff
@@ -30,14 +33,20 @@
...
 import org.drools.core.time.TimerService;
 import org.kie.api.runtime.Calendars;
+import org.kie.api.runtime.rule.EntryPoint;
+import org.kie.api.runtime.rule.FactHandle;
+import org.kie.api.runtime.rule.QueryResults;
+import org.kie.api.runtime.rule.StatefulRuleSession;
 import org.kie.api.time.SessionClock;
 
-public interface ReteEvaluator {
+public interface ReteEvaluator extends StatefulRuleSession {
```
danielezonca: Isn't this change against the idea that `ReteEvaluator` is a new "core"/standalone interface?

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/reteevaluator/ReteEvaluatorTest.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        KieBuilder kieBuilder = ks.newKieBuilder(kfs).buildAll(ExecutableModelProject.class);
+
+        List<Message> messages = kieBuilder.getResults().getMessages();
+        if ( !messages.isEmpty() ) {
+            fail( messages.toString() );
+        }
+
+        return (InternalKnowledgeBase) ks.newKieContainer( releaseId ).getKieBase();
```
danielezonca, line 77: It is fine for now to have a similar test but I think one of the goal of this "new" modularisation is reduce the footprint of the API so for example I don't think we should create a POM as part of the codepath to produce a compiled instance.
I think we will need soon a "lightweight kbuilder" too

drools-core/src/main/java/org/drools/core/impl/RuleUnitExecutorImpl.java
```diff
@@ -0,0 +1,295 @@
...
+
+    @Override
+    public void fireUntilHalt(AgendaFilter agendaFilter) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void halt() {
+        throw new UnsupportedOperationException();
+    }
```
danielezonca: To be removed from `ReteEvaluator` interface?

drools-core/src/main/java/org/drools/core/base/MapGlobalResolver.java
mariofusco: Having changed a bit the integration even with jbpm, I found that in some test cases (which are probably invalid) it calls that method passing a null value (with the purpose of removing the global) which however is not allowed by a `ConcurrentHashMap` and thus raised an Exception.

drools-core/src/main/java/org/drools/core/common/ReteEvaluator.java
mariofusco: The `StatefulRuleSession` is just a small interface in kie-api defining the firing methods. I don't expect any change in the signature of those methods and I don't see this as a legacy thing. I could alternative duplicate those methods in the `ReteEvaluator` interface, but I don't think this would be very useful.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/reteevaluator/ReteEvaluatorTest.java
mariofusco: I agree on that, this test was just the quickest thing that I could write to try the new lightweight implementation.

drools-core/src/main/java/org/drools/core/impl/RuleUnitExecutorImpl.java
mariofusco: Why should it be removed? The lightweight session still doesn't support the `fireUntilHalt` mode but I give for granted that it should and if so we will also need a method to halt that async firing.

drools-core/src/main/java/org/drools/core/base/MapGlobalResolver.java
danielezonca: Sorry but I still don't understand: are these `jbpm` tests new? If not how was it possible that current `main` code, that already has the `CuncurrentHashMap`, is working while now we need this change?
Btw my comment was mainly to understand the reason of this change :)

drools-core/src/main/java/org/drools/core/common/ReteEvaluator.java
danielezonca: My main concern is that it should be possible to use `ReteEvaluator` even when we don't expect just because it also implements this additional interface. If it is only to avoid to duplicate few methods let's duplicate them so that we keep well isolated this new interface for now

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/reteevaluator/ReteEvaluatorTest.java
danielezonca: Do we have a ticket to review this? If not, can you create one and link to Drools 8 epic?

drools-core/src/main/java/org/drools/core/impl/RuleUnitExecutorImpl.java
danielezonca: During our last call you mentioned that all necessary methods were implemented so I'm assuming that these are not necessary :)
Btw I would prefer to apply the opposite approach: add methods only when necessary instead of already add them upfront with an `UnsupportedOperationException`
Moreover `fireUntilHalt` has a quite strong/strict behaviour that I expect we will revisit (it locks the thread where it is invoked)

PR #3984, opened 2021-11-04
BXMSPROD-1540 composite actions centralized github actions

Ginxo:
- https://github.com/kiegroup/kogito-pipelines/pull/313
- https://github.com/kiegroup/kogito-runtimes/pull/1692
- https://github.com/kiegroup/kogito-examples/pull/968
- https://github.com/kiegroup/kogito-apps/pull/1093
- https://github.com/kiegroup/optaplanner/pull/1649
- https://github.com/kiegroup/drools/pull/3984

### DIFF
--- a/.github/workflows/drools-pr.yml
+++ b/.github/workflows/drools-pr.yml
@@ -29,6 +29,7 @@ jobs:
     steps:
       - name: Clean Disk Space
         uses: ginxo/kogito-pipelines/.ci/actions/disk-space@BXMSPROD-1540
+        if: ${{ matrix.os == 'ubuntu-latest' }}
         with:
           java-version: ${{ matrix.java-version }}
           maven-version: ${{ matrix.maven-version }}
--- a/.github/workflows/drools-pr.yml
+++ b/.github/workflows/drools-pr.yml
@@ -28,7 +28,7 @@ jobs:
     name: Maven Build
     steps:
       - name: Clean Disk Space
-        uses: ginxo/kogito-pipelines/.ci/actions/disk-space@BXMSPROD-1540
+        uses: ginxo/kogito-pipelines/.ci/actions/ubuntu-disk-space@BXMSPROD-1540
         if: ${{ matrix.os == 'ubuntu-latest' }}
         with:
           java-version: ${{ matrix.java-version }}
--- a/.github/workflows/drools-pr.yml
+++ b/.github/workflows/drools-pr.yml
@@ -37,13 +37,10 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' }}
         uses: kiegroup/kogito-pipelines/.ci/actions/long-paths@main
       - name: Java and Maven Setup
-        uses: kiegroup/kogito-pipelines/.ci/actions/java@main
+        uses: kiegroup/kogito-pipelines/.ci/actions/maven@main
         with:
           java-version: ${{ matrix.java-version }}
           maven-version: ${{ matrix.maven-version }}
-      - name: Cache Maven
-        uses: kiegroup/kogito-pipelines/.ci/actions/cache-maven@main
-        with:
           key-prefix: ${{ runner.os }}-${{ matrix.java-version }}-maven${{ matrix.maven-version }}
       - name: Build Chain
         uses: kiegroup/kogito-pipelines/.ci/actions/build-chain@main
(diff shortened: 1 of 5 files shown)


Ginxo:
it will obviously not work until https://github.com/kiegroup/kogito-pipelines/pull/313 is merged

.github/workflows/drools-pr.yml
```diff
@@ -22,48 +22,36 @@ jobs:
...
-        uses: actions/setup-java@v1
+      - name: Clean Disk Space
+        uses: kiegroup/kogito-pipelines/.ci/actions/ubuntu-disk-space@main
+        if: ${{ matrix.os == 'ubuntu-latest' }}
         with:
           java-version: ${{ matrix.java-version }}
-      # See https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven#caching-dependencies
-      - name: Cache Maven packages
-        uses: actions/cache@v2
+          maven-version: ${{ matrix.maven-version }}
```
radtriste: why do you need those for `ubuntu-disk-space` ?

.github/workflows/drools-pr.yml
```diff
@@ -22,48 +22,36 @@ jobs:
...
-          restore-keys: ${{ runner.os }}-m2
-      - name: Build Chain ${{ matrix.java-version }}
-        id: build-chain
-        uses: kiegroup/github-action-build-chain@v2.6.13
+          java-version: ${{ matrix.java-version }}
+          maven-version: ${{ matrix.maven-version }}
+      - name: Cache Maven
+        uses: kiegroup/kogito-pipelines/.ci/actions/cache-maven@main
+        with:
+          key-prefix: ${{ runner.os }}-${{ matrix.java-version }}-maven${{ matrix.maven-version }}
```
radtriste: Just thinking that `kiegroup/kogito-pipelines/.ci/actions/java@main` should be named `kiegroup/kogito-pipelines/.ci/actions/maven@main` instead.
And we merge those 2, which means incorporate the `cache` step in the maven one. And cache would be activated automatically if `cache-key-prefix` is set ?

radtriste (commented):
See my comments

Also worrying about the `@main` on actions. We would need the pipeline to update those when branching I guess ... wdyt ?

.github/workflows/drools-pr.yml
Ginxo: @radtriste I don't really know whether drools need to clean up the disk space or not, I just reproduce what's already defined following the new approach. We can treat this as a different PR (and test). WDYT?

.github/workflows/drools-pr.yml
Ginxo: done. The only problem about `cache-key-prefix` condition is they are not allowed at composite actions types. So I would suggest to force to define one. This is the way we propose to treat java-maven action for kogito-pipelines, which is basically the idea behind composite actions, to propose your own way of dealing with common stuff. WDYT @radtriste ?

radtriste: I am good with that

.github/workflows/drools-pr.yml
radtriste: I mean `java-version` and `maven-version` are not needed for that action or ?

.github/workflows/kogito-apps-pr.yml
```diff
@@ -22,51 +22,36 @@ jobs:
...
-        id: build-chain
-        uses: kiegroup/github-action-build-chain@v2.6.13
+          java-version: ${{ matrix.java-version }}
+          maven-version: ${{ matrix.maven-version }}
+          key-prefix: ${{ runner.os }}-${{ matrix.java-version }}-maven${{ matrix.maven-version }}
+      - name: Build Chain
+        uses: kiegroup/kogito-pipelines/.ci/actions/build-chain@main
         with:
-          definition-file: https://raw.githubusercontent.com/${GROUP}/kogito-pipelines/${BRANCH:main}/.ci/pull-request-config.yaml
+          annotations-prefix: ${{ runner.os }}-${{ matrix.java-version }}/${{ matrix.maven-version }}
```
radtriste, line 45: here we will need to set the definition-file with branch mapping (else it will fail on release branch) but that could be done in another PR

radtriste (approved):
minor comment

.github/workflows/drools-pr.yml
Ginxo: Ahhh yes, sorry, my fault. Solved on every job from this PR set

.github/workflows/kogito-apps-pr.yml
Ginxo: yes, I would treat here just the composite-action part please.

danielezonca (approved):

Ginxo:
failed tests from `Kogito Apps Build / Maven Build (ubuntu-latest, 11, 3.8.1) (pull_request)` not related with this change

PR #4035, opened 2021-11-22
[DROOLS-6707] move sessions to the new drools-kiesession module

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6707

Related PRs
https://github.com/kiegroup/drools/pull/4035
https://github.com/kiegroup/kogito-runtimes/pull/1737


drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationBuilder.java
```diff
@@ -267,7 +268,7 @@ protected void createBean( AbstractClassTypeDeclarationDescr typeDescr,
             success = ! kbuilder.hasErrors();
 
             if ( success ) {
-                ClassBuilder classBuilder = kbuilder.getBuilderConfiguration().getClassBuilderFactory().getClassBuilder(type);
```
danielezonca, line 270: Can you remove `getBuilderConfiguration` from [`KnowledgeBuilderImpl`](https://github.com/kiegroup/drools/blob/1433670701f3dbc3d393776078dded4604fd83f9/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java#L1867)?

drools-core/src/main/java/org/drools/core/RuleBaseConfiguration.java
```diff
@@ -1012,72 +984,6 @@ public String toString() {
         }
     }
 
-    public static class LogicalOverride
```
danielezonca, line 1015: Unused class?

drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
```diff
@@ -724,9 +723,8 @@ public TruthMaintenanceSystem getTruthMaintenanceSystem() {
         return tms;
     }
 
-    @Override
     public TraitHelper getTraitHelper() {
-        return new TraitDisabledHelper();
+        throw new UnsupportedOperationException("In order to use traits you must add the drools-traits module to your classpath");
```
danielezonca, line 727: Is it possible to remove this method (probably better with another ticket 🤔 )?

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
```diff
@@ -533,7 +534,7 @@ StatefulSessionPool createKieSessionsPool(String kSessionName, KieSessionConfigu
         return kBase == null ? null : new StatefulSessionPool(kBase, initialSize, () -> {
             SessionConfiguration sessConf = conf != null ? (SessionConfiguration) conf : kBase.getSessionConfiguration();
             StatefulKnowledgeSessionImpl kSession = stateless ?
-                    kBase.internalCreateStatefulKnowledgeSession( env, sessConf, false ).setStateless( true ) :
+                    (StatefulKnowledgeSessionImpl) RuntimeComponentFactory.get().createStatefulSession(kBase, env, sessConf, false).setStateless( true ) :
```
danielezonca: I'm not sure if this cast really is necessary (it is deeply nested inside `StatefulSessionPool` 🤔

drools-core/src/main/java/org/drools/core/common/InternalWorkingMemory.java
```diff
@@ -156,4 +156,6 @@ default FactHandleClassStore getStoreForClass(Class<?> clazz) {
     default boolean isThreadSafe() {
         return getSessionConfiguration().isThreadSafe();
     }
+
+    InternalWorkingMemory setStateless(boolean stateless);
```
danielezonca: This method looks strange to me: as far as I know `InternalWorkingMemory` represents a stateful session so why it should be possible to set `stateless=true`? At the same time why this is the only setter that returns a value?

drools-core/src/main/java/org/drools/core/reteoo/ReteooBuilder.java
```diff
@@ -88,7 +89,7 @@ public ReteooBuilder( final InternalKnowledgeBase  kBase ) {
         this.queries = new HashMap<>();
         this.namedWindows = new HashMap<>();
 
-        this.ruleBuilder = kBase.getConfiguration().getComponentFactory().getRuleBuilderFactory().newRuleBuilder();
+        this.ruleBuilder = new ReteooRuleBuilder();
```
danielezonca, line 92: Why this change? Do you think we don't need to have pluggability (aka alternative rule builder) here?

drools-core/src/main/java/org/drools/core/reteoo/ReteooBuilder.java
```diff
@@ -463,8 +464,7 @@ public void readExternal(ObjectInput in) throws IOException,
 
     public void setRuleBase( InternalKnowledgeBase kBase ) {
         this.kBase = kBase;
-
-        this.ruleBuilder = kBase.getConfiguration().getComponentFactory().getRuleBuilderFactory().newRuleBuilder();
+        this.ruleBuilder = new ReteooRuleBuilder();
```
danielezonca, line 467: Same comment here

drools-core/src/main/java/org/drools/core/reteoo/builder/ReteooRuleBuilder.java
```diff
@@ -112,7 +113,7 @@ public ReteooRuleBuilder() {
         final List<TerminalNode> nodes = new ArrayList<TerminalNode>();
 
         // transform rule and gets the array of subrules
-        final GroupElement[] subrules = rule.getTransformedLhs( kBase.getConfiguration().getComponentFactory().getLogicTransformerFactory().getLogicTransformer(),
+        final GroupElement[] subrules = rule.getTransformedLhs( new DefaultLogicTransformerFactory().getLogicTransformer(),
```
danielezonca, line 116: Similar comment here, no kie.conf/spi approach here?

drools-engine-classic/pom.xml
```diff
@@ -36,6 +36,10 @@
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-kiesession</artifactId>
+    </dependency>
```
danielezonca, line 42: This should not be necessary, `drools-compiler` has a transitive dependency on `drools-kiesession` module. Is this redundant or it is wrong that `drools-compiler` depends on `drools-kiesession`?

drools-engine/pom.xml
```diff
@@ -36,6 +36,10 @@
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-kiesession</artifactId>
+    </dependency>
```
danielezonca, line 42: Same comment here

drools-kiesession/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+  <properties>
+    <java.module.name>org.drools.kiesession</java.module.name>
+    <surefire.forkCount>2</surefire.forkCount>
+  </properties>
+
+  <dependencies>
+    <!--
+      following are required for parser (lang). May be refactored out of
+      core module.
+    -->
```
danielezonca: To be removed?
Copy/paste forever :) 
https://twitter.com/aalmiray/status/1458103592976531457

drools-kiesession/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- Base64 encoding -->
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+    </dependency>
```
danielezonca: Is this used?

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgenda.java
```diff
@@ -87,16 +101,12 @@
...
-    public static final String ON_AFTER_ALL_FIRES_CONSEQUENCE_NAME = "$onAfterAllFire$";
-    public static final String ON_DELETE_MATCH_CONSEQUENCE_NAME = "$onDeleteMatch$";
-
     protected static final transient Logger                      log                = LoggerFactory.getLogger( DefaultAgenda.class );
 
     private static final long                                    serialVersionUID   = 510l;
 
     /** Working memory of this Agenda. */
-    protected InternalWorkingMemory                              workingMemory;
+    protected InternalWorkingMemory workingMemory;
```
danielezonca: Please review class formatting, some fields are aligned while others are not

drools-kiesession/src/main/resources/META-INF/kie/drools-kiesession/kie.conf
```diff
@@ -0,0 +1,2 @@
+org.drools.core.reteoo.RuntimeComponentFactory=org.drools.kiesession.factory.KieComponentFactory
```
danielezonca: Why `KieComponentFactory` and not `RuntimeComponentFactoryImpl` or similar considering that it is the implementation of `RuntimeComponentFactory` interface?

drools-traits/src/main/resources/META-INF/kie/drools-traits/kie.conf
```diff
@@ -1,4 +1,5 @@
-org.drools.core.reteoo.KieComponentFactoryFactory=org.drools.traits.core.reteoo.TraitKieComponentFactoryFactory
 org.drools.compiler.builder.impl.TypeDeclarationBuilderFactory=org.drools.traits.compiler.builder.impl.TraitTypeDeclarationBuilderFactory
 ?org.kie.internal.process.CorrelationKeyFactory = org.jbpm.persistence.correlation.JPACorrelationKeyFactory
+org.drools.core.reteoo.RuntimeComponentFactory=org.drools.traits.core.reteoo.TraitKieComponentFactory;1
```
danielezonca: Why `TraitKieComponentFactory` instead of `TraitRuntimeComponentFactory`?

drools-traits/src/main/java/org/drools/traits/core/definitions/impl/TraitKnowledgePackageImpl.java
```diff
@@ -58,17 +59,21 @@ public TraitKnowledgePackageImpl(String name) {
...
-    public TraitRegistry getTraitRegistry() {
+    public TraitRegistryImpl getTraitRegistry() {
         if (traitRegistry == null) {
             traitRegistry = new TraitRegistryImpl();
         }
         return traitRegistry;
     }
 
+    @Override
+    public void mergeTraitRegistry(InternalKnowledgeBase knowledgeBase) {
```
danielezonca, line 70: As far as I see this method is only invoked [here](https://github.com/kiegroup/drools/blob/36bcc38c82bd032341dccf407eb5c2c7dd7748f6/drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java#L592). 
If so I'm not sure it makes sense to have it declared in `KnowledgePackageImpl` or just move this impl where is invoked 🤔

drools-core/src/main/java/org/drools/core/reteoo/RuntimeComponentFactory.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    static RuntimeComponentFactory get() {
+        return RuntimeComponentFactory.Holder.INSTANCE != null ? RuntimeComponentFactory.Holder.INSTANCE : throwExceptionForMissingRuntime();
+    }
+
+    static <T> T throwExceptionForMissingRuntime() {
+        if (Drools.isNativeImage()) {
+            return null;
+        }
```
danielezonca: Is this correct? I don't think this is an optional component that can be not available at (native) runtime 🤔

drools-core/src/main/java/org/drools/core/reteoo/CoreComponentFactory.java
```diff
@@ -0,0 +1,52 @@
...
+    NodeFactory getNodeFactoryService();
+
+    InternalKnowledgePackage createKnowledgePackage(String name);
+
+    class Holder {
+        private static final CoreComponentFactory INSTANCE = ServiceRegistry.getService( CoreComponentFactory.class );
+    }
+
+    static CoreComponentFactory get() {
+        return CoreComponentFactory.Holder.INSTANCE != null ? CoreComponentFactory.Holder.INSTANCE : DroolsCoreComponentFactory.INSTANCE;
```
danielezonca, line 35: Why this get or else approach? kie.conf already supports priorities so if we want to override the impl it should be enough to tweak kie.conf 🤔

drools-core/src/main/java/org/drools/core/management/StatelessKieSessionMonitoringImpl.java
```diff
@@ -57,7 +71,7 @@ public void reset() {
     public long getTotalSessions() {
         long totalCount = 0;
         for (KieRuntimeEventManager kr : ksessions) {
-            totalCount += ((StatelessKnowledgeSessionImpl) kr).getWorkingMemoryCreatec();
+            totalCount += ((WorkingMemoryCounter) kr).getWorkingMemoryCreated();
```
danielezonca: Is `WorkingMemoryCounter` only here to avoid to use the kieSession?
In general it seems to me that this whole class (`StatelessKieSessionMonitoringImpl`) should be moved to `drools-kiesession` module or as alternative renamed because no kieSession code should remain in `drools-core` (like this `getTotalSessions` method).

drools-core/src/main/java/org/drools/core/management/WorkingMemoryCounter.java
```diff
@@ -14,11 +14,8 @@
...
 
-package org.drools.core.reteoo;
-
-
-public interface RuleBuilderFactory {
-
-    public RuleBuilder newRuleBuilder();
+package org.drools.core.management;
 
+public interface WorkingMemoryCounter {
```
danielezonca: If we move to `drools-kiesession` this interface can be probably removed

drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java
```diff
@@ -342,45 +328,26 @@ KieSession newKieSession(KieSessionConfiguration conf, Environment environment,
...
-    private StatefulKnowledgeSessionImpl internalInitSession( SessionConfiguration sessionConfig, StatefulKnowledgeSessionImpl session ) {
-        if ( sessionConfig.isKeepReference() ) {
-            addStatefulSession(session);
-        }
-        return session;
+    public KieSessionsPool getSessionPool() {
+        return sessionPool;
     }
 
     public Collection<? extends KieSession> getKieSessions() {
```
danielezonca, line 341: All these `*Session*` methods should be removed from this class 🤔 
They are all specific for `KieSession` so we probably need a factory class that accepts `kbase` to produce the session.
I expect this to have a huge impact on the codebase but it is probably necessary if we want to make `KieBase` class generic

drools-core/src/main/java/org/drools/core/common/ObjectStoreWrapper.java
```diff
@@ -0,0 +1,112 @@
...
+package org.drools.core.common;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+
+import org.kie.api.runtime.rule.FactHandle;
+
+public class ObjectStoreWrapper extends AbstractImmutableCollection {
```
danielezonca: Where is it used?

drools-core/src/main/java/org/drools/core/common/AbstractImmutableCollection.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.core.common;
+
+import java.util.Collection;
+
+public abstract class AbstractImmutableCollection
```
danielezonca: As far as I see this is only used by `ObjectStoreWrapper` so we don't probably need also this abstract class 🤔

drools-compiler/src/main/java/org/drools/compiler/rule/builder/RuleBuildContext.java
```diff
@@ -105,7 +105,7 @@ public RuleBuildContext(final DroolsAssemblerContext kBuilder,
             dialect.init(ruleDescr);
         }
 
-        this.compilerFactory = kBuilder.getBuilderConfiguration().getComponentFactory();
+        this.compilerFactory = kBuilder.getBuilderConfiguration().getCompilerComponentFactory();
```
danielezonca: Isn't this factory possible to be loaded via some `*Factory.get()` static method? Is it specific/tied to the specific `kBuilder` instance?

drools-compiler/src/main/java/org/drools/compiler/compiler/BoundIdentifiers.java
```diff
@@ -99,7 +99,7 @@ public void setGlobals( Map<String, Class<?>> globals ) {
         }
 
         if ( cls == null && operators.containsKey( identifier )) {
-            cls = context.getConfiguration().getComponentFactory().getExpressionProcessor().getEvaluatorWrapperClass();
+            cls = context.getConfiguration().getCompilerComponentFactory().getExpressionProcessor().getEvaluatorWrapperClass();
```
danielezonca: Similar comment here, no `CompilerComponentFactory.get()`?

drools-core/src/main/java/org/drools/core/RuleBaseConfiguration.java
mariofusco: Yes, that was totally unused and there for ages. I even have no idea of what it was supposed to do :)

drools-compiler/src/main/java/org/drools/compiler/builder/impl/TypeDeclarationBuilder.java
mariofusco: No, it is still used in tons of places.

image

drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
mariofusco: No, at the moment I don't see how I could remove it.

drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieContainerImpl.java
mariofusco: The `StatefulSessionPool` is only able to specifically pool `StatefulKnowledgeSessionImpl` and this is why the cast is necessary for now. I could relax this constraint and allow pooling any type of session, but this would mean also being obliged to add more methods on the `ReteEvaluator` interface and I want to be conservative on it. Also I'm hoping that the lightweight session would be light enough to not require any pooling, so I'd prefer to be conservative on this and extend the capabilities of the current pool only if strictly required.

drools-core/src/main/java/org/drools/core/common/InternalWorkingMemory.java
mariofusco: I removed this method from the interface where it was actually misleading. Anyway since the stateless session is only a wrapper around the stateful one, that flag is used to say when a stateful is used in this situation, thus allowing some minor optimizations.

drools-core/src/main/java/org/drools/core/reteoo/ReteooBuilder.java
mariofusco: This one of the many overengineered things that we have in our code and that I'm trying to remove. We don't have (and never had) a different implementation so I don't see why we should keep this pluggability.

drools-core/src/main/java/org/drools/core/reteoo/ReteooBuilder.java
mariofusco: See above.

drools-core/src/main/java/org/drools/core/reteoo/builder/ReteooRuleBuilder.java
mariofusco: See above.

drools-engine-classic/pom.xml
mariofusco: That's redundant and I could indeed remove it, but I'd prefer to keep the explicit dependency if you don't see any problem with this.

drools-engine/pom.xml
mariofusco: See above.

drools-kiesession/pom.xml
mariofusco: Fixed.

drools-kiesession/pom.xml
mariofusco: Removed.

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgenda.java
mariofusco: Fixed.

drools-traits/src/main/resources/META-INF/kie/drools-traits/kie.conf
mariofusco: Renamed.

drools-kiesession/src/main/resources/META-INF/kie/drools-kiesession/kie.conf
mariofusco: Renamed.

drools-traits/src/main/java/org/drools/traits/core/definitions/impl/TraitKnowledgePackageImpl.java
mariofusco: Doing so allowed me to remove other methods [here](https://github.com/kiegroup/drools/pull/4035/commits/1433670701f3dbc3d393776078dded4604fd83f9#diff-30df987a825654ab9499b9dac1b0e3f1e773f9edc812df589dc90ea4e8841cfeL846) and [here](https://github.com/kiegroup/drools/pull/4035/commits/1433670701f3dbc3d393776078dded4604fd83f9#diff-92114709b0a59cc960c1c43ab3f57be14b9efc65982d815d206198f15e3bd89aL94).

drools-core/src/main/java/org/drools/core/reteoo/RuntimeComponentFactory.java
mariofusco: No, this was another copy/paste leftover. Removed.

drools-core/src/main/java/org/drools/core/reteoo/CoreComponentFactory.java
mariofusco: We're using the `kie.conf` mostly to load (optional) service implementations from different modules. Since this is not the case and the default implementation is in the same module I preferred to avoid polluting the kie.conf mechanism.

drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java
mariofusco: This is not a factory method and indeed the sessions are already created through a factory. It's a way to keep track of all `KieSession`s created from this `KieBase` and update them accordingly when the `KieBase` is incrementally changed. If we want to keep the possibility of incrementally updating a running `KieSession` (and I think we should) I don't see a better way to do this.

drools-core/src/main/java/org/drools/core/common/AbstractImmutableCollection.java
mariofusco: Removed.

drools-core/src/main/java/org/drools/core/common/ObjectStoreWrapper.java
mariofusco: This is used to return parts of the ObjectStore as an (immutable) collection, for instance [here](https://github.com/kiegroup/drools/blob/main/drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java#L652).

drools-compiler/src/main/java/org/drools/compiler/rule/builder/RuleBuildContext.java
mariofusco: That factory was another overengineering and totally useless, I remove it.

drools-compiler/src/main/java/org/drools/compiler/compiler/BoundIdentifiers.java
mariofusco: See above.

drools-core/src/main/java/org/drools/core/management/WorkingMemoryCounter.java
mariofusco: Removed.

drools-core/src/main/java/org/drools/core/management/StatelessKieSessionMonitoringImpl.java
mariofusco: Done.

drools-engine-classic/pom.xml
danielezonca: Fine for me, the comment was mainly to double check if it is correct that `drools-compiler` also depends on `drools-kiesession` 👍

drools-kiesession/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+      <artifactId>drools-core</artifactId>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>jakarta.inject</groupId>
+      <artifactId>jakarta.inject-api</artifactId>
+      <scope>provided</scope><!-- HACK for OSGi: should be <optional>true</optional> instead -->
+    </dependency>
```
danielezonca: OSGi support has been removed so it should be possible to remove this or at least use `optional` as suggested

drools-core/src/main/java/org/drools/core/impl/KnowledgeBaseImpl.java
mariofusco: This will be managed separately with this ticket https://issues.redhat.com/browse/DROOLS-6720

PR #4038, opened 2021-11-23
[DROOLS-6719] MVELDialect.java::loadImportedClass optimization

HSnicolascerrejon:
When doing a _KieBuilder::buildAll_, mvel dialect tries to import all functions looking for the class in the classpath first, which takes too long, and then in drools, which is pretty faster. I propose to exchange the order in which they are looked for a better performance.

In my tests it's about 33%-50% faster. (in a DRL with 1000 rules and 50 function, _::buildAll_ goes from 10s to 6s)

Can it be applied to _7.62.0.Final_ too, please?

First time pulling, sorry in advance for any inconvenience.

### DIFF
--- a/drools-mvel/src/main/java/org/drools/mvel/builder/MVELDialect.java
+++ b/drools-mvel/src/main/java/org/drools/mvel/builder/MVELDialect.java
@@ -452,10 +452,10 @@ public class MVELDialect
 
     private Class<?> loadImportedClass(String className) {
         try {
-            return pkg.getTypeResolver().resolveType(className);
+            return this.packageRegistry.getPackageClassLoader().loadClass(className);
         } catch (ClassNotFoundException e) { }
         try {
-            return this.packageRegistry.getPackageClassLoader().loadClass(className);
+            return pkg.getTypeResolver().resolveType(className);
         } catch (ClassNotFoundException e) { }
         return null;
     }


tkobayas:
Jenkins retest this

tkobayas:
Hi @HSnicolascerrejon , thank you for the pull request. Could you please file a JIRA (https://issues.redhat.com/projects/DROOLS) for this enhancement so that we can track? Thanks!

HSnicolascerrejon:
Done it! https://issues.redhat.com/browse/DROOLS-6719

tkobayas:
Thanks, @HSnicolascerrejon ! Could you also please edit the followings?

- Edit the commit comment to have the JIRA ID (e.g. `[DROOLS-6719] MVELDialect.java::loadImportedClass optimization`) and push again
- Edit this pull request subject to have the JIRA ID as well (e.g. `[DROOLS-6719] MVELDialect.java::loadImportedClass optimization #4038`)

Thanks!

tkobayas:
@mariofusco , I added you as a reviewer. I did a benchmark for the fix on/off (pkgCLfirst) and confirmed that drl function gets the improvement while java function (which doesn't benefit from the fix) doesn't degrade the performance.

https://github.com/tkobayas/kie-benchmarks/blob/loadImportedClass-7.x/drools-benchmarks/src/main/java/org/drools/benchmarks/buildtime/BuildKJarFromResourceWithFunctionsBenchmark.java

```
Benchmark                                                           (functionType)  (numberOfFunctions)  (numberOfRules)  (pkgCLfirst)  Mode  Cnt    Score    Error  Units
BuildKJarFromResourceWithFunctionsBenchmark.createKJarFromResource             drl                   50             1000          true  avgt   10  343.808 ± 16.717  ms/op
BuildKJarFromResourceWithFunctionsBenchmark.createKJarFromResource             drl                   50             1000         false  avgt   10  374.133 ± 15.429  ms/op
BuildKJarFromResourceWithFunctionsBenchmark.createKJarFromResource            java                   50             1000          true  avgt   10  202.941 ±  7.084  ms/op
BuildKJarFromResourceWithFunctionsBenchmark.createKJarFromResource            java                   50             1000         false  avgt   10  203.813 ±  8.036  ms/op
```

So I think it's good to go with the fix. WDYT?

mariofusco (approved):

PR #4065, opened 2021-12-09
[DROOLS-6744] refactor timer service creation and call it only in str…

mariofusco:
…eam mode

**JIRA**: https://issues.redhat.com/browse/DROOLS-6744

### DIFF
--- a/drools-core/src/main/java/org/drools/core/SessionConfiguration.java
+++ b/drools-core/src/main/java/org/drools/core/SessionConfiguration.java
@@ -21,7 +21,6 @@ import java.util.Map;
 import java.util.Properties;
 
 import org.drools.core.process.instance.WorkItemManagerFactory;
-import org.drools.core.time.TimerService;
 import org.drools.core.time.impl.TimerJobFactoryManager;
 import org.drools.core.util.StringUtils;
 import org.kie.api.KieBase;
@@ -104,8 +103,6 @@ public abstract class SessionConfiguration implements KieSessionConfiguration, E
     public abstract String getSignalManagerFactory();
 
     public abstract ExecutableRunner getRunner( KieBase kbase, Environment environment );
-
-    public abstract TimerService newTimerService();
 
     public abstract QueryListenerOption getQueryListenerOption();
     public abstract void setQueryListenerOption( QueryListenerOption queryListener );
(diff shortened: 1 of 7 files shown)


mariofusco:
@triceo this should fix the native compilation in `optaplanner-quarkus-integration` please double check.

triceo:
@mariofusco Thank you! Any idea why `optaplanner-quickstarts` doesn't have the issue?

mariofusco:
> @mariofusco Thank you! Any idea why `optaplanner-quickstarts` doesn't have the issue?

I didn't check, but probably it brings in some kogito dependency?

tkobayas:
@mariofusco  Regarding failed tests in drools repo are NPE which you will likely fix quickly.

Btw, Drools 8 doesn't allow `timer` attribute in DRL unless STREAM mode? 

https://docs.jboss.org/drools/release/latest/drools-docs/html_single/#drl-timers-calendars-con_drl-rules

We don't have tests with `timer` and CLOUD mode (so tests passed) but anyway, Drools 7 can use `timer` with CLOUD mode (and no restriction is written in drools docs for now).

Regarding failed tests in kogito-runtimes, I see that jbpm depends on TimerService:

https://github.com/kiegroup/kogito-runtimes/blob/main/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java#L89

So we will ask jBPM team to use STREAM mode?

mariofusco:
@tkobayas You're right, my assumption that timers are only used in stream mode was too strong and breaks some tests even in drools. I'll review this and make it lazy regardless of the event processing mode.

tkobayas (approved):
Approved. But probably you still need to fix `MultipleKbasesExampleTest` which I see test failures.
```
[ERROR] Failures: 
[ERROR]   MultipleKbasesExampleTest.testEditSessionModel:118 expected:<class org.drools.kiesession.session.StatefulKnowledgeSessionImpl$DummySessionClock> but was:<class org.drools.core.time.impl.PseudoClockScheduler>
[ERROR]   MultipleKbasesExampleTest.testEditSessionModelAfterFirstCreatedKieSession:137 expected:<class org.drools.kiesession.session.StatefulKnowledgeSessionImpl$DummySessionClock> but was:<class org.drools.core.time.impl.JDKTimerService>
```

PR #4066, opened 2021-12-10
[DROOLS-6746] make drools-core to depend on the static wiring instead…

mariofusco:
… of the dynamic one

**JIRA**: https://issues.redhat.com/browse/DROOLS-6746

Related PRs
https://github.com/kiegroup/drools/pull/4066
https://github.com/kiegroup/kogito-runtimes/pull/1782
https://github.com/kiegroup/kogito-examples/pull/1033

I made `drools-core` to depend on `drools-wiring-static` and fixed the static registry, especially for the missing entries in the dmn area. This should allow a better integration on kogito side and give the possibility to get rid of at least some of the shading modules we created there. 

When necessary (mostly for tests) I brought in the `drools-wiring-dynamic` module. Note that in these cases it is not required to exclude the static one. In fact, when present, the dynamic module simply takes a [higher precedence](https://github.com/kiegroup/drools/blob/main/kie-api/src/main/java/org/kie/api/internal/utils/ServiceRegistry.java#L89) over the static one. Similarly the dependency to `drools-wiring-dynamic` has been added to modules when the dynamic context is evident and unavoidable like `drools-mvel`.

This change will probably require to create user friendly aggregating poms for both pmml and dmn, including the dynamic module in pure drools 8 (no kogito) usage scenarios. I already did the same for both `drools-engine` and `drools-engine-classic`.

### DIFF
--- a/drools-templates/pom.xml
+++ b/drools-templates/pom.xml
@@ -40,10 +40,6 @@
       <artifactId>drools-compiler</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-mvel</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-api</artifactId>
     </dependency>
@@ -54,6 +50,12 @@
     <dependency>
       <groupId>org.mvel</groupId>
       <artifactId>mvel2</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-mvel</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.hsqldb</groupId>
--- a/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/SimpleInstanceCreator.java
+++ b/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/SimpleInstanceCreator.java
@@ -28,9 +28,18 @@ public class SimpleInstanceCreator {
 
     static Object instance(String className) {
         try {
-            return Class.forName(className).newInstance();
+            return Class.forName(className).getConstructor().newInstance();
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
+    }
+
+    static Object tryInstance(String className) {
+        try {
+            return Class.forName(className).getConstructor().newInstance();
+        } catch (Exception e) {
+            // ignore
+        }
+        return null;
     }
 }
--- a/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/SimpleInstanceCreator.java
+++ b/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/SimpleInstanceCreator.java
@@ -16,6 +16,7 @@
 package org.drools.wiring.statics;
 
 import java.lang.reflect.Constructor;
+import java.util.Optional;
 
 public class SimpleInstanceCreator {
     static Constructor<?> constructor(String className) {
@@ -34,12 +35,11 @@ public class SimpleInstanceCreator {
         }
     }
 
-    static Object tryInstance(String className) {
+    static Optional<Object> tryInstance(String className) {
         try {
-            return Class.forName(className).getConstructor().newInstance();
+            return Optional.of( Class.forName(className).getConstructor().newInstance() );
         } catch (Exception e) {
-            // ignore
+            return Optional.empty();
         }
-        return null;
     }
 }
--- a/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticKieAssemblers.java
+++ b/drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticKieAssemblers.java
@@ -28,6 +28,8 @@ import org.kie.api.io.ResourceWithConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.drools.wiring.statics.SimpleInstanceCreator.tryInstance;
+
 public class StaticKieAssemblers implements KieAssemblers {
 
     private static final Logger log = LoggerFactory.getLogger(StaticKieAssemblers.class);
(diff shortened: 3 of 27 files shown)


kie-dmn/kie-dmn-validation/pom.xml
```diff
@@ -113,6 +113,12 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-wiring-dynamic</artifactId>
+      <scope>runtime</scope>
```
tarilabs: sorry but I do NOT concur here.
the Validator rules are produced as exec model, if there was a place I was expecting that
1. -static
2. non runtime scope

to be used, was here.

drools-templates/pom.xml
```diff
@@ -55,6 +51,12 @@
       <groupId>org.mvel</groupId>
       <artifactId>mvel2</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-mvel</artifactId>
+      <scope>test</scope>
```
danielezonca, line 58: Are you sure about this change? Is mvel not necessary at runtime?

drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/SimpleInstanceCreator.java
```diff
@@ -28,9 +28,18 @@
...
         }
     }
+
+    static Object tryInstance(String className) {
+        try {
+            return Class.forName(className).getConstructor().newInstance();
+        } catch (Exception e) {
+            // ignore
+        }
+        return null;
```
danielezonca: Isn't `Optional` a better way to represent this instead of return null?

drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticServiceRegistry.java
```diff
@@ -51,13 +51,14 @@ protected void wireServices() {
         registerService("org.kie.api.io.KieResources", "org.drools.core.io.impl.ResourceFactoryServiceImpl", true);
         registerService("org.kie.api.concurrent.KieExecutors", "org.drools.core.concurrent.ExecutorProviderImpl", true);
```
danielezonca, line 52: Can you create a followup ticket to remove both Static and Dynamic (kie.conf) discovery to use `ServiceLoader` mechanism of Java?
You probably need to tune then also native compilation (see https://github.com/kiegroup/kogito-runtimes/blob/main/quarkus/extensions/kogito-quarkus-predictions-extension/kogito-quarkus-predictions-deployment/src/main/java/org/kie/kogito/quarkus/predictions/deployment/PredictionsAssetsProcessor.java#L62)

drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticKieAssemblers.java
```diff
@@ -28,17 +28,17 @@
...
 
     private Map<ResourceType, KieAssemblerService> assemblers = new HashMap<>();
 
     public StaticKieAssemblers() {
-        // insert here reflective instantiation to assembler services
-        // this is an ad-interim solution
-        // e.g. assemblers.put(ResourceType.DMN, (KieAssemblerService)
-        //         instance("org.kie.dmn.core.assembler.DMNAssemblerService"));
+        assemblers.put(ResourceType.DMN, (KieAssemblerService) tryInstance("org.kie.dmn.core.assembler.DMNAssemblerService"));
+        assemblers.put(ResourceType.PMML, (KieAssemblerService) tryInstance("org.kie.pmml.evaluator.assembler.service.PMMLAssemblerService"));
```
danielezonca: Using ServiceLoader this should not be necessary anymore

drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticServiceRegistry.java
danielezonca: It might be easier, see https://quarkus.io/guides/building-native-image#quarkus-native-pkg-native-config_quarkus.native.auto-service-loader-registration

triceo:
@mariofusco When merged, could you please write up what's the proper way of depending on Drools now? I'm getting a bit lost.

I'm assuming these changes mean that the new default is `drools-engine` + `wiring-dynamic` in OptaPlanner, just `drools-engine` in OptaPlanner Quarkus and `kogito-rules` when I also want DRL.

mariofusco:
@triceo for now I wanted to keep the compatibility of the `drools-engine` module, so it already brings in `drools-wiring-dynamic`. In situations where you want the static behavior please depend on `drools-core`, that already depends on `drools-wiring-static`, plus all other modules required by you application.

drools-templates/pom.xml
mariofusco: I'm doing this to avoid that `drools-decisiontables`, which depends on this, could have a transitive dependency to `drools-wiring-dynamic` thus breaking the native compilation of examples containing DTs. The only alternative that I see is keeping the runtime dependency on `drools-mvel` here and explicitly exclude `drools-wiring-dynamic` in the decisiontables pom. Not sure which of the 2 is preferable.

drools-wiring/drools-wiring-static/src/main/java/org/drools/wiring/statics/StaticServiceRegistry.java
mariofusco: I already had a ticket for that https://issues.redhat.com/browse/DROOLS-6654 
Moved to next sprint.

evacchi (approved):
I haven't followed all the discussion, as far as I am concerned this is ok

danielezonca (approved):

drools-templates/pom.xml
danielezonca: Ok but can this work without dynamic at runtime? If not we should probably split what decision table need (that I assume work with static) and the rest

tarilabs (approved):
https://github.com/kiegroup/drools/pull/4066#pullrequestreview-828991195 was acted

PR #4069, opened 2021-12-13
[DROOLS-6781] keystore file opened ~4 times per request/process

gillyobeast:
targetting 7.x for now as that's what jBPM uses from what I can tell - LMK if I need to target main.

**JIRA**:
[DROOLS-6781](https://issues.redhat.com/browse/DROOLS-6781)

**referenced Pull Requests**:
* https://github.com/kiegroup/droolsjbpm-integration/pull/2664

This patch attempts to solve an issue we've seen with jBPM 7.44.0.Final where a keystore file is kept open many times by a single request/process, taking up FDs on our jBPM servers.

I've raised this Draft PR to start a discussion on the best way to manage this - my approach here is pretty naïve, just making KeyStoreHelper static in PersisterHelper and KieModuleCacheHelper. I also applied a similar change in KeyStoreHelperUtil in https://github.com/kiegroup/droolsjbpm-integration, see https://github.com/kiegroup/droolsjbpm-integration/pull/2664.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-serialization-protobuf/src/main/java/org/drools/serialization/protobuf/PersisterHelper.java
+++ b/drools-serialization-protobuf/src/main/java/org/drools/serialization/protobuf/PersisterHelper.java
@@ -62,6 +62,8 @@ import org.kie.api.marshalling.ObjectMarshallingStrategy.Context;
 
 public class PersisterHelper extends MarshallingHelper {
 
+    private static KeyStoreHelper keyStoreHelper;
+
     public static WorkingMemoryAction readWorkingMemoryAction( MarshallerReaderContext context) throws IOException {
         int type = context.readShort();
         switch ( type ) {
@@ -97,7 +99,7 @@ public class PersisterHelper extends MarshallingHelper {
             case BEHAVIOR_EXPIRE : {
                 return new ProtobufBehaviorExpireWMAction( context, _action );
 
-            }         
+            }
             case SIGNAL : {
                 // need to fix this
             }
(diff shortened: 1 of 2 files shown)


triceo:
ok to test

gillyobeast:
the class that opens the .jceks file is KeyStoreHelper , which does so in its constructor. the issue is that several classes (the targets of my PRs) instantiate this KeyStoreHelper every time they need it, so the file is opened every time it’s needed.

so my approach here is just to make KeyStoreHelper a static member of classes that use it, and to use lazy instantiation. i’ve given it a cursory check and i think it should be thread-safe to do it this way, as KeyStoreHelper’s public methods are free from side effects, but i don’t really have the experience there to say for sure.

there may be a better solution that i haven’t seen but this got us going :slightly_smiling_face:

let me know what you think!

mariofusco (changes requested):
I see what you did there, and caching and reusing the `KeyStoreHelper` makes sense. What I don't understand is why do you need an instance for each class that needs it instead of making the `KeyStoreHelper` itself a singleton and using the same singleton everywhere. Do you think you could change this pull request accordingly? Am I missing something?

gillyobeast:
thanks @mariofusco, that's a fair comment. I simply hadn't noticed that this case was suitable for a singleton, I'll see if I can get the PRs updated shortly. :)

mariofusco:
@gillyobeast I just attempted to make this change against our main branch here https://github.com/kiegroup/drools/pull/4137

gillyobeast:
that'll do for me, as long as it makes it back to 7.x too :) it'll require a change in droolsjbpm-integration/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java too, per my [other PR](https://github.com/kiegroup/droolsjbpm-integration/pull/2664)

mariofusco:
@gillyobeast great thanks. I'll close your PRs then, merge mine when approved and also backport it to 7.x. Thanks again for having noticed and reported this.

PR #4075, opened 2021-12-16
[DROOLS-6707] move tms in its own module

mariofusco:
Related PRs:
https://github.com/kiegroup/drools/pull/4075
https://github.com/kiegroup/kogito-runtimes/pull/1805
https://github.com/kiegroup/optaplanner/pull/1735


drools-compiler/src/main/java/org/drools/compiler/compiler/MissingDependencyError.java
```diff
@@ -0,0 +1,27 @@
+package org.drools.compiler.compiler;
```
danielezonca, line 15: License header

drools-core/src/main/java/org/drools/core/common/MissingDependencyException.java
```diff
@@ -0,0 +1,8 @@
+package org.drools.core.common;
```
danielezonca, line 15: License header

drools-serialization-protobuf/pom.xml
```diff
@@ -33,6 +33,10 @@
             <groupId>org.drools</groupId>
             <artifactId>drools-mvel</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.drools</groupId>
+            <artifactId>drools-tms</artifactId>
```
danielezonca, line 38: Why is TMS necessary for protobuf?

drools-test-coverage/test-integration-notms/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+        <excludedGroups>org.kie.test.testcategory.TurtleTestCategory</excludedGroups>
+        <surefire.forkCount>2</surefire.forkCount>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.drools</groupId>
+            <artifactId>drools-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
danielezonca: I would avoid this "aggregated" dependency here and use the specific dependencies to make sure it is not working only because of some transitive dependency.

drools-test-coverage/test-integration-notms/src/test/java/org/drools/compiler/integrationtests/notms/NoTmsTest.java
```diff
@@ -0,0 +1,52 @@
...
+        this.kieBaseTestConfiguration = kieBaseTestConfiguration;
+    }
+
+    @Parameterized.Parameters(name = "KieBase type={0}")
+    public static Collection<Object[]> getParameters() {
+        return TestParametersUtil.getKieBaseCloudConfigurations(true);
+    }
+
+    @Test
+    public void testUnsupportedTms() {
```
danielezonca, line 49: Please add also a test without insertLogical to see if it works too

drools-test-coverage/test-integration-notms/src/test/java/org/drools/compiler/integrationtests/notms/NoTmsTest.java
```diff
@@ -0,0 +1,52 @@
+package org.drools.compiler.integrationtests.notms;
```
danielezonca, line 15: License Header

drools-engine-classic/pom.xml
```diff
@@ -44,6 +44,10 @@
       <groupId>org.drools</groupId>
       <artifactId>drools-kiesession</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-tms</artifactId>
```
danielezonca, line 49: I would add this dependency to `drools-engine` too: TMS is not deprecated so I don't see a benefit for opt-in with `drools-engine`.
Do we consider this new module like `decisiontable` so something "fewer" users use so it is better to opt-in?

drools-core/src/main/java/org/drools/core/spi/KnowledgeHelper.java
```diff
@@ -72,13 +72,10 @@
      * @param dynamic -
      *            specifies the object implements onPropertyChangeListener
      */
-    InternalFactHandle insert(Object object,
-                boolean dynamic) ;
+    InternalFactHandle insert(Object object, boolean dynamic);
     
     InternalFactHandle insertLogical(Object object) ;
     
-    InternalFactHandle insertLogical(Object object, boolean dynamic) ;
```
danielezonca, line 80: Is `insertLogical` with `dynamic = true` not supported? What about removing it also in `InternalFactHandle insert(Object object, boolean dynamic)`?

drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
```diff
@@ -718,7 +649,7 @@ public void dispose() {
 
     public TruthMaintenanceSystem getTruthMaintenanceSystem() {
```
danielezonca: Is this method necessary? I don't expect `NamedEntryPoint` to expose a similar getter 🤔 
Btw can you review `tms` field lifecycle? It seems initialized only as side effect of this call and I expect it must be in sync with `typeConf.isTMSEnabled()` but I don't know if it is like that

drools-core/src/main/java/org/drools/core/common/LogicalDependency.java
```diff
@@ -21,12 +21,11 @@
 
 public interface LogicalDependency<M extends ModedAssertion<M>> extends LinkedListNode<LogicalDependency<M>> {
```
danielezonca, line 22: Can't this class moved to TMS module? It looks like very strictly related to TMS 🤔

drools-compiler/src/main/java/org/drools/compiler/compiler/MissingDependencyError.java
danielezonca, line 15: License Header

drools-core/src/main/java/org/drools/core/common/TruthMaintenanceSystem.java
```diff
@@ -16,263 +16,60 @@
...
-import static org.drools.core.common.ClassAwareObjectStore.getActualClass;
+public interface TruthMaintenanceSystem {
 
-/**
- * The Truth Maintenance System is responsible for tracking two things. Firstly
- * It maintains a Map to track the classes with the same Equality, using the
- * EqualityKey. The EqualityKey has an internal data structure which references
- * all the handles which are equal. Secondly It maintains another map tracking
- * the justifications for logically asserted facts.
- */
```
danielezonca: Why are you removing the javadoc? It is now on the class while it makes more sense to keep it here

drools-core/src/main/java/org/drools/core/rule/QueryImpl.java
```diff
@@ -77,4 +83,23 @@ public boolean isAbductive() {
...
+        throw new UnsupportedOperationException("Available only for abductive query");
+    }
+
+    public Class<? extends Annotation> getAbductiveAnnotationClass() {
+        throw new UnsupportedOperationException("Available only for abductive query");
+    }
+
+    public <T extends Annotation> Class<?> getAbductionClass(Function<Class<T>, T> annotationReader) {
+        throw new UnsupportedOperationException("Available only for abductive query");
+    }
```
danielezonca: Why are you adding these methods in the base impl? They seem to be related to TMS/Belief only 🤔 
Can you split `QueryBuilder` to have one impl specific for abductive queries?

drools-tms/src/main/java/org/drools/tms/TMSKnowledgeHelper.java
```diff
@@ -0,0 +1,161 @@
+package org.drools.tms;
```
danielezonca: License Header

drools-tms/src/main/java/org/drools/tms/TMSKnowledgeHelper.java
```diff
@@ -0,0 +1,161 @@
...
+import org.drools.core.reteoo.ObjectTypeConf;
+import org.drools.core.reteoo.RuntimeComponentFactory;
+import org.drools.core.rule.EntryPointId;
+import org.drools.core.spi.Activation;
+import org.drools.core.util.LinkedList;
+import org.drools.kiesession.consequence.DefaultKnowledgeHelper;
+import org.drools.tms.beliefsystem.simple.SimpleLogicalDependency;
+import org.kie.api.runtime.rule.Match;
+
+public class TMSKnowledgeHelper<T extends ModedAssertion<T>> extends DefaultKnowledgeHelper {
```
danielezonca: What about using the full name instead of the TMS acronym?

drools-traits/pom.xml
```diff
@@ -81,6 +81,12 @@
       <artifactId>drools-wiring-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-tms</artifactId>
+      <scope>test</scope>
+    </dependency>
+
```
danielezonca, line 89: Is trait using TMS? Or is this necessary only because there is a test with `insertLogical`? 
I think all "addon" module should not have tests that requires another feature so that we can make sure that it works also as standalone component. The final coverage with all the features is provided by `drools-test-coverage`

drools-kiesession/src/main/java/org/drools/kiesession/factory/RuntimeComponentFactoryImpl.java
```diff
@@ -108,8 +100,10 @@ public ClassBuilderFactory getClassBuilderFactory() {
         return ClassBuilderFactory.get();
     }
 
-    public KnowledgeHelper createKnowledgeHelper(ReteEvaluator reteEvaluator) {
-        return new DefaultKnowledgeHelper( reteEvaluator );
+    public final KnowledgeHelper createKnowledgeHelper(ReteEvaluator reteEvaluator) {
+        return TruthMaintenanceSystemFactory.present() ?
+                TruthMaintenanceSystemFactory.get().createKnowledgeHelper( reteEvaluator ) :
+                new DefaultKnowledgeHelper( reteEvaluator );
```
danielezonca: This breaks the abstraction, what about use kie.conf priority to load "a default" or TMS factory and use it?
Or at least make `TruthMaintenanceSystemFactory` implements a generic `KnowledgeHelperFactory`  interface and declare it via kie.conf, then if none impl is loaded, use `DefaultKnowledgeHelper`?

drools-kiesession/src/main/java/org/drools/kiesession/consequence/DefaultKnowledgeHelper.java
```diff
@@ -213,135 +162,33 @@ public InternalFactHandle insert(final Object object, final boolean dynamic) {
...
-                justified.getBlockers().remove( dep.getMode());
-                if (justified.getBlockers().isEmpty() ) {
-                    RuleAgendaItem ruleAgendaItem = justified.getRuleAgendaItem();
-                    toStatefulKnowledgeSession().getAgenda().stageLeftTuple(ruleAgendaItem, justified);
-                }
-                dep = tmp;
-            }
-        }        
-    }
+    public void cancelRemainingPreviousLogicalDependencies() { }
```
danielezonca: Can you please review where this method is used? It should be nice to hide this method inside TMS module

drools-serialization-protobuf/pom.xml
mariofusco: Protobuf explicitly executes tms serialization.

drools-traits/pom.xml
mariofusco: Only necessary for tests using `insertLogical`.

drools-core/src/main/java/org/drools/core/rule/QueryImpl.java
mariofusco: I spllt the `QueryBuilder` and the `AbductiveQueryBuilder`, but cannot move the second to the `drools-tms` module since I don't to make it to depend on `drools-compiler`. Since in the `AbductiveQueryBuilder` I cannot downcast to `AbductiveQuery` those methods need to be declared on the `QueryImpl`.

drools-core/src/main/java/org/drools/core/spi/KnowledgeHelper.java
mariofusco: Users can programattically call an `insert` on an entry-point optionally passing `dynamic = true` with [this semantic](https://github.com/kiegroup/drools/blob/main/drools-core/src/main/java/org/drools/core/WorkingMemoryEntryPoint.java#L54). I suspect that nobody used it (it's not even in our public API) and there is only a test covering that feature. I'm leaving it for now for plain insert, but a since the `insertLogical` can only be performed from a consequence there is no reason to keep it also there.

drools-beliefs/src/main/java/org/drools/beliefs/bayes/BayesBeliefSystem.java
```diff
@@ -15,27 +15,25 @@
 
 package org.drools.beliefs.bayes;
```
danielezonca, line 16: What about create a ticket to rename `drools-beliefs` module? I think it is misleading because this in reality is _bayes beliefs_ and not "only belief"

drools-compiler/src/main/java/org/drools/compiler/compiler/Dialect.java
```diff
@@ -79,7 +80,7 @@
 
     PatternBuilder getPatternBuilder();
 
-    QueryBuilder getQueryBuilder();
+    QueryBuilder getQueryBuilder(QueryImpl query);
```
danielezonca: I find a bit misleading that `QueryBuilder` is not producing a `Query` and actually `QueryImpl` is an input param.

drools-core/src/main/java/org/drools/core/rule/QueryImpl.java
danielezonca: Can you add a comment to explain that this is necessary to avoid to add a dependency between `drools-compiler` and `drools-tms`?

danielezonca (approved):

drools-compiler/src/main/java/org/drools/compiler/compiler/Dialect.java
mariofusco: This builder actually builds the Pattern invoking a query. I renamed the class accordingly.

PR #4102, opened 2022-01-05
[DROOLS-6646] Read/Write accessors cleanup

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6646


drools-core/src/main/java/org/drools/core/base/evaluators/Operator.java
```diff
@@ -25,54 +25,70 @@
...
-    implements
-    Externalizable, org.kie.api.runtime.rule.Operator {
-
-    private static final long                  serialVersionUID = 510l;
+public class Operator implements Externalizable, org.kie.api.runtime.rule.Operator {
 
     // a static private cache so that pluggable operator can register their implementations
     // it is automatically initialized with common operator implementations
-    private static final Map<String, Operator> CACHE            = Collections.synchronizedMap( new HashMap<String, Operator>() );
+    private static final Map<String, Operator> CACHE = Collections.synchronizedMap( new HashMap<String, Operator>() );
```
danielezonca: What it is different to use a `ConcurrentHashMap`?

kie-internal/src/main/java/org/kie/internal/utils/ChainedProperties.java
```diff
@@ -83,6 +83,7 @@ private ChainedProperties(String confFileName, ClassLoader classLoader) {
 
         loadProperties( "META-INF/kie." + confFileName, classLoader, this.props );
         loadProperties( "META-INF/kie.default." + confFileName, classLoader, this.defaultProps);
+        loadProperties( "META-INF/kie.default.mvel." + confFileName, classLoader, this.defaultProps);
```
danielezonca, line 86: How does this work?

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/StrEvaluatorTest.java
```diff
@@ -44,8 +44,7 @@ public StrEvaluatorTest(final KieBaseTestConfiguration kieBaseTestConfiguration)
 
     @Parameterized.Parameters(name = "KieBase type={0}")
     public static Collection<Object[]> getParameters() {
-     // TODO: EM failed with some tests. File JIRAs
-        return TestParametersUtil.getKieBaseCloudConfigurations(false);
+        return TestParametersUtil.getKieBaseCloudConfigurations(true);
```
danielezonca, line 47: Why this change? Is this PR changing some behaviour?

drools-core/src/main/java/org/drools/core/base/evaluators/Operator.java
```diff
@@ -25,54 +25,70 @@
...
-    public static final Operator  SOUNDSLIKE       = Operator.addOperatorToRegistry( "soundslike",
-                                                                                     false );
-    public static final Operator  NOT_SOUNDSLIKE   = Operator.addOperatorToRegistry( "soundslike",
-                                                                                     true );
+    public static final Operator               EQUAL            = addOperatorToRegistry( "==", false );
+    public static final Operator               NOT_EQUAL        = addOperatorToRegistry( "!=", false );
+    public static final Operator               LESS             = addOperatorToRegistry( "<", false );
+    public static final Operator               LESS_OR_EQUAL    = addOperatorToRegistry( "<=", false );
+    public static final Operator               GREATER          = addOperatorToRegistry( ">", false );
+    public static final Operator               GREATER_OR_EQUAL = addOperatorToRegistry( ">=", false );
```
danielezonca: What is the difference between these operators and the `Op` enum?

drools-core/src/main/java/org/drools/core/base/evaluators/Operator.java
mariofusco: Good suggestion. I just changed the formatting without checking it. I'll replace it with a `ConcurrentHashMap`.

kie-internal/src/main/java/org/kie/internal/utils/ChainedProperties.java
mariofusco: I needed another conf file in the `drools-mvel` module since the mvel evaluator have been moved there. I could have used the prefix `kie.default.` also for that one, but I wanted to avoid having the same file with the same name and position in 2 different modules to prevent the problems with uber-jar creation that we had in past.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/StrEvaluatorTest.java
mariofusco: My refactor caused a problem with the `StrEvaluator` and I used this test to debug it. Then I realized that it was working also with the execmodel (nothing related with my refactor I think, I believe it worked even before but we didn't realize it) and then I enabled it.

drools-core/src/main/java/org/drools/core/base/evaluators/Operator.java
mariofusco: To be honest I don't know, I left them there as they were before my refactor, but indeed it would make sense to unify all the operators including them. I will try to also move them into the enum.

danielezonca (approved):
LGTM 👍

PR #4116, opened 2022-01-12
[DROOLS-6646] move ClassFieldAccessorStore out of drools-core

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6646

Related PRs
https://github.com/kiegroup/drools/pull/4116
https://github.com/kiegroup/kogito-runtimes/pull/1862


drools-traits/src/main/resources/META-INF/kie/drools-traits/kie.conf
```diff
@@ -1,5 +1,5 @@
 org.drools.compiler.builder.impl.TypeDeclarationBuilderFactory=org.drools.traits.compiler.builder.impl.TraitTypeDeclarationBuilderFactory
 ?org.kie.internal.process.CorrelationKeyFactory = org.jbpm.persistence.correlation.JPACorrelationKeyFactory
 org.drools.core.reteoo.RuntimeComponentFactory=org.drools.traits.core.reteoo.TraitRuntimeComponentFactoryImpl;1
-org.drools.core.reteoo.CoreComponentFactory=org.drools.traits.core.reteoo.TraitCoreComponentFactory
+org.drools.core.reteoo.CoreComponentFactory=org.drools.traits.core.reteoo.TraitCoreComponentFactory;1
```
danielezonca, line 4: Why this change? Do we have a new CoreComponentFactory impl?

drools-core/src/main/java/org/drools/core/base/AccessorKey.java
```diff
@@ -109,4 +108,32 @@ public String toString() {
...
+     * Utility method to take a string and convert it to normal Java variable
+     * name capitalization.  This normally means converting the first
+     * character from upper case to lower case, but in the (unusual) special
+     * case when there is more than one character and both the first and
+     * second characters are upper case, we leave it alone.
+     * <p>
+     * Thus "FooBah" becomes "fooBah" and "X" becomes "x", but "URL" stays
+     * as "URL".
+     *
+     * Taken from
```
danielezonca: Taken from what?

drools-core/src/main/java/org/drools/core/base/AccessorKey.java
```diff
@@ -109,4 +108,32 @@ public String toString() {
...
+     *
+     * @param  name The string to be decapitalized.
+     * @return  The decapitalized version of the string.
+     */
+    public static String decapitalizeFieldName(String name) {
```
danielezonca: I'm not sure here is a good fit for this method, what about [`StringUtils`](https://github.com/kiegroup/drools/blob/main/drools-core/src/main/java/org/drools/core/util/StringUtils.java)?

drools-core/src/main/java/org/drools/core/definitions/InternalKnowledgePackage.java
```diff
@@ -144,15 +148,30 @@
...
+
+    default Class<?> getFieldType(Class<?> clazz, String leftValue) { return null; }
+
+    default InternalReadAccessor getReader(String className, String fieldName, AcceptsReadAccessor target) { return null; }
+
+    default Collection<KnowledgeBuilderResult> getWiringResults(Class<?> classType, String fieldName) { return Collections.emptyList(); }
+
+    default InternalReadAccessor getFieldExtractor( TypeDeclaration type, String timestampField, Class<?> returnType ) { return null; }
+
+    default void setClassFieldAccessorCache(ClassFieldAccessorCache classFieldAccessorCache) { }
```
danielezonca, line 176: Are all "empty" methods?
Are they used only by MVEL?

drools-traits/src/main/resources/META-INF/kie/drools-traits/kie.conf
mariofusco: Now there is also a `MVELCoreComponentFactory`. If there are only mvel it should use that one, if there are mvel and traits it should use this and that's why it has higher precedence.

drools-core/src/main/java/org/drools/core/definitions/InternalKnowledgePackage.java
mariofusco: Yes.

PR #4130, opened 2022-01-19
[DROOLS-6783] fix SegmentPropagator for update staged tuples

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6783

PR #4134, opened 2022-01-19
[DROOLS-6783] fix SegmentPropagator for update staged tuples (#4130)

mariofusco:


mariofusco:
The failure in `org.drools.compiler.integrationtests.TimerAndCalendarWithRealTimeTest.testIntervalRuleInsertion` is totally unrelated with this commit. That's a well known flaky test and we're working to improve it.

PR #4143, opened 2022-01-21
[DROOLS-6785] merge package descrs coming from different resources du…

mariofusco:
…ring executable model generation

**JIRA**: https://issues.redhat.com/browse/DROOLS-6785

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


tkobayas (approved):
@mariofusco  I confirmed that the fix solves the issue of the reproducer in the JIRA.

You didn't add a unit test because it requires jbpm dependency and jbpm repo doesn't have a test base class for exec-model, right? We may need to think about:

* For 7.x, parameterize jbpm tests which involve rules to be able to test exec-model
* For 8.x/kogito, such tests exist under `kogito-runtimes/jbpm` and `kogito-runtimes/integration-tests`. We may need to review the test coverage (e.g. drools expression on gateway).

This could be considered as a JIRA separated from this PR. WDYT?

mariofusco:
> You didn't add a unit test because it requires jbpm dependency and jbpm repo doesn't have a test base class for exec-model, right?

Correct.

>     * For 7.x, parameterize jbpm tests which involve rules to be able to test exec-model

+1

>     * For 8.x/kogito, such tests exist under `kogito-runtimes/jbpm` and `kogito-runtimes/integration-tests`. We may need to review the test coverage (e.g. drools expression on gateway).

That's trickier. All jbpm tests in kogito have to be reviewed because the biggest part of them use the legacy integration with drools (via KieSession) that we don't want to support anymore there.

> This could be considered as a JIRA separated from this PR. WDYT?

I'd open a JIRA only for the 7.x part.

tkobayas:
Okay, created https://issues.redhat.com/browse/JBPM-9996

PR #4151, opened 2022-01-25
[DROOLS-6797] UnsupportedOperationException when different package ru…

tkobayas:
…les from DRL and RF

**Ports** 
This PR is for 7.x. 

for main -> https://github.com/kiegroup/drools/pull/4155

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-6797

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

* for a <b>full downstream build</b> 
  * for <b>jenkins</b> job: please add comment: <b>Jenkins run fdb</b>
  * for <b>github actions</b> job: add the label `run_fdb`

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
@@ -18,6 +18,7 @@ package org.drools.modelcompiler.builder;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -135,7 +136,12 @@ public class ModelBuilderImpl<T extends PackageSources> extends KnowledgeBuilder
                     if (optPkg.isPresent()) {
                         optPkg.get().addPackageDescr(entry.getValue().getResource(), entry.getValue());
                     } else {
-                        compositePackages.add(entry.getValue());
+                        try {
+                            compositePackages.add(entry.getValue());
+                        } catch (UnsupportedOperationException uoe) {
+                            compositePackages = new HashSet<>(compositePackages);
+                            compositePackages.add(entry.getValue());
+                        }
                     }
                 }
             }
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
@@ -131,17 +131,13 @@ public class ModelBuilderImpl<T extends PackageSources> extends KnowledgeBuilder
     private Collection<CompositePackageDescr> findPackages( Collection<CompositePackageDescr> compositePackages ) {
         if (compositePackages != null && !compositePackages.isEmpty()) {
             if (compositePackagesMap != null) {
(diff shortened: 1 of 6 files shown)


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
```diff
@@ -135,7 +136,12 @@ protected void processOtherDeclarations(PackageRegistry pkgRegistry, PackageDesc
                     if (optPkg.isPresent()) {
                         optPkg.get().addPackageDescr(entry.getValue().getResource(), entry.getValue());
                     } else {
-                        compositePackages.add(entry.getValue());
+                        try {
+                            compositePackages.add(entry.getValue());
+                        } catch (UnsupportedOperationException uoe) {
+                            compositePackages = new HashSet<>(compositePackages);
+                            compositePackages.add(entry.getValue());
+                        }
```
tkobayas: Not sure if this is an elegant solution.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/PackageInMultipleResourcesTest.java
```diff
@@ -0,0 +1,86 @@
...
+        this.kieBaseTestConfiguration = kieBaseTestConfiguration;
+    }
+
+    @Parameterized.Parameters(name = "KieBase type={0}")
+    public static Collection<Object[]> getParameters() {
+        return TestParametersUtil.getKieBaseCloudConfigurations(true);
+    }
+
+    @Test
+    public void testSamePackageRulesInDRLAndRF() {
```
tkobayas, line 52: These 2 tests don't execute ruleflow actually. But I think they are enough to reproduce the issues in DROOLS-6785 and DROOLS-6797

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
mariofusco: I know this may be wasteful in some cases, but I'd prefer to always create a mutable collection in that situation instead of using that try/catch block, something like:

```
    private Collection<CompositePackageDescr> findPackages( Collection<CompositePackageDescr> compositePackages ) {
        if (compositePackages != null && !compositePackages.isEmpty()) {
            if (compositePackagesMap != null) {
                compositePackages = new HashSet<>(compositePackages);
                for (Map.Entry<String, CompositePackageDescr> entry : compositePackagesMap.entrySet()) {
                    Optional<CompositePackageDescr> optPkg = compositePackages.stream().filter(pkg -> pkg.getNamespace().equals(entry.getKey()) ).findFirst();
                    if (optPkg.isPresent()) {
                        optPkg.get().addPackageDescr(entry.getValue().getResource(), entry.getValue());
                    } else {
                        compositePackages.add(entry.getValue());
                    }
                }
            }
            return compositePackages;
        }
        if (compositePackagesMap != null) {
            return compositePackagesMap.values();
        }
        return emptyList();
    }
```

mariofusco (approved):

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/ModelBuilderImpl.java
tkobayas: Sure, updated

PR #4224, opened 2022-02-21
[DROOLS-6841] fix jitting of varargs invocations

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6841

### DIFF
--- a/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
+++ b/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
@@ -172,7 +172,7 @@ public class CoercedExpression {
                 || Number.class.isAssignableFrom(rightClass)
                 || Boolean.class == rightClass
                 || String.class == rightClass
-                || (Object.class == rightClass && equalityExpr)
+                || Object.class == rightClass
                 || (Map.class.isAssignableFrom(leftClass) || Map.class.isAssignableFrom(rightClass));
     }
 
--- a/drools-mvel/src/main/java/org/drools/mvel/ConditionAnalyzer.java
+++ b/drools-mvel/src/main/java/org/drools/mvel/ConditionAnalyzer.java
@@ -538,14 +538,17 @@ public class ConditionAnalyzer {
             return;
         }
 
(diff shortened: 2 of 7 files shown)


drools-mvel/src/main/java/org/drools/mvel/ConditionAnalyzer.java
```diff
@@ -538,14 +538,17 @@ private void readInvocationParams(Invocation invocation, ExecutableStatement[] p
             return;
         }
 
+        // if it's a vararg, but the last argument is already an array there's no need for any special treatment
+        isVarArgs &= !(params.length == paramTypes.length && params[paramTypes.length-1].getKnownEgressType().isArray());
```
mariofusco, line 542: @tkobayas The fix for the jitting problem is just this line.

drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/drlxparse/CoercedExpression.java
```diff
@@ -172,7 +172,7 @@ private boolean canCoerce() {
                 || Number.class.isAssignableFrom(rightClass)
                 || Boolean.class == rightClass
                 || String.class == rightClass
-                || (Object.class == rightClass && equalityExpr)
+                || Object.class == rightClass
```
mariofusco, line 175: @tkobayas I changed this line to also make all other jitting tests to pass even with the executable model ( see https://issues.redhat.com/browse/DROOLS-6152 ) and enable those tests also for it. This change (that however is not related with the fix of the main issue) made a few tests to fail. I believe that the tests are wrong and show an inconsistent behaviour of the drl: sometimes it allows comparison between numeric types and raw objects and sometimes it doesn't. The executable model always allow that type of coercion, so this change just enables it. It could be ideally possible to also allow this in plain drl (thus making it more consistent) but this would require quite a big change in mvel and I don't think it is a good idea to attempt it now. At least this change fixes the inconsistency on the executable model side, which is the part that at the moment is more relevant to us. I had to remove the tests relying on this inconsistency.

tkobayas: Agreed. Thanks!

tkobayas (approved):

PR #4235, opened 2022-03-01
normalize path in predicate information of externelized lambda

mariofusco:


triceo (approved):
I have no way to test this. It requires a bump in Kogito, and then a test execution of OP on Windows which I do not have on my laptop.
That said, it does look like it fixes the issue.

PR #4305, opened 2022-04-11
Revert "[7.x] Fixed potential XXE vulnerabilities"

yesamer:
Reverts kiegroup/drools#4283

The PR introduced 6 tests failures in https://github.com/kiegroup/droolsjbpm-integration tests: 

eg. 
org.kie.server.integrationtests.scenariosimulation.ScenarioSimulationIntegrationTest.executeScenarioTest[JAXB KieServicesConfiguration{transport=REST, serverUrl='http://localhost:40477/kie-server-services/services/rest/server'}]

`org.kie.server.api.exception.KieServicesHttpException: 
Unexpected HTTP response code when requesting URI 'http://localhost:40477/kie-server-services/services/rest/server/containers/input-data-string/scesim'! Error code: 400, message: <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<response type="FAILURE" msg="Test scenario parsing error: Not supported: [http://javax.xml.XMLConstants/property/accessExternalDTD"/](http://javax.xml.xmlconstants/property/accessExternalDTD)>`

As the same patch has been applied in 8.x stream, I suggest to double check if Test Scenario works correctly in main too.

### DIFF
--- a/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/DOMParserUtil.java
+++ b/drools-scenario-simulation/drools-scenario-simulation-backend/src/main/java/org/drools/scenariosimulation/backend/util/DOMParserUtil.java
@@ -352,8 +352,6 @@ public class DOMParserUtil {
     public static String getString(Document toRead) throws TransformerException {
         DOMSource domSource = new DOMSource(toRead);
         TransformerFactory factory = TransformerFactory.newInstance();
-        factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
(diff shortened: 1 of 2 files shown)


mariofusco (approved):

mareknovotny:
we should not revert it blindly. this options are important to have there due XXE. What do we do to get it back and fix or rework the tests better? @yesamer @mariofusco

yesamer:
@mareknovotny Unfortunately that change results in a blocker for Test Scenario: it's not possible to unmarshal the XML file, so it totally brakes the whole functionality. I know @pibizza is trying to find an alternative way to remove the vulnerability. Let's see.

PR #4344, opened 2022-04-27
[DROOLS-6922] create a DSL to programmatically define rules in a RuleUnit

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6922


mariofusco:
This was just a failed experiment. Lesson learned: it's hard to have a DSL that is significantly easier than the executable model one without giving up in expressiveness and features. It's also impossible to define a rule in a completely fluent statement because of variable binding, that cannot be implicit as in optaplanner's DSL. The only thing that seems achievable is replacing a constraint currently expressed as a general lambda predicate with something in the form <leftValue, operator, rightValue> thus also implicitly defining also the indexing for that constraint, but I doubt that this simplification would be enough for users to allow them to programmatically use this DSL.

mariofusco:
A second, more successful, attempt has been implemented with this pull request https://github.com/kiegroup/drools/pull/4416

PR #4345, opened 2022-04-28
Add length check back to EvaluatorWrapper#getFactHandle

gitorres:
Add length check removed in 54c3bf1a0b1297b3d1594cf3229d9e6218dd1ebc to avoid `ArrayIndexOutOfBoundsException` exception.

**JIRA**: https://issues.redhat.com/browse/DROOLS-6925.


mariofusco (commented):
This pull request cannot be accepted since it is meant to fix the effect of the problem and not it's root cause, i.e. figuring out why `declaration.getObjectIndex()` contains the wrong value. Also we cannot bring in a blind fix without a test case reproducing the bug.

PR #4348, opened 2022-04-29
[JBPM-10071] WIH's abortWorkItem method should be called when aborting workItem through kie-server REST

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10071

### DIFF
--- a/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
+++ b/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
@@ -195,8 +195,19 @@ public class JPAWorkItemManager implements WorkItemManager {
         // work item may have been aborted
         if ( workItemInfo != null ) {
             WorkItem workItem = (WorkItemImpl) internalGetWorkItem( workItemInfo );
-            ProcessInstance processInstance = kruntime.getProcessInstance( workItem.getProcessInstanceId() );
+            WorkItemHandler handler = (WorkItemHandler) this.workItemHandlers.get( workItem.getName() );
+            // avoid multiple calls // recursion
+            if (workItem == null || workItem.getState() == WorkItem.ABORTED) {
+                return;
+            }
             workItem.setState( WorkItem.ABORTED );
+
+            handler = (WorkItemHandler) this.workItemHandlers.get( workItem.getName() );
+            if ( handler != null) {
+                handler.abortWorkItem( workItem, this );
+            }
+
+            ProcessInstance processInstance = kruntime.getProcessInstance( workItem.getProcessInstanceId() );
             // process instance may have finished already
             if ( processInstance != null ) {
                 processInstance.signalEvent( "workItemAborted", workItem );


elguardian:
Jenkins run fdb

gmunozfe:
Jenkins run fdb

gmunozfe (changes requested):
An infinite loop is created if the implementation invokes the manager.abortWorkItem, like this:

https://github.com/kiegroup/droolsjbpm-integration/blob/751cc6edcf3421f6454dfdae0a1a9cea96002286/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/filtered-resources/kjars-sources/async-exec-project-101/src/main/java/com/self/exceptionhandbpm/CustomWorkItemHandler.java#L36

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
```diff
@@ -195,6 +195,10 @@ public void abortWorkItem( long id ) {
         // work item may have been aborted
         if ( workItemInfo != null ) {
             WorkItem workItem = (WorkItemImpl) internalGetWorkItem( workItemInfo );
+            WorkItemHandler handler = (WorkItemHandler) this.workItemHandlers.get( workItem.getName() );
+            if ( handler != null ) {
+                handler.abortWorkItem( workItem, this );
+            }
```
gmunozfe: You could set the State to ABORTED before invoking `handler.abortWorkItem`, and also, control first if the state is not ABORTED. I've tested and worked like this:

```suggestion
            if (workItem!=null && workItem.getState() == WorkItem.ABORTED) {
            	return;
            }
            workItem.setState( WorkItem.ABORTED );
           
            WorkItemHandler handler = (WorkItemHandler) this.workItemHandlers.get( workItem.getName() );
	    if ( handler != null) {
	        handler.abortWorkItem( workItem, this );
            }
```

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
```diff
@@ -195,6 +195,10 @@ public void abortWorkItem( long id ) {
...
             ProcessInstance processInstance = kruntime.getProcessInstance( workItem.getProcessInstanceId() );
             workItem.setState( WorkItem.ABORTED );
```
gmunozfe: Move this line above

elguardian: done

drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/processinstance/JPAWorkItemManager.java
elguardian: done

gmunozfe (approved):
Looks good to me, outstanding work @elguardian !
I've also added a PR with a test: https://github.com/kiegroup/droolsjbpm-integration/pull/2829

PR #4441, opened 2022-06-07
DROOLS-6971 Stand-Alone DRL Canonical Model Compiler

evacchi:
Currently there is no clear entry point from file (DRL) to output (.java/.class). Create one such "driver" class with its own tests. Follow up: test cases to load the generated rule base.


drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/tool/ExplicitCanonicalModelCompiler.java
```diff
@@ -0,0 +1,156 @@
...
+ * An alternative compilation flow that generates code starting
+ * from a collection of {@link CompositePackageDescr}s,
+ * skipping the {@link org.kie.internal.builder.KnowledgeBuilder} entirely.
+ *
+ * It explicitly invokes the necessary {@link CompilationPhase}s
+ * one after the other and collects the error.
+ *
+ * @param <T> a subclass of PackageSources to retrieve the generated results
+ */
+public class ExplicitCanonicalModelCompiler<T extends PackageSources> {
```
evacchi, line 52: we should think of a better name...

mariofusco (approved):

PR #4442, opened 2022-06-07
[DROOLS-7007] avoid memory leak in groupby

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7007

@triceo please give a try to this fix. I reproduced the problem on my side also with a pure Drools test, but the fix is simply removing the useless data structure that was causing the leak, so now that test is meaningless simply because I cannot make any assertion against a no longer existing thing :)


triceo (approved):
OP tests look good locally.

PR #4474, opened 2022-06-20
DROOLS-6972 Refactor immutable compilation phases

evacchi:
Compilation Phases that have been refactored from the KnowledgeBuilder and ModelBuilder are still in some cases referring to a kBase and assetFilters that in immutable use cases are always null.

We refactor these phases into "simple" or "immutable" phases, that do no take any filter or kbase as a parameter. The original phases will extend such "basic" phases.

The new basic phases can be instantiated directly in all those case when a kbase and/or an asset filter is irrelevant

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ *
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.drools.compiler.builder.impl.processors;
 
 import org.drools.compiler.builder.impl.AssetFilter;
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
@@ -2,71 +2,36 @@ package org.drools.compiler.builder.impl.processors;
 
 import org.drools.compiler.builder.impl.AssetFilter;
 import org.drools.compiler.builder.impl.GlobalVariableContext;
-import org.drools.compiler.compiler.GlobalError;
 import org.drools.compiler.compiler.PackageRegistry;
 import org.drools.core.definitions.InternalKnowledgePackage;
-import org.drools.drl.ast.descr.GlobalDescr;
 import org.drools.drl.ast.descr.PackageDescr;
 import org.drools.kiesession.rulebase.InternalKnowledgeBase;
 import org.kie.internal.builder.ResourceChange;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
-import java.util.Set;
-
-public class GlobalCompilationPhase extends AbstractPackageCompilationPhase {
-    protected static final transient Logger logger = LoggerFactory.getLogger(GlobalCompilationPhase.class);
+public class GlobalCompilationPhase extends ImmutableGlobalCompilationPhase {
 
     private final InternalKnowledgeBase kBase;
-    private final GlobalVariableContext globalVariableContext;
     private final AssetFilter assetFilter;
 
     public GlobalCompilationPhase(PackageRegistry pkgRegistry, PackageDescr packageDescr, InternalKnowledgeBase kBase, GlobalVariableContext globalVariableContext, AssetFilter filterAcceptsRemoval) {
-        super(pkgRegistry, packageDescr);
+        super(pkgRegistry, packageDescr, globalVariableContext);
         this.kBase = kBase;
-        this.globalVariableContext = globalVariableContext;
         this.assetFilter = filterAcceptsRemoval;
     }
 
-    public void process() {
-        InternalKnowledgePackage pkg = pkgRegistry.getPackage();
-        Set<String> existingGlobals = new HashSet<>(pkg.getGlobals().keySet());
-
-        for (final GlobalDescr global : packageDescr.getGlobals()) {
-            final String identifier = global.getIdentifier();
-            existingGlobals.remove(identifier);
-            String className = global.getType();
-
-            // JBRULES-3039: can't handle type name with generic params
-            while (className.indexOf('<') >= 0) {
-                className = className.replaceAll("<[^<>]+?>", "");
-            }
-
-            try {
-                Class<?> clazz = pkgRegistry.getTypeResolver().resolveType(className);
-                if (clazz.isPrimitive()) {
-                    this.results.add(new GlobalError(global, " Primitive types are not allowed in globals : " + className));
-                    return;
-                }
-                pkg.addGlobal(identifier, clazz);
-                globalVariableContext.addGlobal(identifier, clazz);
-
-                if (kBase != null) {
-                    kBase.addGlobal(identifier, clazz);
-                }
-            } catch (final ClassNotFoundException e) {
-                this.results.add(new GlobalError(global, e.getMessage()));
-                logger.warn("ClassNotFoundException occured!", e);
-            }
+    @Override
+    protected void addGlobal(InternalKnowledgePackage pkg, String identifier, Class<?> clazz) {
+        super.addGlobal(pkg, identifier, clazz);
+        if (kBase != null) {
+            kBase.addGlobal(identifier, clazz);
         }
+    }
 
-        for (String toBeRemoved : existingGlobals) {
-            if (assetFilter != null && AssetFilter.Action.REMOVE.equals(assetFilter.accept(ResourceChange.Type.GLOBAL, pkg.getName(), toBeRemoved))) {
-                pkg.removeGlobal(toBeRemoved);
-                if (kBase != null) {
-                    kBase.removeGlobal(toBeRemoved);
-                }
+    protected void removeGlobal(InternalKnowledgePackage pkg, String toBeRemoved) {
+        if (assetFilter != null && AssetFilter.Action.REMOVE.equals(assetFilter.accept(ResourceChange.Type.GLOBAL, pkg.getName(), toBeRemoved))) {
+            pkg.removeGlobal(toBeRemoved);
+            if (kBase != null) {
+                kBase.removeGlobal(toBeRemoved);
             }
         }
     }
--- a/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
@@ -1,0 +1,67 @@
+package org.drools.compiler.builder.impl.processors;
+
+import org.drools.compiler.builder.impl.GlobalVariableContext;
+import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
+import org.drools.compiler.compiler.GlobalError;
+import org.drools.compiler.compiler.PackageRegistry;
+import org.drools.core.definitions.InternalKnowledgePackage;
+import org.drools.drl.ast.descr.GlobalDescr;
+import org.drools.drl.ast.descr.PackageDescr;
+import org.kie.internal.builder.ResourceChange;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.Set;
+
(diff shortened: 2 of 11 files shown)


drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/phases/ExplicitCompilerTest.java
```diff
@@ -119,14 +118,14 @@ public void testCompile() throws DroolsParserException, IOException {
...
                 /*         packageRegistry.setDialect(getPackageDialect(packageDescr)) */
                 new RuleValidator(packageRegistry, packageDescr, configuration),
-                new FunctionCompiler(packageDescr, packageRegistry, null, rootClassLoader),
-                new RuleCompiler(packageRegistry, packageDescr, kBase, parallelRulesBuildThreshold,
-                        null, attributesForPackage, resource, typeDeclarationContext),
-                new ReteCompiler(packageRegistry, packageDescr, kBase, null),
+                new SimpleFunctionCompiler(packageRegistry, packageDescr, rootClassLoader),
+                new ImmutableRuleCompilationPhase(packageRegistry, packageDescr, parallelRulesBuildThreshold,
+                        attributesForPackage, resource, typeDeclarationContext),
+//                new ReteCompiler(packageRegistry, packageDescr, kBase, null), // no-op when kbase==null
```
evacchi, line 128: I am leaving this commented out code for clarity

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
```diff
@@ -2,71 +2,36 @@
...
-
-                if (kBase != null) {
-                    kBase.addGlobal(identifier, clazz);
-                }
-            } catch (final ClassNotFoundException e) {
-                this.results.add(new GlobalError(global, e.getMessage()));
-                logger.warn("ClassNotFoundException occured!", e);
-            }
+    @Override
+    protected void addGlobal(InternalKnowledgePackage pkg, String identifier, Class<?> clazz) {
```
evacchi, line 40: moved to overridable method (see super)

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
```diff
@@ -2,71 +2,36 @@
...
         }
+    }
 
-        for (String toBeRemoved : existingGlobals) {
-            if (assetFilter != null && AssetFilter.Action.REMOVE.equals(assetFilter.accept(ResourceChange.Type.GLOBAL, pkg.getName(), toBeRemoved))) {
-                pkg.removeGlobal(toBeRemoved);
-                if (kBase != null) {
-                    kBase.removeGlobal(toBeRemoved);
-                }
+    protected void removeGlobal(InternalKnowledgePackage pkg, String toBeRemoved) {
```
evacchi, line 47: same

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
```diff
@@ -0,0 +1,67 @@
...
+                logger.warn("ClassNotFoundException occured!", e);
+            }
+        }
+
+        for (String toBeRemoved : existingGlobals) {
+            removeGlobal(pkg, toBeRemoved);
+        }
+    }
+
+    protected void addGlobal(InternalKnowledgePackage pkg, String identifier, Class<?> clazz) {
```
evacchi, line 76: overridable by subclass

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
```diff
@@ -0,0 +1,67 @@
...
+        pkg.addGlobal(identifier, clazz);
+        globalVariableContext.addGlobal(identifier, clazz);
+    }
+
+    protected void removeGlobal(InternalKnowledgePackage pkg, String toBeRemoved) {
```
evacchi, line 81: same

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableRuleCompilationPhase.java
```diff
@@ -36,41 +32,33 @@
...
     private int parallelRulesBuildThreshold;
-    private final AssetFilter assetFilter;
 
     //This list of package level attributes is initialised with the PackageDescr's attributes added to the assembler.
     //The package level attributes are inherited by individual rules not containing explicit overriding parameters.
     //The map contains a map of AttributeDescr's keyed on the AttributeDescr's name.
     private final Map<String, AttributeDescr> packageAttributes;
     private final Resource resource;
-    private final TypeDeclarationContext kBuilder;
+    private final TypeDeclarationContext typeDeclarationContext;
```
evacchi, line 59: renamed for clarity

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableRuleCompilationPhase.java
```diff
@@ -82,72 +70,10 @@ public void process() {
...
-                    kBase.removeRules(rulesToBeRemoved);
-                }
-            });
-        }
-    }
-
-    private boolean filterAccepts(ResourceChange.Type type, String namespace, String name) {
-        return assetFilter == null || !AssetFilter.Action.DO_NOTHING.equals(assetFilter.accept(type, namespace, name));
+    protected boolean filterAccepts(ResourceChange.Type type, String namespace, String name) {
+        return true;
```
evacchi, line 89: overridable by the subclass; this is still not beautiful because it is kind of ad-hoc for this case, but we can improve it over time

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
```diff
@@ -1,72 +1,54 @@
...
-                    kBase.addGlobal(identifier, clazz);
-                }
-            } catch (final ClassNotFoundException e) {
-                this.results.add(new GlobalError(global, e.getMessage()));
-                logger.warn("ClassNotFoundException occured!", e);
-            }
+    @Override
+    protected void addGlobal(InternalKnowledgePackage pkg, String identifier, Class<?> clazz) {
+        super.addGlobal(pkg, identifier, clazz);
+        if (kBase != null) {
```
mariofusco, line 42: Is it safe to assume that for a mutable phase the `kbase` is always present?

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableRuleCompilationPhase.java
```diff
@@ -1,21 +1,17 @@
 package org.drools.compiler.builder.impl.processors;
```
mariofusco, line 16: Missing header?

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/SimpleFunctionCompiler.java
```diff
@@ -0,0 +1,111 @@
...
+import org.drools.drl.ast.descr.FunctionDescr;
+import org.drools.drl.ast.descr.PackageDescr;
+import org.drools.wiring.api.classloader.ProjectClassLoader;
+
+import java.util.List;
+
+import static org.drools.core.impl.KnowledgeBaseImpl.registerFunctionClassAndInnerClasses;
+import static org.drools.util.StringUtils.isEmpty;
+
+public class SimpleFunctionCompiler extends AbstractPackageCompilationPhase {
```
mariofusco: Why `Simple` instead of `Immutable`?

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
evacchi: nice question 🤔  currently this is a drop-in replacement for the "original" phase, but instead we could add a factory that returns the right one if a kbase is present (?) 🤔 /cc @pibizza a factory here may make sense 🤔

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/SimpleFunctionCompiler.java
evacchi: the answer is very silly: because I was not sure this was really immutable :D i.e. "mutable" versions usually deal with a kbase, in this case it's all about assetfilter; hence SimpleFunctionCompiler+FunctionCompiler instead of ImmutableFunctionCompiler+FunctionCompiler. I admit this is confusing. Maybe the name "Immutable" is wrong... and I don't like Simple either..

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
evacchi: in other words: because of how it is currently used, no. But it's a good observation

mariofusco (approved):

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
evacchi: created jira https://issues.redhat.com/browse/DROOLS-7045

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/SimpleFunctionCompiler.java
evacchi: I have renamed to Immutable for consistency.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/GlobalCompilationPhase.java
pibizza: @evacchi re: factory. It could be a solution, but at the same time I am thinking that for kbase we can create a nice Null KBase that behaves pretends to do some work and really does nothing. This remove the if completely and remove the need for a double Immutable/Mutable.

evacchi: I appreciate the idea in spirit but -1: the goal is to able to abstract the immutable compilation phase from the concept of a kbase entirely. In fact, the factory would be only useful in the old parts of the codebase where kbase may be present at all. In other components we may not even have the kbase concept

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/RuleCompilationPhase.java
```diff
@@ -0,0 +1,157 @@
...
+import org.drools.kiesession.rulebase.InternalKnowledgeBase;
+import org.kie.api.io.Resource;
+import org.kie.internal.builder.ResourceChange;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+public class RuleCompilationPhase extends ImmutableRuleCompilationPhase {
```
danielezonca, line 38: This is not ideal: `RuleCompilationPhase` extends `ImmutableRuleCompilationPhase` so this means that everywhere `ImmutableRuleCompilationPhase` is necessary, I can replace it with the mutable version changing completely the semantic. Said differently it seems to break the substitution principle

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        for (String toBeRemoved : existingGlobals) {
+            removeGlobal(pkg, toBeRemoved);
+        }
+    }
+
+    protected void addGlobal(InternalKnowledgePackage pkg, String identifier, Class<?> clazz) {
+        pkg.addGlobal(identifier, clazz);
+        globalVariableContext.addGlobal(identifier, clazz);
```
danielezonca, line 78: This method is modifying `pkg` and `globalVariableContext` so what _immutable_ means in this context?
I'm not neither sure if `process()` method is idempotent so I don't understand the immutable aspect 🤔

danielezonca (approved):
I understand the idea but I think `Immutable` concept is not very clear (at least from me). Btw approving with two comments because in general the PR looks good to me

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/RuleCompilationPhase.java
evacchi: well in a way you could argue that can be said for any "phase" as they all implement `CompilationPhase`. But at the same point, that's the whole point: you can substitute the equivalent phase, *augmenting* it to support a mutable case (not really "completely changing" the semantics, it's just handling an extra-case)

regardless, I am not attached neither to the name, neither to the approach: it's all about separating the case when a kbase is used, from the case when the kbase is not necessary, thereby allowing to reuse the case regardless of the existence of the _concept_ of kbase entirely.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ImmutableGlobalCompilationPhase.java
evacchi: I have used the prefix "immutable" to indicate that a kbase won't be mutated; data can be still collected in the process (in this case by mutating context objects) -- the naming however is terrible and suggestions are welcome.

PR #4523, opened 2022-07-06
split drl to executable model compilation in 2 steps

mariofusco:


efesto/efesto-core/efesto-compilation-manager/efesto-compilation-manager-core/src/main/java/org/kie/efesto/compilationmanager/core/utils/CompilationManagerUtils.java
```diff
@@ -71,9 +69,9 @@ public static Collection<IndexFile> getIndexFilesWithProcessedResource(EfestoRes
                 populateIndexFile(indexFile, compilationOutput);
                 if (compilationOutput instanceof EfestoCallableOutputClassesContainer) {
                     loadClasses(((EfestoCallableOutputClassesContainer) compilationOutput).getCompiledClassesMap(), memoryCompilerClassLoader);
-                } else if (compilationOutput instanceof EfestoRedirectOutput) {
```
gitgabrio, line 74: Thanks @mariofusco for removing that!

efesto/efesto-core/efesto-compilation-manager/efesto-compilation-manager-core/src/main/java/org/kie/efesto/compilationmanager/core/utils/CompilationManagerUtils.java
```diff
@@ -71,9 +69,9 @@ public static Collection<IndexFile> getIndexFilesWithProcessedResource(EfestoRes
...
-                    toPopulate.addAll(getIndexFilesWithProcessedResource((EfestoRedirectOutput) compilationOutput, memoryCompilerClassLoader));
                 }
+            } else if (compilationOutput instanceof EfestoResource) {
```
gitgabrio, line 73: @mariofusco ... but I'm not sure about that.
The code you removed was used only in some tests, so did not cover the real situation; beside, it was meant for a specific case ("A redirect resource")
But with this new piece, I'm realizing that it may happen that the same Collection (being a List) could ends up containing multiple instances of the same index file; i.e. line 74 could return a list containing an indexfile already present, for example if, inside that recursive loop, the same kind of resource plugin (dmn/drl/pmml/whatever) is invoked multiple times: does this make sense ?
To be sure, I'll verify tomorrow on my branch

efesto/efesto-core/efesto-compilation-manager/efesto-compilation-manager-core/src/main/java/org/kie/efesto/compilationmanager/core/utils/CompilationManagerUtils.java
```diff
@@ -116,14 +114,11 @@ static void populateGeneratedResources(GeneratedResources toPopulate, EfestoComp
     }
 
     static GeneratedResource getGeneratedResource(EfestoCompilationOutput compilationOutput) {
-        if (compilationOutput instanceof EfestoRedirectOutput) {
```
gitgabrio, line 119: @mariofusco 
I have the impression this changes completely the semantic of the method -. I have to check on the other branch

kie-drl/kie-drl-compilation-common/src/main/java/org/kie/drl/engine/compilation/service/KieCompilerServiceDecisionTable.java
```diff
@@ -26,7 +41,7 @@ public <T extends EfestoResource, E extends EfestoCompilationOutput> List<E> pro
                     this.getClass().getName(),
                     toProcess.getClass().getName()));
         }
-        return Collections.singletonList( (E) getDrlCallableClassesContainer((DecisionTableFileSetResource) toProcess, memoryCompilerClassLoader) );
+        return Collections.singletonList( (E) dTableToDrl((DecisionTableFileSetResource) toProcess, memoryCompilerClassLoader) );
```
gitgabrio, line 44: @mariofusco 
what does `dTableToDrl` means ?
IMO I think it would be better to have names understandle for everyone, since the code we write should be easy to understand by anyone else - just an opinion, though

gitgabrio (commented):
@mariofusco Thanks for the PR!
Beside some strongly opinionated comments, I have to check on my branch if the modifications inside "efesto" core does not break something.

kie-drl/kie-drl-compilation-common/src/main/java/org/kie/drl/engine/compilation/service/KieCompilerServiceDecisionTable.java
mariofusco: Naming is always hard, but to be honest I find the one that I'm proposing much more understandable. That method is intended to do exactly what it says: transforming a decision table into drl as an intermediary compilation step, so that the drl will be subsequently processed by the other compiler.

efesto/efesto-core/efesto-compilation-manager/efesto-compilation-manager-core/src/main/java/org/kie/efesto/compilationmanager/core/utils/CompilationManagerUtils.java
gitgabrio: @mariofusco Fine for me - misreading

efesto/efesto-core/efesto-compilation-manager/efesto-compilation-manager-core/src/main/java/org/kie/efesto/compilationmanager/core/utils/CompilationManagerUtils.java
```diff
@@ -57,23 +55,23 @@ public class CompilationManagerUtils {
     private CompilationManagerUtils() {
     }
 
-    public static Collection<IndexFile> getIndexFilesWithProcessedResource(EfestoResource toProcess, KieMemoryCompiler.MemoryCompilerClassLoader memoryCompilerClassLoader) {
+    public static Set<IndexFile> getIndexFilesWithProcessedResource(EfestoResource toProcess, KieMemoryCompiler.MemoryCompilerClassLoader memoryCompilerClassLoader) {
```
gitgabrio, line 58: @mariofusco 
Thanks!

gitgabrio (approved):
@mariofusco 
Many thanks! LGTM

kie-drl/kie-drl-compilation-common/src/main/java/org/kie/drl/engine/compilation/service/KieCompilerServiceDecisionTable.java
gitgabrio: :+1:

PR #4524, opened 2022-07-07
[DROOLS-7056] NullPointerException when sending DeleteCommand for non…

tkobayas:
…-existing fact

**Ports** 
This PR is for 7.x
for main -> https://github.com/kiegroup/drools/pull/4536

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-7056

**referenced Pull Requests**: 
* https://github.com/kiegroup/droolsjbpm-integration/pull/2797

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

- for <b>native checks</b>  
  Run native checks  

  Run native checks 

- for <b>mandrel checks</b>  

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
+++ b/drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
@@ -471,7 +471,7 @@ public class NamedEntryPoint
                     handle = this.objectStore.reconnect(handle);
                 }
                 if (handle == null) {
-                    // the handle doesn't exist. no op
+                    log.warn("The factHandle doesn't exist so cannot be deleted. " + factHandle.toExternalForm());
                     return;
                 }
 
--- a/drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
+++ b/drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
@@ -470,6 +470,10 @@ public class NamedEntryPoint
                 if (handle.isDisconnected()) {
                     handle = this.objectStore.reconnect(handle);
                 }
+                if (handle == null) {
+                    // the handle doesn't exist. no op
+                    return;
+                }
 
                 if (!handle.getEntryPointId().equals( entryPoint )) {
                     throw new IllegalArgumentException("Invalid Entry Point. You updated the FactHandle on entry point '" + handle.getEntryPointId() + "' instead of '" + getEntryPointId() + "'");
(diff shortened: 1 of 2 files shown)


drools-core/src/main/java/org/drools/core/common/NamedEntryPoint.java
```diff
@@ -470,6 +470,10 @@ public void delete(FactHandle factHandle,
                 if (handle.isDisconnected()) {
                     handle = this.objectStore.reconnect(handle);
                 }
+                if (handle == null) {
+                    // the handle doesn't exist. no op
+                    return;
+                }
```
tkobayas: In a usual Java API use case (not disconnected),
```
DefaultFactHandle johnHandle = (DefaultFactHandle)ksession.insert(john);
ksession.fireAllRules();
ksession.delete(johnHandle); // 1st delete
ksession.fireAllRules();
ksession.delete(johnHandle); // 2nd delete
ksession.fireAllRules();
```
No Exception is raised and the 2nd delete doesn't affect the ksession. So I implemented no-op rather than throwing an Exception.

danielezonca: I'm not a fan of a silent "no op". For example the [`remove` method](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#remove(java.lang.Object)) of a `Collection` returns a `boolean` so it is possible to recognise if the element has been really deleted.

I agree that NPE is not the right behaviour (we should use something like `NoSuchElementException`) but if we want to do a no-op, I would at least log a message as `warn` before return.

Wdyt?

tkobayas: @danielezonca  Thank you for pointing this out. I chose no-op just to meet with the usual (not disconnected) use case. TBH, yes, I prefer an Exception (`NoSuchElementException`) because it's usually caused by a client operation mistake. Could you share your thought on this, @mariofusco  ?

mariofusco: @tkobayas I'd prefer to avoid throwing an exception and just logging a warning message in this case. In my opinion throwing exception should be used only for actual failures that compromise the integrity of the session and this is clearly not the case.

tkobayas: @mariofusco Okay. I added warn logging. Thanks!

mariofusco (approved):

PR #4529, opened 2022-07-07
Drools 6987 Sonarqube based fixes

gorky:
**JIRA:**
https://issues.redhat.com/browse/DROOLS-6987

https://jenkins.updater.j2eeguys.com/job/Kiegroup%20Drools

Use List, DeQueue, LinkedList and Hashmap rather than synchronized Stack, Vector, and Hashtable.
With assist from the Honor Systems Updater.

Jenkins execute product fdb


mariofusco:
@gorky In general I wouldn't change internal data structure without a proper profiling assessment and just because "Sonar said so", but in this case I believe that all the changes you suggested make sense and doesn't have a direct impact on the core engine so I'm fine with this.

What I don't like is the extensive use of `LinkedList`s. They proved to perform quite badly in real use case scenarios, mainly for the lack of data locality that in practice implies lot's of cache misses. We are trying to avoid this sort of data structure wherever possible. Please replace them anywhere possible with an [ArrayDeque](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ArrayDeque.html). You can find in that JavaDocs also a few hints on the performance implications.

gorky:
Hi @mariofusco 
In the case of the LinkedList vs ArrayDeque for the Stack Replacement -- that would be a "Whoops!" in my case, sorry.  You are correct.  I will correct this submission, and then fix the Solution Template for dealing with Stacks for later runs.

It should be noted that there are places (DeclarationScopeResolver for instance) that loop through a list using .get(i), which ArrayDeque doesn't provide.  However, using the descendingIterator in this case gives the same functionality.  I will have the fix along shortly (hopefully later today).

gorky:
Let me know if I missed anything.

mariofusco (approved):

PR #4535, opened 2022-07-12
[DROOLS-7065] fix and simplify propagation of right deletes in existe…

mariofusco:
…ntial nodes

**JIRA**: https://issues.redhat.com/browse/DROOLS-7065


triceo (approved):
Fixes the original problem.
Thank you, Mario!

PR #4542, opened 2022-07-14
[DROOLS-7062] ClassNotFoundException when using generics in accumulat…

tkobayas:
…e inline code

**Ports** 
This PR is for 7.x.
Will do for main

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-7062

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

- for <b>pull request checks</b>  
  Please add comment: <b>Jenkins retest this</b>

- for a <b>specific pull request check</b>  
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] tests</b>

- for a <b>full downstream build</b> 
  - for <b>jenkins</b> job: please add comment: <b>Jenkins run fdb</b>
  - for <b>github actions</b> job: add the label `run_fdb`

- <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

- <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

- <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>

- for <b>quarkus branch checks</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins run quarkus-branch</b>

- for a <b>quarkus branch specific check</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] quarkus-branch</b>

- for <b>quarkus main checks</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins run quarkus-main</b>

- for a <b>specific quarkus main check</b>  
[...]

PR #4545, opened 2022-07-18
[DROOLS-7062] ClassNotFoundException when using generics in accumulat…

mariofusco:
…e inline code

**JIRA**: https://issues.redhat.com/browse/DROOLS-7062

@tkobayas I reviewed your original PR ( https://github.com/kiegroup/drools/pull/4542 ) and discussed this with @lucamolteni. As he suggested the actual fix is getting rid of all the `DrlNameExpr` nodes before the printer tries to convert them in strings. Please give a look and let me know if this is ok for you. I sent this PR against main, if ok I'll close yours and then we will need to also backport this fix to 7.x

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/pattern/PatternVisitor.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/pattern/PatternVisitor.java
@@ -21,7 +21,6 @@ import java.util.List;
 
 import org.drools.drl.ast.descr.AccumulateDescr;
 import org.drools.drl.ast.descr.BaseDescr;
-import org.drools.drl.ast.descr.ExprConstraintDescr;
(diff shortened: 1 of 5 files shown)


drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/pattern/PatternVisitor.java
```diff
@@ -89,10 +87,4 @@ private DSLNode parsePatternWithClass(PatternDescr pattern, String className) {
         }
         return null;
     }
-
-    private boolean areAllConstraintsPositional(List<? extends BaseDescr> constraintDescrs) {
```
mariofusco, line 93: Code removal not related with the fix. I just found it was no longer in use :)

tkobayas (approved):
Thanks! I was not aware of `transformDrlNameExprToNameExpr`!

PR #4546, opened 2022-07-18
DROOLS-7003 Refactor addKnowledgeResource, introduce ResourceHandlerManager

nprentza:
refactor KnowledgeBuilderImpl.addKnowledgeResource to use resource handlers.


danielezonca:
ok to test

drools-compiler/src/main/java/org/drools/compiler/builder/impl/PackageDescrBuilder.java
```diff
@@ -34,41 +30,21 @@
...
-
+        this.handlerManager = new ResourceHandlerManager(configuration,releaseId,dslExpander);
     }
 
     public Collection<CompositePackageDescr> build(Map<ResourceType, List<CompositeKnowledgeBuilderImpl.ResourceDescr>> resourcesByType) {
         Map<String, CompositePackageDescr> packages = new HashMap<>();
 
-        for (ResourceType type : this.orderedResourceTypes) {
-            ResourceHandler mapper = handlerForType(type);
+        for (ResourceType type : this.handlerManager.getOrderedResourceTypes()) {
```
evacchi, line 46: it's better not not expose the internal config of such a class

drools-compiler/src/main/java/org/drools/compiler/builder/impl/PackageDescrBuilder.java
```diff
@@ -34,41 +30,21 @@
...
+            ResourceHandler mapper = this.handlerManager.handlerForType(type);
```
evacchi, line 47: I would prefer not to expose `getOrderedResourceTypes()` directly; either it should belong to PackageDescrBuilder or the handlerManager should return the handlers already in the right order.

However, I can't think of a good way to do it right now without further complicating the PR 😬  it's ok for now, let's create a ticket https://issues.redhat.com/browse/DROOLS-7077

drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java
```diff
@@ -522,28 +362,25 @@ public void addKnowledgeResource(Resource resource,
...
-                addPackageFromDslr(resource);
-            } else if (ResourceType.RDSLR.equals(type)) {
-                addPackageFromDslr(resource);
+
+            if ((ResourceType.DRL.equals(type)) || (ResourceType.GDRL.equals(type)) || (ResourceType.RDRL.equals(type))
+                    || (ResourceType.DESCR.equals(type)) || (ResourceType.TDRL.equals(type))
+            || (ResourceType.DSLR.equals(type)) || (ResourceType.RDSLR.equals(type))
+            || ResourceType.TEMPLATE.equals(type)) {
+                ResourceHandlerManager handlerManager = new ResourceHandlerManager(this.getBuilderConfiguration(), this.releaseId, this::getDslExpander);
+                ResourceHandler handler = handlerManager.handlerForType(type);
```
evacchi: formatting is inconsistent here, align everything in the same way.

evacchi: also, we made great progress here, but unfortunately there are still some types that are being handled in a custom way! we should strive to uniform everything, let's see if we can do that; I created a follow-up ticket: https://issues.redhat.com/browse/DROOLS-7078

evacchi (approved):

evacchi:
it's good for me, but I have created a couple of follow-up tickets to further improve. /cc @mariofusco

mariofusco (approved):

PR #4581, opened 2022-08-01
[DROOLS-7085] Rule doesn't match with more than 3 rules of BigDecimal…

tkobayas:
… coercion in non executable model

**Ports** 
This PR if for 7.x
for main -> https://github.com/kiegroup/drools/pull/4613

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-7085

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

- for <b>native checks</b>  
  Run native checks  

  Run native checks 

- for <b>mandrel checks</b>  

</details>

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/CompositeObjectSinkAdapter.java
@@ -42,6 +42,8 @@ import org.drools.core.spi.ReadAccessor;
 import org.drools.core.util.index.AlphaRangeIndex;
 import org.drools.core.util.index.IndexUtil.ConstraintType;
 
+import static org.drools.core.util.index.IndexUtil.isBigDecimalEqualityConstraint;
+
 public class CompositeObjectSinkAdapter implements ObjectSinkPropagator {
 
     //    /** You can override this property via a system property (eg -Ddrools.hashThreshold=4) */
@@ -207,6 +209,7 @@ public class CompositeObjectSinkAdapter implements ObjectSinkPropagator {
     private static boolean isHashable( IndexableConstraint indexableConstraint ) {
         return indexableConstraint.getConstraintType() == ConstraintType.EQUAL && indexableConstraint.getField() != null &&
                 indexableConstraint.getFieldExtractor().getValueType() != ValueType.OBJECT_TYPE &&
+                !isBigDecimalEqualityConstraint(indexableConstraint) &&
                 // our current implementation does not support hashing of deeply nested properties
                 indexableConstraint.getFieldExtractor().getIndex() >= 0;
     }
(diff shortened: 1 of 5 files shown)


tkobayas:
Firstly, this is a hash AlphaIndex issue that is enabled when we have more than 3 same type constraint rules. The original issue in the JIRA (https://issues.redhat.com/browse/DROOLS-7085) is that `10` of the constraint `salary == 10` is created as `BigDecimal(10.0)` so it fails to match with the hash of inserted `BigDecimal("10")` in the unit test. 

Fixing `MathUtils.getBigDecimal` solves this particular issue but it doesn't solve the root issue that hashCode of BigDecimal with different scales (e.g. `10` and `10.0`) are different. The issue is also pointed in another JIRA https://issues.redhat.com/browse/DROOLS-6260

Hence, we should disable hash AlphaIndex for BigDecimal. WDYT? @mariofusco

tkobayas:
Ah, same for hash BetaIndex.

tkobayas:
If we will disable hash AlphaIndex/BetaIndex for BigDecimal, we may introduce an option to enable it for performance as long as users are certain that they use completely the same BigDecimal scale.

mariofusco:
> Fixing `MathUtils.getBigDecimal` solves this particular issue but it doesn't solve the root issue that hashCode of BigDecimal with different scales (e.g. `10` and `10.0`) are different. The issue is also pointed in another JIRA https://issues.redhat.com/browse/DROOLS-6260

In which cases the fix you added to `MathUtils` isn't enough? Can you reproduce also this problem and if so can you please add this reproducer as another test case to this PR?

Also is it true that the executable model doesn't suffer of this problem as reported in https://issues.redhat.com/browse/DROOLS-7085 ? If so why? At the moment I don't understand the differences.

tkobayas:
@mariofusco  Sorry that I haven't added test cases while saying there is a root issue.

> In which cases the fix you added to MathUtils isn't enough? Can you reproduce also this problem and if so can you please add this reproducer as another test case to this PR?

Sure, added `testAlphaIndexWithBigDecimalDifferentScale` and `testBetaIndexWithBigDecimalDifferentScale`. Please see that the BigDecimal scale are different so it fails to match.

Note : in `testAlphaIndexWithBigDecimalDifferentScale`, I still use coercion because `salary == 10B` doesn't create an alpha index in non-exec-model (it may be another issue).

> Also is it true that the executable model doesn't suffer of this problem as reported in https://issues.redhat.com/browse/DROOLS-7085 ? If so why? At the moment I don't understand the differences.

exec-model doesn't suffer of creating `BigDecimal(10.0)` from `salary == 10` because it creates the value with `EvaluationUtil.toBigDecimal(10)` instead of `MathUtils.getBigDecimal`.

However, regarding the additional 2 tests (`testAlphaIndexWithBigDecimalDifferentScale` and `testBetaIndexWithBigDecimalDifferentScale`), these tests fail with both exec-model and non-exec-model.

mariofusco:
The problem here is in the special semantic that we're giving to the `BigDecimal`. In general when in the drl we write `a == b`, we mean `a.equals(b)`. But with `BigDecimal`s we have a different requirement because we want `new BigDecimal("10") == new BigDecimal("10.00")` to return `true` even if in java `new BigDecimal("10").equals(new BigDecimal("10.00"))` returns `false`. Given this very specific situation I agree that the only reasonable solution to fulfill this requirement is disabling alpha indexing for BigDecimals. Please implement it with this PR.

mariofusco (approved):

PR #4585, opened 2022-08-02
[DROOLS-7088] Property reactivity discrepancy between executable mode…

tkobayas:
…l and non executable model with a capitalized property

- unit tests

**Ports** 
This PR is for 7.x
for main -> https://github.com/kiegroup/drools/pull/4596

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-7088

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

- for <b>native checks</b>  
  Run native checks  

  Run native checks 

- for <b>mandrel checks</b>  

</details>

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PropertyReactivityTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/PropertyReactivityTest.java
@@ -27,6 +27,7 @@ import org.drools.modelcompiler.domain.Address;
 import org.drools.modelcompiler.domain.Person;
 import org.drools.modelcompiler.domain.Pet;
 import org.drools.modelcompiler.domain.Result;
+import org.drools.modelcompiler.domain.VariousCasePropFact;
 import org.junit.Test;
 import org.kie.api.builder.KieBuilder;
 import org.kie.api.builder.Message.Level;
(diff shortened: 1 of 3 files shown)


tkobayas:
Hi @mariofusco  , this is a tricky issue.

I added 3 tests but the first test `testOnlyFirstLetterIsUpperCaseProperty` is the problem. It behaves differently between non-exec-model and exec-model.

With non-exec-model, it considers that `MyTarget` is not a property (It collects property names using `ClassUtils.getAccessibleProperties` beforehand so it knows the correct property name is `myTarget`). Hence it results in AllSetButLastBitMask.

https://github.com/kiegroup/drools/blob/7.73.0.Final/drools-mvel/src/main/java/org/drools/mvel/builder/MVELConsequenceBuilder.java#L291-L297

This is correct from JavaBeans spec point of view.

```
8.8 Capitalization of inferred names.

When we use design patterns to infer a property or event name, we need to decide what rules
to follow for capitalizing the inferred name. If we extract the name from the middle of a normal
mixedCase style Java name then the name will, by default, begin with a capital letter.

Java programmers are accustomed to having normal identifiers start with lower case letters.
Vigorous reviewer input has convinced us that we should follow this same conventional rule
for property and event names.

Thus when we extract a property or event name from the middle of an existing Java name, we
normally convert the first character to lower case. However to support the occasional use of all
upper-case names, we check if the first two characters of the name are both upper case and if
so leave it alone. So for example,

    “FooBah” becomes “fooBah”
    “Z” becomes “z”
    “URL” becomes “URL”

We provide a method Introspector.decapitalize which implements this conversion rule.
```

With exec-model, `Consequence.createExecuteCallMvel` converts
```
modify($f) { MyTarget = "123" };
```
to
```
$f.setMyTarget("123");
```
Then, it calculates BitMask in `Consequence.rewriteRHS`.

https://github.com/kiegroup/drools/blob/7.73.0.Final/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/generator/Consequence.java#L364

At this point, it's natural to consider that `setMyTarget` is for a property `myTarget` so it creates a BitMask for `myTarget`, not AllSetButLastBitMask.

Do you think we should fix exec-model to solve the behaviour discrepancy?

mariofusco:
@tkobayas I believe that the answer is in what the code actually does and then decide the correct property reactivity mask from there. When you write 

`modify($f) { MyTarget = "123" };`

is it actually modifying the MyTarget property of the $f object? I believe so, otherwise the rule wouldn't compile at all correct? If I'm right on this, the consequence is setting the MyTarget field of the fact and then the correct behaviour should be the one of the executable model, propagating a bitmask with only the bit corresponding to the MyTarget field set.

Do you agree? I'm of course open to discuss this, and moreover my concern is that there could be some implication specifically related to the Japanese language that I'm possibly ignoring. If so, can you please clarify?

tkobayas:
Thank you, @mariofusco 

> I believe that the answer is in what the code actually does

Ahh, yes. You are right.

```
modify($f) { MyTarget = "123" };
```

> is it actually modifying the MyTarget property of the $f object?

Yes. More precisely, it modifies `myTarget` property (The property name is deduced from the accessors `getMyTarget` `setMyTarget`). The actulal private field name `MyTarget` of `VariousCasePropFact` should not matter (The private field name can be even `data`). But again, yes, it actually modifies the private field `MyTarget` in this test.

> the consequence is setting the MyTarget field of the fact and then the correct behaviour should be the one of the executable model, propagating a bitmask with only the bit corresponding to the MyTarget field set.

Agreed. I'll fix non-exec-model. Thanks!

> moreover my concern is that there could be some implication specifically related to the Japanese language

`testFirstLetterIsMultibyteProperty` runs successfully. So you may completely ignore Japanese language part :)

mariofusco (approved):

PR #4637, opened 2022-08-26
Skip IT tests when using quick build

cristianonicolai:
Overriding `skipITs` property here was blocking the change of value when executing with `-Dquickly`


gitgabrio (approved):
Thanks @cristianonicolai LGTM :+1:

PR #4661, opened 2022-09-05
[DROOLS-7144] fix parsing of a oopath constraint having a string cont…

mariofusco:
…aining a single quote

**JIRA**: https://issues.redhat.com/browse/DROOLS-7144


tarilabs (approved):
I concur is a required fix as an ad-interim solution, this is potentially impacting on many use cases of oopath. thanks for the fix @mariofusco !

PR #4671, opened 2022-09-07
Use the same key for put as get when caching class lookups

jbonzohln:
Using `clazz.getSimpleName()` as the cache key when inserting into `this.cachedImports` results in the cache not working as expected since the lookup is based on the `className` parameter.


mdproctor:
Does the updates to the tests demonstrate the error? does it now fail without you fix?

mariofusco (commented):
This is not a bug, but an intentional behaviour and the cache is working as expected. We use the class `simpleName` as the key in that cache because once a class import is resolved it is (generally) referred in the rest of the drl file to be parsed with its simple name. If you the fully qualified name you will just get tons of cache misses. 

Also note that we can accept pull requests only against our `main` branch.

PR #4694, opened 2022-09-15
[DROOLS-7158] enforce that ProjectClassLoader.CACHE_NON_EXISTING_CLAS…

mariofusco:
…SES is always set to true with a proper test case

**JIRA**: https://issues.redhat.com/browse/DROOLS-7158


tarilabs (approved):
Thank you!

PR #4696, opened 2022-09-19
[DROOLS-7162] create a bundle module for rule units

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7162


tkobayas (approved):
Thanks. Adding `drools-ruleunits-engine` makes sense to me.

tarilabs (approved):
ack. Thanks!

PR #4720, opened 2022-09-28
fix(marshall): recognize fact method with param (DROOLS-7182)

nicko466:


mariofusco:
We won't release any newer version of the 7.x series so it's pointless to bring in a pull request on this branch and the drools-workbench don't exists anymore in Drools 8.

PR #4722, opened 2022-09-30
[DROOLS-7181] allow reuse of RuleUnitDefinition in rule unit Java DSL

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7181


pibizza (approved):
Patch looks good to me.

PR #4731, opened 2022-10-04
[DROOLS-7192] avoid to match any constraint on protypes undefined values

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7192


evacchi (approved):
Approved. I wonder if this should be the default behavior that we want in the core. Let's put a pin on that.

PR #4775, opened 2022-10-26
[DROOLS-7216] automatically evict facts that don't match any rule when using an immutable KieBase

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7216


mariofusco:
The error in building `kogito-example` is an `OOM` not related with this change.

evacchi (approved):

PR #4790, opened 2022-11-03
DROOLS-7228 Prepare to add drools-drl-quarkus-extension to Quarkus Platform

evacchi:
In order to support this, we need to properly publish the integration tests as test jars.

We will also need to automate the pipelines, but I'll probably need @radtriste help there 😅

### DIFF
--- a/bom/drools-bom/pom.xml
+++ b/bom/drools-bom/pom.xml
@@ -595,6 +595,30 @@
 
       <dependency>
         <groupId>org.drools</groupId>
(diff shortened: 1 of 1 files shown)


radtriste:
@evacchi what do you need exactly ? updating the quarkus platform during the release process ?

radtriste:
@evacchi is this PR still needed ?

evacchi:
apparently not? 🤔

PR #4793, opened 2022-11-04
[DROOLS-7226] extend rule unit DSL to programmatically define a rule unit

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7226

### DIFF
--- a/drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
+++ b/drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
@@ -90,7 +90,7 @@ public class RuleDefinition implements RuleFactory {
 
     @Override
     public <A> Pattern1DefImpl<A> on(DataSource<A> dataSource) {
-        FieldDefinition dataSourceField = findUnitField(dataSource);
+        DataSourceFieldDefinition dataSourceField = (DataSourceFieldDefinition) findUnitField(dataSource);
         Pattern1DefImpl<A> pattern1 = new Pattern1DefImpl<>(this,
                 declarationOf(dataSourceField.getDataSourceClass(), entryPoint(asGlobal(() -> dataSourceField, dataSource).getName())));
         if (registerNewPattern) {
@@ -211,10 +211,13 @@ public class RuleDefinition implements RuleFactory {
 
     public interface FieldDefinition {
         Object get(RuleUnitData unit);
+    }
+
+    public interface DataSourceFieldDefinition extends FieldDefinition {
         <A> Class<A> getDataSourceClass();
     }
 
-    static class ReflectiveFieldDefinition implements FieldDefinition {
+    static class ReflectiveFieldDefinition implements DataSourceFieldDefinition {
         private final Field field;
 
         ReflectiveFieldDefinition(Field field) {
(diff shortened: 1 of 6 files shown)


tarilabs (approved):
thank you

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
```diff
@@ -171,26 +174,108 @@ public Rule toRule() {
         return ruleBuilder.build(items.toArray(new RuleItemBuilder[items.size()]));
     }
 
-    private <A> Class<A> findDataSourceClass(Field field) {
-        Type dsType = field.getGenericType();
-        if (dsType instanceof ParameterizedType) {
-            return (Class<A>) ((ParameterizedType) dsType).getActualTypeArguments()[0];
+    private FieldDefinition findUnitField(Object object) {
+        assert(object != null);
```
danielezonca: Do we use assertion in drools codebase?

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
```diff
@@ -171,26 +174,108 @@ public Rule toRule() {
...
+
+        private final String name;
+
+        SyntheticGlobalFieldDefinition(String name) {
+            this.name = name;
+        }
+
+        @Override
+        public Object get(RuleUnitData unit) {
+            return ((SyntheticRuleUnit) unit).getGlobals().get(name);
```
danielezonca, line 277: This cast can fail

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
```diff
@@ -171,26 +174,108 @@ public Rule toRule() {
...
+        }
+
+        @Override
+        public <A> Class<A> getDataSourceClass() {
+            throw new UnsupportedOperationException();
```
danielezonca: When this method is used? Is it safe to fail?

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
mariofusco: I think so, why not?

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
mariofusco: No, it can't, this object is created only for `SyntheticRuleUnit`.

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
mariofusco: It should be never called, it's necessary only for `DataSource`s field and this class is used for a plain global.

drools-ruleunits/drools-ruleunits-dsl/src/main/java/org/drools/ruleunits/dsl/util/RuleDefinition.java
tarilabs: Honestly this answer surprises me. Do you have an example of somewhere else is used, please?

evacchi (approved):

PR #4856, opened 2022-12-06
[DROOLS-7249] make eager segment initialization opt-it

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7249

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakBuilder.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakBuilder.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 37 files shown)


drools-core/src/main/java/org/drools/core/phreak/PhreakBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+
+    static boolean isEagerSegmentCreation() {
+        return Holder.EAGER_SEGMENT_CREATION;
+    }
+
+    void addRule(TerminalNode tn, Collection<InternalWorkingMemory> wms, RuleBase kBase);
+    void removeRule(TerminalNode tn, Collection<InternalWorkingMemory> wms, RuleBase kBase);
+
+    class Holder {
+        private static final boolean EAGER_SEGMENT_CREATION = System.getProperty("drools.useEagerSegmentCreation") != null;
```
whiskeysierra, line 38: This would still evaluate to `true` even if someone would use it like this:

```
-Ddrools.useEagerSegmentCreation=false
```

Maybe, in the interest of the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), it would be better to use one of the following two options instead?

```java
"true".equalsIgnoreCase(System.getProperty("drools.useEagerSegmentCreation"))
```

or 

```java
Boolean.getBoolean("drools.useEagerSegmentCreation")
```

PR #4939, opened 2023-01-24
[kie-issues#81] Set en_US locale for tests

### LINKED ISSUE #81: Address string comparison failures in tests that rely on formatted numbers
Some tests are based on string comparison of Numbers/Date/Currency. E.g.

`[ERROR] Failures: 
[ERROR]   FEELFunctionDefinitionTest>BaseFEELTest.testExpression:72->BaseFEELTest.assertResult:89 [Evaluating: '{ 
    string format : function( mask, value ) external {
                      java : {
                          class : "java.lang.String",
                          method signature : "format( java.lang.String, [Ljava.lang.Object; )"
                      }
                  },
    format currency : function( amount ) 
                 string format( "$%,4.2f", amount )
    ,
   result : format currency( 76499.3456 )
}.result'] expected:<"$76[,499.]35"> but was:<"$76[.499,]35">
`
Formatting is dependent on OS locale, and failure is related to a Locale different then EN.
Date formatting may also fails between US_en and UK_en.
Find a way to set this (possibly at pom level) to have tests always working regardless on OS locale, while keepng base code OS-locale dependent (i.e. - any dependent project should keep the possibility to use a locally formatted numer/date...)

gitgabrio:
@yesamer @Josephblt  @baldimir @mariofusco @danielezonca 

Fixes https://github.com/kiegroup/kie-issues/issues/81

* paste the link(s) from GitHub here
* link 2
* link 3 etc.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] quarkus-main</b>

- for <b>quarkus lts checks</b>  
  Run checks against Quarkus lts branch  
  Please add comment: <b>Jenkins run quarkus-lts</b>

- for a <b>specific quarkus lts check</b>  
  Run checks against Quarkus lts branch  
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] quarkus-lts</b>

- for <b>native checks</b>  
  Run native checks  

  Run native checks 

- for <b>mandrel checks</b>  

 - for <b>mandrel lts checks</b>  
  Run native checks against Mandrel image and quarkus lts branch
  Please add comment: <b>Jenkins run mandrel-lts</b>

- for a <b>specific mandrel lts check</b>  
  Run native checks against Mandrel image and quarkus lts branch
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] mandrel-lts</b>
</details>

<details>
<summary>
How to backport a pull request to a different branch?
</summary>

In order to automatically create a **backporting pull request** please add one or more labels having the following format `backport-<branch-name>`, where `<branch-name>` is the name of the branch where the pull request must be backported to (e.g., `backport-7.67.x` to backport the original PR to the `7.67.x` branch).

> **NOTE**: **backporting** is an action aiming to move a change (usually a commit) from a branch (usually the main one) to another one, which is generally referring to a still maintained release branch. Keeping it simple: it is about to move a specific change or a set of them from one branch to another.

Once the original pull request is successfully merged, the automated action will create one backporting pull request per each label (with the previous format) that has been added.

If something goes wrong, the author will be notified and at this point a manual backporting is needed.

> **NOTE**: this automated backporting is triggered whenever a pull request on `main` branch is labeled or closed, but both conditions must be satisfied to get the new PR created.
</details>


baldimir:
Hi @gitgabrio, There is a failing test on Windows, which also fails for me on Windows. As I know you have an arm Mac, I will take a look and let you know what is causing problems.

gitgabrio:
> Hi @gitgabrio, There is a failing test on Windows, which also fails for me on Windows. As I know you have an arm Mac, I will take a look and let you know what is causing problems.

Thanks @baldimir ! With that info alone
`kiegroup_drools/drools/drools-ecj/src/test/java/org/drools/ecj/JavaCompilerI18NTest.java:46 | expected:<[1]> but was:<[0]>`
I really could not have a clue (anyway, both the failing test and the failure itself are really strange for me)

yesamer (approved):

Josephblt (approved):

gitgabrio:
@danielezonca @mariofusco Hi! Could you please merge this, squashing commits? Thanks!

danielezonca:
@gitgabrio 

I'm not against this PR in principle, I just find strange that nobody noticed it until now 🤔 
How can you reproduce [the issue described](https://github.com/kiegroup/kie-issues/issues/81)? 
I'm asking because we have people more or less ~everywhere so I guess why nobody noticed it 🤔

gitgabrio:
> @gitgabrio
> 
> I'm not against this PR in principle, I just find strange that nobody noticed it until now 🤔 How can you reproduce [the issue described](https://github.com/kiegroup/kie-issues/issues/81)? I'm asking because we have people more or less ~everywhere so I guess why nobody noticed it 🤔

It is the same question I asked myself. First, I did not have this issue when using my old laptop, but I hit that when I begun working with my new laptop; I ignored it because it also seemed strange to me. Then, last week, @Josephblt reported a very similar issue on the Stunner project. So, we realized that the problem is related to underlying OS locale. Almost anyone developing with drools have the locale set to en_US. But both my new laptop and @Josephblt one have different locale. The issue is also more hidden because it appears only when executing build with tests, and all our CI env always uses en_US.
To complete the picture, here's a discussion I had when I asked for a not-en_US CI build: https://mail.google.com/chat/u/0/#chat/space/AAAATlh7wNo

danielezonca:
> > @gitgabrio
> > I'm not against this PR in principle, I just find strange that nobody noticed it until now 🤔 How can you reproduce [the issue described](https://github.com/kiegroup/kie-issues/issues/81)? I'm asking because we have people more or less ~everywhere so I guess why nobody noticed it 🤔
> 
> It is the same question I asked myself. First, I did not have this issue when using my old laptop, but I hit that when I begun working with my new laptop; I ignored it because it also seemed strange to me. Then, last week, @Josephblt reported a very similar issue on the Stunner project. So, we realized that the problem is related to underlying OS locale. Almost anyone developing with drools have the locale set to en_US. But both my new laptop and @Josephblt one have different locale. The issue is also more hidden because it appears only when executing build with tests, and all our CI env always uses en_US. To complete the picture, here's a discussion I had when I asked for a not-en_US CI build: https://mail.google.com/chat/u/0/#chat/space/AAAATlh7wNo

Ok thanks for the clarification. My only concern with this solution is that it might hide similar issues in the future: now the build is going to be stable because we force to use `en_US` (or whatever) at maven level but what if there is a real bug that happens only with a different locale? 
Currently the main issue is the assertion in the test that is not portable but now we can hide scenarios where the impl is not portable but this is never visible and it appears only with end user projects (where this property of corse doesn't exist).

I would argue that we might keep the configuration "broken" (aka without forcing the locale), fix the assertion and see if other people will spot similar bugs/regression in the future.

Wdyt?

gitgabrio:
> > > @gitgabrio
> > > I'm not against this PR in principle, I just find strange that nobody noticed it until now 🤔 How can you reproduce [the issue described](https://github.com/kiegroup/kie-issues/issues/81)? I'm asking because we have people more or less ~everywhere so I guess why nobody noticed it 🤔
> > 
> > 
> > It is the same question I asked myself. First, I did not have this issue when using my old laptop, but I hit that when I begun working with my new laptop; I ignored it because it also seemed strange to me. Then, last week, @Josephblt reported a very similar issue on the Stunner project. So, we realized that the problem is related to underlying OS locale. Almost anyone developing with drools have the locale set to en_US. But both my new laptop and @Josephblt one have different locale. The issue is also more hidden because it appears only when executing build with tests, and all our CI env always uses en_US. To complete the picture, here's a discussion I had when I asked for a not-en_US CI build: https://mail.google.com/chat/u/0/#chat/space/AAAATlh7wNo
> 
> Ok thanks for the clarification. My only concern with this solution is that it might hide similar issues in the future: now the build is going to be stable because we force to use `en_US` (or whatever) at maven level but what if there is a real bug that happens only with a different locale? Currently the main issue is the assertion in the test that is not portable but now we can hide scenarios where the impl is not portable but this is never visible and it appears only with end user projects (where this property of corse doesn't exist).
> 
> I would argue that we might keep the configuration "broken" (aka without forcing the locale), fix the assertion and see if other people will spot similar bugs/regression in the future.
> 
> Wdyt?

I understand your concern, but I think the point is slightly different.
First, about "hiding issue": issues are already hidden, that's why we never found that before. They are hidden because "by default" developers actually working on that have en_US on their machines, and because all our CI is working on machine with en_US.  This is inderectly demonstrated by the statement _"we should develop for En locale until we consider it as Multilang env"_  (see chat I pointed above). And also by the fact that, actually, other people hit them, and found different workarounds (i.e. skipping them, momentary change of the machine locale, etc etc)

Second, about changing tests: this is actually what this PR does, somehow. Those failing tests fails because there is a String comparison between an hardcoded value (en_US formatted) and the evaluated one. Since the tests are clearly written in an "en_US" context, this modification verify that they are always executed in the context they are meant to work. Beside that, there are different places where tests fails (I also found some in PMML that I wrote!), even outside drools repo: changing all of them one by one would be extremely painfull and would require to address two issues:

1. how to rewrite such tests in a locale-sensitive way ?
2. how to prevent that, in future, similar hardcoded-en_US based tests will be written ?

Third, about bug that happens only with a different locale: to found them we should write specially-crafted tests, that by design uses different locales and makes assertions/comparisons. But this would be a different kind of test, that should be written by scratch with a different goal/approach..

Last: this PR does not modify the behavior of the core code. What you are pointing at is: what if our core code is broken with different locale ? But this is not affected at all by the current PR. Moreover, actually, it already could be, potentially. We are "blind" regarding any possible problem with locale != than en_US because we do not test them in any way (AFAIK), and that's why I asked for a pipeline/something in the CI with a different locale. Exactly as we have for different java versions and OS, we should build/verify code with different locales, but my understanding is that this would lead to an increase of resource needs that currently can't be afforded.

Does this answer your doubts ?

gitgabrio:
@danielezonca @mariofusco 
If all doubts are sorted out, could this be merged ? Thanks!

gitgabrio:
@yesamer @baldimir @Josephblt @danielezonca 
After some discussion, we agreed to the following modification:

1. this PR won't change in any way the default maven behavior
2. the activation of a not-en_US setup is demanded to a specific profile (`test-en`)
3. this profile may be activated with the profile-based parameter (`-Ptest-en`) or the property-based one (`-DTestEn`)
4. those changes are explained in the paragraph `Test execution tip` added to the `Readme.md` file

danielezonca (approved):
LGTM 👍

PR #4964, opened 2023-02-03
Link cach name to PersistedSessionOption.sessionId

nprentza:
changes in this PR:
- a PersistedSessionOption is created from a sessionId. If it's a new session we initiate sessionId to the working memory counter.
- SessionConfiguration updated to include the option PersistedSessionOption
- a cache is created as "cacheSession_sessionId"
- ReliabilityTest was revised to test the functionality of creating three sessions: first-session is new, second-session is created from the first-session, third-session is new. At the end of this test we expect to have only two cache folders on the disk, cacheSession_0 and cacheSession_2.

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/Person.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/Person.java
@@ -46,4 +46,32 @@ public class Person implements Serializable {
     public String toString() {
         return "Person [name=" + name + ", age=" + age + "]@" + Integer.toHexString(System.identityHashCode(this));
     }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + age;
+        result = prime * result + ((name == null) ? 0 : name.hashCode());
+        return result;
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


kie-api/src/main/java/org/kie/api/runtime/conf/PersistedSessionOption.java
```diff
@@ -43,6 +43,10 @@ public static PersistedSessionOption newSession() {
         return new PersistedSessionOption();
     }
 
+    public static PersistedSessionOption newSession(long sessionId) {
```
mariofusco: I don't understand what is the purpose of this third factory method for this option. I think that either you want a new empty persisted session or you want create a session from an already persisted one. When this should be used?

nprentza: When we create a new empty persisted session we want this to be linked to the sessionId of the session that will be created. If we don't pass this value to the constructor then a new persisted session will have sessionId=-1 while a new KieSession will have sessionId=kiebase.getWorkingMemoryCounter.

PR #4997, opened 2023-02-14
DROOLS-7337 Options and Configuration improvement

mdproctor:
The PR does the following
1) Splits a configuration into core, rule and flow. Anything on core is neither rule or flow specific. The split uses a composite and type lookup to make it appear as one. The Option delegates to the correct entry in the lookup. 
2) Introduces the KEY static property which provides additional typesafety and allow allows for switch/case statements instead of chained if statements.
3) Moved shared code into base classes so less code duplication.
4) Public API compatibility kept, although internal you cannot cast directly - you must use as() instead. This unwraps the correct implementation, although all implementations delegate the Option or Property call back to the composite if it is no the target.


mariofusco:
Replaced by https://github.com/kiegroup/drools/commit/cf1fa88a3bf93087ed3f936a385d580d2a0b623f

PR #5021, opened 2023-02-16
Person: hashCode, equals. ReliabilityTest: more tests

nprentza:
Changes in this PR:
- Person class : hashCode, equals added
- ReliabilityTest class:: 
	- try/finally{tearDown()} added to all test functions
	- new test method for testing fireAllRules before creating secondSession
	- new test method for testing fireAllRules with updates before creating secondSession

Remarks:
- Reliability.test() is now functioning properly, @Disable annotation removed.
- ReliabilityTest.testSessionFromCacheFireAllRules(): when we execute firstSession.fireAllRules() before creating secondSession  (from firstSession), then secondSession.fireAllRules() returns 0. I tried to change the ReliableRuntimeComponentFactoryImpl.updateFactsWithCache() with deleting/retracting the fact handle and re-inserting the object, but code execution throws exception. 
- In ReliabilityTest.testSessionFromCacheWithUpdates(): rule X also updates object $p. This results in two copies of the updated object in the cache.

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
@@ -37,7 +37,6 @@ public class ReliabilityTest {
         CacheManager.INSTANCE.removeCache("cacheSession_0");
     }
 
-    @Disabled("Still fails at the assertion because of cache replica")
     @Test
     public void test() {
         String drl =
@@ -49,24 +48,28 @@ public class ReliabilityTest {
                 "  System.out.println( $p.getAge() );\n" +
                 "end";
 
-        KieBase kbase = new KieHelper().addContent(drl, ResourceType.DRL).build();
-        KieSessionConfiguration conf = KieServices.get().newKieSessionConfiguration();
-        conf.setOption(PersistedSessionOption.newSession());
-        KieSession firstSession = kbase.newKieSession(conf, null);
+        try {
+            KieBase kbase = new KieHelper().addContent(drl, ResourceType.DRL).build();
+            KieSessionConfiguration conf = KieServices.get().newKieSessionConfiguration();
+            conf.setOption(PersistedSessionOption.newSession());
+            KieSession firstSession = kbase.newKieSession(conf, null);
 
-        long id = firstSession.getIdentifier();
+            long id = firstSession.getIdentifier();
 
-        firstSession.insert("M");
-        firstSession.insert(new Person("Mark", 37));
+            firstSession.insert("M");
+            firstSession.insert(new Person("Mark", 37));
 
-        KieSessionConfiguration conf2 = KieServices.get().newKieSessionConfiguration();
-        conf2.setOption(PersistedSessionOption.fromSession(id));
-        KieSession secondSession = kbase.newKieSession(conf2, null);
+            KieSessionConfiguration conf2 = KieServices.get().newKieSessionConfiguration();
+            conf2.setOption(PersistedSessionOption.fromSession(id));
+            KieSession secondSession = kbase.newKieSession(conf2, null);
 
-        secondSession.insert(new Person("Edson", 35));
-        secondSession.insert(new Person("Mario", 40));
+            secondSession.insert(new Person("Edson", 35));
+            secondSession.insert(new Person("Mario", 40));
 
-        assertThat(secondSession.fireAllRules()).isEqualTo(2);
+            assertThat(secondSession.fireAllRules()).isEqualTo(2);
+        } finally {
+            this.tearDown();
+        }
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
```diff
@@ -49,24 +48,28 @@ public void test() {
...
-        secondSession.insert(new Person("Edson", 35));
-        secondSession.insert(new Person("Mario", 40));
+            secondSession.insert(new Person("Edson", 35));
+            secondSession.insert(new Person("Mario", 40));
 
-        assertThat(secondSession.fireAllRules()).isEqualTo(2);
+            assertThat(secondSession.fireAllRules()).isEqualTo(2);
+        } finally {
+            this.tearDown();
+        }
```
tkobayas: Hi @nprentza , What I meant in the previous call was `ksession.dispose()` in finally block. Then if you will implement removing cache inside `dispose()` later, you would be able to make sure that the cache is removed. Sorry if confused. You don't need to call `tearDown()` in test methods. junit always calls `tearDown()` even if an Exception is thrown.

tkobayas: So the code would be like
```
        KieSession secondSession = kbase.newKieSession(conf2, null);
        try {
            secondSession.insert(new Person("Edson", 35));
            secondSession.insert(new Person("Mario", 40));

            assertThat(secondSession.fireAllRules()).isEqualTo(2);
        } finally {
            secondSession.dispose();
        }
```

We don't care about `firstSession` in these tests because we simulate that `firstSession` is lost.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
```diff
@@ -80,20 +83,30 @@ public void testReliableObjectStore() {
...
+            firstSession.insert(new Person("Nicole", 27));
+
+            assertThat(firstSession.fireAllRules()).isEqualTo(0);
+
+        } finally {
+            this.tearDown();
+        }
     }
 
+    @Disabled("It fails at the assertion, secondSession.fireAllRules() returns 0.")
```
tkobayas, line 108: Probably `ReliableRuntimeComponentFactoryImpl.updateFactsWithCache` is not enough to propagate the facts. I think it's okay to merge this test with `@Disabled`, and then we will investigate further.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
```diff
@@ -103,27 +116,73 @@ public void testSessionFromCache() {
...
-        KieSession secondSession = kbase.newKieSession(conf2, null);
+            secondSession.insert(new Person("John", 22));
+            secondSession.insert(new Person("Mary", 42));
 
-        System.out.println("secondSession getObjects.size (from firstSession) = " + secondSession.getObjects().size());
+            secondSession.fireAllRules();
 
-        secondSession.insert(new Person("John", 22));
-        secondSession.insert(new Person("Mary", 42));
+            assertThat(secondSession.fireAllRules()).isEqualTo(1);
```
tkobayas: @nprentza You call `fireAllRules` twice consecutively here. Is this intentional?

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
```diff
@@ -103,27 +116,73 @@ public void testSessionFromCache() {
...
+    public void testSessionFromCacheWithUpdates() {
+        String drl =
+                "import " + Person.class.getCanonicalName() + ";" +
+                        "rule X when\n" +
+                        "  $s: String()\n" +
+                        "  $p: Person( getName().startsWith($s) )\n" +
+                        "then\n" +
+                        "  System.out.println( $p.getAge() );\n" +
+                        "  $p.setName(\"-\");\n" +
+                       "end";
```
tkobayas: @nprentza You modify the fact, but don't call `update($p);`. Is it intentional?

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
nprentza: Oh got it, thanks Toshiya.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
nprentza: No, it's a mistake, one call of secondSession.fireAllRules() is needed.  Thanks.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
nprentza: No it's not intentional, I thought modifying the object property is enough .. So why do we need to call the update function?

tkobayas: If you want to re-evaluate rules based on the modified facts, you need to use `modify` or `update` to notify the change to the engine.

FYI:  https://docs.drools.org/7.73.0.Final/drools-docs/html_single/#drl-rules-THEN-methods-ref_drl-rules

tkobayas: Btw, I noticed that the section is missing in the Drools 8 docs. Filed a JIRA to bring it back.
https://issues.redhat.com/browse/DROOLS-7344

nprentza: Right, I have read about this before. For the particular test-case I think we just want the secondSession to use the updated version of the object $p. I modified the drl to use the update function but the result of test is the same.

tkobayas: > we just want the secondSession to use the updated version of the object $p

Do you mean you want to test without re-evaluation (= `update($p)`)? Probably it's an irregular usage, but maybe worth testing. Please add a comment in the test method to describe the intention. Thanks!

> I modified the drl to use the update function but the result of test is the same.

If you want to add `update($p)`, it would be also okay. It means a test case of re-evaluation. We can merge the test with `@Disable` and investigate later.

tkobayas (approved):
Thanks!

mariofusco (approved):

PR #5037, opened 2023-02-27
first reliable propagation list implementation

mariofusco:
With this PR I'm implementing a first version of a reliable propagation list. In this way it is no longer necessary to repropagate the objects from the object store into the persisted session created from an existing one (and I commented out the invocation of that method). However consider that now the existing test case works only because when the first session is replaced we never called a fireAllRules on it yet and then all the inserted objects are still in the propagation queue. I expect that the test fails if you issue a first fireAllRules before moving to the second session. I will add one more test for this scenarion and this will be fixed when also the other necessary data structures will be persisted.

Just a couple of notes:

1. I found that normally the session id is only unique per kbase, but we need an universally unique id, that's why I reimplemented the generation in the reliable components.
2. At the moment I'm putting in the infinispan cache the whole propagation queue which is a modifiable object and just letting infinispan to deal with the changes on it. I'm not sure if this is a correct way to use infinispan though.

### DIFF
--- a/drools-reliability/src/main/java/org/drools/reliability/ReliableRuntimeComponentFactoryImpl.java
+++ b/drools-reliability/src/main/java/org/drools/reliability/ReliableRuntimeComponentFactoryImpl.java
@@ -15,7 +15,10 @@
 
 package org.drools.reliability;
 
+import java.util.concurrent.atomic.AtomicLong;
+
 import org.drools.core.SessionConfiguration;
+import org.drools.core.common.AgendaFactory;
 import org.drools.core.common.EntryPointFactory;
 import org.drools.core.common.InternalWorkingMemory;
 import org.drools.core.impl.RuleBase;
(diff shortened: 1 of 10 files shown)


drools-reliability/src/main/java/org/drools/reliability/ReliableRuntimeComponentFactoryImpl.java
```diff
@@ -28,6 +31,10 @@ public class ReliableRuntimeComponentFactoryImpl extends RuntimeComponentFactory
 
     public static final ReliableRuntimeComponentFactoryImpl DEFAULT = new ReliableRuntimeComponentFactoryImpl();
 
+    private static final AtomicLong RELIABLE_SESSIONS_COUNTER = new AtomicLong(0);
```
tkobayas, line 34: 〉I found that normally the session id is only unique per kbase, but we need an universally unique id, that's why I reimplemented the generation in the reliable components.

I understand the point. Btw, in case of a process down scenario, `RELIABLE_SESSIONS_COUNTER` could be also lost. So we need to store it in Infinispan cache as well? Or it can be determined by iterating existing Cache key names. Whichever, It can be implemented by a separated PR.

tkobayas:
> At the moment I'm putting in the infinispan cache the whole propagation queue which is a modifiable object and just letting infinispan to deal with the changes on it. I'm not sure if this is a correct way to use infinispan though.

When we use `SoftIndexFileStore`, the cache object is retrieved from memory, so we can use the same modified object. However, the modification is not persisted. If the process crashed and the new process retrieves the object from the cache through the FileStore, the object is in the state of the time when it was "put". So we need to call `put` every time the object changes.

https://infinispan.org/docs/stable/titles/configuring/configuring.html#cache-modes_creating-embedded-caches

```
When updating cached values, please use the standard put operation
```

We would need to change the test cases to reflect the fail-over situation more precisely (e.g. `CacheManager` is recreated). It can be done in a separated PR.

tkobayas (approved):

PR #5057, opened 2023-03-05
ReliableObjectStore with separate cache.

nprentza:
changes in this PR:

### (a) ReliableObjectStore with separate cache: 
- private Map<InternalFactHandle, Object> cache; added to ReliableObjectStore class
- MapObjectStore.fhMap is instantiated as  new IdentityHashMap<>();
- a FactHandle is added to both fhMap and cache.

### (b) ReliableNamedEntryPoint initObjectStore: 
- If cache is not empty, then FactHandle objects from cache are inserted into the new session

### (c) ReliableRuntimeComponentFactoryImpl internalInitSession:
- If session contains facts (from cache) then we call fireAllRules with an AgendaFilter that disables rule match.

### (d) ReliabilityTest.separateCache_insert_update_fire_insert_fire:
- tests the scenario : 
create firstSession >> insert facts >> fireAllRules >> create secondSession from firstSession >> insert facts >> fireAllRules
- it also tests updating a fact already in a session to avoid the creation of extra entries in the cache on updates

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
@@ -1,0 +1,178 @@
+package org.drools.reliability;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.kie.api.KieBase;
+import org.kie.api.KieServices;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.KieSessionConfiguration;
+import org.kie.api.runtime.conf.PersistedSessionOption;
+import org.kie.api.runtime.rule.FactHandle;
+import org.kie.internal.utils.KieHelper;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.drools.reliability.ReliabilityTestUtils.failover;
+
+@ExtendWith(BeforeAllMethodExtension.class)
+public class StoresOnlyStrategyTest {
+
+    private static final String BASIC_RULE =
+            "import " + Person.class.getCanonicalName() + ";" +
+                    "global java.util.List results;" +
+                    "rule X when\n" +
+                    "  $s: String()\n" +
+                    "  $p: Person( getName().startsWith($s) )\n" +
+                    "then\n" +
+                    "  results.add( $p.getAge() );\n" +
+                    "end";
+
+    @AfterEach
+    public void tearDown() {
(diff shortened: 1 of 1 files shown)


drools-reliability/src/main/java/org/drools/reliability/ReliableObjectStore.java
```diff
@@ -19,9 +19,39 @@
...
+import java.util.Map;
+
 public class ReliableObjectStore extends MapObjectStore {
 
-    public ReliableObjectStore(Cache<Object, InternalFactHandle> fhCache) {
-        super(fhCache);
+    private Map<InternalFactHandle, Object> cache;
+
+    public ReliableObjectStore(Cache<InternalFactHandle,Object> fhCache) {
+        super();
```
tkobayas: If you call `super(new IdentityHashMap<>())` instead of `super()`, you don't need to change `MapObjectStore`, right? If you can scope changes within only `drools-reliability`, generally, it would be greater.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTest.java
```diff
@@ -184,4 +185,37 @@ void updateInRHS_insert_fire_failover_insert_fire() {
...
+        KieSession firstSession = getKieSession(BASIC_RULE, PersistedSessionOption.newSession());
+        long id;
+        id = firstSession.getIdentifier();
+        try {
+            // 1st round
+            firstSession.insert("L");
+            Person pMark = new Person("Mark", 37);
+            FactHandle fhMark = firstSession.insert(pMark);
+            firstSession.insert(new Person("Liza", 42));
+            pMark.setAge(27);
```
tkobayas: I guess you want to add `firstSession.update(fhMark, pMark);` here, don't you? (Or you wanted to test updating a fact without `update()` call?) I confirmed that the test is still successful after the adding the update call.

> it also tests updating a fact already in a session to avoid the creation of extra entries in the cache on updates

Ah, I see that it's intentional. Then it's okay to leave it as you wrote.

tkobayas:
This PR uses `DefaultFactHandle` as a Cache key, so we can remove `equals` and `hashCode` from `Person` class, right? I confirmed that the test is successful after removing these methods.

tkobayas:
`DefaultFactHandle` could be serialized when putting into Infinispan Cache so there is serialization cost. We may use `DefaultFactHandle.id` as a key instead so it would improve the performance? We can do it in the later stage where we prepare some benchmark. Filed a JIRA  https://issues.redhat.com/browse/DROOLS-7361

drools-reliability/src/main/java/org/drools/reliability/ReliableObjectStore.java
nprentza: Yes,  you are right, that way we don't need to change MabObjectStore.

PR #5065, opened 2023-03-06
first implementation of persisted stores only strategy

mariofusco:
This pull request is a first implemetation of lightweight reliable session that only relies on persistance of object store. It includes:

- Fix to serialization of `FactHandle`s.
- A mechanism to [listen when a working memory action is performed](https://github.com/kiegroup/drools/pull/5065/files#diff-f120d1c752fce374911be2ae53336d91edfad7132e83535cd654445a1d158317R151). This is necessary to [refresh the cache](https://github.com/kiegroup/drools/pull/5065/files#diff-0ddb008ecafa6fd81b44ca5e3ea5e1372a21dca06f80b761d102d66d43d78c9bR76) when a `FactHandle` is modified. At the moment this is acting only on the Insert action, but it will need to be extended to many other actions.
- A mechanism to [flush the objects from the persisted object stores](https://github.com/kiegroup/drools/pull/5065/files#diff-0ddb008ecafa6fd81b44ca5e3ea5e1372a21dca06f80b761d102d66d43d78c9bR89) to the new empty session. It distinguishes between facts that have been already propagated, that shouldn't cause the firing of a consequence and others not yet propagated, for which the consequence should fire at the first explicit `fireAllRules` invocation.

The following is instead still outstanding:

- A mechanism to also persist and restore `global`s.
- Something also dealing with timers (more in general we are not considering CEP yet and I'm afraid that this will be tricky).
- A lighter persisted object store specific for this use case: for this scenario it is not necessary to persist the whole `FactHandle`s, also because [they are all discarded](https://github.com/kiegroup/drools/pull/5065/files#diff-0ddb008ecafa6fd81b44ca5e3ea5e1372a21dca06f80b761d102d66d43d78c9bR94) and recreated. It will be enough to keep track of only the objects present in a store in a given moment.

As first improvement I will concentrate on this last point (light persisted object store) with a subsequent pull request.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/SessionConfiguration.java
+++ b/drools-core/src/main/java/org/drools/core/SessionConfiguration.java
@@ -202,6 +202,10 @@ public class SessionConfiguration extends BaseConfiguration<KieSessionOption, Si
         return this.keepReference;
     }
 
+    public boolean hasPersistedSessionOption() {
+        return this.persistedSessionOption != null;
+    }
+
     public PersistedSessionOption getPersistedSessionOption() {
         return this.persistedSessionOption;
     }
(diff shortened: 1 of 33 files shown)


tkobayas (approved):
Thanks!

PR #5069, opened 2023-03-07
simple reliable object store

mariofusco:
This is a first implementation of the lighter `ObjectStore` that we discussed. In all honesty I'm still not convinced that this is a good idea so please give a look, feedback are welcome.

@tkobayas in all cases we need to try this with more use cases, so please let me know when you will have those tests ready.


tkobayas (approved):
LGTM, thanks!

tkobayas:
> @tkobayas in all cases we need to try this with more use cases, so please let me know when you will have those tests ready.

Sure, I'll raise a PR for test cases.

nprentza (approved):
I think I understand the idea of the SimpleReliableObjectStore, need to run some test cases to see how it works in action.

PR #5070, opened 2023-03-08
[DROOLS-7359] Improve ReliabilityTest

tkobayas:
- Parameterized test
- Add CacheManagerTest
- Refactored for readability

Not adding  new tests.

**JIRA**: 
https://issues.redhat.com/browse/DROOLS-7359

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

- for <b>pull request checks</b>  
  Please add comment: <b>Jenkins retest this</b>

- for a <b>specific pull request check</b>  
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] tests</b>

- for a <b>full downstream build</b> 
  - for <b>jenkins</b> job: please add comment: <b>Jenkins run fdb</b>
  - for <b>github actions</b> job: add the label `run_fdb`

- <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

- <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

- <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>

- for <b>quarkus branch checks</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins run quarkus-branch</b>

- for a <b>quarkus branch specific check</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins (re)run [drools|kogito-runtimes|kogito-apps|kogito-examples] quarkus-branch</b>

- for <b>quarkus main checks</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins run quarkus-main</b>

- for a <b>specific quarkus main check</b>  
[...]

PR #5074, opened 2023-03-09
SimpleReliableStore tests - basic fact updates

nprentza:
The PR contains a new set of tests for testing fact updates before / after propagation, and updates of facts already propagated.

(a)  fireInsertUpdateFailover_CacheImmutableKey():
- a new fact is inserted & updated after facts propagation (fireAllRules).
- test run with success.
(b) insertUpdateFireFailover_CacheImmutableKey():
- a new fact is inserted & updated before facts propagation (fireAllRules).  
- test run with success.
(c) insertFireUpdateFailover_RePropagateUpdates():
- a fact is inserted, propagated (fireAllRules) and then updated (before failover simulation).
- the fact is populated in the new session in its updated version, but its handle.hasMatches() gives true. 
- in the next propagation of facts (secondSession.fireAllRules) the updated fact is not propagated.
- test failed.

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
@@ -1,0 +1,178 @@
+package org.drools.reliability;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.kie.api.KieBase;
+import org.kie.api.KieServices;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.KieSessionConfiguration;
+import org.kie.api.runtime.conf.PersistedSessionOption;
+import org.kie.api.runtime.rule.FactHandle;
+import org.kie.internal.utils.KieHelper;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


tkobayas:
Thank you for the PR, @nprentza 

Firstly, when you raise a PR, please use a branch name different from `d7247r2` for example, `d7247r2_fact_update_tests`. Actually, `d7247r2` in drools repo has diverged from your repo, so applyig the PR locally using `gh` had a small problem.

Secondly, test methods in `StoresOnlyStrategyTest` can be added to `ReliabilityTest` with these annotations instead of adding a new class (I can understand that at the time when you wrote the test, my PR was not yet merged...).

```
    @ParameterizedTest
    @MethodSource("strategyProviderStoresOnly")
```

You would also need some modifications to the test methods (e.g. have `PersistedSessionOption.Strategy strategy` as an argument, etc.), but I think you can understand by looking at other test methods.

nprentza:
thank you @tkobayas for your comments. 

Just to clarify that with your first comment you mean using a different brach name for each PR. Is that correct?

I created this new class for tests temporarily because I was aware that your were working on some changes on the ReliabilityTest class. I will of course move these tests properly into the appropriate class when available.

tkobayas:
> you mean using a different brach name for each PR. Is that correct?

Yes, for example, Mario does like `d7247r2_light_store` : https://github.com/kiegroup/drools/pull/5069

I prefer using JIRA ID like `DROOLS-7359` : https://github.com/kiegroup/drools/pull/5070

> I will of course move these tests properly into the appropriate class when available.

Thanks!

nprentza:
PR to be replaced by https://github.com/kiegroup/drools/pull/5077

PR #5077, opened 2023-03-09
Drools 7362 [SimpleReliableStore: test fact updates]

nprentza:
Testing SimpleReliableObjectStore when creating a new session from a persisted store that contains fact updates not processed (propagated).

The PR also contains two basic tests to ensure that infinispan cache immutable key limitation is not causing any issues (insertUpdateFireFailover_CacheImmutableKey(), fireInsertUpdateFailover_CacheImmutableKey).

With regards to fact updates and creating a session from a persisted store,
I noticed that depending on how we update facts, that we have previously inserted into the wm, affects the way a fact is populated into a new session (fireUpdateFactByFactHandleFailover(), fireUpdateFactWithNewObjectFailover()).
Please see jira  [DROOLS-7362](https://issues.redhat.com/browse/DROOLS-7362).

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
@@ -1,6 +1,8 @@
 package org.drools.reliability;
 
+import org.drools.core.ClassObjectFilter;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.kie.api.KieBase;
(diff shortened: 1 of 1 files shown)


mariofusco (approved):

tkobayas:
I notice the error when I run `mvn test`. It doesn't affect the test results though.
```
10:51:06.071 [blocking-thread--p13-t3] ERROR o.i.persistence.sifs.Compactor.lambda$schedule$2:229 - ISPN029021: File id 0 encountered an exception while compacting, file may be orphaned
java.lang.NullPointerException: No index info found for key: M
```
I can fix it with `BeforeAllMethodExtension`, so will do in my next PR. -> https://github.com/kiegroup/drools/pull/5085

tkobayas (approved):
Thank you for the test coverage, @nprentza 

I leave some comments, but they could be done in a separate PR.

tkobayas:
https://github.com/kiegroup/drools/pull/5074#issuecomment-1461484408
> > I will of course move these tests properly into the appropriate class when available.
>
> Thanks!

Note that my PR is already merged in `d7247r2` so you can move the tests. Anyway, moving can be done in a separate PR.

drools-reliability/src/test/java/org/drools/reliability/StoresOnlyStrategyTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+            Person pMark = (Person) firstSession.getObject(fhMark);
+            pMark.setName("Mark");
+            firstSession.update(fhMark, pMark);
+
+            firstSession.getObjects( new ClassObjectFilter( Person.class ) ).forEach(p ->{
+                if (((Person) p).getAge()==47){
+                    assertThat(((Person) p).getName().equals("Mark")).isTrue();
+                }
+            });
```
tkobayas, line 78: As you often use this logic, extracting a private method like this could be useful. It can be done in a separate PR.
```
    private Optional<Person> getPersonByAge(KieSession kieSession, int age) {
        return kieSession.getObjects(new ClassObjectFilter(Person.class))
                .stream()
                .map(Person.class::cast)
                .filter(p -> p.getAge() == age)
                .findFirst();
    }
```

PR #5093, opened 2023-03-14
stores only reliable session to support updates and deletes

mariofusco:
This pull request makes the stores only strategy to also support updates and deletes. The main problem that was necessary to solve to do so was having an immutable and stable key for the persisted cache, instead of using directly the users domain objects that are generally mutable. I figured out that I didn't need to introduce another map to do so but I could reuse the `IdentityMap` that the `SimpleReliableObjectStore` inherits from the `IdentityObjectStore` and use as key for the persisted cache the `FactHandle`'s id.

### DIFF
--- a/drools-reliability/src/main/java/org/drools/reliability/ReliableSessionInitializer.java
+++ b/drools-reliability/src/main/java/org/drools/reliability/ReliableSessionInitializer.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


drools-reliability/src/main/java/org/drools/reliability/ReliableSessionInitializer.java
```diff
@@ -0,0 +1,108 @@
...
+    interface SessionInitializer {
+        InternalWorkingMemory init(InternalWorkingMemory session);
+    }
+
+    private static class StoresOnlySessionInitializer implements SessionInitializer {
+
+        @Override
+        public InternalWorkingMemory init(InternalWorkingMemory session) {
+            // re-propagate objects from the cache to the new session
+            populateSessionFromCache(session);
```
tkobayas: `populateSessionFromCache` can be skipped when `persistedSessionOption.isNewSession() == true`? This can be done in a separate PR.

tkobayas (approved):
Thank you, @mariofusco !

tkobayas:
Btw, I confirmed that after removing `equals`/`hashCode` from `Person`, tests still pass. Probably we will go without `equals`/`hashCode`as we have overcome the Cache key issue?

drools-reliability/src/main/java/org/drools/reliability/ReliableSessionInitializer.java
mariofusco: @tkobayas Done, thanks a lot for suggestion.

PR #5097, opened 2023-03-15
[DROOLS-7363] demonstrate SyntheticRuleUnit reuse

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7363

@tarilabs With this PR I'm just adding a test case demonstrating how to reuse the definition of a synthetic rule unit. In this case I don't think that it is possible to create multiple `RuleUnitInstance` from the same `RuleUnit`, because the `DataSource`s are inherently bound to the `RuleUnit` itself. Also note that this is consistent with what happens for non-synthetic `RuleUnit`s defined through the Java DSL, see https://github.com/kiegroup/drools/blob/f41aa536dd47263fb83b737153122739fc074438/drools-ruleunits/drools-ruleunits-dsl/src/test/java/org/drools/ruleunits/dsl/RuleUnitsTest.java#L276

Please let me know if this is ok for your use case and if you have any suggestion to improve this implementation.

### DIFF
--- a/drools-ruleunits/drools-ruleunits-dsl/src/test/java/org/drools/ruleunits/dsl/SyntheticRuleUnitsTest.java
+++ b/drools-ruleunits/drools-ruleunits-dsl/src/test/java/org/drools/ruleunits/dsl/SyntheticRuleUnitsTest.java
@@ -107,6 +107,32 @@ public class SyntheticRuleUnitsTest {
         unitInstance.close();
     }
 
+    @Test
+    public void syntheticRuleUnitDefinitionReuse() {
+        // DROOLS-7363
+        SyntheticRuleUnit unit = createSyntheticRuleUnit();
+        RuleUnitInstance<SyntheticRuleUnit> unitInstance = RuleUnitProvider.get().createRuleUnitInstance(unit);
+
+        unit.getDataStore("strings", String.class).add("Hello World");
+
+        assertThat(unitInstance.fire()).isEqualTo(2);
+        assertThat(unit.getGlobal("results", List.class)).containsExactlyInAnyOrder("it worked!", "it also worked with HELLO WORLD");
+
+        SyntheticRuleUnit unit2 = createSyntheticRuleUnit();
+        RuleUnitInstance<SyntheticRuleUnit> unitInstance2 = RuleUnitProvider.get().createRuleUnitInstance(unit2);
+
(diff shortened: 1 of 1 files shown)


drools-ruleunits/drools-ruleunits-dsl/src/test/java/org/drools/ruleunits/dsl/SyntheticRuleUnitsTest.java
```diff
@@ -107,6 +107,32 @@ public void testSeparatedUnitCreation() {
...
+        // DROOLS-7363
+        SyntheticRuleUnit unit = createSyntheticRuleUnit();
+        RuleUnitInstance<SyntheticRuleUnit> unitInstance = RuleUnitProvider.get().createRuleUnitInstance(unit);
+
+        unit.getDataStore("strings", String.class).add("Hello World");
+
+        assertThat(unitInstance.fire()).isEqualTo(2);
+        assertThat(unit.getGlobal("results", List.class)).containsExactlyInAnyOrder("it worked!", "it also worked with HELLO WORLD");
+
+        SyntheticRuleUnit unit2 = createSyntheticRuleUnit();
```
tarilabs, line 121: unfortunately I believe the problem is here, once I have defined a `SyntheticRuleUnit`, I would like to re-use it, not re-define it again by calling `createSyntheticRuleUnit()` additional times.

mariofusco: I see your point, but however, as I showed in the other linked test case, this is exactly the same thing that you so also using the plain Java DSL. I will check if I can split the rule unit definition and creation part. This means that you could define via DSL a `SyntheticRuleUnitDefinition` that you could reuse to instance multiple rule units out of it. Would this be better?

tarilabs: > Would this be better?

I believe worthy to discuss in a dedicated meeting, than in the constraints of a PR textfield 🙃

tarilabs (approved):
discussed offline and the lambda of the rulefactory is cached internally by the ruleunitprovider, so I could invoke the syntethicgeneration twice and the second time I could even pass it a different lambda that throws so just to be sure the cache behaviour hasn't changed.

PR #5014, opened 2023-02-16
[8.34.x] Update project version to 8.34.1-SNAPSHOT

kie-ci4:
Generated by build jenkins-KIE-drools-8.34.x-release-drools-promote-1: https://eng-jenkins-csb-business-automation.apps.ocp-c1.prod.psi.redhat.com/job/KIE/job/drools/job/8.34.x/job/release/job/drools-promote/1/.
Please do not merge, it should be merged automatically.

PR #5101, opened 2023-03-16
DROOLS-7363 refactor tests

nprentza:
the PR introduces a new class for testing with a 2-rule drl, the basic rule plus one that updates a fact under conditions.
the PR also suggests the introduction of the abstract class ReliabilityTestBasics for the common functions the various test classes will share.

### DIFF
--- a/drools-reliability/src/test/java/org/drools/reliability/ReliabilityTestBasics.java
+++ b/drools-reliability/src/test/java/org/drools/reliability/ReliabilityTestBasics.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.reliability;
+
+import org.drools.core.ClassObjectFilter;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.kie.api.KieBase;
+import org.kie.api.KieServices;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.KieSessionConfiguration;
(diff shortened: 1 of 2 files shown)


drools-reliability/src/test/java/org/drools/reliability/ReliabilityTestBasics.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    static Stream<PersistedSessionOption.Strategy> strategyProviderStoresOnly() {
+        return Stream.of(PersistedSessionOption.Strategy.STORES_ONLY);
+    }
+
+    @AfterEach
+    public void tearDown() {
+        // We can remove this when we implement ReliableSession.dispose() to call CacheManager.removeCachesBySessionId(id)
+        CacheManager.INSTANCE.removeAllSessionCaches();
```
mariofusco, line 52: I think that this method should also call `failover()` or at least reset the counter of sessions id. If you don't do so I experienced a coupling between different tests, succeeding in isolation but failing when launched the whole test class.
@tkobayas can you please provide your feedback about this?

tkobayas: @mariofusco Ahh, okay. The root problem is `savedSessionId`. It is changed every time `ReliabilityTest.getKieSession` is called, so even if a test starts with sessionId = 1, after `failover` + `restoreSession`, `savedSessionId` becomes 0. It broke the tests. I raised a fix PR.

https://github.com/kiegroup/drools/pull/5104

The reason why `failover` in `tearDown` solved the issue was it keeps tests start with sessionId = 0, so the problem wouldn't arise (same for single method test running).

with the PR fix, we no longer need  `failover` in `tearDown`.

drools-reliability/src/test/java/org/drools/reliability/ReliabilityTestBasics.java
```diff
@@ -0,0 +1,111 @@
...
+    protected KieSession restoreSession(String drl, PersistedSessionOption.Strategy strategy) {
+        return getKieSession(drl, PersistedSessionOption.fromSession(savedSessionId, strategy));
+    }
+
+    protected KieSession getKieSession(String drl, PersistedSessionOption option) {
+        KieBase kbase = new KieHelper().addContent(drl, ResourceType.DRL).build();
+        KieSessionConfiguration conf = KieServices.get().newKieSessionConfiguration();
+        conf.setOption(option);
+        session = kbase.newKieSession(conf, null);
+        savedSessionId = session.getIdentifier();
```
tkobayas: @nprentza Similar to my PR https://github.com/kiegroup/drools/pull/5104 , please let `savedSessionId` change only in `createSession`.

nprentza: @tkobayas I have applied the changes of PR #5014 into ReliabilityTestBasics. I will issue now a new PR.

PR #5114, opened 2023-03-22
DROOLS-7369: refactor ReliabilityTest to extend ReliabilityTestBasics

nprentza:
refactor ReliabilityTest to extend ReliabilityTestBasics class so that all ReliabilityTest* classes will use the same supporting methods (insertString, etc).


tkobayas (approved):
LGTM, Thanks!

pibizza (approved):
Looks good to me.

mariofusco (approved):

PR #5116, opened 2023-03-22
create a shaded module containing both javax and jakarta inject API

mariofusco:


mariofusco:
We decide to avoid introducing a new module only to wrap the inject dependencies.

PR #5127, opened 2023-03-27
[DROOLS-7377] enforce kie-ci to always depend on javax.inject

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7377

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -367,6 +367,12 @@
       </dependency>
 
       <dependency>
+        <groupId>javax.inject</groupId>
+        <artifactId>javax.inject</artifactId>
+        <version>1</version>
+      </dependency>
+
+      <dependency>
         <groupId>info.picocli</groupId>
         <artifactId>picocli</artifactId>
         <version>${version.info.picocli}</version>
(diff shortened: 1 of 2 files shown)


build-parent/pom.xml
```diff
@@ -1360,6 +1366,8 @@
                 <rules>
                   <banDuplicateClasses>
                     <ignoreClasses>
+                      <!-- Duplication of javax.inject classes is necessary to also support migration to jakarta namespace. -->
```
gitgabrio, line 1369: hi @mariofusco Thanks for PR. Is it needed to have this here ? Does this duplication happen on different modules ?

gitgabrio (approved):
HI @mariofusco thanks for PR. Just a question, anyway LGTM

danielezonca (approved):

build-parent/pom.xml
mariofusco: @gitgabrio Unfortunately this is not possible because the ban exclusion is not transitive and then other modules depending on kie-ci will make this enforcer to fail.

PR #5141, opened 2023-03-30
improve globals for reliable session + minor refactor

mariofusco:
Related PRs:

https://github.com/kiegroup/drools/pull/5141
https://github.com/kiegroup/kogito-runtimes/pull/2912

### DIFF
--- a/drools-kiesession/src/main/java/org/drools/kiesession/session/StatefulKnowledgeSessionImpl.java
+++ b/drools-kiesession/src/main/java/org/drools/kiesession/session/StatefulKnowledgeSessionImpl.java
@@ -618,7 +618,6 @@ public class StatefulKnowledgeSessionImpl extends AbstractRuntime
 
         try {
             if (!calledFromRHS) {
-                startOperation();
                 this.lock.lock();
             }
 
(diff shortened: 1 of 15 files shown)


drools-kiesession/src/main/java/org/drools/kiesession/session/StatefulKnowledgeSessionImpl.java
```diff
@@ -1578,17 +1567,8 @@ public long getTimeToNextJob() {
...
-                startOperation();
-            }
-            agenda.addPropagation( propagationEntry );
-        } finally {
-            if (register) {
-                endOperation();
-            }
-        }
+    public void addPropagation(PropagationEntry propagationEntry) {
+        agenda.addPropagation( propagationEntry );
```
tkobayas, line 1571: Hi @mariofusco Am I correct to understand that we no longer need `register` which was used with `ExpireJob` to trigger `startOperation`/`endOperation` to sync `StatefulKnowledgeSessionImpl.lastIdleTimestamp` with `timerService`, so we don't consider that a job expiration is "busy"? (and `getIdleTime` is not important so removed in `CepEspTest`?)

tkobayas (approved):

drools-kiesession/src/main/java/org/drools/kiesession/session/StatefulKnowledgeSessionImpl.java
mariofusco: Yes, that's no longer relevant (not sure if it has ever been tbh, just a useless overcomplication).

PR #5222, opened 2023-05-09
relational operators for BD

quadrik:
Hi,

imho the relational operators , like `<`, should be translated to `compareTo` in the context of BigDecimals. Moreover wouldn't it be better to also translate `==` to `compareTo` instead of `equals` ? Because otherwise `new BigDecimal("10.0") != new BigDecimal("10.00")` would be true.

Regards

### DIFF
--- a/drools-model/drools-mvel-compiler/src/test/java/org/drools/mvelcompiler/MvelCompilerTest.java
+++ b/drools-model/drools-mvel-compiler/src/test/java/org/drools/mvelcompiler/MvelCompilerTest.java
@@ -611,6 +611,66 @@ public class MvelCompilerTest implements CompilerTest {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


mariofusco:
@tkobayas can you please give a look? I believe you already addressed this in a recent commit?

tkobayas:
Sure, I'll have a look.

tkobayas:
BigDecimal relational operators in mvel LHS is already implemented using `EvaluationUtil`, but they are not yet implemented in mvel RHS (and it's also not affected by my recent fix in DROOLS-7414). Thank you for the test cases, @quadrik 

I filed a JIRA and will work on it.

https://issues.redhat.com/browse/DROOLS-7440

tkobayas:
Replaced with https://github.com/kiegroup/drools/pull/5226 , thanks!

PR #5259, opened 2023-05-30
DROOLS-7453: Add Sonarcloud nightly

radtriste:
https://issues.redhat.com/browse/DROOLS-7453


danielezonca (approved):
Approving but I don't understand why it has been disabled, I hope the configuration in the repo is valid so fine for me to proceed

PR #5120, opened 2023-03-22
[DROOLS-7371] remove leftover usages of javax.inject annotations

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7371

PR #5263, opened 2023-05-31
[DROOLS-7371] Restoring KieActivator 

yesamer:
Restoring the KieActivator annotation, requested for [Import Java Class functionality](https://blog.kie.org/2022/05/dmn-types-from-java-classes.html) in kie-tools and removed in this PR https://github.com/kiegroup/drools/pull/5120
`Qualifier` annotation has been removed

### DIFF
--- a/kie-api/src/main/java/org/kie/api/project/KieActivator.java
+++ b/kie-api/src/main/java/org/kie/api/project/KieActivator.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


gitgabrio (changes requested):
HI @yesamer 
I understand the needs of this interface, but if its possible usage is only inside kie-tools, I think it would be better to implement it in that repository.
If it may be used by all `kogito` projects, then it would be better to have it in kogito-runtimes.
Is there any dopcumentation/blog describe its usage for a plain (i.e. no kogito/vscode) drools project ?
I think we should try to make a clear separation between the core of drools itself (that should be in this repository) and all possible extensions: wdyt ?

yesamer:
@gitgabrio @mariofusco @danielezonca I'm going to manage this on kie-tools side, in order to better decouple the feature architecture. We can close here. Thanks!

PR #5330, opened 2023-06-19
rework drools-reliability tests to make them work seamlessly with one or more sessions

mariofusco:
I tried to avoid the need of different methods for the cases when you're using one or more sessions. If you're using only one you can use all the tests as before and the base class will assume to be working with the only session available. Conversely when you're using more than one session you need to always pass the session on which you want to perform a specific action. The only exception to this is in the failover method because we assume that all sessions are running on the same cloud node, so when there is a failure on that node all the sessions in execution on it are to be considered lost at the same time.


nprentza (approved):
I thought about overloading ReliabilityTestBasics methods with KieSession as additional parameter but wasn't sure if its a good idea, that's why I decided to use sessionId instead. I remember in the past that we tried to avoid using large objects as kiesession as parameters, but I guess for testing purposes there should be no problem.

PR #5381, opened 2023-07-10
[DROOLS-7488] tests - update rule action method in DRL

nprentza:
**JIRA**: [DROOLS-7488](https://issues.redhat.com/browse/DROOLS-7488)

### DIFF
--- a/drools-reliability/drools-reliability-infinispan/src/test/java/org/drools/reliability/infinispan/ReliabilityUpdateInDrlTest.java
+++ b/drools-reliability/drools-reliability-infinispan/src/test/java/org/drools/reliability/infinispan/ReliabilityUpdateInDrlTest.java
@@ -15,11 +15,15 @@
 
 package org.drools.reliability.infinispan;
 
-import org.test.domain.Person;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.kie.api.runtime.KieSession;
 import org.kie.api.runtime.conf.PersistedSessionOption;
+import org.kie.api.runtime.rule.FactHandle;
+import org.test.domain.Person;
+
+import java.util.Optional;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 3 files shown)


tkobayas (approved):
Thanks, good coverage.

drools-reliability/drools-reliability-infinispan/src/test/java/org/drools/reliability/infinispan/ReliabilityUpdateInDrlTest.java
```diff
@@ -44,37 +48,215 @@ class ReliabilityUpdateInDrlTest extends ReliabilityTestBasics {
...
+
+        failover();
+
+        session1 = restoreSession(session1.getIdentifier(), RULE_UPDATE, persistenceStrategy,safepointStrategy);
+        clearResults(session1);
+        session2 = restoreSession(session2.getIdentifier(), RULE_UPDATE, persistenceStrategy,safepointStrategy);
+        clearResults(session2);
+
+        Optional<FactHandle> getFactHandleForPerson = getFactHandle(session2, pEleven);
+        if (!getFactHandleForPerson.isEmpty()){
```
mariofusco: If I'm not missing anything we always expect this `Optional` not to be empty, right? In case this is the expectation, the test should reflect it and fail with an empty `Optional` instead of simply silencing the problem as it does now.

mariofusco (changes requested):
Good job, please only review my comment and change the test accordingly in case I'm right.

drools-reliability/drools-reliability-infinispan/src/test/java/org/drools/reliability/infinispan/ReliabilityUpdateInDrlTest.java
nprentza: Yes, we expect that the Optional FactHandle will not be empty. Thanks for the suggestion, I will replace if-statement with an assertion so that the test will fail if the FactHandle is empty.

nprentza:
> Good job, please only review my comment and change the test accordingly in case I'm right.

Thanks Mario, I changed the test and updated the PR.

mariofusco (approved):

PR #5440, opened 2023-08-03
[DROOLS-7523] allow parallel evaluation of LHS part of rules with sequential firing of consequences

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7523

This pull request is still a draft. The idea is to introduce a third option for parallel execution, something in the middle between the fully sequential (not leveraging in any way multicore CPUs) and fully parallel (not allowing any control mechanism in rules firing like salience and agenda-group) that we already have. I called it parallel evaluation, because the goal is to allow parallel evaluation of the LHS part of rules, while keeping the firing of consequences (execution) sequential and consistent with our conflict resolution strategy.


kie-internal/src/main/java/org/kie/internal/conf/MultithreadEvaluationOption.java
```diff
@@ -1,58 +0,0 @@
...
-import org.kie.api.conf.SingleValueRuleBaseOption;
-
-/**
- * An Enum for multithread evaluation option.
- *
- * drools.multithreadEvaluation = &lt;true|false&gt;
- *
- * DEFAULT = false
- */
-public enum MultithreadEvaluationOption implements SingleValueRuleBaseOption {
```
mariofusco, line 29: Since this option still belongs to our internal API I removed it and replaced with the 3-values `ParallelExecutionOption`. If required (but I don't think so) I could keep this option, flag it as deprecated and map it into 2 of the values of the `ParallelExecutionOption` when using it in the `RuleBaseConfiguration`. I believe that this option is only used in some benchmarks of `kie-benchmarks`, so they will need to be fixed accordingly. @lucamolteni I'm not entirely sure if OptaPlanner ever used it (but I remember some discussion around this in the not so recent past), please give a look.

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgendaFactory.java
```diff
@@ -33,9 +33,9 @@ public static AgendaFactory getInstance() {
 
     private DefaultAgendaFactory() { }
 
-    public InternalAgenda createAgenda(InternalRuleBase kBase, boolean initMain) {
-        return kBase.hasParallelEvaluation() ?
-               new CompositeDefaultAgenda( kBase, initMain ) :
-               new DefaultAgenda( kBase, initMain );
+    public InternalAgenda createAgenda(InternalWorkingMemory workingMemory) {
```
mariofusco, line 36: The agenda was always created from a `kBase` and then its `workingMemory` was immediately set after its creation. I find this mechanism brittle and confusing, so I'm now creating it directly passing the `workingMemory`. This, together with the removal of no longer necessary `Externalizable` interface, also allowed to declare `final` the biggest part of the agenda fields.

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgenda.java
```diff
@@ -527,45 +491,6 @@ public void clearAndCancelRuleFlowGroup(final String name) {
...
-     * that are not allowed by the agenda filter.
-     *
-     * @return true if an activation was fired. false if no more activations
-     *              to fire
-     *
-     * @throws ConsequenceException
-     *             If an error occurs while firing an agenda item.
-     */
-    @Override
-    public int fireNextItem(final AgendaFilter filter,
```
mariofusco, line 541: This method was only used in a test. I removed it and changed the test.

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgendaFactory.java
```diff
@@ -33,9 +33,9 @@ public static AgendaFactory getInstance() {
...
+        return workingMemory.getKnowledgeBase().getRuleBaseConfiguration().isParallelExecution() ?
+                new CompositeDefaultAgenda( workingMemory ) :
```
mariofusco, line 38: A composite agenda is a set of agendas evaluating and firing rules independently and in parallel so this is used only for full parallel execution (parallel evaluation + firing).

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgendaFactory.java
```diff
@@ -33,9 +33,9 @@ public static AgendaFactory getInstance() {
...
+                new DefaultAgenda( workingMemory );
```
mariofusco, line 39: When sequential consequences firing is required all activations must be placed in the same single agenda in order to have a coordinated conflict resolution strategy. The sequential or parallel LHS evaluation is implemented internally to the agenda.

drools-kiesession/src/main/java/org/drools/kiesession/agenda/DefaultAgenda.java
```diff
@@ -90,18 +87,15 @@
  * Non-invalidated actions are left on the agenda, and are executed in turn.
  * </p>
  */
-public class DefaultAgenda implements Externalizable, InternalAgenda {
+public class DefaultAgenda implements InternalAgenda {
```
mariofusco, line 90: see https://github.com/kiegroup/drools/pull/5440/files#r1282897577

kie-internal/src/main/java/org/kie/internal/conf/MultithreadEvaluationOption.java
lucamolteni: The option `MultithreadEvaluationOption` is not used in OP

drools-core/src/main/java/org/drools/core/concurrent/ParallelRuleEvaluator.java
```diff
@@ -0,0 +1,61 @@
...
+                                int fireLimit,
+                                InternalAgendaGroup group ) {
+
+        Map<RuleBasePartitionId, List<RuleAgendaItem>> partitionedActivations = group.getActivations().stream().collect(Collectors.groupingBy(RuleAgendaItem::getPartition));
+        if (partitionedActivations.size() < MIN_PARALLEL_THRESHOLD) {
+            RuleAgendaItem item = group.peek();
+            return item != null ? internalEvaluateAndFire( filter, fireCount, fireLimit, item ) : 0;
+        }
+
+        partitionedActivations.values().parallelStream().forEach( items -> items.forEach( item -> item.getRuleExecutor().evaluateNetworkIfDirty(activationsManager) ) );
```
lucamolteni: is it a good idea to use an non configured parallel stream in the hot path?

mariofusco: Done.

drools-base/src/main/java/org/drools/base/common/PartitionsManager.java
```diff
@@ -51,8 +52,12 @@ private static class ForkJoinPoolHolder {
         private static final ForkJoinPool RULES_EVALUATION_POOL = new ForkJoinPool(); // avoid common pool
     }
 
-    public static ForkJoinPool getFireAllExecutors() {
-        return ForkJoinPoolHolder.RULES_EVALUATION_POOL;
+    public static void doOnForkJoinPool(Runnable task) {
```
mariofusco, line 55: This method and the next one are intended to be used to submit a task to a parallel Stream.

mariofusco:
I believe that this pull request is completed and ready to be reviewed now. I figured out that with the `ParallelRuleEvaluator` I was actually evaluating (possibly in parallel) an entire group and in reality I could do the same also with the sequential version instead of ping-ponging between it and the agenda for each and every rule evaluation. In essence I turned the rule evaluators into complete group evaluators also hoping that this could have a (minimal) positive impact on performances, for both sequential and parallel evaluations. In fact the change of focus of a group can only happen when executing the corresponding `PropagationEntry` that now [halts the group evaluator](https://github.com/kiegroup/drools/pull/5440/files#diff-f236e828a5a8516db9e3efe18ceebccec886ee3b7c25e623dbefa8ccc5cb3c86R142) when this happens.

lucamolteni (approved):

tkobayas (approved):

PR #5449, opened 2023-08-07
[DROOLS-7526] avoid emitting a misleading warning on a contains mvel expression

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7526

I understand this is a sort of a hack, but I didn't find a better alternative. @tkobayas please advice if you can find another solution.


tkobayas (approved):
Thanks. That makes sense.

PR #5450, opened 2023-08-08
[DROOLS-7520] fix query OTN lookup

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-7520

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/RuntimeSegmentUtilities.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/RuntimeSegmentUtilities.java
@@ -15,29 +15,14 @@
 
 package org.drools.core.phreak;
 
+import org.drools.base.reteoo.NodeTypeEnums;
+import org.drools.base.rule.constraint.QueryNameConstraint;
 import org.drools.core.common.Memory;
 import org.drools.core.common.MemoryFactory;
 import org.drools.core.common.ReteEvaluator;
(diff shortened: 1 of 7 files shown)


drools-core/src/main/java/org/drools/core/phreak/RuntimeSegmentUtilities.java
```diff
@@ -103,8 +101,9 @@ private static SegmentMemory restoreSegmentFromPrototype(ReteEvaluator reteEvalu
         return smem;
     }
 
-    public static SegmentMemory getQuerySegmentMemory(ReteEvaluator reteEvaluator, LeftTupleSource segmentRoot, QueryElementNode queryNode) {
-        LeftInputAdapterNode liaNode = getQueryLiaNode(queryNode.getQueryElement().getQueryName(), getQueryOtn(segmentRoot));
+    public static SegmentMemory getQuerySegmentMemory(ReteEvaluator reteEvaluator, QueryElementNode queryNode) {
+        ObjectTypeNode queryOtn = reteEvaluator.getDefaultEntryPoint().getEntryPointNode().getQueryNode();
```
mariofusco, line 105: The `ObjectTypeNode` from where the internal drools query objects are inserted is always under the default entry-point, so the old method to look for it was completely wrong, other than unnecessarily convoluted.

tkobayas (approved):

lucamolteni (approved):

PR #5529, opened 2023-09-19
[kie-issues-249] Data Index Events changes for allowing historical data

elguardian:
Issue: https://github.com/kiegroup/kie-issues/issues/249

### DIFF
--- a/kie-api/src/main/java/org/kie/api/event/usertask/Attachment.java
+++ b/kie-api/src/main/java/org/kie/api/event/usertask/Attachment.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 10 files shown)


porcelli (dismissed):
all new files requires the use of Apache headers and not Red Hat

mariofusco (dismissed):
Is it necessary to have this in Drools? It seems very Kogito specific so why not move it there?

porcelli:
@mariofusco I hear you, and it makes sense.

however if you check all events related to process are already here, so I'd argue that we could discuss in a different thread and eventually PR moving the process related content from drools to kogito.

elguardian:
@mariofusco unfortunately is how the design works. the idea is to strip the jbpm engine in the medium term. so for now we cannot change the dependencies just to acommodate this. it is out of the scope of the jira.

mariofusco:
@elguardian @porcelli That's fine, for now it's ok for me. If possible please open another ticket to better isolate this stuff in a second step.

mariofusco (approved):

pefernan (approved):
+1

porcelli (approved):

PR #5539, opened 2023-10-02
kie-issues#595: revert accidental change of credentials

jstastny-cz:
apache/incubator-kie-issues#595

Move back original credentials that are working fine for PR checks.

Earlier they were not working for nightlies to download buildchain config, but that should be handled differently.


jstastny-cz:
@baldimir when you have a minute.

baldimir (approved):

PR #5550, opened 2023-10-10
[kie-issues#624] Fix coercing double type in Mvel compiler

### LINKED ISSUE #624: MvelCompiler is not coercing type double
When having a double type in an mvel expression and using executable model, the type is not coerced. This leads to problem in using a parameter of type double in BigDecimal.valueOf method (when such method is used in an expression), in an expression like `$someBigDecimalVariable = $someBigDecimalVariable + BigDecimal.valueOf(0.5);`. In this case, the executable model doesn't generate properly BigDecimal addition using `.add` method, because it cannot resolve an appropriate version of valueOf method, because of missing type information about the parameter.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/624

The double type was not handled in the mvel compiler.


baldimir:
@tkobayas could you please take a look too?

lucamolteni (approved):

tkobayas:
@baldimir LGTM, thanks!

tkobayas (approved):

PR #5552, opened 2023-10-12
DROOLS-7552: improve updateObjectTypesList()

nprentza:
[DROOLS-7552](https://issues.redhat.com/browse/DROOLS-7552)

This improvement is related to benchmark FireAndAlarmBenchmark findings ([DROOLS-7545](https://issues.redhat.com/browse/DROOLS-7545)).

### DIFF
--- a/drools-reliability/drools-reliability-core/src/main/java/org/drools/reliability/core/SimpleSerializationReliableRefObjectStore.java
+++ b/drools-reliability/drools-reliability-core/src/main/java/org/drools/reliability/core/SimpleSerializationReliableRefObjectStore.java
@@ -122,7 +122,7 @@ public class SimpleSerializationReliableRefObjectStore extends SimpleSerializati
     }
 
     private void updateObjectTypesList(Object object) {
-        if (inverseStorage.keySet().stream().filter(sObject -> sObject.getClass().equals(object.getClass())).findAny()!=null){
+        if (!inverseStorage.keySet().stream().filter(sObject -> sObject.getClass().equals(object.getClass())).findAny().isEmpty()){
             uniqueObjectTypesInStorage.add(object.getClass().getName());
         }else{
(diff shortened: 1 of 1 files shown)


drools-reliability/drools-reliability-core/src/main/java/org/drools/reliability/core/SimpleSerializationReliableRefObjectStore.java
```diff
@@ -119,32 +116,25 @@ private Long fromObjectToFactHandleId(Object object) {
...
                 .collect(Collectors.toList());
     }
 
     private void updateObjectTypesList(Object object) {
-        uniqueObjectTypesInStore.put(object.getClass().getName(),
-                storage.values().stream().filter(sObject -> sObject.getObject().getClass().equals(object.getClass())).count());
-        // if count==0 then remove entry
-        if (uniqueObjectTypesInStore.get(object.getClass().getName()) == 0) {
-            uniqueObjectTypesInStore.remove(object.getClass().getName());
+        if (inverseStorage.keySet().stream().filter(sObject -> sObject.getClass().equals(object.getClass())).findAny()!=null){
```
tkobayas: `findAny()` returns `Optional.empty()` when there is no match, so comparing with `null` is not right, I think. Please check the execution path when the object class doesn't exist in `inverseStorage`.

nprentza: @tkobayas your comment is correct, comparing with `null` is not right, it always returns true. So, I will replace line 125 with 
`if (!inverseStorage.keySet().stream().filter(sObject -> sObject.getClass().equals(object.getClass())).findAny().isEmpty())`

mariofusco (approved):

tkobayas (approved):
LGTM, thanks!

PR #5562, opened 2023-10-24
[kie-issues#649] Use appropriate BigDecimal method to convert to primitive type

### LINKED ISSUE #649: Use appropriate BigDecimal method to convert to primitive type
When having an mvel expression in a DRL rule, that assigns a BigDecimal to a primitive type value, e.g. long (like `$someFact.longValue = $someOtherFact.bigDecimalValue`), the generated code fails on Java compilation, with an exception like `incompatible types: java.math.BigDecimal cannot be converted to long`. The mvel compiler needs to generate the assign expression with an appropriate BigDecimal conversion method based on the primitive type to which the value is assigned to. An example of a correctly generated code: `$someFact.longValue = $someOtherFact.bigDecimalValue.longValue()`

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/649. 

Please squash before merging.


porcelli (approved):
lgtm! 

(although black magic always get us on trouble)

mariofusco (commented):
The fix looks ok to me (but I'd prefer that also @lucamolteni  and @tkobayas could give a look first). I appreciate the extensive unit tests, but it would be nice to also add an integration test demonstrating the problem with the simplest possible rule.

baldimir:
> The fix looks ok to me (but I'd prefer that also @lucamolteni and @tkobayas could give a look first). I appreciate the extensive unit tests, but it would be nice to also add an integration test demonstrating the problem with the simplest possible rule.

Ok, will add some.

baldimir:
I added integration tests.

mariofusco (approved):

mariofusco:
Ok, for me, please @lucamolteni and/or @tkobayas give a look.

baldimir:
@lucamolteni I changed the BigDecimalArgumentCoercion class as discussed - I removed the typed class to a separate util class. Otherwise, based on my investigation, my fix should be correct. Please let me know if that is not correct and what is your suggestion then. Based on my observation, doing it in the AssignExprT is late, same with doing it in ReProcessRHSPhase. I tried to follow how you did it here https://github.com/apache/incubator-kie-drools/blob/0cc9229c54f036767c45381ad36ee980a0588d52/drools-model/drools-mvel-compiler/src/main/java/org/drools/mvelcompiler/RHSPhase.java#L256.

tkobayas (approved):
I'm not fully understanding the discussion in zulip. I leave the final approval to Luca.

lucamolteni:
See https://github.com/apache/incubator-kie-drools/pull/5567 for a smaller fix

PR #5565, opened 2023-10-26
[kie-issues#653] Don't use System.out and logger debug mode in test output in traits

### LINKED ISSUE #653: Do not use logger level debug or System.out.println in tests in Drools repository
There are places in the Drools repository, where logger level debug is used or an output is written using System.out commands. E.g. https://github.com/apache/incubator-kie-drools/blob/18047e93c81c3502cd759fbe3e8ce0c8f2e796f4/drools-traits/src/test/resources/logback-test.xml#L33. That should be avoided and logger level debug should be used only when debugging locally on a person's computer. In the repository itself, there should be reasonable logger levels set, so the output doesn't flood the Ci automation pipelines output.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/653

- Changes logger debug mode in tests in the whole repository
- Replaces or removes the use of System.out or System.err in drools-traits tests. When an occurence is replaced, it is replaced with logger.

@jstastny-cz this should fix the huge logs you reported from drools-traits in the CI logs.

### DIFF
--- a/drools-compiler/src/test/resources/logback-test.xml
+++ b/drools-compiler/src/test/resources/logback-test.xml
@@ -30,10 +30,10 @@
   <logger name="org.kie" level="warn"/>
   <logger name="org.drools" level="warn"/>
   
-  <logger name="org.drools.core.management" level="debug"/>
+  <logger name="org.drools.core.management" level="info"/>
   <!--<logger name="org.drools.ancompiler" level="debug"/>-->
 
-  <logger name="org.drools.metric.util.MetricLogUtils" level="trace"/>
+  <logger name="org.drools.metric.util.MetricLogUtils" level="info"/>
 
   <root level="warn">
     <appender-ref ref="consoleAppender" />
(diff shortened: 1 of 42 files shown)


mariofusco (approved):

jstastny-cz:
I will review the PR check logs in the morning, but seems promising ;-) Thanks.

jstastny-cz (approved):
Thanks, looks good.

Just for the record I see some seemingly sysout prints still:
`[INFO] Running org.drools.decisiontable.SpreadsheetCompilerUnitTest`
`[INFO] Running org.drools.model.codegen.execmodel.AccumulateTest`
`[INFO] Running org.drools.model.codegen.execmodel.CepTest`
`[INFO] Running org.drools.model.codegen.execmodel.CompilationFailuresTest`
`[INFO] Running org.drools.model.codegen.execmodel.FromTest`
(and generally codegen tests seem to be quite verbose)

`[INFO] Running org.drools.mvel.integrationtests.concurrency.ConcurrentInsertionsToSubnetworksTest`
`[INFO] Running org.drools.mvel.integrationtests.QueryInRHSCepTest` - this one has exception stack trace printed

etc.

But the drools-traits tests seem alright, which was the scope tackled in this one.

PR #5568, opened 2023-10-27
[kie-issues#656] Change Overpass font to more widely used Arial in ReteDiagram generator.

### LINKED ISSUE #656: Use a more widely used font in Rete diagram generator
The Rete diagram generator located in https://github.com/apache/incubator-kie-drools/blob/main/drools-retediagram/src/main/java/org/drools/retediagram/ReteDiagram.java#L167 uses Overpass font, which is not widely used and often missing in operating systems. A more widely used font should be used there instead. 

Example: This change was initiated because Overpass was needed to be installed in the CI image because of Rete diagram generator tests. It is not available in the package managers of various operating systems, so it had to be installed in a non-standard way.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/656

### DIFF
--- a/drools-retediagram/src/main/java/org/drools/retediagram/ReteDiagram.java
+++ b/drools-retediagram/src/main/java/org/drools/retediagram/ReteDiagram.java
@@ -164,9 +164,9 @@ public class ReteDiagram {
         File pngFile = new File(outputPath, pngFileName);
         try (PrintStream out = new PrintStream(new FileOutputStream(gvFile));) {
             out.println("digraph g {\n" +
-                    "graph [fontname = \"Overpass\" fontsize=11];\n" + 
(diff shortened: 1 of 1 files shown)


tarilabs (approved):
I'm a bit surprised some tests are failing due to missing Fonts, I was expecting simply it would encode a SVG which does not render Overpass but another similar typeface.

You can consider instead of _replacing_ Overpass with Arial, to make it a fallback with something similar to:

```
Overpass,Arial
```

but up to you really

jstastny-cz:
@tarilabs it was not failing without the fonts, but rather producing ERROR or WARN logs. Though better not to have those :-)

jstastny-cz (approved):
LGTM.

mariofusco (approved):

PR #5620, opened 2023-12-11
[kie-issues#754] Use UTF-8 on important places during Quarkus build

### LINKED ISSUE #754: Fix problems on Japanese encoded envornments during Kogito DRL Quarkus build
Currently, UTF-8 is hardcoded on various places during Kogito DRL Quarkus application build, however on some places, the process uses environment encoding settings. This raises problems with non-standard characters, like Japanese. UTF-8 must be used on all important places during code generation.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/754

[Dependant kogito-runtimes PR](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3323) 

Uses UTF-8 on important places during the Quarkus build. I suggest we should do a broader encoding review during next year. There are more places, like e.g. calling String.getBytes() without encoding and similar on many places. Unfortunately this is very hard to test, as the behaviour I was fixing only happens on an environment that doesn't use UTF-8, so there are no new test cases added. If you know about some option how to add such tests, please let me know and I will add them.

### DIFF
--- a/kie-memory-compiler/src/main/java/org/kie/memorycompiler/jdknative/NativeJavaCompiler.java
+++ b/kie-memory-compiler/src/main/java/org/kie/memorycompiler/jdknative/NativeJavaCompiler.java
@@ -87,7 +87,7 @@ public class NativeJavaCompiler extends AbstractJavaCompiler {
         DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
         JavaCompiler compiler = getJavaCompiler();
 
-        try (StandardJavaFileManager jFileManager = compiler.getStandardFileManager(diagnostics, null, null)) {
+        try (StandardJavaFileManager jFileManager = compiler.getStandardFileManager(diagnostics, null, Charset.forName(pSettings.getSourceEncoding()))) {
             try {
                 jFileManager.setLocation( StandardLocation.CLASS_PATH, pSettings.getClasspathLocations() );
                 jFileManager.setLocation( StandardLocation.CLASS_OUTPUT, Collections.singletonList(new File("target/classes")) );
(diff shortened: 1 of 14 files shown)


kie-memory-compiler/src/main/java/org/kie/memorycompiler/jdknative/NativeJavaCompiler.java
```diff
@@ -87,7 +87,7 @@ public CompilationResult compile( String[] pResourcePaths,
         DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
         JavaCompiler compiler = getJavaCompiler();
 
-        try (StandardJavaFileManager jFileManager = compiler.getStandardFileManager(diagnostics, null, null)) {
+        try (StandardJavaFileManager jFileManager = compiler.getStandardFileManager(diagnostics, null, Charset.forName(pSettings.getSourceEncoding()))) {
```
gitgabrio, line 90: hi @baldimir : why here it is not used the hardcoded "UTF_8" charset ? Could there be any issue when this encoding it is not UTF_8 ?

baldimir: UTF-8 is used by default in the configuration, so I wanted to promote its use as it is used also elsewhere. I plan a larger review of encoding next year, I would say we could solve this problem in a more global better way then (maybe having some drools.encoding property or similar). What do you think?

gitgabrio: Fine for me!
Side note: is there any test/pr check with different encoding ? If not, I think, in the future, we should try to also have this kind of checkl

gitgabrio (approved):

kie-memory-compiler/src/main/java/org/kie/memorycompiler/jdknative/NativeJavaCompiler.java
baldimir: There are some tests, as far as I know, that use a resource in different encoding. However this fix is for a case where the environment is in different encoding. That is why I was not able to add any new test case unfortunately. If someone knows how we could simulate that reliably, please let me know.

mariofusco (approved):

PR #5630, opened 2023-12-17
Flat Lambdas and Multi Variable Indexes [apache/incubator-kie-issues#770]

### LINKED ISSUE #770: Flat Lambdas and Multi Variable Indexes
Join indexing does not support multi-variable inputs. i.e. you cannot index "str1.lenght + i == str2.length", as the left operand requires both the str1 variable and the i variable.

The object graph to reach the lambda is quite high, especially the BetaConstraints wrapper that provides old optimisations for single, double, and triple beta constraints - redundancy with Lambdas.

mdproctor:
Closes https://github.com/apache/incubator-kie-issues/issues/770

Flat Lambdas and Multi Variable Indexes #770 (All details in the ticket).

Do not merge this yet, I haven't checked all tests pass. Just uploading it, so others can look at the implementation.


mdproctor:
[apache/incubator-kie-issues#770]

mdproctor:
Green build first time :)

drools-core/src/main/java/org/drools/core/common/BetaConstraints.java
```diff
@@ -22,37 +22,36 @@
...
+import org.drools.base.rule.constraint.BetaConstraint;
 import org.drools.core.RuleBaseConfiguration;
 import org.drools.core.reteoo.BetaMemory;
 import org.drools.core.reteoo.Tuple;
 import org.drools.core.reteoo.builder.BuildContext;
 import org.drools.util.bitmask.BitMask;
 import org.kie.api.runtime.rule.FactHandle;
 
-public interface BetaConstraints
+public interface BetaConstraints<C>
```
mariofusco, line 35: I doubt if this generics is necessary, you don't seem to actually use it?

drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
```diff
@@ -141,7 +141,7 @@ private void doLeftInserts(AccumulateNode accNode,
         BetaMemory bm = am.getBetaMemory();
         TupleMemory ltm = bm.getLeftTupleMemory();
         TupleMemory rtm = bm.getRightTupleMemory();
-        ContextEntry[] contextEntry = bm.getContext();
+        Object contextEntry = bm.getContext();
```
mariofusco, line 144: Having to declare this as `Object` is also the bad side-effect of introducing that not necessary generics on the `BetaConstraint`.

drools-core/src/main/java/org/drools/core/reteoo/builder/BuildUtils.java
```diff
@@ -203,9 +203,14 @@ private boolean areNodesCompatibleForSharing(BuildContext context, BaseNode node
      * @param list the list of constraints
      */
     public BetaConstraints createBetaNodeConstraint(final BuildContext context,
-                                                    final List<BetaNodeFieldConstraint> list,
+                                                    final List<BetaConstraint> list,
                                                     final boolean disableIndexing) {
         BetaConstraints constraints;
+        if ( list.size() == 1 && list.get(0) instanceof BetaConstraints) {
```
mariofusco, line 209: Not understanding this, the list is already a `List<BetaConstraint>` so it is already guaranteed that the first item (if any) is a `BetaConstraint`, what am I missing?

drools-core/src/main/java/org/drools/core/util/index/IndexSpec.java
```diff
@@ -1,108 +1,17 @@
...
-import org.drools.base.rule.constraint.BetaNodeFieldConstraint;
-import org.drools.base.util.FieldIndex;
+import org.drools.base.util.IndexedValueReader;
 import org.drools.base.util.index.ConstraintTypeOperator;
-import org.drools.core.RuleBaseConfiguration;
-import org.kie.internal.conf.IndexPrecedenceOption;
-
-import static org.drools.base.util.index.IndexUtil.isEqualIndexable;
-
-public class IndexSpec {
```
mariofusco, line 33: What's the point of turning this into an `interface`? I see only one implementation and I cannot foresee other possible alternatives.

drools-core/src/main/java/org/drools/core/util/index/DefaultIndexSpec.java
```diff
@@ -0,0 +1,159 @@
...
+        return constraintType;
+    }
+
+    @Override
+    public IndexedValueReader[] getIndexes() {
+        return indexes;
+    }
+
+    @Override
+    public Supplier<Index> getIndexSupplier() {
```
mariofusco: Why a `Supplier` instead of a simple factory returning the index?

drools-core/src/main/java/org/drools/core/util/index/IndexSpec.java
mdproctor: For a while, I had two implementations. I went back to one implementation but left the interface/impl split. I'll revert this back to class only now.

drools-core/src/main/java/org/drools/core/reteoo/builder/BuildUtils.java
mdproctor: Look at the instanceof, notice the "s" on the end ;) BetaConstraints. It's checking if we have a BetaConstraint, that also implements BetaConstraints

drools-core/src/main/java/org/drools/core/phreak/PhreakAccumulateNode.java
mdproctor: Constraints no longer return an array, so it has to be an Object. It cannot just return a market interface, as its sometimes an array and sometimes it is not.

mdproctor: Note that in reality, with the flat PositionalConstraint, it doesn't actually need a Context any more. But it was too much work to figure out how to make that optional, but in a future release it might be nice.

drools-core/src/main/java/org/drools/core/common/BetaConstraints.java
mdproctor: Its not strictly necessary, but I found in a few places it avoided casts, so less code.

drools-core/src/main/java/org/drools/core/util/index/DefaultIndexSpec.java
mdproctor: Agreed, stale code from an earlier attempt. I've revered this back to getIndex.

tkobayas (approved):
I'll look into it tomorrow. If in hurry, it's okay to merge if Mario approves.

mariofusco (approved):

PR #5637, opened 2023-12-20
[NO ISSUE] Shade kie-dmn-xls2dmn-cli runner separately.

baldimir:
The Maven deploy of the shaded kie-dmn-xls2dmn-cli artifact in the default configuration caused org.apache:xmlbeans maven-metadata.xml to be deployed separately for some reason. The only configuration fixing the problem is the one in the PR. This may need further investigation. Submitting this to fix the nightly builds deployment of snapshot artifacts.


mariofusco (approved):

gitgabrio (approved):
Thanks, LGTM!

PR #5650, opened 2024-01-12
[KIE-820] add timestamp for reproducible builds

mariofusco:
See https://github.com/apache/incubator-kie-issues/issues/820

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,6 +140,25 @@
     </mailingList>
   </mailingLists>
 
+  <properties>
+    <project.build.outputTimestamp>2024-01-12T00:00:00Z</project.build.outputTimestamp>
+  </properties>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-artifact-plugin</artifactId>
+          <version>3.3.0</version>
+          <configuration>
+            <outputTimestamp>${project.build.outputTimestamp}</outputTimestamp>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
   <profiles>
     <profile>
       <id>default</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -140,6 +140,25 @@
...
+    <project.build.outputTimestamp>2024-01-12T00:00:00Z</project.build.outputTimestamp>
+  </properties>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-artifact-plugin</artifactId>
+          <version>3.3.0</version>
```
gitgabrio: could you please move that 3.3.0 version inside the `<properties>` tag, and also pick the most recent one ?
Generally speaking, we try to define actual version as property, to allow easier modification (if needed)

pom.xml
```diff
@@ -140,6 +140,25 @@
     </mailingList>
   </mailingLists>
 
+  <properties>
+    <project.build.outputTimestamp>2024-01-12T00:00:00Z</project.build.outputTimestamp>
```
gitgabrio, line 144: Out of curiosity: is this fixed/hardcoded timestamp something that we have to modify, in future, maybe with some dinamc value ?

porcelli (approved):
thank you for that @mariofusco !

gitgabrio (changes requested):
Thanks @mariofusco !
I would ask you just a small modification

pom.xml
mariofusco: Done. I'm using version `3.4.1` now. Actually there is also a `3.5.0` release available, but the reproducible build check doesn't work if I try to use it.

pom.xml
mariofusco: Yes, as I wrote on zulip the solution using `${commit.time.iso}` looks [very convoluted](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/master/pom.xml#667) and I wanted to avoid such complication. I'm looking for a reasonable alternative, the hardcoded timestamp is only a temporary solution to at least have a reproducible build.

gitgabrio (approved):
LGTM!

PR #5665, opened 2024-01-25
[KIE-858] make executable model code generation optionally reproducible

### LINKED ISSUE #858: Make executable model code generation optionally reproducible
Part of https://github.com/apache/incubator-kie-issues/issues/840

mariofusco:
Closes: https://github.com/apache/incubator-kie-issues/issues/858

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -44,7 +45,6 @@ import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.InitializerDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
-import com.github.javaparser.ast.comments.JavadocComment;
 import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.ClassExpr;
 import com.github.javaparser.ast.expr.Expression;
@@ -95,6 +95,7 @@ import org.drools.util.TypeResolver;
 import org.kie.api.builder.ReleaseId;
 import org.kie.api.conf.PrototypesOption;
 import org.kie.api.runtime.rule.AccumulateFunction;
+import org.kie.internal.builder.conf.ReproducibleExecutableModelGenerationOption;
 import org.kie.internal.ruleunit.RuleUnitDescription;
 import org.kie.internal.ruleunit.RuleUnitVariable;
 
@@ -155,7 +156,7 @@ public class PackageModel {
     private final List<TypeDeclaration> generatedPOJOs = new ArrayList<>();
     private final List<GeneratedClassWithPackage> generatedAccumulateClasses = new ArrayList<>();
 
-    private final Set<Class<?>> domainClasses = new HashSet<>();
+    private final Set<Class<?>> domainClasses = new LinkedHashSet<>();
     private final Map<Class<?>, ClassDefinition> classDefinitionsMap = new HashMap<>();
     
     private final Set<Class<?>> otnsClasses = new HashSet<>();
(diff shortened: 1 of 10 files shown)


kie-internal/src/main/java/org/kie/internal/builder/conf/ReproducibleExecutableModelGenerationOption.java
```diff
@@ -0,0 +1,55 @@
...
+ * An Enum to force the reproducible executable model generation.
+ *
+ * drools.reproducibleExecutableModelGeneration = &lt;true|false&gt;
+ *
+ * DEFAULT = false
+ */
+public enum ReproducibleExecutableModelGenerationOption implements SingleValueRuleBuilderOption {
+
+    ENABLED(true),
+    DISABLED(false);
```
porcelli, line 33: what would be the reason to have reproducible builds disabled?

lucamolteni: I agree, there's no need for a feature flag here. Let's keep it always reproducible

lucamolteni (changes requested):
Most of the changes are related to the feature flag which it's useless

mariofusco:
@porcelli @lucamolteni To enforce the reproducibility of the generated executable model it is necessary to serialize many of the tasks that are currently performed in parallel thus greatly reducing the time required to the code generation itself. I don't see why we should give up to this optimization when a reproducible executable model is not required, that in reality is the 99.999...% of the cases.

lucamolteni (approved):
After having discussed the flag enables parallel builds so let's keep it

porcelli (approved):

baldimir (approved):

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
```diff
@@ -233,10 +234,21 @@ public static void initPackageModel(TypeDeclarationContext typeDeclarationContex
      * @param packageName
      * @return
      */
-    public static String getPkgUUID(ReleaseId releaseId, String packageName) {
+    public static String getPkgUUID(KnowledgeBuilderConfigurationImpl configuration, ReleaseId releaseId, String packageName) {
+        if (isReproducibleExecutableModelGeneration(configuration)) {
+            return StringUtils.getPkgUUID(releaseId != null ? releaseId.toString() : "", packageName);
```
gitgabrio, line 239: @mariofusco 
Why we do not use always that snippet ? If it is possible to avoid "randomness" somehow, I think we should always do, regardless of "reproduciblebuild" flag.
If, on he other side, that snippet is potentially unsafe in some context, we should not add it at all IMO

kie-dmn/kie-dmn-validation-bootstrap/src/main/java/org/kie/dmn/validation/bootstrap/GenerateModel.java
```diff
@@ -63,10 +64,15 @@ public void generate() throws IOException {
         KieServices ks = KieServices.Factory.get();
         final KieBuilderImpl kieBuilder = (KieBuilderImpl) ks.newKieBuilder(kieDmnValidationBaseDir);
 
-        kieBuilder.buildAll(ValidationBootstrapProject::new,
-                            s -> !s.contains("src/test/java") && !s.contains("src\\test\\java") &&
-                                 !s.contains("DMNValidator") && // <- to break circularity which is only caused by the KieBuilder trying to early compile everything by itself  
-                                 !s.contains("dtanalysis"));
+        System.setProperty(ReproducibleExecutableModelGenerationOption.PROPERTY_NAME, "true");
```
gitgabrio: @mariofusco 
Could you please set that property in the pom itself, instead of this class? This property is not relevant for that specific class, but for the module as a whole (even if _currently_ is used only there)

gitgabrio (changes requested):
Hi @mariofusco 
Thanks for PR. I think the option to have a fast build vs a reproducible one is a cool feature, especially because it can be used by users/customers.
I have just a couple of modifications to ask for

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
mariofusco: In general for (nor reproducible) SNAPSHOTs we need a random UUID instead, otherwise this breaks the incremental compilation when using a snapshot release id.

gitgabrio (approved):
Thanks @mariofusco LGTM!

PR #5671, opened 2024-02-02
[kie-issues#903] Add reflect-config.json files related to drools-commands.

### LINKED ISSUE #903: Add reflect-config.json configuration to drools-commands and relevant other modules
When using JAXB classes from the drools-commands module in a native binary build, it fails, because some of the classes are not declared for reflection. New reflect-config.json configuration file need to be created, with classes added.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/903.

### DIFF
--- a/drools-commands/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/drools-commands/src/main/resources/META-INF/native-image/reflect-config.json
@@ -1,0 +1,245 @@
+[
+  {
+    "name" : "org.drools.commands.runtime.BatchExecutionCommandImpl",
+    "allDeclaredConstructors" : true,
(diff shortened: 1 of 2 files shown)


yesamer (approved):

mariofusco (approved):

gitgabrio (changes requested):
thanks @baldimir 
THose changes seems fine, but we need a test to verify
1) they does not break anything
2) that solve the issue they are required for

Could you please create a kogito-example that makes use of such classes and pair with this PR ? THanks

baldimir:
After discussion with Gabriele, I am merging this, he is fine merging this as it is.

PR #5672, opened 2024-02-03
[incubator-kie-issues#875] DMN - Do not create an exception when property is not defined.

jmreardon-yoppworks:
This PR addresses this issue https://github.com/apache/incubator-kie-issues/issues/875 by removing the creation of an exception when a property is not defined. 

This change is only observable by calling the `getValueResult`, but none of the uses of that method are affected by this change. There are only 4 uses of this method in the code base, and each case is either:

* guarded by a call to `isDefined` so that `getValueResult` is not called
* or transforms not defined results to `null` anyways.

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
@@ -281,6 +281,11 @@ public class EvalHelper {
 
     public static class PropertyValueResult implements FEELPropertyAccessible.AbstractPropertyValueResult {
 
+        // This exception is used to signal an undefined property for notDefined(). This method may be many times when
+        // evaluating a decision, so a single instance is being cached to avoid the cost of creating the stack trace
+        // each time.
+        private static final Exception undefinedPropertyException = new UnsupportedOperationException("Property was not defined.");
+
         private final boolean defined;
         private final Either<Exception, Object> valueResult;
 
@@ -290,7 +295,7 @@ public class EvalHelper {
         }
 
         public static PropertyValueResult notDefined() {
-            return new PropertyValueResult(false, Either.ofRight(null));
+            return new PropertyValueResult(false, Either.ofLeft(undefinedPropertyException));
         }
 
         public static PropertyValueResult of(Either<Exception, Object> valueResult) {
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/EvalHelper.java
@@ -290,7 +290,7 @@ public class EvalHelper {
         }
 
         public static PropertyValueResult notDefined() {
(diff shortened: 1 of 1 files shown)


gitgabrio (commented):
HI @jmreardon-yoppworks 
Thanks for the PR. I'm going to analyze locally, just to double check.

gitgabrio:
Hi @jmreardon-yoppworks 
Firtst of all, thatnks again for your PR and your analysys on that.
I'm looking at your modification, and I agree with your observations about actual usage of the generated exception.
But, here's my small concerns.
On one side, the definition of the class clearly define the semantic of the `valueResult` as 
`Either<Exception, Object>` 
On the other side, this modification would introduce a small inconsistency, i.e. in some cases the `getDefinedValue` would return an` Either of righ(null) `, and in other cases an `Either of left (Exception)`.
Moreover, those methods are `public static`, so maybe used elsewhere, with the defined semantic.
What if, instead, we use a singleton of the `UnsupportedOperationException`, so that
1. the semantic of the class is preserved
2. the exception is created only once, so it should reduce the overall impact

Wdyt ?

FYI @baldimir @yesamer

jmreardon-yoppworks:
@gitgabrio , I'm fine using a singleton of the `UnsupportedOperationException`. I started with the null approach because singleton exceptions are a little obscure, but the singleton is safer if we have downstream dependencies to worry about.

gitgabrio:
> @gitgabrio , I'm fine using a singleton of the `UnsupportedOperationException`. I started with the null approach because singleton exceptions are a little obscure, but the singleton is safer if we have downstream dependencies to worry about.

Thanks @jmreardon-yoppworks ! Please let us know if that approach would solve the performance issue

jmreardon-yoppworks:
> > @gitgabrio , I'm fine using a singleton of the `UnsupportedOperationException`. I started with the null approach because singleton exceptions are a little obscure, but the singleton is safer if we have downstream dependencies to worry about.
> 
> Thanks @jmreardon-yoppworks ! Please let us know if that approach would solve the performance issue

I'll rerun my bench with it this weekend to verify. I would be shocked if it had a different profile from the null fix. The only expensive part was filling the trace and the singleton avoids that too.

jmreardon-yoppworks:
Static exception comes out slightly ahead of my null benchmark [but that could be my computer having fewer Intellij instances open today].

## Static Exception Benchmarks
```
[info] Benchmark                               Mode  Cnt    Score   Error   Units
[info] DmnListFilterBench.evaluate1Function   thrpt   10   72.720 ± 0.278  ops/ms
[info] DmnListFilterBench.evaluate2Functions  thrpt   10   49.537 ± 0.375  ops/ms
[info] DmnListFilterBench.evaluateBool        thrpt   10  149.365 ± 0.891  ops/ms
[info] DmnListFilterBench.evaluateBoolItem    thrpt   10  126.287 ± 0.452  ops/ms

gitgabrio (approved):
Thanks @jmreardon-yoppworks LGTM!

baldimir (approved):

baldimir:
Thank you very much for the contribution. Looks good to me. When the PR checks are done and pass, it can be merged.

PR #5675, opened 2024-02-06
[kie-issues#839] Remove comments script - ignore nonexistant files

baldimir:
This change makes the RemoveCommentsMain class to ignore non-existant files. This is a prerequisite for configuration in kogito-runtimes.

### DIFF
--- a/drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
+++ b/drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
@@ -2,6 +2,7 @@ package org.drools.util;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.util.stream.Collectors;
 
@@ -12,7 +13,10 @@ public class RemoveCommentsMain {
             try {
                 Files.write(Path.of(fileName), removeComments(fileName).getBytes());
             } catch (IOException e) {
-                throw new RuntimeException(e);
+                // Ignore non-existant files.
+                if (!(e instanceof NoSuchFileException)) {
+                    throw new RuntimeException(e);
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
```diff
@@ -10,9 +11,15 @@ public class RemoveCommentsMain {
...
             try {
-                Files.write(Path.of(fileName), removeComments(fileName).getBytes());
+                final String result = removeComments(fileName);
+                if (result != null) {
+                    Files.write(Path.of(fileName), result.getBytes());
+                }
             } catch (IOException e) {
-                throw new RuntimeException(e);
+                // Ignore non-existant files.
+                if (!(e instanceof NoSuchFileException)) {
```
mariofusco: Shouldn't we at least log that the file didn't exist and therefore we ignored it?

drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
```diff
@@ -21,7 +28,11 @@ static String removeComments(String fileName) {
         try (var lines = Files.lines(Path.of(fileName))) {
             return lines.filter(line -> !line.startsWith("#")).collect(Collectors.joining("\n"));
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            // Ignore non-existant files.
+            if (!(e instanceof NoSuchFileException)) {
```
mariofusco: Is it necessary to duplicate this check? Isn't one of the 2 just enough?

drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
baldimir: Fixed.

drools-util/src/main/java/org/drools/util/RemoveCommentsMain.java
baldimir: Fixed.

mariofusco (approved):

lucamolteni (approved):

yesamer (approved):

PR #5677, opened 2024-02-08
[DROOLS-7591] Experiment branch : migrate a new drools-lsp parser int…

tkobayas:
- New parser is migrated in `drools-drl/drools-drl-parser` with org.drools.drl10.parser package
- To enable the new parser, run `mvn test -PDRL10`
- At the moment, we get lots of test failures especially in `drools-model/drools-model-codegen` and `drools-test-coverage/test-compiler-integration`

https://github.com/apache/incubator-kie-drools/issues/5678

This PR is an experiment to migrate a new DRL parser in a feature branch `dev-new-parser`.

Currently, we get test failures with the new parser enabled `mvn test -PDRL10`.

Especially,

`drools-model/drools-model-codegen`
[ERROR] Tests run: 2433, Failures: 592, Errors: 19, Skipped: 9

`drools-test-coverage/test-compiler-integration`
[ERROR] Tests run: 5396, Failures: 1249, Errors: 269, Skipped: 71

We will be able to split and fix the issues.


kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
```diff
@@ -25,7 +25,7 @@
  */
 public enum LanguageLevelOption implements SingleValueRuleBuilderOption {
 
-    DRL5(false), DRL6(false), DRL6_STRICT(true);
+    DRL5(false), DRL6(false), DRL6_STRICT(true), DRL10(false);
```
tkobayas, line 28: @mariofusco @yurloc Do you think it's fine to call the new parser `DRL10`? This doesn't introduce a new syntax, but we need a clear separation. So I use `LanguageLevelOption`. Feel free to share your thoughts.

drools-drl/drools-drl-parser/src/main/java/org/drools/drl10/parser/Drl10ExprParser.java
```diff
@@ -0,0 +1,87 @@
...
+import org.drools.drl.ast.descr.ConstraintConnectiveDescr;
+import org.drools.drl.parser.DrlExprParser;
+import org.drools.drl.parser.DroolsParserException;
+import org.kie.internal.builder.conf.LanguageLevelOption;
+
+/**
+ * This is a helper class that provides helper methods to parse expressions
+ * using both the DRLExpressions parser and the DRLExprTree parser.
+ */
+public class Drl10ExprParser implements DrlExprParser {
```
tkobayas, line 37: If we accept that the new parser is called `DRL10`, then I would rename other classes with `DRL10` prefix.

drools-drl/drools-drl-parser/pom.xml
```diff
@@ -66,7 +66,32 @@
             <groupId>org.antlr</groupId>
             <artifactId>antlr-runtime</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
+        </dependency>
```
tkobayas, line 72: At first, I tried to create a separated module `drools-drl10-parser` for the new parser, but it resulted in circular dependency with `drools-drl-parser`, so I ended up with migrating the new parser into `drools-drl-parser`. I used a new package `org.drools.drl10.parser` for a better separation.

drools-drl/drools-drl-parser-tests/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.drools</groupId>
+        <artifactId>drools-drl</artifactId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.drools</groupId>
+    <artifactId>drools-drl-parser-tests</artifactId>
```
tkobayas, line 33: I created a separate module for `drools-drl-parser-tests`. Having tests inside `drools-drl-parser` is also okay from Maven point of view, but Intellij IDEA complained about the circular dependency with `drools-compiler` and `drools-mvel`. Anyway, these tests were originally copied from `drools-test-coverage/test-compiler-integration`, so finally, we would be able to remove this module. If we can distill pure parser tests, we would move them to `drools-drl-parser` test.

tkobayas:
Do not forget to set a profile `DRL10`. -> `mvn test -PDRL10`. Otherwise, you would run default `DRL6`. You may activate `DRL10` by default by editing pom.xml.

drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/DrlParser.java
```diff
@@ -164,10 +166,35 @@ public PackageDescr parse(final boolean isEditor,
                               final InputStream is) throws DroolsParserException, IOException {
         this.resource = resource;
         String encoding = resource instanceof InternalResource ? ((InternalResource) resource).getEncoding() : null;
-
-        lexer = DRLFactory.buildLexer(is, encoding, languageLevel);
-        DRLParser parser = DRLFactory.buildParser(lexer, languageLevel);
-        return compile(isEditor, parser);
+        System.out.println("### parse : languageLevel = " + languageLevel);
```
tkobayas, line 169: While running unit tests, you would see STDOUT everytime
```
### parse : languageLevel = DRL10
```
I print this just to be sure that I'm running DRL10, but if you think it's verbose, feel free to remove on your side.

mariofusco (approved):

tkobayas:
I know that `DrlParser`, `DRLFactory` integration with the new parser is not enough in this PR. We can incrementally improve using unit tests.

For example) https://github.com/apache/incubator-kie-drools/issues/5681

kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
yurloc: Temporarily, yes. But it should be removed in the final step in my opinion.
1. The language level stays the same. We're just changing an implementation detail that deals with how the source code in DRL6 language is being parsed.
2. Is the language level option a public API? If yes, it's one more reason not to introduce DRL10. Implementation details should not proliferate to the public API.

drools-drl/drools-drl-parser/src/main/java/org/drools/drl10/parser/Drl10ExprParser.java
yurloc: Temporarily, definitely yes. Ideally, if could drop DRL5 and DRL6_STRICT as part of the final step, this should become something like `Drl6ExpressionParser`.

kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
tkobayas: > Implementation details should not proliferate to the public API.

@yurloc Totally agreed. I just couldn't come up with other Option to switch this implementation. Any idea? (of course, we can change it after merging this PR)

drools-drl/drools-drl-parser/src/main/java/org/drools/drl10/parser/Drl10ExprParser.java
tkobayas: > Ideally, if could drop DRL5 and DRL6_STRICT as part of the final step

True, but it may take some time to drop the old implementation (antlr3 based)

tkobayas:
Replaced by https://github.com/apache/incubator-kie-drools/pull/5682 to preserve commit history in drools-lsp.

mariofusco:
Replaced by https://github.com/apache/incubator-kie-drools/pull/5682

kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
tkobayas: Even system property like `drools.new.parser.enabled` = true, may work.

Then a consideration would be a "name" which we will use for class name or package name while we keep the existing classes like `DRL6Parser`. `NewDRL6Parser`? :D

PR #5685, opened 2024-02-13
[incubator-kie-issues-861] PIM for kogito workflow / remove unused kie api / kie internal api

### LINKED ISSUE #861: PIM for kogito workflow.
[001 - Process Instance Migration Proposal.pdf](https://github.com/apache/incubator-kie-issues/files/14236985/001.-.Process.Instance.Migration.Proposal.pdf)

The idea is to create a system to migrate kogito process from one version to the other. The requirments are:

- Need to support the same persistence pluggable tier from workflow engine
- needs to support batch processing and paralelization (avoid one by one execution)
- need to implement mapping between one version and the other.
- some modifications in the code to support container localization (2 or more versions of the same process and know which one is responsible for executing the workflow)
- Endpoints to implement
- cloud event to indicate the system has beeing migrated (this should go to the data audit as well) 
- revisit the idea of migrated logs (in v7 there were a bit bad)

elguardian:
Closes https://github.com/apache/incubator-kie-issues/issues/861#issue-2101810651


mariofusco (approved):
Great work, thanks a lot. There are a few tests failures that however seem unrelated to this. I gave a second run to the github actions just in case. As agreed if you see other things in this area that could be removed or refactored please let me know and I'll be glad of taking care of this.

mariofusco:
@elguardian can you please double check if the failures in `kogito-examples` are in any way related with this PR? (I don't think so)

elguardian:
@mariofusco they are related to the PR in kogito runtimes

pefernan (approved):
+1 when green

PR #25, opened 2011-05-31
Forbid/force binding in accumulat function calls

ghost:
As discussed with etirelli: forbid in "from accumulate", enforce in accumulate CE.

Some minor updates in "Expert" related to Calendars & Timers

PR #26, opened 2011-06-03
Edits from JIRA https://issues.jboss.org/browse/JBRULES-1908

ghost:
Edits proposed and requested in JIRA https://issues.jboss.org/browse/JBRULES-1908

PR #5691, opened 2024-02-15
[DROOLS-7608] fix KieProject lookup on classpath for EAP 7.4.15 VFS

mariofusco:
@tkobayas This is only a rewritten and cleaned up version of the solution that you found [here](https://github.com/kiegroup/drools/pull/25).

I dug into VFS sources a lot and couldn't find any better solution. I understand that using reflection is not ideal, but in the end we were already doing so, even if just because we didn't want a direct dependency on VFS. Now it's true that this is a bit worse than before because we're also accessing package private methods (that are probably subject to change in future), but other than asking the Jboss team to make those methods public, thus avoiding to use `setAccessible(true)`, I don't see any possible improvement.


mariofusco:
Replaced by https://github.com/kiegroup/drools/pull/26

PR #5693, opened 2024-02-19
[incubator-kie-drools#5681] Experiment: New DRL Parser : Test failure…

tkobayas:
… : Unknown language level

- Revert the introduction of LanguageLevelOption.DRL10. Use DRL6 as usual. Instead, intorduced system property "drools.drl.antlr4.parser.enabled" for parser impl switch
- Use a package name 'antlr4' instead of 'drl10' for the new parser separation

https://github.com/apache/incubator-kie-drools/issues/5681


pom.xml
```diff
@@ -262,10 +262,10 @@
     </profile>
 
     <profile>
-      <id>DRL10</id>
+      <id>antlr4parser</id>
```
tkobayas: Please use `mvn test -Pantlr4parser` instead of `mvn test -PDRL10` for the new parser testing.

tkobayas:
As we don't introduce new DRL syntax, introducing language level `DRL10` was unnatural. @yurloc had already pointed out that (https://github.com/apache/incubator-kie-drools/pull/5677#discussion_r1484046003).

So I reverted the introduction of `LanguageLevelOption.DRL10` in this PR, and use `DRL6` as usual. It naturally solves the `Unknown language level` errors.

Instead, I introduced a system property "drools.drl.antlr4.parser.enabled" for the new parser impl switch.

I have also moved java sources and test resources from `org/drools/drl10/parser` package to `org/drools/drl/parser/antlr4` package. If you are working on `dev-new-parser` branch now, please make sure to rebase your work after this PR is merged.

tkobayas:
@mariofusco @yurloc Please review. Thanks!

kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
```diff
@@ -25,7 +25,7 @@
  */
 public enum LanguageLevelOption implements SingleValueRuleBuilderOption {
 
-    DRL5(false), DRL6(false), DRL6_STRICT(true), DRL10(false);
+    DRL5(false), DRL6(false), DRL6_STRICT(true);
```
mariofusco, line 28: @tkobayas @yurloc at this point I believe that this option is totally obsolete and we can get rid of it. For now I will open a ticket to do so. I'm not sure if it makes sense to do this work now, because I'm afraid it will be heavily in conflict with this branch. Any objection or something that I'm missing?

mariofusco (approved):

kie-internal/src/main/java/org/kie/internal/builder/conf/LanguageLevelOption.java
tkobayas: @mariofusco TBH, I don't know the background why we had `DRL5` (probably backward compatibility, so it would be fine to drop now) and `DRL6_STRICT` (I haven't seen someone uses this). No objection for dropping.

yurloc:
I would like to propose that we make the new parser default on the feature branch to make all the failures visible in the CI. But that can be done in a separate PR, feel free to merge.

Otherwise LGTM.

tkobayas:
@yurloc Thanks, updated this PR:
- Enabled the new parser by default
    - If you want to disable, use `mvn test -PdisableNewParser`
- Suppressed some GHAs which don't work with a feature branch

tkobayas:
GHA gets lots of test failures as expected. I think `drools-model-codegen` is a good place to look around.

```
2024-02-20T02:13:39.3232373Z [INFO] --- surefire:3.1.2:test (default-test) @ drools-model-codegen ---
...
2024-02-20T02:14:46.2835379Z [ERROR] Tests run: 2433, Failures: 592, Errors: 16, Skipped: 9
```

tkobayas:
@mariofusco Could you merge this? Thanks!

PR #5746, opened 2024-02-26
[incubator-kie-issues#968] Support ConditionalExpr in Drools executable model

### LINKED ISSUE #968: Support ConditionalExpr in Drools executable model
When not using executable model, it is possible to write an expression like this in a rule when section: 
`$p : Person($booleanVariable: (name != null ? true : false))`

This doesn't work in executable model as it ends with UnsupportedOperationException. 
Test case for this: https://gist.github.com/baldimir/d3fc199aaecdaec9616b3b6759edf2e2

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/968

Note: this is the `when` condition used as example
` $p : Person($booleanVariable: (name != null ? true : false))`

In both kind of executions (legacy vs executablemodel) the condition that populate the `$booleanVariable` is ignored for the evaluation of the rule, it is used only for the population of the variable itself

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
@@ -44,6 +44,7 @@ import com.github.javaparser.ast.expr.BinaryExpr.Operator;
 import com.github.javaparser.ast.expr.CastExpr;
 import com.github.javaparser.ast.expr.CharLiteralExpr;
 import com.github.javaparser.ast.expr.ClassExpr;
+import com.github.javaparser.ast.expr.ConditionalExpr;
 import com.github.javaparser.ast.expr.DoubleLiteralExpr;
 import com.github.javaparser.ast.expr.EnclosedExpr;
 import com.github.javaparser.ast.expr.Expression;
@@ -378,6 +379,10 @@ public class ExpressionTyper {
                 }
             }
             return of(new TypedExpression(drlxExpr, type));
+        }
+
+        if (drlxExpr instanceof ConditionalExpr) {
+            return of(new TypedExpression(drlxExpr, Boolean.class));
         }
 
         if (drlxExpr.isAssignExpr()) {
--- a/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ConditionalExprTest.java
+++ b/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ConditionalExprTest.java
@@ -1,0 +1,87 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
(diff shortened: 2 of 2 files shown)


yesamer (approved):

baldimir:
I don't think it can be ignored for the rule evaluation. If it is a boolean condition, it is a constraint in my opinion. What do you think please? Also how does it behave please for non-Boolean typed expressions?

gitgabrio:
> I don't think it can be ignored for the rule evaluation. If it is a boolean condition, it is a constraint in my opinion. What do you think please? Also how does it behave please for non-Boolean typed expressions?

As stated in the comment, that modification on the executable model EXACTLY MIMIC what happen in the legacy one.
Please look at the test, that verify the two different conditions (true/false) for both implementations.
IINW drools currently does not support the nested assignment before "when" evaluation. 
My understanding is that (if there is an assignment inside an object):
1. drools evaluate the condition based on presence of the object itself, and possibly properties already defined for the object
2. after that, it evaluate the nested property assignment, populating it

IIUC, the request would be to
1. first, evaluate the inner property assignement
2. then, evaluate the when condition

I have the impression it would be quite a change, to be done at both compile and execution time, for both model implementations.
This PR ensure the behaviour in the given cases is the same for both implementations, and this is required to allow migration by users.
Changing the logic itself would require more effort, and I think it should be delegated to a specific ticket.

@mariofusco Of course, all my previous assumptions may be wrong, feel free to correct/extend, thanks!

mariofusco (approved):
I discussed this with @baldimir since this is also related to another fix that he developed recently. We agreed that those bound boolean variables are also predicate and then should also act as constraint. But more important drools should always behave exactly in the same way regardless if you're using the executable model or not. Since at the moment making the plain drl rules evaluation to run these as proper constraints will require a bigger rework that will also impact the current parser, we agreed that this is a nice to have, but we can put it aside for now. I will approve and merge this pull request since it guarantees the same behavior for both evaluation modes as expected.

gitgabrio:
Thanks @mariofusco !

The caveat I found is that in the following expression

`$p : Person($booleanVariable: (name != null ? true : false))` 

the 

`$booleanVariable: (name != null ? true : false)` 

part is actually an _**assignment**_, so semantically the same as

`$p : Person($anotherVariable: ("name"+"somethingelse"))`

so, changing the behavior based on the type of the variable, seems unconsistent to me.

In that very example, the rule should be evaluated only when 
`name != null`, and only in that case the `$booleanVariable` should be populated.
The concept applies beside that simple example, of course, whenever a boolean property is assigned _on the fly_ in the `when` clause while, actually, this should happen in the `then`:
if the rule is executed based on the `predicate` itself, then the variable will be populated (in that context, for me the `name != null ? true : false` is the predicate, and the `$booleanVariable` is the result of it)

AFAIK, this should be the correct syntax

```
global Boolean booleanVariable
rule \"r1\"
when 
    $p : Person(name != null);
then
   booleanVariable = true;
   System.out.println(booleanVariable); 
   System.out.println($p); 
   booleanListGlobal.add(booleanVariable); 
end
```

I have the impression that we are looking for a modification around a wrong usage of the language, putting a `consequence` semantic inside an `evaluating` one. 
I'm not the responsible of drl language, but this seems strange to me, from a logical point of view.

gitgabrio:
> I don't think it can be ignored for the rule evaluation. If it is a boolean condition, it is a constraint in my opinion. What do you think please? Also how does it behave please for non-Boolean typed expressions?

TBH I do not think so: from logical point of view, it is an "assignment" what we are talking about. "By chance, in this case the variable is a boolean, that of course can be populated only by a condition. BUt dealing with that case in a different way then the other kind of "assignment" would introduce a pretty evident unconsistency

PR #5751, opened 2024-02-28
[incubator-kie-drools#5743] Update Developer_Notes.md to be community…

tkobayas:
…-friendly

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5743

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


tkobayas:
@gitgabrio @baldimir @yurloc @mariofusco Please review, thanks!

mariofusco (approved):

baldimir (approved):

drools-drl/drools-drl-parser/Developer_Notes.md
```diff
@@ -0,0 +1,70 @@
+# drools-drl-parser
+
+In this `dev-new-parser` branch, `drools-drl-parser` has 2 implementations of the DRL parser. The new parser is under development, and the old parser is used in the current drools code base.
+
+The old DRL parser is based on Antlr3 and contains lots of hard-coded logic in generated java codes, so it's hard to maintain. e.g. [DRL6Parser](https://github.com/apache/incubator-kie-drools/blob/main/drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/lang/DRL6Parser.java). It is the reason why the new parser is being developed. The old parser resources is placed under `src/main/java/org/drools/drl/parser/lang` and `src/main/resources/org/drools/drl/parser`.
```
yurloc:
```suggestion
The old DRL parser is based on Antlr3 and contains lots of hard-coded logic in generated java codes, so it's hard to maintain e.g. [DRL6Parser](https://github.com/apache/incubator-kie-drools/blob/main/drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/lang/DRL6Parser.java). It is the reason why the new parser is being developed. The old parser resources are placed under `src/main/java/org/drools/drl/parser/lang` and `src/main/resources/org/drools/drl/parser`.
```

drools-drl/drools-drl-parser/Developer_Notes.md
```diff
@@ -0,0 +1,70 @@
...
+
+## How to contribute to the development
+
+**Contribution would be highly appreciated!** Here is a rough guide to contribute to the development.
+
+1. Look at the parent issue [Experiment: New DRL Parser](https://github.com/apache/incubator-kie-drools/issues/5678) and choose a child issue which you want to work on.
+2. If you have an ASF committer role, assign yourself to the child issue. If you don't have the role, post a comment on the child issue that you are working on it so that we can avoid duplicated work.
+3. Create a feature branch based on the `dev-new-parser` branch.
+4. Fix the issue and make sure that the test in problem is passed and also all tests under `drools-drl-parser-tests` are passed.
+5. Add new tests to `drools-drl-parser-tests` to cover the issue. Hopefully, such a test would be a Descr comparison test. See `MiscDRLParserTest`.
```
yurloc:
```suggestion
5. Add new tests to `drools-drl-parser-tests` to cover the issue. Hopefully, such a test would be a Descr comparison test. See `MiscDRLParserTest` and `DRLExprParserTest`.
```

tkobayas:
Thank you for the suggestions, @yurloc !

gitgabrio:
Hi @tkobayas , many thanks, that's great!
The only think I noticed is that this document dive deep in the current implementation (e.g old vs new approach) which is very good ( 👍  ) but reading it, I could not find a description of the "design" itself (beside that 

`DRLParser processes a DRL text and produces an AST(abstract syntax tree). Then apply DRLVisitorImpl to generate PackageDescr following the visitor pattern. So the main work would be implementing DRLParser.g4 and DRLVisitorImpl`

).
What I mean is that this document is very good to describe the current implementation, but does not provide a sort of mapping between the "technical needs"and the current code. Reason for raising this is that, in the future,
1. as we moved from different Antlr version, maybe we could switch to a completely different technology, and this document could help to distinguish what was the scope of different piece of code from what was the consequence
2. from general quality POV, it would help to verify if the current implementation, that is the result of long works and refactory, does exactly reflect the needs, or if it is flawed by some improper design choices that have been introduced long time ago and we are now keep building on top of.

Does this make sense ? Do you think it could be usefull/feasible to also provide this kind of information ?
Of course, it could be deferred to other ticket (and, of course, it is very possible that I misread/misundestood that document 😄  )

tkobayas:
@gitgabrio Thank you for the suggestion. I updated with "Design Notes".

gitgabrio (approved):
Thanks @tkobayas LGTM!

PR #5757, opened 2024-03-01
enable eager segment initialization by default

mariofusco:


mdproctor:
Don't forget to document the change in the release notes.

Issue #5721, opened 2024-02-21
[new-parser] Array initialization fails


yurloc:
## Parent issue
- #5678 

## Failing tests
- `org.drools.model.codegen.execmodel.ComplexRulesTest#test1`

## Notes
- `STANDARD_FROM_DRL` only.

## Rule code snippet
```
not ( 
        $policySet_not : ChildFactWithObject ( 
            id == $childFactWithObject.id , 
            eval(true == functions.arrayContainsInstanceWithParameters((Object[])$policySet_not.getObjectValue(), new Object[]{"getMessageId", "42103"}))
        )
    )
```

## Error output
```
### parse : ANTLR4_PARSER_ENABLED = true
line 1:97 extraneous input 'new' expecting {TIME_INTERVAL, DRL_STRING_LITERAL, '?/', DECIMAL_LITERAL, HEX_LITERAL, FLOAT_LITERAL, BOOL_LITERAL, STRING_LITERAL, 'null', '(', '[', '.', '<', '!', '~', '++', '--', '+', '-', '*', '/', IDENTIFIER}
line 1:109 no viable alternative at input 'Object[]{'
21:45:15.378 [main] ERROR o.d.c.k.b.impl.AbstractKieProject.buildKnowledgePackages:280 - Unable to build KieBaseModel:defaultKieBase
Unable to Analyse Expression true == functions.arrayContainsInstanceWithParameters((Object[])$policySet_not.getObjectValue(), new:
[Error: unbalanced braces ( ... )]
[Near : {... stanceWithParameters((Object[])$policySet_not.getO ....}]
                                                         ^ : [Rule name='R1']



java.lang.AssertionError: [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=r0.drl, line=-1, column=0
   text=Unable to Analyse Expression true == functions.arrayContainsInstanceWithParameters((Object[])$policySet_not.getObjectValue(), new:
[Error: unbalanced braces ( ... )]
[Near : {... stanceWithParameters((Object[])$policySet_not.getO ....}]
                                                         ^]]

	at org.drools.model.codegen.execmodel.BaseModelTest.createKieBuilder(BaseModelTest.java:157)
	at org.drools.model.codegen.execmodel.BaseModelTest.createKieBuilder(BaseModelTest.java:132)
	at org.drools.model.codegen.execmodel.BaseModelTest.getKieContainer(BaseModelTest.java:121)
[...]

Issue #5704, opened 2024-02-20
[new-parser] casting in expressions doesn't work


yurloc:
## Parent issue
- #5678 

## Failing tests
- `org.drools.model.codegen.execmodel.CompilerTest#testBetaCast`

## Notes
- `STANDARD_FROM_DRL` only.

## Rule code snippet
```
  $a : Address(shortNumber == (short)$intField)
```

## Error output
```
### parse : ANTLR4_PARSER_ENABLED = true
line 1:16 no viable alternative at input '(short'
line 1:16 mismatched input 'short' expecting {'not', 'matches', '(', '>', '<', '~', '==', '<=', '>=', '!=', IDENTIFIER}
21:44:56.387 [main] ERROR o.d.c.k.b.impl.AbstractKieProject.buildKnowledgePackages:280 - Unable to build KieBaseModel:defaultKieBase
Unable to Analyse Expression shortNumber == :
[Error: Malformed expression]
[Near : {... shortNumber == ....}]
                           ^ : [Rule name='R']



java.lang.AssertionError: [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=r0.drl, line=-1, column=0
   text=Unable to Analyse Expression shortNumber == :
[Error: Malformed expression]
[Near : {... shortNumber == ....}]
                           ^]]

	at org.drools.model.codegen.execmodel.BaseModelTest.createKieBuilder(BaseModelTest.java:157)
	at org.drools.model.codegen.execmodel.BaseModelTest.createKieBuilder(BaseModelTest.java:132)
	at org.drools.model.codegen.execmodel.BaseModelTest.getKieContainer(BaseModelTest.java:121)
	at org.drools.model.codegen.execmodel.BaseModelTest.getKieContainer(BaseModelTest.java:114)
	at org.drools.model.codegen.execmodel.BaseModelTest.getKieSession(BaseModelTest.java:110)
	at org.drools.model.codegen.execmodel.BaseModelTest.getKieSession(BaseModelTest.java:106)
	at org.drools.model.codegen.execmodel.CompilerTest.testBetaCast(CompilerTest.java:1893)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:577)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
[...]

PR #5759, opened 2024-03-04
[new-parser] Fix the rest of the keywords

yurloc:
Follow up on #5748.

Fixes:
- #5704 (casting to all kinds of primitive types)
- #5721 (`new` keyword)

Now that we're using a specialized token to match each keyword, it's redundant to use `helper.validateIdentifierKey()` so I'm removing that.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL6Expressions.g4
+++ b/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL6Expressions.g4
@@ -770,72 +770,72 @@ assignmentOperator
 //                      KEYWORDS
 // --------------------------------------------------------
 extends_key
-    :      {(helper.validateIdentifierKey(DroolsSoftKeywords.EXTENDS))}? id=IDENTIFIER { helper.emit($id, DroolsEditorType.KEYWORD); }
+    : id=EXTENDS { helper.emit($id, DroolsEditorType.KEYWORD); }
     ;
 
 super_key
-    :      {(helper.validateIdentifierKey(DroolsSoftKeywords.SUPER))}? id=SUPER { helper.emit($id, DroolsEditorType.KEYWORD); }
+    : id=SUPER { helper.emit($id, DroolsEditorType.KEYWORD); }
     ;
 
 instanceof_key
-    :      {(helper.validateIdentifierKey(DroolsSoftKeywords.INSTANCEOF))}? id=IDENTIFIER { helper.emit($id, DroolsEditorType.KEYWORD); }
+    : id=INSTANCEOF { helper.emit($id, DroolsEditorType.KEYWORD); }
     ;
 
 boolean_key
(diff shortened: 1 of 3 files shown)


yurloc:
@tkobayas @gitgabrio @mariofusco please review.

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL6Expressions.g4
```diff
@@ -770,72 +770,72 @@ assignmentOperator
...
+    : id=EXTENDS { helper.emit($id, DroolsEditorType.KEYWORD); }
     ;
 
 super_key
-    :      {(helper.validateIdentifierKey(DroolsSoftKeywords.SUPER))}? id=SUPER { helper.emit($id, DroolsEditorType.KEYWORD); }
+    : id=SUPER { helper.emit($id, DroolsEditorType.KEYWORD); }
     ;
 
 instanceof_key
-    :      {(helper.validateIdentifierKey(DroolsSoftKeywords.INSTANCEOF))}? id=IDENTIFIER { helper.emit($id, DroolsEditorType.KEYWORD); }
```
gitgabrio, line 781: HI @yurloc many thanks for PR!
May you explain, just to understand, what this sort of check
`{(helper.validateIdentifierKey(DroolsSoftKeywords.INSTANCEOF))}?`
is for, and why it could be removed now ?

yurloc: Sure. Please read the first two paragraphs about [semantic predicates](https://github.com/antlr/antlr4/blob/master/doc/predicates.md). Semantic predicate at the start of an alternative is used to "kill" that alternative although it is viable from the syntactic perspective.

The semantic predicate guard was necessary here in the previous version of the grammar because the lexer grammar didn't have specific tokens for the individual keywords (e.g. `SUPER` token for the `super` keyword). Instead, it used the next most specific token (`IDENTIFIER`) in all the different keyword rules and it used the semantic predicate to check whether the `IDENTIFIER`-matching word is actually the keyword that the rule is about.

On this line, `helper.validateIdentifierKey(DroolsSoftKeywords.INSTANCEOF)` simply looked at the next token in the token stream and checked whether the string value of the token was equal to `DroolsSoftKeywords.INSTANCEOF` (which is `"instanceof"`).

In the next parser version, we have `JavaLexer.g4` that has these specific tokens (or [lexer rules](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md)) so we can use them instead of `IDENTIFIER`, and that makes the syntactic predicate redundant.

tkobayas (approved):
@yurloc Thank you for the clean fix.

mariofusco (approved):

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL6Expressions.g4
gitgabrio: Thanks, @yurloc !!!

yurloc:
Fixing progress: There were 520 failed tests (-29).

PR #5761, opened 2024-03-05
[kie-issues#986] Coerce object to String in executable model codegen

### LINKED ISSUE #986: toString method is not applied to an object in a String rule constraint
When having constraints like this,
```
"    $p: Person() \n" +
"    String(this == \"someString\" + $p)\n"
```
the build ends with 
`Failed to coerce : InvalidExpressionErrorResult: Arithmetic operation requires compatible types. Found class java.lang.String and class org.drools.model.codegen.execmodel.domain.Person`
and 
`Unable to Analyse Expression this == "someString" + $p`

This works without using executable model. 

Test case for this: https://gist.github.com/baldimir/c382da571b53f6faf2df96b245ff2add

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/986

For binary expression, only numeric arithmetic operations were considered during code generation.

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
@@ -95,6 +95,7 @@ import org.drools.mvel.parser.printer.PrintUtil;
 import org.drools.mvelcompiler.CompiledExpressionResult;
 import org.drools.mvelcompiler.ConstraintCompiler;
 import org.drools.mvelcompiler.util.BigDecimalArgumentCoercion;
+import org.drools.util.ClassUtils;
 import org.drools.util.MethodUtils;
 import org.drools.util.TypeResolver;
 import org.slf4j.Logger;
@@ -224,16 +225,19 @@ public class ExpressionTyper {
             TypedExpression left = optLeft.get();
             TypedExpression right = optRight.get();
 
-            ArithmeticCoercedExpression.ArithmeticCoercedExpressionResult coerced;
-            try {
-                coerced = new ArithmeticCoercedExpression(left, right, operator).coerce();
-            } catch (ArithmeticCoercedExpression.ArithmeticCoercedExpressionException e) {
-                logger.error("Failed to coerce : {}", e.getInvalidExpressionErrorResult());
-                return empty();
-            }
+            // Can be a binary expression of non-numeric types, like String + customType, etc.
+            if (ClassUtils.isNumericClass(left.getRawClass()) || ClassUtils.isNumericClass(right.getRawClass())) {
+                ArithmeticCoercedExpression.ArithmeticCoercedExpressionResult coerced;
+                try {
+                    coerced = new ArithmeticCoercedExpression(left, right, operator).coerce();
+                } catch (ArithmeticCoercedExpression.ArithmeticCoercedExpressionException e) {
+                    logger.error("Failed to coerce : {}", e.getInvalidExpressionErrorResult());
+                    return empty();
+                }
 
-            left = coerced.getCoercedLeft();
-            right = coerced.getCoercedRight();
+                left = coerced.getCoercedLeft();
+                right = coerced.getCoercedRight();
+            }
 
             final BinaryExpr combo = new BinaryExpr(left.getExpression(), right.getExpression(), operator);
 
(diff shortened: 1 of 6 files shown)


tkobayas (approved):

yesamer (approved):

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
```diff
@@ -224,16 +225,19 @@ private Optional<TypedExpression> toTypedExpressionRec(Expression drlxExpr) {
...
 
-            ArithmeticCoercedExpression.ArithmeticCoercedExpressionResult coerced;
-            try {
-                coerced = new ArithmeticCoercedExpression(left, right, operator).coerce();
-            } catch (ArithmeticCoercedExpression.ArithmeticCoercedExpressionException e) {
-                logger.error("Failed to coerce : {}", e.getInvalidExpressionErrorResult());
-                return empty();
-            }
+            // Can be a binary expression of non-numeric types, like String + customType, etc.
+            if (ClassUtils.isNumericClass(left.getRawClass()) || ClassUtils.isNumericClass(right.getRawClass())) {
```
gitgabrio: Thanks @baldimir 
But IIUC the `ArithmeticCoercedExpression` requires one part to be a number and the other a String. With the above if, the code checks only if one of the elements is a number, but does not verify if the other is a String. If it is not, the following
`coerced = new ArithmeticCoercedExpression(left, right, operator).coerce();`
would throw an exception (IINW)

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/TypeCoercionTest.java
```diff
@@ -612,4 +612,30 @@ public void testFloatOperation() {
         assertThat(list.size()).isEqualTo(1);
         assertThat(list.get(0)).isEqualTo("Mario");
     }
+
+    @Test
```
gitgabrio, line 616: Hi @baldimir 
COuld you please add a similar test where one of the two components is a number ?

gitgabrio (changes requested):
Thanks @baldimir 
I think a couple of minor modifications may be required

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
baldimir: This is just a check to find out if one of the inputs is a number. Other type checks are within the ArithmeticCoercedExpression.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/TypeCoercionTest.java
baldimir: There are existing tests added as part of the PR that introduced the ArithmeticCoercedExpression. https://github.com/apache/incubator-kie-drools/commit/5512f7294ea9b54e15d60786e3e42f5486b46ee4

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
gitgabrio: Yes, but `ArithmeticCoercionExpression` should be returned only with number/string elements; otherwise, a different expression is needed. 
With the above, when one of the element is a number and the other is not a number nor a string, the code will throw an exception and return an empty Optional.
Instead, when one element is number and the other is not a number nor a string, I think it should anyway return a BynaryExpr (as at line 242).
Is that clear ? Does this make sense ?

baldimir: I am simplifying this more. I will push new changes in few mins.

baldimir: Updated.

baldimir:
Based on feedback, I redid the changes a bit. There were concerns, that it is not clear what side effects the original ArithmeticCoercionExpression class has, when reading/debugging the code. So I reorganized it a bit.

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
baldimir: Please see my latest changes.

mariofusco (approved):

tkobayas (approved):

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
gitgabrio: Thanks, good to me!

gitgabrio (approved):
LGTM, thanks!

PR #5762, opened 2024-03-05
[kie-issues#969]: Negated value in accumulate function ends with Invalid expression error

### LINKED ISSUE #969: Negated value in accumulate function ends with Invalid expression error
When negating a value in a DRL rule, in an accumulate function, like this: 
```
accumulate ($p: Person (getName().startsWith("M")); 
                     $sum : sum(-$p.getAge()))
```
The build of the executable model ends with error: 
`Invalid expression -$p.getAge()`

Test case that could be added to org.drools.model.codegen.execmodel.AccumulateTest class: 
https://gist.github.com/baldimir/dd594f0b6e789a61284e134b7ee8f529

yesamer:
Closes: https://github.com/apache/incubator-kie-issues/issues/969

```
accumulate ($p: Person (getName().startsWith("M")); 
                     $sum : sum(-$p.getAge()))
```
This case leads to an `Invalid expression -$p.getAge()` error.

That is because the UnaryExpression types are not managed in the `AccumulateVisitor` logic.
To fix the issue, I managed that case relying on the existing BinaryExpression logic management, which fits well for the `UnaryExpression` case according to my analysis. 
Please correct me if I'm wrong @lucamolteni @mariofusco.

In addition, we added a check to ensure that the accumulate function always has no more than one parameter (kudos to @gitgabrio ). I added that check for the executable Model case only

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/accumulate/AccumulateVisitor.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/accumulate/AccumulateVisitor.java
@@ -209,7 +209,7 @@ public class AccumulateVisitor {
         final Expression accumulateFunctionParameter = DrlxParseUtil.parseExpression(accumulateFunctionParameterStr).getExpr();
 
         if (accumulateFunctionParameter instanceof BinaryExpr || accumulateFunctionParameter instanceof UnaryExpr) {
-            return binaryOrUnaryExprParameter(basePattern, function, functionDSL, bindingId, accumulateFunctionParameterStr);
+            return bindingParameter(basePattern, function, functionDSL, bindingId, accumulateFunctionParameterStr);
         }
 
         if (parameterNeedsConvertionToMethodCallExpr(accumulateFunctionParameter)) {
@@ -414,7 +414,7 @@ public class AccumulateVisitor {
         return accumulateFunctionParameter.isMethodCallExpr() || accumulateFunctionParameter.isArrayAccessExpr() || accumulateFunctionParameter.isFieldAccessExpr();
     }
 
-    private Optional<NewBinding> binaryOrUnaryExprParameter(PatternDescr basePattern, AccumulateDescr.AccumulateFunctionCallDescr function, MethodCallExpr functionDSL, String bindingId, String accumulateFunctionParameterStr) {
+    private Optional<NewBinding> bindingParameter(PatternDescr basePattern, AccumulateDescr.AccumulateFunctionCallDescr function, MethodCallExpr functionDSL, String bindingId, String accumulateFunctionParameterStr) {
         final DrlxParseResult parseResult = ConstraintParser.defaultConstraintParser(context, packageModel).drlxParse(Object.class, bindingId, accumulateFunctionParameterStr);
 
         return parseResult.acceptWithReturnValue(new ParseResultVisitor<>() {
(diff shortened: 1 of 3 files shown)


lucamolteni (approved):

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/AccumulateTest.java
```diff
@@ -4204,4 +4206,80 @@ public void testJoinInAccumulate() {
...
+                "  accumulate ( $i : Integer() and $p: Person ( name.length >= $i ); \n" +
+                "                $sum : sum($p.getAge(), $p.getName())  \n" +
+                "              )                          \n" +
+                "then\n" +
+                "  insert(new Result($sum));\n" +
+                "end";
+
+        Results results = createKieBuilder(str).getResults();
+        if (testRunType.isExecutableModel()) {
+            System.out.println(results.getMessages(Message.Level.ERROR).get(0).getText());
```
gitgabrio: Thanks @yesamer 
Could you please remove this System.out.println ?

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/accumulate/AccumulateVisitor.java
```diff
@@ -406,10 +414,10 @@ private boolean parameterNeedsConvertionToMethodCallExpr(Expression accumulateFu
         return accumulateFunctionParameter.isMethodCallExpr() || accumulateFunctionParameter.isArrayAccessExpr() || accumulateFunctionParameter.isFieldAccessExpr();
     }
 
-    private Optional<NewBinding> binaryExprParameter(PatternDescr basePattern, AccumulateDescr.AccumulateFunctionCallDescr function, MethodCallExpr functionDSL, String bindingId, String accumulateFunctionParameterStr) {
+    private Optional<NewBinding> binaryOrUnaryExprParameter(PatternDescr basePattern, AccumulateDescr.AccumulateFunctionCallDescr function, MethodCallExpr functionDSL, String bindingId, String accumulateFunctionParameterStr) {
```
gitgabrio: HI @yesamer 
Would it make sense to rename it as `bindingParameter` or similar ?
1. if, by chance, we discover that another kind of expression has to be managed here, the method name would become awfull
2. usually, the method name express the returned object
3. IINW, nothing prevent this method to be invoked by somehting different then a `BinaryExpr` or a `UnaryExpr` (if it is implicitily so, I think there should be a check/guard somehow)

WDYT ?

gitgabrio (changes requested):
HI @yesamer 
There is a System.out.println to remove. And maybe a small refactoring/enforcement

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/accumulate/AccumulateVisitor.java
yesamer: Thank you, done!

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/AccumulateTest.java
yesamer: Done

gitgabrio (approved):
Thanks @yesamer LGTM!

baldimir (approved):

PR #5764, opened 2024-03-05
[KIE-990] Replace hibernate with openjpa in persistence tests

mariofusco:


lucamolteni (approved):

drools-persistence/drools-persistence-jpa/pom.xml
```diff
@@ -76,6 +75,29 @@
...
+          <enforcePropertyRestrictions>true</enforcePropertyRestrictions>
+        </configuration>
+        <executions>
+          <execution>
+            <id>enhancer</id>
+            <phase>process-test-classes</phase>
+            <goals>
+              <goal>test-enhance</goal>
+            </goals>
+          </execution>
```
yurloc: @mariofusco I'm a bit concerned whether this can't negatively affect the included classes. Not sure what the enhancing does, but if it modified the class files' byte code permanently and polluted them with some OpenJPA APIs it would introduce a runtime dependency on it, which is bad if these drools persistence classes are a public API that should be JPA implementation-agnostic, right?

mariofusco: @yurloc This is indeed a good point. In general I prefer to do stuff at build time rather then at runtime when possible, but in this specific case I agree that this will introduce in those instrumented classes some OpenJPA specific code. This could hopefully be harmless but I share your concern, will check if I can do the enhancement at runtime.

mariofusco: @yurloc Fixed, thanks for suggestion.

yurloc (approved):

tkobayas (approved):

drools-persistence/drools-persistence-jpa/pom.xml
```diff
@@ -193,12 +183,18 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging</artifactId>
+      <scope>test</scope>
+    </dependency>
```
yurloc, line 190: @mariofusco Is this necessary? Why? Is this change related to switching from Hibernate to OpenJPA?

PR #5779, opened 2024-03-12
Improve AppPaths

yurloc:
Hi @gitgabrio, I came across https://github.com/apache/incubator-kie-drools/pull/5777 and noticed a few things that could be improved a little further (in my opinion). So I decided to take a stab at it while it's fresh. Let's see if you find this useful as well.

- The `isJar` flag was useless because it was `true` on every invocation of the method.
- I find it simpler to feed the transformation method with just a path than repeating the same lambda over and over.
- I noticed that `Paths.get()` is [deprecated](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/nio/file/Paths.html) so I replaced it with `Path.of()`.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-model/drools-codegen-common/src/main/java/org/drools/codegen/common/AppPaths.java
+++ b/drools-model/drools-codegen-common/src/main/java/org/drools/codegen/common/AppPaths.java
@@ -20,14 +20,12 @@ package org.drools.codegen.common;
 
 import java.io.File;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


gitgabrio:
Thanks @yurloc , but I just modified this class and, for the moment being, I think it is better to keep it as it is.
Some side notes:
1. some of the refactoring (mostly, the static methods) are aimed at improve testability (this class has never been tested before), even if currently I had no time to extend them
2. `Path.of` and `path.resolve` have completely different semantics
3. I saw that the `isJar` boolean is always true, but for the moment being I preferred to kept it in the static method (see above about tests)

I spent enough time on that and, regardless of test/downstream results (that, we are discovering, are missing lot of corner cases), I would prefer (for the moment being) to avoid further modification on it: I hope it would make sense

PR #5793, opened 2024-03-18
[incubator-kie-drools#5708] [new-parser] function definition causes a…

tkobayas:
…n error

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5708

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
+++ b/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
@@ -164,7 +164,7 @@ drlRelationalOperator
     | DRL_NOT? DRL_STR LBRACK IDENTIFIER RBRACK ;
 
 /* function := FUNCTION type? ID parameters(typed) chunk_{_} */
-functiondef : DRL_FUNCTION typeTypeOrVoid? IDENTIFIER formalParameters block ;
+functiondef : DRL_FUNCTION typeTypeOrVoid? IDENTIFIER formalParameters drlBlock ;
 
 
 /* extending JavaParser qualifiedName */
@@ -512,3 +512,37 @@ drlVariableInitializer
  drlArrayInitializer
      : LBRACE (drlVariableInitializer (COMMA drlVariableInitializer)* (COMMA)? )? RBRACE
      ;
+
+/* extending JavaParser block */
+drlBlock
+    : LBRACE drlBlockStatement* RBRACE
+    ;
+/* extending JavaParser blockStatement */
+drlBlockStatement
+    : localVariableDeclaration SEMI
+    | drlStatement
+    | localTypeDeclaration
+    ;
+
+/* extending JavaParser statements */
+drlStatement
+    : blockLabel=block
+    | ASSERT drlExpression (COLON drlExpression)? SEMI
+    | IF parExpression drlStatement (ELSE drlStatement)?
+    | FOR LPAREN forControl RPAREN drlStatement
+    | WHILE parExpression drlStatement
+    | DO drlStatement WHILE parExpression SEMI
+    | TRY block (catchClause+ finallyBlock? | finallyBlock)
+    | TRY resourceSpecification block catchClause* finallyBlock?
+    | SWITCH parExpression LBRACE switchBlockStatementGroup* switchLabel* RBRACE
+    | SYNCHRONIZED parExpression block
+    | RETURN drlExpression? SEMI
+    | THROW drlExpression SEMI
+    | BREAK drlIdentifier? SEMI
+    | CONTINUE drlIdentifier? SEMI
+    | YIELD drlExpression SEMI // Java17
+    | SEMI
+    | statementExpression=drlExpression SEMI
+    | switchExpression SEMI? // Java17
+    | identifierLabel=drlIdentifier COLON drlStatement
+    ;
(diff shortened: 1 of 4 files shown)


tkobayas:
`org.drools.model.codegen.execmodel.MvelOperatorsTest#testMatchesWithFunction` is fixed.

Before PR
`drools-model-codegen`
```
[ERROR] Tests run: 2433, Failures: 419, Errors: 2, Skipped: 9
```

After PR
`drools-model-codegen`
```
[ERROR] Tests run: 2433, Failures: 409, Errors: 2, Skipped: 9

File: drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4 [hunk 1]
```diff
@@ -512,3 +512,37 @@ drlVariableInitializer
  drlArrayInitializer
      : LBRACE (drlVariableInitializer (COMMA drlVariableInitializer)* (COMMA)? )? RBRACE
      ;
+
+/* extending JavaParser block */
+drlBlock
```
tkobayas, line 517: String literal (e.g. `"*"`) is `DRL_STRING_LITERAL`, so we need to extend JavaParser rules to accept it (eventually. `DRL_STRING_LITERAL` is in `drlLiteral`)

tkobayas:

tkobayas:
Rebased to resolve conflicts.

mariofusco (approved):

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
yurloc: Quick thought: we keep overriding the JavaParser rules again and again... that adds a lot of duplicate code. Is it possible to remove the original rule from JavaParser every time we override one? Simply remove JavaParser rules that are no longer being used since we use the overriding rules.

We could end up with a pretty minimal JavaParser at the end or perhaps get rid of it completely.

Or is this impossible because we still need to parse plain old Java code in the RHS?

tkobayas: @yurloc That's a great point. Let me try...

> Or is this impossible because we still need to parse plain old Java code in the RHS?

It's okay. RHS is just retrieved as text by the parser. For RHS execution, non-exec-model uses `JavaParser.java` in `drools-compiler`, exec-model uses the java code directly with some modification.

For DRL editor, we may need another parser rules specific to RHS, but it would be a later story.

tkobayas: @yurloc Ahh, I can't. For example, if I remove `block` in `JavaParser.g4`, other rules in `JavaParser.g4` which depend on `block` fail to process. Even if I use the same name `block` in `DRLParser.g4`, the result is the same because `JavaParser.g4`doesn't look at `DRLParser.g4`.

A possible approach would be to fully copy the content of `JavaParser.g4` into `DRLParser.g4` and modify rules in the single file. It would be worth trying, but it would be another separate issue, WDYT?

yurloc: I understand. We could only remove "entry-point" rules that are not depended on by other rules, which is currently not the case.

Let's revisit this idea when we're done fixing the failing tests and we see the final shape of DRLParser and JavaParser.

drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
```diff
@@ -193,7 +194,7 @@ public FunctionDescr visitFunctiondef(DRLParser.FunctiondefContext ctx) {
                 functionDescr.addParameter(typeTypeContext.getText(), variableDeclaratorIdContext.getText());
             });
         }
-        functionDescr.setBody(getTextPreservingWhitespace(ctx.block()));
+        functionDescr.setBody(stripBracesFromBlock(getTextPreservingWhitespace(ctx.drlBlock())));
```
yurloc:
```suggestion
        functionDescr.setBody(getTextPreservingWhitespace(ctx.drlBlock().drlBlockStatement()));
```
@tkobayas by using block statements instead of the whole block, we could avoid "manually" trimming away the braces.

Of course, this requires a new whitespace-text-preserving function that works on a list of contexts, something like this:
```java
    public static String getTextPreservingWhitespace(List<? extends ParserRuleContext> ctx) {
        return ctx.stream().map(Antlr4ParserStringUtils::getTextPreservingWhitespace).collect(Collectors.joining());
    }
```
I find this simpler to understand and less error-prone (and maybe also reusable for other use cases) than the low-level string manipulation in `stripBracesFromBlock()`.

Wdyt?

yurloc (approved):
One suggestion to avoid low-level string manipulation. Otherwise LGTM.

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
tkobayas: Filed https://github.com/apache/incubator-kie-drools/issues/5799

drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
tkobayas: @yurloc Thank you! Applied the suggestion.

yurloc:
Overall fixing progress: There were 396 failed tests.

PR #5794, opened 2024-03-19
[incubator-kie-drools#5709] [new-parser] Some rules do not fire in Mu…

tkobayas:
…ltiKieBaseTest

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5709

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
+++ b/drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
@@ -268,6 +268,7 @@ public class DRLVisitorImpl extends DRLParserBaseVisitor<Object> {
             ruleDescr.setConsequence(trimThen(getTokenTextPreservingWhitespace(ctx.rhs(), tokenStream))); // RHS is just a text
         }
 
+        populateStartEnd(ruleDescr, ctx);
         return ruleDescr;
     }
 
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -780,6 +780,9 @@ class MiscDRLParserTest {
 
         assertThat(rule.getName()).isEqualTo("simple_rule");
 
+        assertThat(rule.getStartCharacter()).isEqualTo(803);
+        assertThat(rule.getEndCharacter()).isEqualTo(996);
+
         assertThat(rule.getConsequenceLine()).isEqualTo(22);
         assertThat(rule.getConsequencePattern()).isEqualTo(2);
 


tkobayas:
Fixed `org.drools.model.codegen.execmodel.MultiKieBaseTest#testHelloWorldWithPackagesAnd2KieBases`. Now all tests in `MultiKieBaseTest` are successful.

Before PR
`drools-model-codegen`
```
[ERROR] Tests run: 2433, Failures: 336, Errors: 0, Skipped: 9
```

After PR
`drools-model-codegen`
```
[ERROR] Tests run: 2433, Failures: 335, Errors: 0, Skipped: 9
```

drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
```diff
@@ -268,6 +268,7 @@ public RuleDescr visitRuledef(DRLParser.RuledefContext ctx) {
             ruleDescr.setConsequence(trimThen(getTokenTextPreservingWhitespace(ctx.rhs(), tokenStream))); // RHS is just a text
         }
 
+        populateStartEnd(ruleDescr, ctx);
```
tkobayas, line 271: The root cause was that `RuleDescr.getStartCharacter()` returned `-1` (missed in `DRLVisitorImpl`), so the kbase was not updated (the error was logged). So fixed.

https://github.com/apache/incubator-kie-drools/blob/dev-new-parser/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java#L247

tkobayas:

drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
gitgabrio: Hi @tkobayas , thanks for the PR!
IIUC the comment, it means that a [piece of code](https://github.com/apache/incubator-kie-drools/blob/dev-new-parser/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java#L247) does not populate correctly a given object (`ruleDescr`), and your modification somehow "workaround" that.
I read the [method](https://github.com/apache/incubator-kie-drools/blob/5f83f08f169c0c4922e7bd842ae612c607f96dc8/drools-compiler/src/main/java/org/drools/compiler/kie/util/ChangeSetBuilder.java#L223), and TBH is somehow "brittle" by itself (I may give detailed explanation of this remark, if you want): would not be better to fix the root problem itself ? Eventually, I would be glad to help there

gitgabrio (changes requested):
Thanks @tkobayas 
but IIUC, the problem is elsewhere (please correct me if I'm wrong)

mariofusco (approved):

gitgabrio:
@mariofusco @baldimir @yesamer 
Hi folks. This PR, accidentally, uncover two kinds of issues we are currently ignoring:

1. technical debt, that instead of being addressed is workarounded
2. increase of technical debt, polluting the code of such "workarounds"

IMO we have to to change the approach and the mindset, striving to improve the code whenever we have the chance: wdyt ?

tkobayas:
@gitgabrio Thank you for the suggestion.

`startCharacter` and `endCharacter` are Descr's properties, so populating them is not a workaround, rather a right thing to do from the parser point of view.

Having said that,

- This PR is okay as-is

and

- It's great to refactor/improve the `ChangeSetBuilder.diffDescrs` method, so could you file it as a separate issue?

Does it make sense?

Thanks!

gitgabrio:
@tkobayas 
if 

> startCharacter and endCharacter are Descr's properties

why they are not already populated at that specific point, and the proposed modification has to populate them  ? 

> populating them is not a workaround, rather a right thing to do from the parser point of view.


Yes, I understand that "from the parser point of view" this could be the right fix, but IIUC the problem is somewhere else (please correct me if I'm wrong, I'm simply inferring from your own [comment](https://github.com/apache/incubator-kie-drools/pull/5794#discussion_r1529778339)) and we, as communtiy, have to manage all the codebase as a whole.
As usual, I may be wrong on all my remarks, and I can't wait to talk directly with you, to get to a better understanding.

tkobayas:
> if
> 
> >   startCharacter and endCharacter are Descr's properties
> 
> why they are not already populated at that specific point, and the proposed modification has to populate them ?

`startCharacter` and `endCharacter` should be populated when the RuleDescr is created, and the PR fix does it --- in the same method `DRLVisitorImpl.visitRuledef`.

Talking about the root cause "why we had this bug", I think of

- We can improve the way to create Descr objects not to miss populating the properties in `DRLVisitorImpl`
- Add more test cases

In this regard, I filed https://github.com/apache/incubator-kie-drools/issues/5798

> but IIUC the problem is somewhere else (please correct me if I'm wrong, I'm simply inferring from your own comment) 

My comment:

> The root cause was that RuleDescr.getStartCharacter() returned -1 (missed in DRLVisitorImpl), so the kbase was not updated (the error was logged). So fixed.

meant that the `RuleDescr.startCharacter` was `-1` at the point. It had to be populated by `DRLVisitorImpl` beforehand. So I wanted to explain that the root cause was `DRLVisitorImpl`, not `ChangeSetBuilder.diffDescrs`.

> I can't wait to talk directly with you, to get to a better understanding.

Sure, we can talk on meet or zoom.

gitgabrio (approved):
@tkobayas 
Many thanks for the chat and the explanation: it clarifies things a lot!
With that in place, I see this PR is good to merge.
Thanks again!

tkobayas:
@gitgabrio Thanks!

gitgabrio:
@mariofusco @tkobayas @baldimir @yesamer 
Follow up [ticket](https://github.com/apache/incubator-kie-issues/issues/1026)

yurloc:
Overall fixing progress: There were 407 failed tests.

Issue #5818, opened 2024-04-02
[new-parser] Parsing fails if a Java keyword appears in a qualified name


yurloc:
## Parent issue
- #5678 

## Failing tests
- `org.kie.declarativetypes.JavaBeansEventRoleTest#testImportBean`
- `org.drools.mvel.integrationtests.ExtendsTest#testDeclareExtendsWithFullyQualifiedName`

## Description
- `package` translates to the `PACKAGE` token, which is unexpected in the package's qualified name.
- The rest of the input is ignored and the parser **does not report an error**!
  - This is no longer true thanks to #5930. The parser now reports an error as expected.

## Rule code snippet
```
package some.other.package;
import org.drools.compiler.Bean;
declare Bean
  @role(event)
end
```

## Error output
Updated:
```
### parse : ANTLR4_PARSER_ENABLED = true
line 1:18 mismatched input '.' expecting {<EOF>, 'unit', 'function', 'global', 'declare', 'rule', 'query', 'salience', 'enabled', 'no-loop', 'auto-focus', 'lock-on-active', 'refract', 'direct', 'agenda-group', 'activation-group', 'ruleflow-group', 'date-effective', 'date-expires', 'dialect', 'calendars', 'timer', 'duration', 'import'}
14:07:57.000 [main] ERROR o.d.c.k.b.impl.AbstractKieProject.buildKnowledgePackages:280 - Unable to build KieBaseModel:defaultKieBase
[1,18]: mismatched input '.' expecting {<EOF>, 'unit', 'function', 'global', 'declare', 'rule', 'query', 'salience', 'enabled', 'no-loop', 'auto-focus', 'lock-on-active', 'refract', 'direct', 'agenda-group', 'activation-group', 'ruleflow-group', 'date-effective', 'date-expires', 'dialect', 'calendars', 'timer', 'duration', 'import'}
[0,0]: Parser returned a null Package


org.kie.api.builder.CompilationErrorsException: Unable to create KieModule, Errors Existed: [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=bean1.drl, line=1, column=0
[...]

PR #5819, opened 2024-04-02
[new-parser] Allow package to be used as an identifier

yurloc:
Fixes
- #5818 

### DRAFT
Let's discuss whether this is a good approach.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
+++ b/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
@@ -198,6 +198,7 @@ drlIdentifier
     | RECORD
     | VAR
     | THIS
+    | PACKAGE
     ;
 
 drlKeywords
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -3978,4 +3978,24 @@ class MiscDRLParserTest {
         assertThat(query.getParameterTypes()).isEmpty();
         assertThat(query.getParameters()).isEmpty();
     }
+
+    @Test
+    public void packageInQualifiedName() throws Exception {
+        String pkgName = "org.drools.package";
+        String text = "package " + pkgName + "\n" +
+                "rule R\n" +
+                "when\n" +
+                "  $p : Person()\n" +
+                "then\n" +
+                "end\n";
+
+
+        PackageDescr pkg = parser.parse(text);
+        assertThat(parser.hasErrors()).as(parser.getErrorMessages().toString()).isFalse();
+
+        assertThat(pkg.getName()).isEqualTo(pkgName);
+        assertThat(pkg.getRules()).hasSize(1);
+
+        assertThat(pkg.getRules().get(0).getName()).isEqualTo("R");
+    }
 }


drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
```diff
@@ -198,6 +198,7 @@ drlIdentifier
     | RECORD
     | VAR
     | THIS
+    | PACKAGE
```
yurloc, line 201: @tkobayas @mariofusco @gitgabrio Together with `THIS` on the line above, this is just the tip of the iceberg. If we want to keep the new parser 100% backward compatible, we need to A) also add all the other Java keywords, for example, `long`, `break`, `for`, etc. or B) revert closer to the old approach and stop using JavaLexer that defines all these keyword tokens.

B) means using just a few tokens for operators, parentheses, and literals of various types, and `IDENTIFIER` for everything else on the Lexer level. The parser then needs to use semantic predicates (e.g. `helper.validateIdentifierKey(DroolsSoftKeywords.PACKAGE)` to see whether the next `IDENTIFIER` token actually contains `package`).

A) means listing all JavaLexer tokens for various Java keywords here and also replacing nearly all usages of `IDENTIFIER` with `drlIdentifier`.

tkobayas: I think the original test failure https://github.com/apache/incubator-kie-drools/issues/5818 `org.kie.declarativetypes.JavaBeansEventRoleTest#testImportBean` is a very exceptional case.

```
        String drl = "package some.other.package;\n" +
                "import org.drools.compiler.Bean;\n" +
                "declare Bean\n" +
                "  @role(event)\n" +
                "end";
```

The test doesn't fail with the old parser, because the package `some.other.package` doesn't have anything to compile actually. If a user adds a rule to the package, it would fail on the java compilation phase even with the old parser. 

DRL is more or less based on Java syntax, so I think, theoretically, it's fine to fail such a java keyword usage (which normally fails with java compiler) at the parse phase with the new parser (As written in #5818, no-error-report is another issue, though). In other words, fix the test case not to use `package` in `package`.

C) Keep excluding java keywords from `drlIdentifier`

It's against the policy "100% backward compatible with the old parser", but sounds acceptable. WDYT?

(We may face a situation that we will need to choose option A or B in the future, though)

gitgabrio: Hi @yurloc @tkobayas @mariofusco 
sorry I could provide only very general observations:

1. I would stick with 100% backward compatibility; without that, something that was working previously, simply stop to work, and it is not clear who/how would fix the issues that will popup, nor how much time this could take
2. I wonder (again, just asking) if there is not a programmatic way to automatically add all the java keywords from JavaLexer to `drlIdentifier : drlKeywords` - and then keep going with option `A` (IIUC)

Wdyt ? Am I missing something ?

tkobayas: > 1. I would stick with 100% backward compatibility; without that, something that was working previously, simply stop to work, and it is not clear who/how would fix the issues that will popup, nor how much time this could take

Okay, I'm not opinionated on this. We can go with Option A and leave Option C as a future enhancement possibility.

> 2. I wonder (again, just asking) if there is not a programmatic way to automatically add all the java keywords from JavaLexer to `drlIdentifier : drlKeywords` - and then keep going with option A (IIUC)

There is no programmatic way because those Java keywords are individually declared in `JavaLexer.g4`. I think explicitly listing them in `drlIdentifier` is readable enough.

Thanks,

gitgabrio: ok @tkobayas 
Anyway, just remember that copying/duplicating stuff creates problem of maintanability, not readability.
Anyway, fine for me 👍

yurloc:
Superseded by #5958.

PR #5827, opened 2024-04-04
[incubator-kie-drools-5798] [new-parser] Ensure Descr properties

tkobayas:
- Use DescrHelper

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5798

No change in `drools-model-codegen` test result.

Just refactoring and adding unit tests.

* paste the link(s) from GitHub here
* link 2
* link 3 etc.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/DescrCommonPropertyTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/DescrCommonPropertyTest.java
@@ -1,0 +1,405 @@
+package org.drools.drl.parser.antlr4;
+
+import org.drools.drl.ast.descr.AccumulateDescr;
+import org.drools.drl.ast.descr.AccumulateImportDescr;
+import org.drools.drl.ast.descr.AndDescr;
+import org.drools.drl.ast.descr.AnnotationDescr;
+import org.drools.drl.ast.descr.BaseDescr;
+import org.drools.drl.ast.descr.BehaviorDescr;
+import org.drools.drl.ast.descr.CollectDescr;
+import org.drools.drl.ast.descr.EntryPointDeclarationDescr;
+import org.drools.drl.ast.descr.EntryPointDescr;
+import org.drools.drl.ast.descr.EvalDescr;
+import org.drools.drl.ast.descr.ExistsDescr;
+import org.drools.drl.ast.descr.ExprConstraintDescr;
+import org.drools.drl.ast.descr.ForallDescr;
+import org.drools.drl.ast.descr.FromDescr;
+import org.drools.drl.ast.descr.FunctionDescr;
+import org.drools.drl.ast.descr.FunctionImportDescr;
+import org.drools.drl.ast.descr.GlobalDescr;
+import org.drools.drl.ast.descr.ImportDescr;
+import org.drools.drl.ast.descr.NotDescr;
+import org.drools.drl.ast.descr.OrDescr;
+import org.drools.drl.ast.descr.PackageDescr;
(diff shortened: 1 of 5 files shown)


tkobayas:
Hi @yurloc  @gitgabrio  @mariofusco 

I think of some approaches for this refactoring and not very sure which one is the best. I'd like to hear your thoughts.

The point is to make sure `Descr`'s common properties (e.g. `startCharacter`) are set. The current implementation explicitly calls `DRLVisitorImpl.populateStartEnd()` to populate them. So we'd like to do it implicitly.

A) Introduce a constructor `BaseDescr(BasicDescrContext ctx)` where `BasicDescrContext` is a new interface to hold basic information like `startCharacter`. We can specify a super class of generated Context classes by an option `contextSuperClass=DRLParserRuleContext` (https://github.com/antlr/antlr4/blob/master/doc/options.md#contextsuperclass). So we can pass the context class to the constructor without making `drools-drl-ast` depending on antlr4. The PR is https://github.com/apache/incubator-kie-drools/pull/5826

B) Introduce a helper method like `DescrHelper.init(new PackageDescr(), ctx)`. This doesn't hide a Descr constructor, so is not much different from the current code (= calling `populateStartEnd()`). However, new contributors will understand how to initialize the Descr object and make sure common properties are set. It is this PR. https://github.com/apache/incubator-kie-drools/pull/5827

C) Introduce a factory method like `DescrHelper.createPackageDescr(ctx)` replacing the option B's `DescrHelper.init(new PackageDescr(), ctx)`. It can hide the Descr constructor. Instead, `DescrHelper` would have a bunch of similar factory methods. e.g.
```
pablic static PackageDescr createPackageDescr(DRLParser.CompilationUnitContext ctx) {
    return populateCommonProperties(new PackageDescr(), ctx);
}

pablic static RuleDescr createRuleDescr(DRLParser.RuledefContext ctx) {
    return populateCommonProperties(new RuleDescr(), ctx);
}
...
```
I haven't wrote this version yet, but it's easy to write it on top of option B.

---

Option A looks smarter. But actually, I had to modify many codes and also I still need `DescrHelper` for some special cases (root AndDescr, PatternBind). 

Option B and C may not be smart, but I think it's easier to understand/maintain. I'm leaning towards option C.

WDYT?

gitgabrio:
Thanks @tkobayas , good stuff!

I've made a slight modification on top of you B) proposal that somehow mix B) and C)
1. a "Factory" class (with builder pattern) 
2. remove instantiation concern from `DescrHelper`

Reason for that:
1. use common and well-known design pattern
2. clearly define and separate concerns - the factory is responsible of instantiation, the Helper is responsible of manipulation - the two should be somehow independent
3. extensibility: if/ever, there will be need to add a "needed" parameter at instantiation time, the builder pattern allows easy extension (the init method would requires overrides or parameters explosion)


My proposal is [here](https://github.com/gitgabrio/drools/tree/incubator-kie-drools-5798-descr-props-DescrHelper)

@tkobayas @mariofusco @yurloc wdyt ?

tkobayas:
Thank you, @gitgabrio ! I merged your suggestion and it's much better.

Please review this PR, thanks! @yurloc @mariofusco @gitgabrio

gitgabrio (approved):
Thanks @tkobayas LGTM!

mariofusco:
@tkobayas I must admit that when I read the options you listed my first reaction was "Option A) looks smarter and way better, let's go for it". However now I also see that it would introduce some additional complexity and more importantly will require much more time that in this moment we could better use moving forward the parser development faster. This pull request, also thanks to @gitgabrio's improvements, is more pragmatic so it is ok for me.

mariofusco (approved):

PR #5840, opened 2024-04-10
Alter code generation for Custom Operators to reduce new class instances based on Evaluator methods / interfaces

JaredDavis22:
This change allows Custom Operators to avoid new wrapper and new instances being created on each usage.

The test class is weak as it does not verify the wrapper instances are not created. There are a lot of copy/paste/edit blocks in the test code.

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/operatorspec/CustomOperatorSpec.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/operatorspec/CustomOperatorSpec.java
@@ -24,9 +24,17 @@ import com.github.javaparser.ast.expr.MethodCallExpr;
 import org.drools.model.functions.Operator;
 import org.drools.model.codegen.execmodel.generator.RuleContext;
 
+
+
 public class CustomOperatorSpec extends NativeOperatorSpec {
     public static final CustomOperatorSpec INSTANCE = new CustomOperatorSpec();
 
+    /*
+     Three different cases for generating code:
+       1) if EvaluatorDefinition evalDef contains the method getInstance and implements Operator.SingleValue use the getInstance(opName) and do not create a CustomOperatorWrapper
+       2) if EvaluatorDefinition evalDef contains the method getInstance use getInstance(opName) vs creating a new instance of the class
+       3) none of the above :   new CustomOperatorWrapper( .. new evalDef)
+    */
     @Override
     protected Operator addOperatorArgument( RuleContext context, MethodCallExpr methodCallExpr, String opName ) {
         EvaluatorDefinition evalDef = context.getEvaluatorDefinition( opName );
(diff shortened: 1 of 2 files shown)


tkobayas:
Hi @JaredDavis22  , Thank you for raising the PR.

@mariofusco  , could you share your thoughts on this? This is a continuation of https://kie.zulipchat.com/#narrow/stream/381961-drools-dev/topic/Custom.20Evaluator.20Implementation.20Changes

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/operatorspec/CustomOperatorSpec.java
```diff
@@ -24,18 +24,41 @@
...
-        String arg = "new " + CustomOperatorWrapper.class.getCanonicalName() + "( new " + evalDef.getClass().getCanonicalName() + "().getEvaluator(" +
-                ValueType.class.getCanonicalName() + ".OBJECT_TYPE, \"" + opName + "\", false, null), \"" + opName + "\")";
+        String arg;
+        try {
+            evalDef.getClass().getMethod("getInstance",String.class);
+            boolean needWrapper = !Operator.SingleValue.class.isAssignableFrom(evalDef.getClass());
+            if (needWrapper) {
+                // case 2
+                arg = "new " + CustomOperatorWrapper.class.getCanonicalName() + "( "+ evalDef.getClass().getCanonicalName() + ".getInstance(\"" + opName+ "\").getEvaluator(" +
+                        ValueType.class.getCanonicalName() + ".OBJECT_TYPE, \"" + opName + "\", false, null), \"" + opName + "\")";
```
tkobayas, line 52: From my point of view, do we need this `case 2`? I understand that there might be a case where a user creates a custom operator with `getInstance` and not implementing `Operator.SingleValue`. But I guess it's very rare... @JaredDavis22 Do you expect it will happen in your application?

tkobayas:
Hmm, while playing with the test case, I found an easier way to register a custom operator (exec-model only).
https://github.com/apache/incubator-kie-drools/pull/5842/files

@mariofusco , Am I doing something wrong?

mariofusco (commented):
If we want to avoid the repeated creation of those operator instances I believe there's a better (even if maybe slightly more complicated) solution which is creating the instance field for that operator in the package model and always pass that same instance as argument to the MethodCallExpr. In this way you're ensured that it is created only once and you don't do that ugly reflection on the INSTANCE field and it works regardless if such field exists or not, because you generate your own field. 

Is the solution that I'm suggesting clear? Should I try to implement it on my own or do you want to give it a try?

mariofusco (changes requested):
If we want to avoid the repeated creation of those operator instances I believe there's a better (even if maybe slightly more complicated) solution which is creating the instance field for that operator in the package model and always pass that same instance as argument to the MethodCallExpr. In this way you're ensured that it is created only once and you don't do that ugly reflection on the INSTANCE field and it works regardless if such field exists or not, because you generate your own field.

Is the solution that I'm suggesting clear? Should I try to implement it on my own or do you want to give it a try?

mariofusco:
I implemented the alternative solution that I suggested here https://github.com/apache/incubator-kie-drools/pull/5844

mariofusco:
Replaced by https://github.com/apache/incubator-kie-drools/pull/5844

PR #5860, opened 2024-04-22
[KIE-1106] allow to configure KieBase and KieSession also via applica…

### LINKED ISSUE #1106: Allow to configure KieBase and KieSession also via application.properties in quarkus extension

mariofusco:
Closes https://github.com/apache/incubator-kie-issues/issues/1106


tkobayas (approved):

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/project/KieSessionModelBuilder.java
```diff
@@ -45,8 +45,12 @@ public class KieSessionModelBuilder {
     private DroolsModelBuildContext context;
 
     public KieSessionModelBuilder(DroolsModelBuildContext context, Collection<CodegenPackageSources> packageSources) {
+        this(context, packageSources, null);
```
gitgabrio: hi @mariofusco 
One question: why this new constructor is needed, now ?
Beside, could you pls refactor those two constructors ?
1. instead of passing `null`, invokes ` KieModuleModelWrapper.fromResourcePaths(context.getAppPaths().getPaths()).kieBaseModels();`
2. in the next constructor, instead of putting an `if` logic, enforce the needs of `Map<String, KieBaseModel> kieBaseModels`
That way
1. the code is clearer and less `if null` dependent
2. the meaning of the two constructors is better defined
wdyt ?

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/project/RuleCodegen.java
```diff
@@ -110,12 +110,19 @@ public RuleCodegen withHotReloadMode() {
...
+        return this;
+    }
+
     public Collection<PackageModel> getPackageModels() {
         return packageModels;
     }
     
     public Collection<KieBaseModel> getKmoduleKieBaseModels() {
-        return kmoduleKieBaseModels;
+        return kieBaseModels == null ? null : kieBaseModels.values();
```
gitgabrio: Would it be possible to avoid returning `null`, that it is extremely `NullPointerException` prone, and instead returns an empty collection ?

drools-quarkus-extension/drools-quarkus-integration-test/src/main/resources/application.properties
```diff
@@ -17,4 +17,33 @@
 # under the License.
 #
 
-drools.prototypes=allowed
\ No newline at end of file
+drools.prototypes=allowed
```
gitgabrio, line 20: @mariofusco 
Would it be possible to have two integration tests, one that verify kmodule.xml, and the other that verifies application.properties ?
Beside - what would happen if both application.properties AND kmodule.xml are defined ?

gitgabrio (changes requested):

baldimir (approved):

gitgabrio (approved):
thanks, LGTM!

PR #5869, opened 2024-04-25
[incubator-kie-drools-5868] exec-model : in operator with BigDecimal …

tkobayas:
…and int

- Enhance InOperator to handle numeric equals
- Convert BigDecimalLiteralExpr for OperatorSpec

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5868


drools-model/drools-canonical-model/src/main/java/org/drools/model/operators/InOperator.java
```diff
@@ -20,16 +20,16 @@
...
 
     INSTANCE;
 
     @Override
     public boolean eval( Object a, Object[] bs ) {
         for (Object b : bs) {
-            if (a == null && b == null) {
-                return true;
-            } else if (a != null && a.equals( b )) {
+            if (areEqual(a, b)) {
```
tkobayas, line 32: `OperatorUtils.areEqual` supports numeric equal.

baldimir (approved):

mariofusco (approved):

PR #5877, opened 2024-04-29
[KIE-1133] when available always use the fully qualified name of a type during parsing

### LINKED ISSUE #1133: Wrong exec model generation of consequence when using 2 classes with same names in different packages
When the following conditions are all met

- use of mvel dialect
- enough rules to trigger parallel exec model generation
- 2 classes with same names in different packages are used in the same consequence 

the exec model generation may fail attributing the wrong fully qualified name to one of those 2 classes.

mariofusco:
Closes https://github.com/apache/incubator-kie-issues/issues/1133


baldimir (approved):

tkobayas (approved):
Thanks!

PR #5884, opened 2024-05-04
[new-parser] Allow OOPath as a LHS pattern

### LINKED ISSUE #5878: [new-parser] OOPath not supported as a LHS pattern
## Parent issue
- #5678 

## Failing tests
- `org.drools.ruleunits.impl.RuleUnitProviderImplTest#notWithAndWithoutSingleQuote`

## Rule code snippet
```
rule PlainNot when
    not( /strings [ this == "It Does Work" ] )
then
    strings.add("It Does Work");
end
```

## Error output
```
### parse : ANTLR4_PARSER_ENABLED = true
13:16:15.451 [main] ERROR o.d.c.k.b.impl.AbstractKieProject.buildKnowledgePackages:280 - Unable to build KieBaseModel:defaultKieBase
InvalidExpressionErrorResult: Unable to find class: not


org.kie.api.builder.CompilationErrorsException: Unable to create KieModule, Errors Existed: [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=/var/home/jlocker/src/github.com/kiegroup/drools/drools-ruleunits/drools-ruleunits-impl/target/test-classes/org/drools/ruleunits/impl/NotTestUnit.drl, line=-1, column=0
   text=Unable to find class: not]]

	at org.drools.compiler.kie.builder.impl.KieBuilderImpl.getKieModule(KieBuilderImpl.java:526)
	at org.drools.compiler.kie.builder.impl.KieBuilderImpl.getKieModule(KieBuilderImpl.java:512)
	at org.drools.ruleunits.impl.RuleUnitProviderImpl.createRuleUnitKieModule(RuleUnitProviderImpl.java:118)
	at org.drools.ruleunits.impl.RuleUnitProviderImpl.generateRuleUnit(RuleUnitProviderImpl.java:86)
	at org.drools.ruleunits.impl.RuleUnitProviderImpl.getRuleUnit(RuleUnitProviderImpl.java:81)
	at org.drools.ruleunits.api.RuleUnitProvider.createRuleUnitInstance(RuleUnitProvider.java:44)
	at org.drools.ruleunits.impl.RuleUnitProviderImplTest.notWithAndWithoutSingleQuote(RuleUnitProviderImplTest.java:63)
```

yurloc:
- Fixes #5878.
- Makes `drools-ruleunits/drools-ruleunits-impl` build successful.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -4091,6 +4091,28 @@ class MiscDRLParserTest {
     }
 
     @Test
+    void ooPathLhsPattern() {
+        final String text = "package org.drools\n"
+                + "rule PlainNot when\n"
+                + "    not( /strings [ this == \"It Does Work\" ] )\n"
+                + "then\n"
+                + "end\n";
+        PackageDescr packageDescr = parseAndGetPackageDescr(text);
+        RuleDescr ruleDescr = packageDescr.getRules().get(0);
+        assertThat(ruleDescr.getLhs().getDescrs().get(0)).isInstanceOfSatisfying(NotDescr.class, notDescr -> {
+            assertThat(notDescr.getDescrs()).hasSize(1);
+            assertThat(notDescr.getDescrs().get(0)).isInstanceOfSatisfying(PatternDescr.class, patternDescr -> {
+               assertThat(patternDescr.getConstraint().getDescrs()).hasSize(1);
+               assertThat(patternDescr.getConstraint().getDescrs().get(0)).isInstanceOfSatisfying(ExprConstraintDescr.class, exprConstraintDescr -> {
+                   assertThat(exprConstraintDescr.getExpression()).isEqualTo("/strings [ this == \"It Does Work\" ]");
+                   assertThat(exprConstraintDescr.getType()).isEqualTo(ExprConstraintDescr.Type.NAMED);
+                   assertThat(exprConstraintDescr.getPosition()).isEqualTo(0);
+               });
+            });
+        });
+    }
+
+    @Test
     void inlineCast() {
         final String text = "package org.drools\n" +
                 "rule R1\n" +
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -26,6 +26,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


yurloc:
@mariofusco @tkobayas @gitgabrio please review.

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
```diff
@@ -4090,6 +4092,28 @@ void ooPathMixedWithStandardConstraint() {
...
+    @Test
+    void ooPathLhsPattern() {
+        final String text = "package org.drools\n"
+                + "rule PlainNot when\n"
+                + "    not( /strings [ this == \"It Does Work\" ] )\n"
+                + "then\n"
+                + "end\n";
+        PackageDescr packageDescr = parseAndGetPackageDescr(text);
+        RuleDescr ruleDescr = packageDescr.getRules().get(0);
+        assertThat(ruleDescr.getLhs().getDescrs().get(0)).isInstanceOfSatisfying(NotDescr.class, notDescr -> {
```
gitgabrio, line 4103: HI @yurloc 
out of curiosity:
1. what will be the printed message in case of failure in some of those nested assertions ?
2. IINW, this test verify only the positive case (i.e. it must not fails): is there a "negative" test to verify the behavior in case of "wrong" situations  (this is a theoretical POV question, not sure if it make sense in that context) ?


Pls notice the above questions are not strictly related to this test, just a general observation

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
```diff
@@ -4605,4 +4629,46 @@ public void enumDeclaration() {
...
+                pkg.getFunctionImports(),
+                pkg.getAccumulateImports(),
+                pkg.getGlobals(),
+                pkg.getFunctions(),
+                pkg.getRules(),
+                pkg.getAttributes()
+        ).flatMap(Collection::stream);
+
+        assertThat(childrenWithNamespace)
+                .hasSize(9)
```
gitgabrio: Hi @yurloc 
I know this is something common, but this `9`is a "magic number" without any description/indication of where it comes from - for readibility improvement it could be better to declare it as a variable with a meaningful name, or description: wdyt ?

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
```diff
@@ -4605,4 +4629,46 @@ public void enumDeclaration() {
...
+
+        Stream<? extends BaseDescr> childrenWithoutNamespace = Stream.of(
+                pkg.getTypeDeclarations(),
+                pkg.getEnumDeclarations(),
+                pkg.getEntryPointDeclarations(),
+                pkg.getWindowDeclarations()
+        ).flatMap(Collection::stream);
+
+        assertThat(childrenWithoutNamespace)
+                .hasSize(4)
```
gitgabrio: see previous comment

gitgabrio (commented):
Thanks @yurloc - I have just some minor comments/questions

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
yurloc: > 1. what will be the printed message in case of failure in some of those nested assertions ?

The nested assertions work just like regular top-level assertions. There is no difference in the failure message.

The benefit of this nesting is that the lambda parameter (`notDescr ->`) type is the same class I'm using to do the instanceof check. So it saves me from casting the checked expression and declaring a new variable with the narrowed type if I want to do more assertions on it that require the specific type. The test code _should_ be smaller and easier to read.

This is an equivalent code without nesting:
```java

        PackageDescr packageDescr = parseAndGetPackageDescr(text);
        RuleDescr ruleDescr = packageDescr.getRules().get(0);
        NotDescr notDescr = (NotDescr) ruleDescr.getLhs().getDescrs().get(0);
        assertThat(notDescr.getDescrs()).hasSize(1);
        PatternDescr patternDescr = (PatternDescr) notDescr.getDescrs().get(0);
        assertThat(patternDescr.getConstraint().getDescrs()).hasSize(1);
        ExprConstraintDescr exprConstraintDescr = (ExprConstraintDescr) patternDescr.getConstraint().getDescrs().get(0);
        assertThat(exprConstraintDescr.getExpression()).isEqualTo("/strings [ this == \"It Does Work\" ]");
        assertThat(exprConstraintDescr.getType()).isEqualTo(ExprConstraintDescr.Type.NAMED);
        assertThat(exprConstraintDescr.getPosition()).isEqualTo(0);
```
Note that I've left out `assertThat(...).isInstanceOf(Some.class);` which are kind of optional. They would simply give a nice failure message instead of a CCE. The same could be said about the `hasSize()` assertions.

**TLDR;** Nested assertions' messages are identical. Using `.isInstanceOfSatisfying()` is a syntactic sugar. Probably just a matter of taste, and can be easily avoided.

yurloc: > 2\. IINW, this test verify only the positive case (i.e. it must not fails): is there a "negative" test to verify the behavior in case of "wrong" situations  (this is a theoretical POV question, not sure if it make sense in that context) ?

Not sure if I get your point 100% correctly but I would say a negative test is not applicable in this context. Yes, this is a positive test in the sense that given this specific input (`...not( /strings [ this == "It Does Work" ] )...`), the parser and the rule context visitor should produce this exact descriptor structure with these exact properties.

Given that, a negative test would make assumptions about the parser output if the input is somehow incorrect or unexpected, right? But that's already covered:
- If the input is wrong (erroneous), the parser should fail and we have tests for that[1].
- If the input is correct but different (unexpected), then it's a different test case.

[1] We have negative (parser error) tests for the expression parser but not for the general DRL parser, which I have found out thanks to your question. I'll file a ticket for that. Thanks!

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
yurloc: Good point. Your question made me think again about the purpose of the assertion and find a different but much better solution. The goal here is to make sure that all children are covered.

mariofusco (approved):

tkobayas (approved):

PR #5888, opened 2024-05-06
[kie-issues#1153] Dispose session in tests.

### LINKED ISSUE #1153: Fix memory leak in ConsequenceTypeTest
The ConsequenceTypeTest(1) test triggers parallel build. When the KieSession is not disposed in the test, it may trigger Out of memory errors during the builds.

Acceptance criteria: 
- KieSession is disposed in the test. 

(1) https://github.com/apache/incubator-kie-drools/blob/946067e2cb03a5065ac11c4e313df731f924c497/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/ConsequenceTypeTest.java#L17

baldimir:
Fixes: https://github.com/apache/incubator-kie-issues/issues/1153

Kie Session needs to be disposed in tests to avoid memory leaks.

### DIFF
--- a/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/ConsequenceTypeTest.java
+++ b/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/ConsequenceTypeTest.java
@@ -43,9 +43,12 @@ public class ConsequenceTypeTest {
 
         final KieBase kbase = KieBaseUtil.getKieBaseFromKieModuleFromDrl("test", kieBaseTestConfiguration, drl.toString());
         final KieSession wm = kbase.newKieSession();
-
-        wm.insert("test");
-        assertThat(wm.fireAllRules()).isEqualTo(ruleNr);
+        try {
+            wm.insert("test");
+            assertThat(wm.fireAllRules()).isEqualTo(ruleNr);
+        } finally {
+            wm.dispose();
+        }
     }
 
     private String generatedRule(int seed) {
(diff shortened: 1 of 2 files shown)


yesamer (approved):

mariofusco (approved):
That's ok, but I wonder if it makes sense to make the `KieSession` to also implement `Autocloseable` and maybe implement the `close()` method just like an alias for the `dispose()` one. I did something similar for the [RuleUnitInstance](https://github.com/apache/incubator-kie-drools/blob/79a21195741c75da2270693c0a933c8d4d31b2b8/drools-ruleunits/drools-ruleunits-api/src/main/java/org/drools/ruleunits/api/RuleUnitInstance.java#L30) and this allows you to use try-with-resource like [this](https://github.com/apache/incubator-kie-drools/blob/79a21195741c75da2270693c0a933c8d4d31b2b8/drools-ruleunits/drools-ruleunits-impl/src/test/java/org/drools/ruleunits/impl/CepTest.java#L41). WDYT?

baldimir:
> That's ok, but I wonder if it makes sense to make the `KieSession` to also implement `Autocloseable` and maybe implement the `close()` method just like an alias for the `dispose()` one. I did something similar for the [RuleUnitInstance](https://github.com/apache/incubator-kie-drools/blob/79a21195741c75da2270693c0a933c8d4d31b2b8/drools-ruleunits/drools-ruleunits-api/src/main/java/org/drools/ruleunits/api/RuleUnitInstance.java#L30) and this allows you to use try-with-resource like [this](https://github.com/apache/incubator-kie-drools/blob/79a21195741c75da2270693c0a933c8d4d31b2b8/drools-ruleunits/drools-ruleunits-impl/src/test/java/org/drools/ruleunits/impl/CepTest.java#L41). WDYT?

Yes, that sounds fine. However I think it would be better done in a separate PR. I will take a look.

baldimir:
Here is another PR with the AutoCloseable functionality https://github.com/apache/incubator-kie-drools/pull/5890.

baldimir:
Jenkins PR check failed on kogito-examples on some Jenkins environment issue. It is unrelated to these changes. So I think this can be merged.

gitgabrio (approved):

PR #5890, opened 2024-05-06
[kie-issues#1154] Make KieSession auto-closeable

### LINKED ISSUE #1154: Make KieSession auto-closeable
KieSession needs to be properly disposed by calling dispose() method on it, when done using it. To make this easier, it would be good if the KieSession supports being used as AutoCloseable in a try-with-resources code block. 

```
try (KieSession kieSession = kieBase.newKieSession()) {
... some code using the KieSession
} 
```

baldimir:
Fixes: https://github.com/apache/incubator-kie-issues/issues/1154

Makes KieSession implement the AutoCloseable interface, so it can be used in try-with-resources code blocks. No new tests added, because KieSession does not have any listener for dispose events. Therefore, some tests were changed to use try-with-resources to dispose the KieSession.


yesamer (approved):

mariofusco (approved):

baldimir:
Jenkins failure is not related to my changes, it is caused by environment.

PR #5921, opened 2024-05-10
[incubator-kie-drools-5918] [new-parser] Annotations can appear in lh…

tkobayas:
…sOr and lhsAnd

**Issue**:
* https://github.com/apache/incubator-kie-drools/issues/5918

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -3991,8 +3991,7 @@ class MiscDRLParserTest {
     }
 
     @Test
-    void andDescrAnnotation() {
-        ParserTestUtils.enableOldParser();
+    void prefixAndDescrAnnotation() {
         final String text =
                 "rule R\n" +
                         "    when\n" +
@@ -4013,7 +4012,7 @@ class MiscDRLParserTest {
     }
 
     @Test
-    void orDescrAnnotation() {
+    void prefixOrDescrAnnotation() {
         final String text =
                 "rule R\n" +
                         "    when\n" +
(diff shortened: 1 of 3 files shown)


drools-drl/drools-drl-parser/src/main/java/org/drools/drl/parser/antlr4/DRLVisitorImpl.java
```diff
@@ -1032,7 +1038,13 @@ private BaseDescr createAndDescr(ParserRuleContext ctx) {
...
             // So, we need to flatten it so that AndDescr has A(), B() and C() as children.
             List<BaseDescr> flattenedDescrs = flattenAndDescr(descrList);
-            flattenedDescrs.forEach(andDescr::addDescr);
+            flattenedDescrs.forEach(descr -> {
+                if (descr instanceof AnnotationDescr annotationDescr) {
+                    andDescr.addAnnotation(annotationDescr);
+                } else {
+                    andDescr.addDescr(descr);
+                }
+            });
```
tkobayas, line 1048: I know there is a duplication in this part and the above one (`visitLhsOr`). But `addDescr` method is not defined in a common parent class of `AndDescr` and `OrDescr`, so I cannot refactor them to a reusable clean private method. (Changing Descr class hierarchy sounds also too much).

tkobayas:

mariofusco (approved):

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
```diff
@@ -3990,6 +3990,50 @@ void lhsPatternAnnotation() {
         assertThat(annotationDescr.getSingleValueAsString()).isEqualTo("!*, age");
     }
 
+    @Test
+    void andDescrAnnotation() {
+        ParserTestUtils.enableOldParser();
```
yurloc: Committed unintentionally, I believe.

yurloc (changes requested):
To be compliant with the old parser, we also need to process annotations from each expression joined by infix OR/AND:
```
rule R
    when
       ( Double() 
         and @Annot1 String() 
         and @Annot2 Integer() ) 
    then
end
```

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
tkobayas: Thank you! Fixed.

tkobayas:
> To be compliant with the old parser, we also need to process annotations from each expression joined by infix OR/AND:

Thank you! I added test cases and also fixed the implementation to collect the annotations.

yurloc (approved):
Perfect! Let's merge.

PR #5941, opened 2024-05-14
[incubator-kie-drools-5924] [new-parser] Remove DRLIncompleteCodeTest

tkobayas:
## Issue
- https://github.com/apache/incubator-kie-drools/issues/5924

```
        DrlParser parser = new DrlParser(LanguageLevelOption.DRL5);
        PackageDescr descr = parser.parse(true, input);
```

`parse(true, ` means that it's in Editor mode. When Editor mode:

-  `parser` method returns non-null PackageDescr even in case of parse error
-  parser provides information for Editor through `parser.getEditorSentences()`.  e.g. the current position is at `Location.LOCATION_RHS`

Here "Editor" means Eclipse plugin and web workbench, which had already retired in Drools 8.

The next DRL Editor aims at LSP based ones. So I think there is no reason to implement the Editor mode and pass `DRLIncompleteCodeTest`. Implementing the Editor mode may not be a large effort, but it would pollute `DRLVisitorImpl` code.


tkobayas:

mariofusco (changes requested):
I'd rather completely remove this test instead of simply flag it as ignored. I don't see how it could be useful in future.

yurloc:
@tkobayas Please consider also removing/ignoring
- `org.drools.mvel.compiler.lang.RuleParserTest#testPackageWithError`
- `org.drools.mvel.compiler.lang.RuleParserTest#testPackageWithError2`

which use the parser in editor mode.

tkobayas:
> I'd rather completely remove this test instead of simply flag it as ignored. I don't see how it could be useful in future.

> Please consider also removing/ignoring
>     * `org.drools.mvel.compiler.lang.RuleParserTest#testPackageWithError`
>     * `org.drools.mvel.compiler.lang.RuleParserTest#testPackageWithError2`
> which use the parser in editor mode.

@mariofusco @yurloc Thanks, both done.

yurloc (approved):

mariofusco (approved):

PR #5953, opened 2024-05-16
[KIE-DROOLS-5943] fix property reactivity in mvel constraint when using redundant pattern binding in constraint

### LINKED ISSUE #5943: Difference when using redundant variable qualifier and not using it (with MVEL runtime, MVEL syntax)
## Description
While testing #5925 I identified a case in which:
```java
    when
        $v:drools.issues.model.vehicles.GasolineVehicle(motor.highOctane == true, score<=1, $v.motor.serialNumber > 50000)
```
and
```java
    when
        $v:drools.issues.model.vehicles.GasolineVehicle(motor.highOctane == true, score<=1, motor.serialNumber > 50000)
```
don't behave the same. Strangely enough, only `motor.serialNumber > 50000` has an effect, qualifying or un-qualifying the other clauses do not make any difference.

Also, only happens with the dynamically generated model. In the process of #5925 I added identical test for both dynamic (byte-buddy) and normally-compiled code, and there was no difference in that issue, but in this one there is.

## How to reproduce
This test
https://github.com/flozano/drools-issues/blob/main/src/test/java/drools/issues/DynamicClassTest.java
Uses these two DRL files:
- https://github.com/flozano/drools-issues/blob/main/src/test/resources/drools/issues/DynamicClassTest.drl
- https://github.com/flozano/drools-issues/blob/main/src/test/resources/drools/issues/DynamicClassTest_static.drl

In the second one (java compiled code), the $v can be added/removed with no issue. In the first one (byte-buddy model, currently ` $v.motor.serialNumber > 50000` is used - and the test pass. If `$v.` is removed from that clause, the test does not pass.

(https://github.com/flozano/drools-issues/commit/3f9c53172e9998905b0ed9e82ce449edda544f2d)

mariofusco:
Fixes https://github.com/apache/incubator-kie-drools/issues/5943

This time the executable model was behaving correctly so I fixed the property reactivity introspection in the `MvelConstraint`, to make the interpreted drl to behave in the same way.

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
@@ -117,6 +117,7 @@ import static java.util.stream.Collectors.toList;
 import static org.drools.model.codegen.execmodel.generator.DslMethodNames.PATTERN_CALL;
 import static org.drools.model.codegen.execmodel.generator.DslMethodNames.isDslTopLevelNamespace;
 import static org.drools.model.codegen.execmodel.generator.expressiontyper.ExpressionTyper.findLeftLeafOfNameExprTraversingParent;
+import static org.drools.util.ClassUtils.actualTypeFromGenerics;
 import static org.drools.util.ClassUtils.toRawClass;
 import static org.drools.util.MethodUtils.findMethod;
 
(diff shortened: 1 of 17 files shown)


drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
```diff
@@ -161,7 +162,7 @@ public static TypedExpression nameExprToMethodCallExpr(String name, java.lang.re
         Method accessor = getAccessor(clazz, name, context);
         if (accessor != null) {
             MethodCallExpr body = new MethodCallExpr( scope, accessor.getName() );
-            return new TypedExpression( body, accessor.getGenericReturnType() );
+            return new TypedExpression( body, actualTypeFromGenerics( type, accessor.getGenericReturnType() ) );
```
mariofusco, line 165: This change is unrelated with the fix of the property reactivity problem but it's a follow up of this issue https://github.com/apache/incubator-kie-drools/issues/5925 fixing it also when the redundant biding is used.

lucamolteni (approved):

tkobayas (approved):

PR #5968, opened 2024-05-22
`for`  expression with a `null` leads to unexpected NPE

### LINKED ISSUE #1246: `for`  expression with a `null` leads to unexpected NPE
In case of a Decision with an empty `iterationVariable` in the `for` expression, a NPE is thrown that shades the real error Message (ie. the `iterationVariable` is missing)

Screenshot 2024-05-22 at 16 38 00

Screenshot 2024-05-22 at 16 38 42

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1246

Screenshot 2024-05-22 at 16 38 00

Screenshot 2024-05-22 at 17 12 14

The NPE is thrown because of ´expression.getTypeRef().toString().toLowerCase()`, in such a case the typeRef is null

Considering that: 
- The `typeRef` is an optional parameter  in the `tExpression` 
Screenshot 2024-05-22 at 17 13 56

- the `in` and `return` variables are passing a constant for creating the error message ([here](https://github.com/apache/incubator-kie-drools/pull/5968/files#diff-2acea59f51fe53327d046e6af624076137435f8d9ff245ebbe96306f6962b8fcR1001)) and ([here](https://github.com/apache/incubator-kie-drools/pull/5968/files#diff-2acea59f51fe53327d046e6af624076137435f8d9ff245ebbe96306f6962b8fcR1008))

To fix the issue it's enough to pass the constants that represent the variable name (`variableIterator`)

@gitgabrio @baldimir Do you have any idea how I can effectively test this change?

After:
Screenshot 2024-05-22 at 16 38 42

Before:
Screenshot 2024-05-22 at 16 54 31


baldimir (approved):

gitgabrio (approved):
Thx! 👍

PR #5969, opened 2024-05-23
[incubator-kie-drools-5938] [new-parser] Tests using wrong duration a…

tkobayas:
…ttribute format

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/5937

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -5216,4 +5216,18 @@ class MiscDRLParserTest {
         assertThat(accumulateFunction.getFunction()).isEqualTo("average");
         assertThat(accumulateFunction.getParams()).containsExactly("$a + $b");
     }
+
+    @Test
+    void durationChunk() {
+        final String text =
+                "rule R\n" +
+                        "  duration (30s 10s) \n" +
+                        "when\n" +
+                        "then\n" +
+                        "end";
+        RuleDescr rule = parseAndGetFirstRuleDescr(text);
+        assertThat(rule.getAttributes()).containsKey("duration");
+        assertThat(rule.getAttributes().get("duration").getType()).isEqualTo(AttributeDescr.Type.EXPRESSION);
+        assertThat(rule.getAttributes().get("duration").getValue()).isEqualTo("30s 10s");
+    }
 }
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -5221,13 +5221,15 @@ class MiscDRLParserTest {
     void durationChunk() {
         final String text =
                 "rule R\n" +
-                        "  duration (30s 10s) \n" +
+                        "  duration (wrong input) \n" +
                         "when\n" +
                         "then\n" +
                         "end";
         RuleDescr rule = parseAndGetFirstRuleDescr(text);
         assertThat(rule.getAttributes()).containsKey("duration");
         assertThat(rule.getAttributes().get("duration").getType()).isEqualTo(AttributeDescr.Type.EXPRESSION);
-        assertThat(rule.getAttributes().get("duration").getValue()).isEqualTo("30s 10s");
+
+        // At the moment, the parser accepts any input and let the compile phase validate it.
+        assertThat(rule.getAttributes().get("duration").getValue()).isEqualTo("wrong input");
     }
 }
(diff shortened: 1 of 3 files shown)


tkobayas:
I choose the approach to meet the same approach as old parser --- use `intOrChunkAttribute`.

tkobayas:

drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
```diff
@@ -5216,4 +5216,18 @@ void doubleAmpersandInfixAndInAccumulate() {
         assertThat(accumulateFunction.getFunction()).isEqualTo("average");
         assertThat(accumulateFunction.getParams()).containsExactly("$a + $b");
     }
+
+    @Test
```
gitgabrio, line 5220: Hi @tkobayas 
IIUC a method has been removed related to this test.
Is there also a "fails" test or similar, i.e. something that demonstrates a managed excpetion, or error, when the duration input si "wrong" ?

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
```diff
@@ -445,7 +445,7 @@ attribute : name=( 'salience' | 'enabled' ) conditionalAttributeValue #expressio
           | name=( 'agenda-group' | 'activation-group' | 'ruleflow-group' | 'date-effective' | 'date-expires' | 'dialect' ) DRL_STRING_LITERAL #stringAttribute
           | name='calendars' DRL_STRING_LITERAL ( COMMA DRL_STRING_LITERAL )* #stringListAttribute
           | name='timer' ( DECIMAL_LITERAL | LPAREN chunk RPAREN ) #intOrChunkAttribute
-          | name='duration' ( DECIMAL_LITERAL | TIME_INTERVAL | LPAREN TIME_INTERVAL RPAREN ) #durationAttribute
+          | name='duration' ( DECIMAL_LITERAL | LPAREN chunk RPAREN ) #intOrChunkAttribute
```
yurloc, line 448: > Hi @tkobayas IIUC a method has been removed related to this test. Is there also a "fails" test or similar, i.e. something that demonstrates a managed excpetion, or error, when the duration input si "wrong" ?


@gitgabrio 

1. Using the `chunk` rule makes the parser extremely benevolent wrt. the duration attribute's value. The input basically cannot be wrong because the `chunk` rule accepts _anything_ between the two parentheses. So no need for a negative test, IMO.
2. Wrt. the removed method - the method name corresponds to the rule alternative's [label](https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#alternative-labels) (string after `#`). In this case, the label has changed from `durationAttribute` to `intOrChunkAttribute` making the `visitDurationAttribute()` method unused. The `visitIntOrChunkAttribute()` method already exists (it handles for example the timer subrule one line above) so no need to add a new method.

yurloc (approved):
I'm fine with the old parser approach.

I would maybe slightly change the duration value used in the test. Provided that tests often play the role of a specification, `30s 10s` might be misleading for a newcomer as it is not clear whether that represents a valid or an invalid value. It could be mistakenly interpreted as if the duration attribute accepted multiple time literals.

Since this specific test proves that any kind of input is accepted on top of valid time literals, I would maybe choose something that is more obviously wrong, for example, `wrong input`.

drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRLParser.g4
gitgabrio: Thanks @yurloc , but now I'm slightly confused. IIUC, the parser does not do any kind of data validation, i.e. anything can be parsed as "DURATION", delegating the verification to the client code; and I have the impression it should not be that way (mind you, I may be wrong, of course, especially in that context)
@mariofusco wdyt ?

tkobayas: Hi @gitgabrio 

> anything can be parsed as "DURATION", delegating the verification to the client code; 

Indeed, the old Antlr3 parser delegates the verification to the compile phase, so the rule build reports an error anyway.

At the moment (= until this new parser passes all existing unit tests), I'd like to follow the same approach as the old Antlr3 parser as possible, so that we can minimize the behaviour difference now.

Then, we will consider any possible improvements. I'll note them in https://github.com/apache/incubator-kie-drools/issues/5972

Does it sound good?

Hi @yurloc . Thank you for the follow-up for the Gabriele's question.

> I would maybe choose something that is more obviously wrong, for example, wrong input.

Thanks, I'll do that.

gitgabrio: HI @tkobayas , thanks for explanation, fine for me!

gitgabrio (approved):

PR #5973, opened 2024-05-24
NO_ISSUE: use release shared library

jstastny-cz:
Depends on:
* apache/incubator-kie-kogito-pipelines#1206

Where `release` share pipeline library was added with gpg specific methods to import keys,etc.


rodrigonull (approved):
LGTM

baldimir (approved):

PR #5976, opened 2024-05-31
[incubator-kie-drools-5920] [new-parser] Strict annotation parsing

tkobayas:
- Skip the test for new parser

**Issue**: 
* https://github.com/apache/incubator-kie-drools/issues/5920

### DIFF
--- a/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/StrictAnnotationTest.java
+++ b/drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/StrictAnnotationTest.java
@@ -32,10 +32,13 @@ import java.util.Properties;
 
 import org.drools.core.common.DefaultEventHandle;
 import org.drools.core.impl.RuleBaseFactory;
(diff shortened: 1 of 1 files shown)


tkobayas:
`LanguageLevelOption.DRL6_STRICT` has 2 points:
1) Parser phase: Annotation should be placed at the beginning of the statement. e.g. 
```
@Xyz rule yyy
when
  ...
```
2) Compiler phase: Annotation should be resolved as Java annotation (= custom annotation java class has to actually exist).

`LanguageLevelOption.DRL6_STRICT`  is not written in document and I haven't seen users using it. Also implementing it would cause complexity which doesn't likely pay off (Old parser maintains 2 parser classes `DRL6Parser` and `DRL6StrictParser`). So I think we can skip the test at the moment, and we will discuss a deprecation plan in 	dev@kie.apache.org later. Of course, if there is a strong need, we will be able to implement it. WDYT? @mariofusco @yurloc @gitgabrio @baldimir

mariofusco (approved):

baldimir (approved):

gitgabrio (approved):

PR #5994, opened 2024-06-17
[incubator-kie-issues#1326] `string join` function with `list` parameter in string format invocation is missing

yesamer:
The function `string join (list <string>, delimiter <string>)` is missing.

Example `string join("a", ".")`

This is not present in the specs, but there are a couple of TCK tests that require it, probably to manage the case of a single-item list. 

Honestly, I don't see the need to manage such a case (i.e. a string join("a", ".") will produce "a") - Adding it just to fix the TCK tests.

Could make sense to discuss in TCK if that function is required? @gitgabrio @baldimir  

@baldimir Provided the original fix.

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StringJoinFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StringJoinFunction.java
@@ -55,6 +55,10 @@ public class StringJoinFunction extends BaseFEELFunction {
         return FEELFnResult.ofResult(sj.toString());
     }
 
+    public FEELFnResult<String> invoke(@ParameterName("list") String stringValue, @ParameterName("delimiter") String delimiter) {
+        return invoke(List.of(stringValue), delimiter);
+    }
+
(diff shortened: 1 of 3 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/extended/StringJoinFunction.java
```diff
@@ -55,6 +55,10 @@ public FEELFnResult<String> invoke(@ParameterName("list") List<?> list, @Paramet
         return FEELFnResult.ofResult(sj.toString());
     }
 
+    public FEELFnResult<String> invoke(@ParameterName("list") String stringValue, @ParameterName("delimiter") String delimiter) {
```
gitgabrio, line 58: Hi @yesamer 
I'm confused about this implementation.
Reading the tests, the Join function works only  when two (or more) Strings are provided. But this method receives only one string and the delimeter, in which case (IINW) it should not do anything...

yesamer:
Ok, let's put this on hold and raise the point at the TCK meeting

yesamer:
Closing, it will be managed in a separate PR.

PR #5999, opened 2024-06-20
[KIE-1339] fix Drools reproducible build also when running the complete test-suite

### LINKED ISSUE #1339: Fix Drools reproducible build also when running the complete test-suite

mariofusco:


drools-core/src/test/java/org/drools/core/util/KeyStoreHelperTest.java
```diff
@@ -52,6 +54,16 @@ public class KeyStoreHelperTest {
     private static final String KEY_PASSWORD = "keypwd";
     private static final String KEY_PHRASE = "secretkey";
 
+    @AfterClass
+    public static void cleanup() {
+        try {
+            new File(KEYSTORE_JCEKS_FILENAME).delete();
```
mariofusco, line 60: This file is created when this test runs, so it has to be deleted in order to avoid that it mistakenly ends up in the test jar.

drools-examples-api/default-kiesession-from-file/src/main/java/org/drools/example/api/defaultkiesessionfromfile/DefaultKieSessionFromFileExample.java
```diff
@@ -80,14 +79,13 @@ public static File getFile(String exampleName) {
...
-            File targetFolder = new File(exampleFolder,
-                                         "target");
+            File targetFolder = new File(exampleFolder, "target");
             if (!targetFolder.exists()) {
                 throw new RuntimeException("The target folder does not exist, please build project " + exampleName + " first");
             }
 
             for (String str : targetFolder.list()) {
-                if (str.startsWith(exampleName) && !str.endsWith("-sources.jar") && !str.endsWith("-tests.jar") && !str.endsWith("-javadoc.jar")) {
+                if (str.startsWith(exampleName) && str.endsWith(".jar") && !str.endsWith("-sources.jar") && !str.endsWith("-tests.jar") && !str.endsWith("-javadoc.jar")) {
```
mariofusco, line 88: This and all subsequent fixes are all identical: the maven task checking the reproducible build adds more files to the target folder, so they have to be filtered away here in order to pick up the correct jar file to be used here.

lucamolteni (approved):

PR #6019, opened 2024-07-09
kie-issues#1381: Increase setup-branch job timeout

rodrigonull:
As part of our tests for the Apache 10 release we found out that the setup-branch Jenkins job was timing out.

This PR increases the job timeout to 180 minutes to avoid such errors.


jstastny-cz (approved):
LGTM.

PR #6024, opened 2024-07-12
[incubator-kie-drools-6016] Emit a warning when an eval is improperly…

### LINKED ISSUE #6016: Emit a warning when an eval is improperly and unnecessarily used
Emit a warning when an eval is improperly and unnecessarily used.

e.g.
```
when
    $a: FirstObject( )
    $b: SecondObject( )
    eval (  finalCheck( $a.someIntValue, $b.anotherInt  ) )
```
to be
```
when
    $a: FirstObject( )
    $b: SecondObject( finalCheck( $a.someIntValue, anotherInt  ) )
```

tkobayas:
… and unnecessarily used

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6016

### DIFF
--- a/drools-base/src/main/java/org/drools/base/rule/EvalCondition.java
+++ b/drools-base/src/main/java/org/drools/base/rule/EvalCondition.java
@@ -32,11 +32,18 @@ import org.drools.base.reteoo.BaseTuple;
 import org.drools.base.rule.accessor.CompiledInvoker;
 import org.drools.base.rule.accessor.EvalExpression;
 import org.drools.base.rule.accessor.Wireable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class EvalCondition extends ConditionalElement
     implements
     Externalizable,
         Wireable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(EvalCondition.class);
+
+    private static long warnLogCounter = 0;
+
     private static final long          serialVersionUID   = 510l;
 
     protected EvalExpression             expression;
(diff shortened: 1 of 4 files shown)


drools-base/src/main/java/org/drools/base/rule/EvalCondition.java
```diff
@@ -223,4 +230,36 @@ public void setCloned(List<EvalCondition> cloned) {
...
     }
+
+    public static void logWarnIfImproperEval(EvalCondition evalCondition, String evalExpression) {
+        if (warnLogCounter == 10) {
+            warnLogCounter++;
+            LOG.warn("More eval warnings will be suppressed...");
+            return;
+        } else if (warnLogCounter > 10) {
+            return; // avoid flooding the logs
+        }
```
tkobayas, line 241: Log a maximum of 10 warnings, because I guess if a user likes to use eval, there would be tons of warnings. 10 logs would be enough to get attention.

drools-base/src/main/java/org/drools/base/rule/EvalCondition.java
```diff
@@ -223,4 +230,36 @@ public void setCloned(List<EvalCondition> cloned) {
...
+        for (Declaration declaration : evalCondition.getRequiredDeclarations()) {
+            if (declaration.getPattern() != null) {
+                sb.append("'");
+                sb.append(declaration.getIdentifier());
+                sb.append("' comes from previous pattern '");
+                String className = declaration.getPattern().getObjectType().getClassName();
+                sb.append(className.substring(className.lastIndexOf('.') + 1));
+                sb.append("'. ");
+            }
+        }
```
tkobayas, line 257: The criteria of the warning is "If the eval uses a declaration which comes from previous pattern". I guess, most of eval usages meet the case. Do you think of other criteria? @mariofusco

mariofusco: No, this is also what I had in mind. I cannot think to any other condition at the moment.

tkobayas: Thanks, I'll go with it.

mariofusco (approved):

baldimir (approved):

pibizza (approved):
LGTM

PR #6030, opened 2024-07-23
[10.0.x] kie-issues#1397: Adjust release pipelines for the Apache 10 release

rodrigonull:
Part of: https://github.com/apache/incubator-kie-issues/issues/1397

### DIFF
--- a/.ci/jenkins/Jenkinsfile.deploy
+++ b/.ci/jenkins/Jenkinsfile.deploy
@@ -26,7 +26,7 @@ deployProperties = [:]
 
 pipeline {
     agent {
-        docker { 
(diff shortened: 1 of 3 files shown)


tkobayas (approved):
LGTM, but one concern, `Jenkinsfile.deploy` calls `commitAndCreatePR()`, so a PR will be created (commit is okay. It's the **R** commit, right?). We don't call `promote` which merges the PR automatically. It means there will be a PR left. I guess we can close it manually (**not merging, according to the release procedure docs**), though. The same would be applied to other repos.

jstastny-cz (approved):
LGTM.

PR #6040, opened 2024-08-01
[Incubator-kie-issues#1345] It shouldn't be possible to add\subtract a DateTime with a number

### LINKED ISSUE #1345: It shouldn't be possible to add\subtract a DateTime with a number
According to the TCK tests, an Add\Substract operation between a DateTime with a number should result to an error

`@"2021-01-01T10:10:10" + 10` 

Currently, the engine returns `"2021-01-111T10:10:10"`

Please notice, `10 + @"2021-01-01T10:10:10"` works correctly (it returns an error)

bncriju:
Fixes https://github.com/apache/incubator-kie-issues/issues/1345

It was possible to add a number with a Temporal and was returning the result, which was not right. According to DMN specification, addition of Temporal with a number shouldn't be supported. This PR has the changes to bring the behavior supported by DMN spec.

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
@@ -21,6 +21,7 @@ package org.kie.dmn.feel.runtime;
 import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -45,15 +46,11 @@ public class FEEL12ExtendedForLoopTest extends BaseFEELTest {
     }
 
     private static Collection<Object[]> data() {
-        List<Integer> nullObject = new ArrayList<>();
-        nullObject.add(null);
-        nullObject.add(null);
-        nullObject.add(null);
         final Object[][] cases = new Object[][] {
           //normal:
           {"for x in [1, 2, 3] return x+1", Stream.of(1, 2, 3 ).map(x -> BigDecimal.valueOf(x + 1 ) ).collect(Collectors.toList() ), null},
                 {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x + @\"P1D\"", Stream.of("2021-01-02", "2021-01-03", "2021-01-04" ).map(LocalDate::parse).collect(Collectors.toList() ), null},
-                {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x+1", nullObject, FEELEvent.Severity.ERROR},
+                {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x+1", Arrays.asList(null, null, null), FEELEvent.Severity.ERROR},
 
           //extended:
           {"for x in 1..3 return x+1", Stream.of(1, 2, 3).map(x -> BigDecimal.valueOf(x + 1)).collect(Collectors.toList()), null},
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
@@ -45,10 +45,15 @@ public class FEEL12ExtendedForLoopTest extends BaseFEELTest {
     }
 
     private static Collection<Object[]> data() {
+        List<Integer> nullObject = new ArrayList<>();
+        nullObject.add(null);
+        nullObject.add(null);
+        nullObject.add(null);
         final Object[][] cases = new Object[][] {
           //normal:
           {"for x in [1, 2, 3] return x+1", Stream.of(1, 2, 3 ).map(x -> BigDecimal.valueOf(x + 1 ) ).collect(Collectors.toList() ), null},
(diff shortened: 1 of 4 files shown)


gitgabrio (approved):
Thx @bncriju well done!!!

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
```diff
@@ -45,10 +45,15 @@ protected void instanceTest(String expression, Object result, FEELEvent.Severity
...
+        List<Integer> nullObject = new ArrayList<>();
+        nullObject.add(null);
+        nullObject.add(null);
+        nullObject.add(null);
         final Object[][] cases = new Object[][] {
           //normal:
           {"for x in [1, 2, 3] return x+1", Stream.of(1, 2, 3 ).map(x -> BigDecimal.valueOf(x + 1 ) ).collect(Collectors.toList() ), null},
-          {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x+1", Stream.of("2021-01-02", "2021-01-03", "2021-01-04" ).map(LocalDate::parse).collect(Collectors.toList() ), null},
+                {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x + @\"P1D\"", Stream.of("2021-01-02", "2021-01-03", "2021-01-04" ).map(LocalDate::parse).collect(Collectors.toList() ), null},
+                {"for x in @\"2021-01-01\"..@\"2021-01-03\" return x+1", nullObject, FEELEvent.Severity.ERROR},
```
yesamer: @gitgabrio Is it expected this test case returns a List of null? WDYT?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELDateTimeDurationTest.java
```diff
@@ -51,6 +51,8 @@ protected void instanceTest(String expression, Object result, FEELEvent.Severity
     private static Collection<Object[]> data() {
         final Object[][] cases = new Object[][] {
                 // date/time/duration function invocations
+                { "@\"2021-01-01\" + 10", null , FEELEvent.Severity.ERROR},
+                { "@\"2021-01-01T10:10:10\" + 10", null , FEELEvent.Severity.ERROR},
```
yesamer, line 55: @bncriju Can you please add here those additional test cases:
- number + Date
- number + DateTime
- number - Date
- number - DateTime
- Date - number 
- DateTime - number

They should all return null with an error.

Just for the sake of completeness :)
Thank you

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
gitgabrio: @yesamer Because with the removal of that

```java
           if (right instanceof BigDecimal bigDecimal) {
                Period toAdd = Period.ofDays(bigDecimal.intValue());
                return temporal.plus(toAdd);
     }
 ```
 
 the case "date + number" is unmanaged, and the evaluation returns `null` - if this was your question.

Comment (yesamer) [reply kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java]: @gitgabrio No, my question wasn't clear, sorry.
The point is:
The `"for x in @\"2021-01-01\"..@\"2021-01-03\" return x+1"`  expression returns a List `(null, null, null)`, but I expected to have a single `null` here, and not a List of nulls. Is that correct?

File: kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java [hunk 3]
```diff
@@ -45,10 +45,15 @@ protected void instanceTest(String expression, Object result, FEELEvent.Severity
     }
 
     private static Collection<Object[]> data() {
+        List<Integer> nullObject = new ArrayList<>();
```
gitgabrio: @bncriju 
Sorry for late review, but could you pls remove this 
`List<Integer> nullObject = new ArrayList<>();`
 and follow the same approach we did elsewhere (i.e.

`Stream.of(null, null, null)` inside the test itself)

Beside: `nullObject` is an extemely misleading term, because
1. this is not a generic object, but a List
2. this is not a `null` `Object`, but a `List` filled with `null`.

gitgabrio (changes requested):
Sry @bncriju - I missed one detail

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEEL12ExtendedForLoopTest.java
gitgabrio: @yesamer 
The `for x....` iterates over all the elements, and for each element the `+1` is executed.
Since the `date + number`  operation returns `null`, every single iteration is mapped to `null`.
So, I'm expecting to have a list of `null`
Very similar to `stream.map`: does this make sense ? Am I clear ?

Please, correct me if I'm wrong

yesamer: @gitgabrio Yes it makes sense, is not the `for` expression that is wrong but its internal statement. Thank you!

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/FEELDateTimeDurationTest.java
yesamer: @bncriju Don't forget this please :)

yesamer (approved):

gitgabrio (approved):

PR #6055, opened 2024-08-22
[incubator-kie-issues#1409]-Fixed matches function behavior

### LINKED ISSUE #1409: `matches()` function wrongly behaves - checks and unicode
TCK Tests revealed some cases where `matches()` wrongly behaves.

1.  `matches("abracadabra", "bra", "p")` should return an error, as the "p" flag doesn't exist; (Please check all the functions where the flags parameter is used e.g. `replace()`)
2. `matches("\u212A", "k", "i")` should return true. \u212A is the [Kelvin Sign](https://codepoints.net/U+212A?lang=en), basically a K. Before fixing that, double-check in the specs that this is the expected behavior 
3. `matches("input", "pattern", " ")` same as point 1.
4. `matches("input", "pattern", "X")` same as point 1.
5.  `matches("hello world", "hello\ sworld", "x")`The `x` is not supported. (Please check all the functions where the flags parameter is used e.g. `replace()`)
6. `matches("h", "(.)\3")` `matches("h", "(.)\2")` and `matches("input", "\3")` should fail. Please investigate why they should fail

bncriju:
Fixes https://github.com/apache/incubator-kie-issues/issues/1409

<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/MatchesFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/MatchesFunction.java
@@ -18,6 +18,7 @@
  */
 package org.kie.dmn.feel.runtime.functions;
 
+import java.security.InvalidParameterException;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


gitgabrio (approved):
Many thanks @bncriju great job!!!

baldimir (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/MatchesFunction.java
```diff
@@ -18,16 +18,21 @@
  */
 package org.kie.dmn.feel.runtime.functions;
 
+import java.security.InvalidParameterException;
```
yesamer, line 21: @bncriju is this import used / correct?

bncriju: @yesamer Whenever input/pattern is null, this is the most appropriate one to be used. it seems. This is used to catch this condition and whenever we verify this case in test as well.

yesamer: @bncriju My point is: that expection is part of the `java.security` package. That means it should be used in the security management scope (as described here https://docs.oracle.com/javase/8/docs/api/java/security/InvalidParameterException.html). 
In our context, [IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) should be used

yesamer:
@bncriju  I run your changes against the TCK tests, and cases 5. and 6. are still not working. Is that intended?

PR #6086, opened 2024-09-17
[DROOLS-7631] unify coercion checks between plain drl and executable model

mariofusco:
Fix for https://issues.redhat.com/browse/DROOLS-7631

The plain drl compilation already reports the compile time error

`Error: Comparison operation requires compatible types. Found class java.time.LocalDateTime and class java.util.Date`

but this doesn't happen when generating the executable model than then fails at runtime with the `ClassCastException` reported in that jira.

The purpose of this pull request is reusing the same code for coercion checks for both plain drl compilation and executable model generation.

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/CoercedExpression.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/CoercedExpression.java
@@ -52,6 +52,8 @@ import static org.drools.model.codegen.execmodel.generator.DrlxParseUtil.toClass
 import static org.drools.model.codegen.execmodel.generator.DrlxParseUtil.toJavaParserType;
 import static org.drools.model.codegen.execmodel.generator.DrlxParseUtil.toStringLiteral;
 import static org.drools.util.ClassUtils.toNonPrimitiveType;
+import static org.drools.util.CoercionUtil.areComparisonCompatible;
+import static org.drools.util.CoercionUtil.areEqualityCompatible;
 
 public class CoercedExpression {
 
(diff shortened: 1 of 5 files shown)


drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/CoercedExpression.java
```diff
@@ -151,9 +153,22 @@ public CoercedExpressionResult coerce() {
             coercedLeft = left;
         }
 
+        checkCoercion(coercedLeft, coercedRight, leftClass, rightClass);
         return new CoercedExpressionResult(coercedLeft, coercedRight, rightAsStaticField);
     }
 
+    private void checkCoercion(TypedExpression coercedLeft, TypedExpression coercedRight, Class<?> leftClass, Class<?> rightClass) {
+        if (equalityExpr) {
```
pibizza, line 161: Since this is known at creation time you can inline the check and put instead a lambda inside the expression? so you would not repeat the check for every invocation of the cohercion.

mariofusco: I believe that this will reduce a bit the readability, and if you are suggesting this only for performance reason I'm pretty sure that invoking a bimorphic capturing lambda will be order of magnitude slower than accessing a local boolean variable.

pibizza: Performance wise I am ignorant and you are probably right. My point is that if something that affects the behavious of an object is known at creation time of an object you should make the object behave accordingly. But I admit it that this is a minor point and we can keep it as it is. thank you.

tkobayas (approved):

PR #6089, opened 2024-09-18
[incubator-kie-drools-6088] Performance issue with many kbase models …

tkobayas:
…and many large spreadsheets

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6088

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/AbstractKieProject.java
@@ -32,6 +32,7 @@ import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
 import org.drools.compiler.kproject.models.KieBaseModelImpl;
 import org.drools.compiler.kproject.models.KieModuleModelImpl;
 import org.drools.compiler.kproject.models.KieSessionModelImpl;
+import org.drools.io.InternalResource;
 import org.drools.util.StringUtils;
 import org.kie.api.builder.Message;
 import org.kie.api.builder.model.KieBaseModel;
@@ -65,6 +66,8 @@ public abstract class AbstractKieProject implements KieProject {
 
     private final Map<String, KieSessionModel>   kSessionModels             = new HashMap<>();
 
+    private Map<InternalResource, String>        packageNameCache           = new HashMap<>();
+
     private static final Predicate<String> BUILD_ALL = s -> true;
 
     public ResultsImpl verify() {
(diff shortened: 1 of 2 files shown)


pibizza (commented):
Ignore comment. Wrong PR:

tkobayas:
Hi @mariofusco 

The performance issue only arises when a KieProject has many kbase models and many large spreadsheets.

In this case, `AbstractKieProject.buildKnowledgePackages` is called per kbase model. 

`AbstractKieProject.buildKnowledgePackages` calls `AbstractKieProject.addFiles` -> `KieBuilderImpl.filterFileInKBase` -> `KieBuilderImpl.filterFileInKBase` (called per file) -> `KieBuilderImpl.isFileInKieBase` -> `KieBuilderImpl.packageNameForFile` -> `KieBuilderImpl.packageNameFromAsset` -> `KieBuilderImpl.packageNameFromDtable` -> `DecisionTableProviderImpl.loadFromResource` -> `SpreadsheetCompiler.compile`.

If a KieProject has 20 kbase models and 20 spreadsheets, `packageNameFromDtable` is called 400 times. One `packageNameFromDtable`  may not take very long time (1 or 2 seconds), but 400 times are significant.

https://github.com/apache/incubator-kie-drools/blob/main/drools-compiler/src/main/java/org/drools/compiler/kie/builder/impl/KieBuilderImpl.java#L438

This PR introduces `packageNameCache` and brings the cache across static methods. This doesn't look nice, but firstly I wanted to share the idea.

Another possible approach would be caching the DRL generated by SpreadsheetCompiler, but I wonder where is the appropriate place to have the cache (and when we should invalidate the cache).

---

Workaround for non-fixed version: Split KieContainer/ReleaseId per kbaseModel so you can avoid the combinatorial explosion.

tkobayas:
GHA  ubuntu-latest: flaky
```
2024-09-18T09:04:13.7047738Z [ERROR] Tests run: 8, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 16.53 s <<< FAILURE! -- in org.drools.compiler.integrationtests.TimerAndCalendarFireUntilHaltTest
```

mariofusco:
Thanks a lot for your analysis of the problem @tkobayas. I implemented a [very similar alternative solution](https://github.com/apache/incubator-kie-drools/pull/6097), that however moves that same cache in the dtable module, trying to explain why I believe this would be a better solution. Let me know what you think about it.

mariofusco:
Replaced by https://github.com/apache/incubator-kie-drools/pull/6097

PR #6099, opened 2024-09-25
[10.0.x] [incubator-kie-drools-6098] apache-rat-plugin license check for drools

tkobayas:
**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6098

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,7 +136,11 @@
           <configuration>
             <excludes>
               <exclude>.git-blame-ignore-revs</exclude>
+              <exclude>**/dependency-reduced-pom.xml</exclude>
               <exclude>**/lunr-2.3.9.min.js</exclude>
+              <exclude>**/search-ui.js</exclude>
+              <exclude>**/branch.yaml</exclude>
+              <exclude>**/main.yaml</exclude>
               <exclude>**/mvel.jj</exclude>
               <exclude>**/*.csv</exclude>
               <exclude>**/*.sdo</exclude>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -136,7 +136,11 @@
           <configuration>
             <excludes>
               <exclude>.git-blame-ignore-revs</exclude>
+              <exclude>**/dependency-reduced-pom.xml</exclude>
               <exclude>**/lunr-2.3.9.min.js</exclude>
+              <exclude>**/search-ui.js</exclude>
+              <exclude>**/branch.yaml</exclude>
+              <exclude>**/main.yaml</exclude>
```
tkobayas, line 143: `dependency-reduced-pom.xml` is a generated file
`search-ui.js` has Mozilla Public License header
`branch.yaml` and `main.yaml` are overwritten by kogito-pipelines, so if we add a License header, it will be lost in the distribution sources.zip (let's treat them as generated files)

pom.xml
```diff
@@ -159,7 +163,6 @@
               <exclude>**/*.log</exclude>
               <exclude>**/*.lst</exclude>
               <exclude>**/checkstyle-cachefile</exclude>
-              <exclude>**/smoke*.drl.txt</exclude>
```
tkobayas, line 162: `smoke*.drl.txt` and `ruleunit.drl.txt` are checked by the apache-rat-plugin. They are not generated files, so I removed the exclusion and added license headers.

tkobayas:
Locally confirmed that `mvn apache-rat:check` was successful.

@baldimir @LightGuard @porcelli @mariofusco Please review. If merged, I'll forward port to main.

tkobayas:
For GHA: Drools Ansible RuleBook Integration Check, filed a GH issue:
https://github.com/apache/incubator-kie-drools/issues/6100

LightGuard:
Looks fine. We don't distribute the docs with a release, so we don't have to have the license there, but not a problem if it is there.

LightGuard (approved):

mariofusco (approved):

PR #6103, opened 2024-09-25
[KIE-1492] Allow KieRuntimeBuilder to also create and provide StatelessKieSession

### LINKED ISSUE #1492: Allow KieRuntimeBuilder to also create and provide StatelessKieSession
At the moment the [`KieRuntimeBuilder`](https://github.com/apache/incubator-kie-drools/blob/main/kie-api/src/main/java/org/kie/api/runtime/KieRuntimeBuilder.java) only allow to retrieve `KieSession` s, but not `StatelessKieSession` s. Even worse when a KieSession configured as stateless is required through the current API it creates a stateful one anyway, making things even more confused.

Finally the impossibility of creating a `StatelessKieSession` from `KieRuntimeBuilder` is also the indirect cause of a memory leak with the Kogito integration as reported [here](https://github.com/apache/incubator-kie-issues/issues/1491).

mariofusco:
Closes: https://github.com/apache/incubator-kie-issues/issues/1492

### DIFF
--- a/drools-quarkus-extension/drools-quarkus-integration-test/src/test/java/org/drools/quarkus/test/RuntimeTest.java
+++ b/drools-quarkus-extension/drools-quarkus-integration-test/src/test/java/org/drools/quarkus/test/RuntimeTest.java
@@ -24,11 +24,16 @@ import java.util.stream.Collectors;
 import io.quarkus.test.junit.QuarkusTest;
 import jakarta.inject.Inject;
 import org.junit.jupiter.api.Test;
+import org.kie.api.KieServices;
+import org.kie.api.command.KieCommands;
 import org.kie.api.definition.KiePackage;
+import org.kie.api.internal.utils.KieService;
 import org.kie.api.prototype.PrototypeFact;
 import org.kie.api.prototype.PrototypeFactInstance;
 import org.kie.api.runtime.KieRuntimeBuilder;
 import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.RuntimeSession;
+import org.kie.api.runtime.StatelessKieSession;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.kie.api.prototype.PrototypeBuilder.prototype;
@@ -42,30 +47,59 @@ public class RuntimeTest {
     @Test
     public void testDrlEvaluation() {
         // canDrinkKS is the default session
-        testSimpleDrl(runtimeBuilder.newKieSession(), "org.drools.drl");
+        testSimpleDrl(runtimeBuilder.newKieSession(), "org.drools.drl", true);
     }
 
     @Test
     public void testDTableEvaluation() {
-        testSimpleDrl(runtimeBuilder.newKieSession("canDrinkKSDTable"), "org.drools.dtable");
+        testSimpleDrl(runtimeBuilder.newKieSession("canDrinkKSDTable"), "org.drools.dtable", true);
     }
 
     @Test
     public void testYamlEvaluation() {
-        testSimpleDrl(runtimeBuilder.newKieSession("canDrinkKSYaml"), "org.drools.yaml");
+        testSimpleDrl(runtimeBuilder.newKieSession("canDrinkKSYaml"), "org.drools.yaml", true);
     }
 
-    private void testSimpleDrl(KieSession ksession, String assetPackage) {
-        List<String> pkgNames = ksession.getKieBase().getKiePackages().stream().map(KiePackage::getName).collect(Collectors.toList());
+    @Test
+    public void testStatelessDrlEvaluation() {
+        // statelessCanDrinkKS is the default stateless session
+        testSimpleDrl(runtimeBuilder.newStatelessKieSession(), "org.drools.drl", false);
+    }
+
+    @Test
+    public void testStatelessDTableEvaluation() {
+        testSimpleDrl(runtimeBuilder.newStatelessKieSession("statelessCanDrinkKSDTable"), "org.drools.dtable", false);
+    }
+
+    @Test
+    public void testStatelessYamlEvaluation() {
+        testSimpleDrl(runtimeBuilder.newStatelessKieSession("statelessCanDrinkKSYaml"), "org.drools.yaml", false);
+    }
+
+    private void testSimpleDrl(RuntimeSession session, String assetPackage, boolean stateful) {
+        if (stateful) {
+            assertThat(session).isInstanceOf(KieSession.class);
+        } else {
+            assertThat(session).isInstanceOf(StatelessKieSession.class);
+        }
+
+        List<String> pkgNames = session.getKieBase().getKiePackages().stream().map(KiePackage::getName).collect(Collectors.toList());
         
         assertThat(pkgNames).hasSize(2).containsExactlyInAnyOrder("org.drools.quarkus.test", assetPackage);
 
+        KieCommands kieCommands = KieServices.get().getCommands();
+
         Result result = new Result();
-        ksession.insert(result);
-        ksession.insert(new Person("Mark", 17));
-        ksession.fireAllRules();
+        session.execute( kieCommands.newBatchExecution(
+                kieCommands.newInsert(result),
+                kieCommands.newInsert(new Person("Mark", 17)),
+                kieCommands.newFireAllRules() ) );
 
         assertThat(result.toString()).isEqualTo("Mark can NOT drink");
+
+        if (stateful) {
+            ((KieSession) session).dispose();
+        }
     }
 
     @Test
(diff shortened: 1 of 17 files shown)


kie-api/src/main/java/org/kie/api/runtime/RuntimeSession.java
```diff
@@ -0,0 +1,67 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.kie.api.runtime;
+
+import org.kie.api.KieBase;
+
+import java.util.Map;
+
+public interface RuntimeSession extends CommandExecutor {
```
mariofusco, line 25: I introduced this interface to have a usable common ground between `KieSession` and `StatelessKieSession`. Note that they still exists with the same methods as before, so this change is totally backward compatible.

kie-dmn/kie-dmn-core-jsr223/pom.xml
```diff
@@ -112,21 +112,10 @@
       <artifactId>kie-dmn-core-jsr223-jq</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.openjdk.nashorn</groupId>
+      <artifactId>nashorn-core</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
-  <profiles>
```
mariofusco, line 116: These and other changes to poms file are actually not related to this fix, but they are meant to fix some tech debt to support old JVM versions that we don't need anymore.

drools-quarkus-extension/drools-quarkus-integration-test/src/test/java/org/drools/quarkus/test/RuntimeTest.java
```diff
@@ -42,30 +47,59 @@ public class RuntimeTest {
...
+    public void testStatelessDTableEvaluation() {
+        testSimpleDrl(runtimeBuilder.newStatelessKieSession("statelessCanDrinkKSDTable"), "org.drools.dtable", false);
+    }
+
+    @Test
+    public void testStatelessYamlEvaluation() {
+        testSimpleDrl(runtimeBuilder.newStatelessKieSession("statelessCanDrinkKSYaml"), "org.drools.yaml", false);
+    }
+
+    private void testSimpleDrl(RuntimeSession session, String assetPackage, boolean stateful) {
```
mariofusco, line 79: Here I'm using the new `RuntimeSession` interface to test both `KieSession` and `StatelessKieSession` with the same method.

kie-api/src/main/java/org/kie/api/command/KieCommands.java
```diff
@@ -115,6 +117,10 @@ Command newQuery(String identifier,
                      String name,
                      Object[] arguments);
 
+    default BatchExecutionCommand newBatchExecution(Command... commands) {
```
gitgabrio, line 120: HI @mariofusco 
is this related to the overall task ? If so, could. you please explain why ?

kie-api/src/main/java/org/kie/api/runtime/StatelessKieSession.java
```diff
@@ -106,51 +103,6 @@
...
-     *
-     * @param name
-     */
-    void unregisterChannel(String name);
-
-    /**
-     * @return a map with all registered channels.
-     */
-    Map<String, Channel> getChannels();
+public interface StatelessKieSession extends StatelessRuleSession, StatelessProcessSession, RuntimeSession, KieRuntimeEventManager {
```
gitgabrio, line 106: I did not notice that before - but at very first sight I have the impression the hierarchies are somehow turned upside:
StatelessRuleSession and StatelessRuleSession should be "specialization" of StatelessKieSession, but as it is, StatelessKieSession is an aggregation of the two, creating an hard-binding: wdyt ?

gitgabrio (commented):
Thanks @mariofusco !!!
For the moment being I left some minor comments.
As side-note, and not specifically related to this PR, I have the impression that the whole "KieSession" (statefull/stateless/etc) hierarchy would benefit from a good review, but surely not in this PR
I need slighlty more time to dig in the code itself

kie-api/src/main/java/org/kie/api/command/KieCommands.java
mariofusco: Not strictly related to this fix, it's just a convenience method that I added to create batch commands in a single line that I used in my test.

kie-api/src/main/java/org/kie/api/runtime/StatelessKieSession.java
mariofusco: Unfortunately you're right, or at least the name of those interfaces are quite misleading, but we cannot change them if we want to keep backward compatibility.

tkobayas:
Thanks, Mario.

GHA kogito-runtimes (and other downstreams):
```
2024-09-25T16:38:02.4742156Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project jbpm-flow: Compilation failure
2024-09-25T16:38:02.4744813Z [ERROR] /home/runner/work/incubator-kie-drools/incubator-kie-drools/apache_incubator-kie-kogito-runtimes/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java:[65,1] org.jbpm.process.instance.DummyKnowledgeRuntime is not abstract and does not override abstract method <T>execute(org.kie.api.command.Command<T>) in org.kie.api.runtime.CommandExecutor
```
`DummyKnowledgeRuntime` needs to implement `execute` or `CommandExecutor.execute` should have a default implementation.

mariofusco:
> Thanks, Mario.
> 
> GHA kogito-runtimes (and other downstreams):
> 
> ```
> 2024-09-25T16:38:02.4742156Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project jbpm-flow: Compilation failure
> 2024-09-25T16:38:02.4744813Z [ERROR] /home/runner/work/incubator-kie-drools/incubator-kie-drools/apache_incubator-kie-kogito-runtimes/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java:[65,1] org.jbpm.process.instance.DummyKnowledgeRuntime is not abstract and does not override abstract method <T>execute(org.kie.api.command.Command<T>) in org.kie.api.runtime.CommandExecutor
> ```
> 
> `DummyKnowledgeRuntime` needs to implement `execute` or `CommandExecutor.execute` should have a default implementation.

This is the result of a **HUGE** technical debt caused by the tight coupling between Drools and Kogito's processes that has no longer reason to exist. I spent a couple of hours trying to refactor Kogito and lower this coupling, but unfortunately this is clearly a multi-days effort, so I gave up for now and fixed the issue in the most straightforward way with this pull request https://github.com/apache/incubator-kie-kogito-runtimes/pull/3681

That said, I'm still willing to solve this debt with more calm when I will have time to dedicate to this, possibly with the help of @elguardian and @fjtirado. I already have a very preliminary question for you 2: is that `DummyKnowledgeRuntime` still necessary for some reasons (and why) or could it be totally removed? I'm asking because this seems to be a hack that it is also a source of a big part of the issues that I found during this first refactor attempt.

@tkobayas @gitgabrio please review again.

fjtirado:
Hi Mario,
DummyKnowledgeRuntime should be a goner for good.
Besides that, I think that  the final result of the refactor (which I agree
is a lengthy task) should be that the process engine (which is currently
used by BPMN and Sonataflow) no longer depends on drools*.
If we agree that's the goal, I believe it is better for you to work on that
alone (meaning quietly, without continuous interference by me ;)). Once you
have the PR, let's review it also quietly.


* The dependency with drools should be in the RuleNode, which is the part
of the engine related with rules.

gitgabrio:
@mariofusco @fjtirado @elguardian 
There is an ongoing task for that: I started make the analysis to remove all the *processes references from drools repository, but I have to finish another ticket before.
@elguardian ? 
Anyway, let's avoid duplicate the effort, and stay sync with that

tkobayas (approved):

mariofusco:
> @mariofusco @fjtirado @elguardian There is an ongoing task for that: I started make the analysis to remove all the *processes references from drools repository, but I have to finish another ticket before. @elguardian ? Anyway, let's avoid duplicate the effort, and stay sync with that

I agree on avoid duplicating the effort, and if possible I would like to work on this on small-ish steps. For now I will try to remove that `DummyKnowledgeRuntime` class as suggested by @fjtirado or is there anybody else working on this? /cc @elguardian 

@gitgabrio If this pull request ok for you? If so can you please approve so I could merge and move forward with that rework?

mariofusco:
I just realized there are a few failing tests, another side effect of that coupling I guess :(
I will fix them immediately.

gitgabrio:
@mariofusco 

> If this pull request ok for you? 

Fine for me as soon as the build is green 👍 

> For now I will try to remove that DummyKnowledgeRuntime class as suggested by @fjtirado or is there anybody else working on this? /cc @elguardian

Fine for me

PR #6104, opened 2024-09-27
Enhance ConstraintOperator capability

tkobayas:
…ares incompatible types

This PR is to enhance `ConstraintOperator` to be able to use alpha/beta index.

### DIFF
--- a/drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
+++ b/drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
@@ -22,4 +22,12 @@ import java.util.function.BiPredicate;
 
 public interface ConstraintOperator {
     <T, V> BiPredicate<T, V> asPredicate();
+
+    default boolean containsConstraintType() {
+        return false;
+    }
+
+    default Index.ConstraintType getConstraintType() {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
+++ b/drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
@@ -23,11 +23,11 @@ import java.util.function.BiPredicate;
 public interface ConstraintOperator {
     <T, V> BiPredicate<T, V> asPredicate();
 
-    default boolean containsConstraintType() {
+    default boolean hasIndex() {
         return false;
     }
 
-    default Index.ConstraintType getConstraintType() {
+    default Index.ConstraintType getIndexType() {
         throw new UnsupportedOperationException();
     }
 }
--- a/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/CustomConstraintOperatorTest.java
+++ b/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/CustomConstraintOperatorTest.java
@@ -67,12 +67,12 @@ public class CustomConstraintOperatorTest {
         }
 
         @Override
-        public boolean containsConstraintType() {
+        public boolean hasIndex() {
             return true;
         }
 
         @Override
-        public Index.ConstraintType getConstraintType() {
+        public Index.ConstraintType getIndexType() {
             return Index.ConstraintType.EQUAL;
         }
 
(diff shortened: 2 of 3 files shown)


tkobayas:
@mariofusco This PR is not a clear separation as you mentioned in https://github.com/kiegroup/drools-ansible-rulebook-integration/pull/112#discussion_r1776835572

This PR enhances `ConstraintOperator` to contain `Index.ConstraintType`, so a custom `ConstraintOperator` can have its own predicate while make use of alpha index. (I haven't implemented for beta index yet. I can do it later in this PR). Bare `Index.ConstraintType` can be both an operator and an index type. I leave it so that I can minimize the change. If you want further separation, feel free to let me know.

tkobayas:
Updated with beta index support and unit test

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/CustomConstraintOperatorTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        ksession.insert(testFact);
+        assertThat(ksession.fireAllRules()).isEqualTo(1);
+
+        Index index = getFirstAlphaNodeIndex((InternalKnowledgeBase) kieBase, testPrototype);
+        assertThat(index.getIndexType()).isEqualTo(Index.IndexType.ALPHA);
+        assertThat(index.getConstraintType()).isEqualTo(Index.ConstraintType.EQUAL);
+
+        // alpha index hashing is effective, so the predicated is not called
+        assertThat(customConstraintOperator.counter).isZero();
```
tkobayas, line 167: If alpha index is effective, the predicate in the `CustomConstraintOperator` is not executed.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/CustomConstraintOperatorTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        ksession.fireAllRules();
+        assertThat(result.getValue()).isEqualTo("Paul is the same age as Mark");
+
+        Index index = getFirstBetaNodeIndex((InternalKnowledgeBase) kieBase, personFact);
+        assertThat(index.getIndexType()).isEqualTo(Index.IndexType.BETA);
+        assertThat(index.getConstraintType()).isEqualTo(Index.ConstraintType.EQUAL);
+
+        // When beta index is used, the predicate in the custom operator is not actually called
+        assertThat(customConstraintOperator.counter).isZero();
```
tkobayas, line 228: If beta index is effective, the predicate in the `CustomConstraintOperator` is not executed.

tkobayas:
As commented above, if alpha/beta index is effective, the predicate in the `CustomConstraintOperator` is not executed. So implementing type check and error logging in the predicate doesn't work.

I guess alpha index is a more possible use case.

For example, 

A user mistakenly wrote wrong rules like this.
```
condition: event.meta.headers == "AAA"
```
```
condition: event.meta.headers == "BBB"
```
```
condition: event.meta.headers == "CCC"
```
Then, an incoming event has `meta.headers` as a Map.

We can do type check in a predicate in a custom `ConstraintOperator`, but in the alpha index hashing case, the predicate is not executed and simply no rule would be fired. The user has no clue what is wrong in the rules.

If we want to rescue the situation, we need to implement type check at `CompositeObjectSinkAdapter.propagateAssertObject`, WDYT? @mariofusco

drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
```diff
@@ -22,4 +22,12 @@
 
 public interface ConstraintOperator {
     <T, V> BiPredicate<T, V> asPredicate();
+
+    default boolean containsConstraintType() {
+        return false;
+    }
+
+    default Index.ConstraintType getConstraintType() {
```
mariofusco: Rename the method as `getIndexType`.

drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
```diff
@@ -22,4 +22,12 @@
 
 public interface ConstraintOperator {
     <T, V> BiPredicate<T, V> asPredicate();
+
+    default boolean containsConstraintType() {
```
mariofusco: Maybe rename the method as `isIndexable` or `hasIndex`.

drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
tkobayas: Thanks, done.

drools-model/drools-canonical-model/src/main/java/org/drools/model/ConstraintOperator.java
tkobayas: Thanks, done.

mariofusco (approved):

lucamolteni (approved):

PR #6105, opened 2024-09-30
[incubator-kie-issues#1478] Register DRG callbacks to assigned models from DMNCompilerImpl

### LINKED ISSUE #6974: DMNCompilerImpl calls DRG (Decision requirements graph) callbacks on all loaded models
Currently one of the responsibilities of the `DMNCompilerImpl` is to call all registered `AfterProcessDrgElements` after each model's evaluation. This behaviour is leading to some issues, since it doesn't validate of the DRG (Decision requirements graph) Compiler would `accept` the implementation or not? Or even if the current model is in fact the model that registered the callback at all. Instead it assumes two things, first, that the `AfterProcessDrgElements` or the `DRGElementCompiler` would check if the model is accepted by that DMN compiler, second, that the call back will be idempotent as well.

For the first assumption, my suggestion would be to actually check if the `DRGElementCompiler` accepts the model. For the second assumption we could do one of two things:
1. Document the `AfterProcessDrgElements` to mention it must be ready to accept many calls that are not related. And/Or,
2. `DMNCompilerImpl` should register the callback to the model that is related to it, and only call the callback on a valid and related model.

The current behaviour as is, was causing an issue that was partially addressed in a PR fixing [incubator-kie-issues#1474](https://github.com/apache/incubator-kie-drools/pull/6080).

samuel-beniamin:
Closes [incubator-kie-issues/1478](https://github.com/apache/incubator-kie/issues/6974)


gitgabrio:
👀

gitgabrio:
Hi @samuel-beniamin thanks for the PR.
I tried to understand better the overall context, and I'm sharing my thoughts: please correct me if/where I'm wrong: I may misread something.

1. by itself, I think the fix could be right
2. but those AfterProcessDrgElements / addCallback /  addCallbackForModel are used/needed only by the signavio module
3. but being defined/ implemented in the "core"  DMNCompilerImpl, those somehow "pollute" it with something that, IMO, has no reason to be there

So, I was thinking to

1. change `DMNCompilerImpl.processDrgElements` from private to protected, to be overridable
2. get rid of all signavio-specific stuff from it (double check also the tests)
4. create a signavio-specific subclass of `DMNCompilerImpl` that add the behavior required by signavio
5. put your modifications in that subclass

The idea would be to have `DMNCompilerImpl` dealing only with "core" responsibilities, and move implementation-specific ones (e.g. signavio) to dedicated subclasses

Wdyt ? Does this make sense ? Do you think you could do that ?

@baldimir @yesamer

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -491,9 +493,30 @@ private void processDrgElements(DMNCompilerContext ctx, DMNModelImpl model, Defi
     public interface AfterProcessDrgElements {
         void callback(DMNCompilerImpl compiler, DMNCompilerContext ctx, DMNModelImpl model);
     }
-    
+
+    /**
```
gitgabrio, line 498: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

samuel-beniamin:
> Hi @samuel-beniamin thanks for the PR. I tried to understand better the overall context, and I'm sharing my thoughts: please correct me if/where I'm wrong: I may misread something.
> 
> 1. by itself, I think the fix could be right
> 2. but those AfterProcessDrgElements / addCallback /  addCallbackForModel are used/needed only by the signavio module
> 3. but being defined/ implemented in the "core"  DMNCompilerImpl, those somehow "pollute" it with something that, IMO, has no reason to be there
> 
> So, I was thinking to
> 
> 1. change `DMNCompilerImpl.processDrgElements` from private to protected, to be overridable
> 2. get rid of all signavio-specific stuff from it (double check also the tests)
> 3. create a signavio-specific subclass of `DMNCompilerImpl` that add the behavior required by signavio
> 4. put your modifications in that subclass
> 
> The idea would be to have `DMNCompilerImpl` dealing only with "core" responsibilities, and move implementation-specific ones (e.g. signavio) to dedicated subclasses
> 
> Wdyt ? Does this make sense ? Do you think you could do that ?
> 
> @baldimir @yesamer

Hello @gitgabrio,

Thank you for taking the time to review the code and for your suggestions.

I think your suggestions make sense, I believe I understood what need to be done, and I will be happy to make these changes in this PR as well, if that is okay.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -113,8 +111,8 @@ public class DMNCompilerImpl implements DMNCompiler {
     private static final Logger logger = LoggerFactory.getLogger( DMNCompilerImpl.class );
 
     private final DMNDecisionLogicCompiler evaluatorCompiler;
-    private DMNCompilerConfiguration dmnCompilerConfig;
-    private Deque<DRGElementCompiler> drgCompilers = new LinkedList<>();
+    private final DMNCompilerConfiguration dmnCompilerConfig;
+    private final Deque<DRGElementCompiler> drgCompilers = new LinkedList<>();
```
gitgabrio, line 116: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -123,6 +121,7 @@ public class DMNCompilerImpl implements DMNCompiler {
         drgCompilers.add( new KnowledgeSourceCompiler() ); // keep last as it's a void compiler
     }
     private final List<AfterProcessDrgElements> afterDRGcallbacks = new ArrayList<>();
+    private final Map<DMNModel, List<AfterProcessDrgElements>> afterDRGcallbacksForModel = new HashMap<>();
```
gitgabrio, line 125: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java
```diff
@@ -491,9 +493,30 @@ private void processDrgElements(DMNCompilerContext ctx, DMNModelImpl model, Defi
...
     public void addCallback(AfterProcessDrgElements callback) {
-        this.afterDRGcallbacks.add(callback);
+        afterDRGcallbacks.add(callback);
+    }
+
+
+    /**
+     * Adds a callback that will be invoked after the DRG elements have been processed, for a given model.
+     */
+    public void addCallbackForModel(AfterProcessDrgElements callback, DMNModel model) {
```
gitgabrio, line 511: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DMNCompilerImplTest.java
```diff
@@ -18,29 +18,51 @@
...
+    private static DMNCompilerImpl.AfterProcessDrgElements mockCallbackForModel;
 
     @BeforeAll
     static void setup() {
         String modelName = "LocalHrefs";
         parent = new TDefinitions();
         parent.setName(modelName);
         parent.setNamespace(nameSpace);
+
+        DMNCompilerConfiguration config = DMNFactory.newCompilerConfiguration();
```
gitgabrio, line 61: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DMNCompilerImplTest.java
```diff
@@ -76,4 +98,27 @@ void getRootElement() {
         retrieved = DMNCompilerImpl.getRootElement(elementReference);
         assertThat(retrieved).isNotNull().isEqualTo(parent);
     }
+
```
gitgabrio, line 101: Hi @samuel-beniamin 
IINW this is signavio-only behaviour that get to the core code: if so, could you please remove and move to signavio module ? Thanks!

gitgabrio (changes requested):

samuel-beniamin:
Hello @gitgabrio 

Well, I tried the approach of the a separate subclass for Signavio specific things, I have some observation and a conclusion: 
1. I ended up not able to inject that Signavio specific compiler easily into the code.
2. `DMNCompilerConfiguration` is not Signavio specific, this is related to the core compiler for extensions in general. 
3. `drgCompilers` are not Signavio specific, they are also used by other extensions, (e.g. `trisotech` is also using that).
    * This makes me think that `AfterProcessDrgElements` follows the same reasoning, is not Signavio specific (since `drgCompilers` are not), `AfterProcessDrgElements` is indeed used only now by Signavio, but that doesn't limit the need to Signavio only.

I have a question, do you know a way with which I can integrate the new compiler into the code by any means, I had the impression that the compiler is not intended to be changed but extended through the `DMNCompilerConfiguration` only, am I correct?

If I am correct on the previous point, I would actually close this ticket after rethinking this part and also close the ticket, since this is the way the compiler was meant to be extended actually.

gitgabrio:
First of all, many thanks @samuel-beniamin for all that.
I'll need a bit of time to double check, but it is very possible that you are right.
I've answered immediately to first of all thanks you again, I'll let you know ASAP!

samuel-beniamin:
I think this PR is not relative anymore. I will close it for now, and if needed we can re-open it in the future.

PR #6129, opened 2024-10-17
[kie-issues#1546] DMN TCK failure: range equality test fail for unary test ranges - treat unary test ranges boundaries as undefined.

### LINKED ISSUE #1546: DMN TCK failure: range equality test fail for unary test ranges
When having a FEEL expression like this:
(<10) = (null..10)
The engine should return false, instead of the current result true. The unary test range has an undefined boundary, which is, based on the DMN specification, not equal to having a null boundary.

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/1546 and also fixes 4 DMN TCK tests from 0068-feel-equality. 

- For unary test ranges defines the boundaries as undefined, instead of treating them as null.


baldimir:
TCK results with these fixes: [ERROR] Tests run: 6573, Failures: 0, Errors: 37, Skipped: 0

baldimir:
Closing, as it won't go through review. @gitgabrio wants to do it differently, he will open a new PR soon.

baldimir:
Closing, as it won't go through review. @gitgabrio will open a new PR soon.

PR #6155, opened 2024-11-18
[incubator-kie-issues-1616] notice file for drools is not correct

tkobayas:
- To solve license and binary issues, removing drools-docs module, which is not required in the source distribution

**Issue**:

* As a part of work https://github.com/apache/incubator-kie-issues/issues/1616 , removing `drools-docs` module, which is not required in the source distribution.


porcelli:
@tkobayas I'll close this, as we don't need to remove completely the content from the github, we need only adjust the pom profile and I can exclude during source packaging script

tkobayas:
> we need only adjust the pom profile and I can exclude during source packaging script

Thanks!

porcelli (approved):
in the end, this would be easier :) - thx again @tkobayas !

PR #6172, opened 2024-11-26
[incubator-kie-issues#1627] Change log level from error to debug inside ImportDMNResolverUtil

### LINKED ISSUE #1627: Cannot include models: I get a "failed to import" when running a test with imported models
- Both models exist in the same directory
- Both models open with no errors
- The model will import with no errors
- The test works if I simply change the drools version to 9.4.4.0.Final with no other changes
- The scesim test will fail with the following output:

``` 
2024-11-19 11:14:23,499 [main] INFO  No executable model found for artifact ca.queensu:sror:1.0-SNAPSHOT. Falling back to resources parsing.
2024-11-19 11:14:23,925 [main] INFO  Start creation of KieBase: defaultKieBase
2024-11-19 11:14:24,350 [main] ERROR DMN Model with name=gror and namespace=https://kiegroup.org/dmn/_676F9720-3A0C-4EAD-9D0A-A68E7016CED7 failed to import a DMN with namespace=https://kiegroup.org/dmn/_691425E9-D743-4F6F-9019-2607F6766BFD name=SROR locationURI=sror.dmn, modelName=null.
2024-11-19 11:14:25,066 [main] ERROR Error compiling FEEL expression 'Consecutive Years Taught(Term Adjunct.courses_Taught,SROR.Academic Year(today()))' for name 'Served 4 Consecutive Years as Term Adjunct' on node 'Served 4 Consecutive Years as Term Adjunct': Unknown variable 'SROR'
```

(I had to change the extension to .txt to attach)
[gror.txt](https://github.com/user-attachments/files/17818260/gror.txt)
[sror.txt](https://github.com/user-attachments/files/17818262/sror.txt)

[SRORDecision.txt](https://github.com/user-attachments/files/17818288/SRORDecision.txt)

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1627

Follow up of https://github.com/apache/incubator-kie-drools/pull/6150, discovered during https://github.com/apache/incubator-kie-issues/issues/1627 testing. 

Based on the current architecture, the `resolveImportDMN(Import, Collection<T>, Function<T, QName>)` is called twice, which means that an error logged in the first call is inconsistent with the second call, where the behavior works as expected.

As a consequence, all log levels should be set as DEBUG, until a definitive fix in the architecture.

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/ImportDMNResolverUtil.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/ImportDMNResolverUtil.java
@@ -82,13 +82,13 @@ public class ImportDMNResolverUtil {
                         importerDMNName, importerDMNNamespace, importNamespace, importName, importLocationURI, importModelName);
                 return Either.ofRight(usingNSandName.get(0));
             } else if (usingNSandName.isEmpty()) {
-                LOGGER.error("DMN Model with name={} and namespace={} failed to import a DMN with namespace={} name={} locationURI={}, modelName={}.",
+                LOGGER.debug("DMN Model with name={} and namespace={} failed to import a DMN with namespace={} name={} locationURI={}, modelName={}.",
                         importerDMNName, importerDMNNamespace, importNamespace, importName, importLocationURI, importModelName);
                 return Either.ofLeft(String.format(
                         "DMN Model with name=%s and namespace=%s failed to import a DMN with namespace=%s name=%s locationURI=%s, modelName=%s. ",
                         importerDMNName, importerDMNNamespace, importNamespace, importName, importLocationURI, importModelName));
             } else {
-                LOGGER.error("DMN Model with name={} and namespace={} detected a collision ({} elements) trying to import a DMN with namespace={} name={} locationURI={}, modelName={}",
+                LOGGER.debug("DMN Model with name={} and namespace={} detected a collision ({} elements) trying to import a DMN with namespace={} name={} locationURI={}, modelName={}",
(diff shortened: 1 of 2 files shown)


gitgabrio (approved):

gitgabrio (changes requested):
Hi @yesamer 
on second thought, I think we have to change the overall approach here.
That method get's invoked twice, once with an empty `Collection<T> dmns` and once with a populated one.
The "misleading" logs are related to the first one.
So, I think it is better that, when an empty dmns collections is recieved, the method immediately returns back with an Either.of left, and debug messaging explaining it.
If, instead, a populated collection is received, all failures on finding imported model should be logged as error: wdyt ?

yesamer:
@gitgabrio The improvement you suggested makes sense, but I would manage it in this ticket https://github.com/apache/incubator-kie/issues/7067 (please feel free to update it) and provide a complete refactoring of the feature

gitgabrio:
@yesamer 
The ticket you mention has a more broader impact and it is not sure when it could be done.
On the other side, reviewing this PR, I realized that we are dealing two different uses cases (empty or populated dmns) with the same logic.
In one case, the "error" level is required, but now (even with [my previous one](https://github.com/apache/incubator-kie-drools/pull/6150)) we are logging everything as debug, which is wrong, IMO.
Since it is just a matter of  a couple of lines in the very same method already touched by this PR, I think it would be better to fix that with this PR (and reverting the "debug" messages to "error").

yesamer:
@gitgabrio Done, thank you

gitgabrio (approved):
Thanks @yesamer 
lgtm!

yesamer:
@baldimir Can you please review it?

baldimir (approved):

PR #6174, opened 2024-11-27
[NO-ISSUE] License check

tkobayas:
- Remaining Category A license
- Remove duplicate headers

Not urgent. This is not a blocker of 10.0.0

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -381,6 +381,283 @@ event that testing suites are implemented or approved by Object Management Group
 specification may claim compliance or conformance with the specification only if the software satisfactorily completes
 the testing suites.
 
- ------------------------------------------------------------------------------------
- for drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
-     JavaParser is dual licensed under LGPL license and Apache License Version 2.0 license. We choose the Apache License Version 2.0.
+------------------------------------------------------------------------------------
+for drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
+    JavaParser is dual licensed under LGPL license and Apache License Version 2.0 license. We choose the Apache License Version 2.0.
+
+------------------------------------------------------------------------------------
+for drools-model/drools-canonical-model/src/main/java/org/drools/model/bitmask/OpenBitSet.java
+    drools-util/src/main/java/org/drools/util/bitmask/OpenBitSet.java
+
+A. HISTORY OF THE SOFTWARE
+==========================
+
+Python was created in the early 1990s by Guido van Rossum at Stichting
+Mathematisch Centrum (CWI, see http://www.cwi.nl) in the Netherlands
+as a successor of a language called ABC.  Guido remains Python's
+principal author, although it includes many contributions from others.
+
+In 1995, Guido continued his work on Python at the Corporation for
+National Research Initiatives (CNRI, see http://www.cnri.reston.va.us)
+in Reston, Virginia where he released several versions of the
+software.
+
+In May 2000, Guido and the Python core development team moved to
+BeOpen.com to form the BeOpen PythonLabs team.  In October of the same
+year, the PythonLabs team moved to Digital Creations (now Zope
+Corporation, see http://www.zope.com).  In 2001, the Python Software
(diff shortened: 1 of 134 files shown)


LICENSE
```diff
@@ -381,6 +381,283 @@ event that testing suites are implemented or approved by Object Management Group
...
- ------------------------------------------------------------------------------------
- for drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
-     JavaParser is dual licensed under LGPL license and Apache License Version 2.0 license. We choose the Apache License Version 2.0.
+------------------------------------------------------------------------------------
+for drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
+    JavaParser is dual licensed under LGPL license and Apache License Version 2.0 license. We choose the Apache License Version 2.0.
+
+------------------------------------------------------------------------------------
+for drools-model/drools-canonical-model/src/main/java/org/drools/model/bitmask/OpenBitSet.java
+    drools-util/src/main/java/org/drools/util/bitmask/OpenBitSet.java
```
mariofusco, line 390: I'm not understanding why the Python license is necessary for our OpenBitSet, did we take the implementation from Python? Also is the whole history of Python really part of the license?

tkobayas: Sorry that I didn't provide the information first.

https://github.com/apache/incubator-kie-drools/blob/main/drools-model/drools-canonical-model/src/main/java/org/drools/model/bitmask/OpenBitSet.java#L901-L919

```
     *  NOTE: this was originally borrowed from Python 2.4.2
     *  listobject.c sources (attribution in LICENSE.txt), but
     *  has now been substantially changed based on
     *  discussions from java-dev thread with subject "Dynamic
     *  array reallocation algorithms", started on Jan 12
     *  2010.
```

I'm not sure how we should treat "borrowed" and "substantially changed" and it's just one method of the class.

@pjfanning Do we need to list this file as Python Software Foundation License? (although this is not urgent --- PSF is Category A)

> is the whole history of Python really part of the license?

It's the content of the LICENSE.

https://github.com/python/cpython/blob/main/LICENSE

@pjfanning If we can drop the "history" part, please let me know.

LICENSE
```diff
@@ -381,6 +381,283 @@ event that testing suites are implemented or approved by Object Management Group
...
+FITNESS, IN NO EVENT SHALL STICHTING MATHEMATISCH CENTRUM BE LIABLE
+FOR ANY SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT
+OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+------------------------------------------------------------------------------------
+for drools-util/src/main/java/org/drools/util/StringUtils.java
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
tkobayas: https://github.com/apache/incubator-kie-drools/blob/main/drools-util/src/main/java/org/drools/util/StringUtils.java#L627-L629
```
     * Borrowed from Spring, under the ASL2.0 license.
```
8 methods have this comment.

So I added this entry. I cannot find "copyright" of spring framework, so just noted the Apache License.

LICENSE
pjfanning: * if the code is borrowed then we need to include details about its LICENSE in our LICENSE
* this 'history' seems to be part of the LICENSE so we probably need to include it
* we could raise a LEGAL JIRA to discuss this but in mean time, it seems best to include full text of cpython license

LICENSE
pjfanning: I think it is best to mention Spring here. It means it is easier for people to track down the original source code.

mariofusco (approved):

LICENSE
tkobayas: Thanks.

LICENSE
tkobayas: Thanks. Done.

baldimir (approved):

PR #6186, opened 2024-12-05
[incubator-kie-drools-6180] accumulate min doesn't evaluate correctly…

tkobayas:
… with more than 18 digits BigDecimal

**Issue**:

* https://github.com/apache/incubator-kie-drools/issues/6180

At first, I thought I can use `MinAccumulateFunction` for `BigDecimal` and `BigInteger`. However, its `getResultType()` returns `Comparable.class`, so it causes a compile error when using methods like `$min.intValue()`. We may change the strategy to cache accumulate function instances (currently, per function name) to make `MinAccumulateFunction` work for various types, but I followed the current approach which is easier to understand, so I added 4 function classes `BigDecimalMinAccumulateFunction` (minBD), `BigIntegerMinAccumulateFunction` (minBI), `BigDecimalMaxAccumulateFunction` (maxBD), `BigIntegerMaxAccumulateFunction` (maxBI)


mariofusco (approved):

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3941,4 +3943,156 @@ public void testPeerCollectWithEager(KieBaseTestConfiguration kieBaseTestConfigu
...
+    void minWithBigIntegerHighAccuracy(KieBaseTestConfiguration kieBaseTestConfiguration) {
+        final String drl =
+                "import " + Primitives.class.getCanonicalName() + ";\n" +
+                        "global java.util.List results;\n" +
+                        "rule R1 when\n" +
+                        "    accumulate(Primitives($bi : bigInteger), $min : min($bi))\n" +
+                        "then\n" +
+                        "    results.add($min);\n" +
+                        // to confirm if $min is BigInteger at build time (Not Comparable)
+                        // The return value isn't important to assert
```
pibizza: Are these comments relevant? it seems that we are asserting the result values.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
```diff
@@ -3941,4 +3943,156 @@ public void testPeerCollectWithEager(KieBaseTestConfiguration kieBaseTestConfigu
...
+                        "    System.out.println($min.nextProbablePrime());\n" +
```
pibizza: Do we need to printout this value? if possible a test should not print values, only check them in assertions.

pibizza (commented):
Check comments to see if they are relevants for the tests.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
tkobayas: `nextProbablePrime()` is specific to `BigInteger` class. So, this line is to check that drl build doesn't raise a compile error, so we can confirm that `$min` is treated as `BigInteger` in the internally generated java code. The return value of  `nextProbablePrime()`  ~~is not guaranteed, so we cannot assert the value~~ is not important, I didn't assert.

Having said that, I can avoid `System.out.println` anyway. I'll fix the PR. Thanks.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/compiler/integrationtests/AccumulateTest.java
tkobayas: The comment is to explain the below code `"    System.out.println($min.nextProbablePrime());\n" `. I'll rewrite it better. Thanks.

tkobayas:
re-running GHA

tkobayas:
Test failures on CallbackStateTimeoutsIT, MultipleTimerInstancesIT etc. are being solved by https://github.com/apache/incubator-kie-kogito-runtimes/pull/3814

PR #6187, opened 2024-12-06
Improve assertJ tests by using containsExactlyInAnyOrder

everbrightw:
### Description:
- The modified two tests could sometimes fail. Improved AssertJ tests by using `containsExactlyInAnyOrder` to account for `HashSet` 's unordered nature. 
- `containsExactly` inherently relies on stream to populate the `collections` and test if the target values are the same and **in order**
- According to AssertJ's [documentation](https://github.com/assertj/assertj/blob/1d6614aeef9b68f0e828bc828b3bcb538ac87a25/assertj-core/src/main/java/org/assertj/core/api/ObjectEnumerableAssert.java#L154C4-L155C21):
  - `containsExactly` verifies that the actual group contains exactly the given values **in order**.
  - `containsExactlyInAnyOrder` verifies that the actual group contains exactly the given values **in any order**.


mariofusco (approved):

pibizza (approved):
Correct, thanks a lot.

PR #6195, opened 2024-12-12
#1708 - Replace original implementation and remove copyright info from license file

### LINKED ISSUE #1708: Replace Drools CsvLineParser.java implementation
[drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java](https://github.com/apache/incubator-kie-drools/blob/fa989e50f0e71e2bd9d54adcb7d1781a41c27ace/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java)

This would help remove the following reference in LICENSE file:

https://github.com/apache/incubator-kie-drools/blob/main/LICENSE#L238-L247

porcelli:
Closes: https://github.com/apache/incubator-kie-issues/issues/1708

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -235,17 +235,6 @@ for drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/Ja
  THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  ------------------------------------------------------------------------------------
- for drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
-
- Inner logic adapted from a C++ original that was Copyright (C) 1999
- Lucent Technologies Excerpted from 'The Practice of Programming' by Brian
- Kernighan and Rob Pike.
-
- Included by permission of the http://tpop.awl.com/ web site, which says:
- "You may use this code for any purpose, as long as you leave the
- copyright notice and book citation attached."
-
- ------------------------------------------------------------------------------------
  for kie-dmn/kie-dmn-backend/src/test/resources/DC.xsd
      kie-dmn/kie-dmn-backend/src/test/resources/DI.xsd
      kie-dmn/kie-dmn-backend/src/test/resources/DMN11.xsd
--- a/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
+++ b/drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
@@ -19,30 +19,15 @@
 package org.drools.decisiontable.parser.csv;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
  * A simple parser for CSV (Comma-Separated Values) format.
- * Supports quoted fields, escaped quotes, and custom delimiters.
+ * Supports quoted fields and escaped quotes.
  */
 public class CsvLineParser {
-    private final char delimiter;
-
-    /**
-     * Creates a CSV parser with comma as the default delimiter.
-     */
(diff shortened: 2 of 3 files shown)


tkobayas (approved):
Thank you!

tkobayas:
kogito-apps GHA test failure can be ignored as it's being worked on.

LICENSE
```diff
@@ -234,17 +234,6 @@ for drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/Ja
  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
  THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
- ------------------------------------------------------------------------------------
- for drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
```
jomarko, line 238: Please, update also https://github.com/apache/incubator-kie-drools/blob/main/.rat-excludes, we need to remove the entry `CsvLineParsrer`  from there

drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
```diff
@@ -1,150 +1,93 @@
...
+    public List<String> parse(CharSequence input) {
+        String line = input != null ? input.toString() : "";
+        List<String> fields = new ArrayList<>();
 
-        /** Construct a CSV parser, with the default separator (','). */
-        public CsvParserImpl() {
-            this( CsvParserImpl.DEFAULT_SEP );
+        if (line.isEmpty()) {
+            fields.add("");
+            return fields;
```
jomarko: In past, we used often `Collections.singletoneList`, what is your opinion about:
```
        if (line.isEmpty()) {
            return Collections.singletonList("");
        }
```

drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
```diff
@@ -1,150 +1,93 @@
...
-     * This is insurance incase I need to replace it with more complex Csv
-     * handlers in the future.
+     * Creates a CSV parser with a custom delimiter.
+     *
+     * @param delimiter The character to use as field delimiter
      */
-    static interface ICsvParser {
-        public List<String> parse(String line);
+    public CsvLineParser(char delimiter) {
+        this.delimiter = delimiter;
```
jomarko: Do we want to keep this constructor? I was not able to find its usage, to me it seems we use only the default one.

jomarko (changes requested):
Thank you for the PR @porcelli . Always good to see we are able to simplify the codebase we maintain.

I have some comments and questions before merge.

mariofusco (approved):

drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
porcelli: good point, I wasn't completely sure and that's why cheated and sent this way so to have the CI tell me that if in the codebase we had this used somehow :)

drools-decisiontables/src/main/java/org/drools/decisiontable/parser/csv/CsvLineParser.java
porcelli: sure, I'll adjust

LICENSE
porcelli: good catch

jomarko (approved):
Thank you for updates @porcelli

PR #1710, opened 2018-01-22
RHBA-382 - Redundant records are inserted by JPAPlaceholderResolverSt…

mswiderski:
…rategy

test case is included in jbpm PR for this jira

PR #6196, opened 2024-12-12
#1710 - library still present and requires LICENSE file update

porcelli:
Related: https://github.com/apache/incubator-kie-issues/issues/1710

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -740,3 +740,385 @@ OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
 OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS RELEASE, EVEN IF ADVISED
 OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------------------------------------------------------------------------------------
+for drools-docs/supplemental-ui/js/search-ui.js
+
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
+
+1.2. "Contributor Version"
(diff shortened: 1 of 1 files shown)


tkobayas:
Hi @porcelli ,

https://github.com/apache/incubator-kie-issues/issues/1710#issuecomment-2540393408
> `drools-docs` is not a part of the software, so the plan is to exclude it from the source distribution on subsequent releases, I thought. We removed it in 10.0.x branch manually, but we would be able to automate it with CI. Isn't it the goal of this GH issue?

I think we don't need to bring back the license statement.

We'd like to:

A) Automate `drools-docs` exclusion in setup-branch CI for release
B) If A is not implemented yet for the next release, we would manually remove it like we did for 10.0.x

WDYT?

jomarko:
Not approving, waiting for response to @tkobayas comment. That was my understanding, that it is not needed as `drools-docs` will be not part of the sources release.

porcelli:
my perspective is that, planned to be removed during release is a temporary solution.. and we have to make sure that we won't overlook it... and to be honest.. the violation is in this repository and should be documented.

actually the real release LICENSE file even ignores this file... but we have it - as based on Apache policies - to have in every repository the LICENSE reflecting the codebase

LICENSE
```diff
@@ -739,4 +739,386 @@ PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
 OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
 OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS RELEASE, EVEN IF ADVISED
-OF THE POSSIBILITY OF SUCH DAMAGE.
\ No newline at end of file
+OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------------------------------------------------------------------------------------
+for drools-docs/supplemental-ui/js/search-ui.js
```
tkobayas, line 745:
```suggestion
for drools-docs/supplemental-ui/js/search-ui.js
    drools-docs/supplemental-ui/css/search.css
```

tkobayas: One more file to add: `drools-docs/supplemental-ui/css/search.css`. Thanks!

mariofusco (approved):

tkobayas (approved):
I agree that we will eventually fix the license problem in `drools-docs`, so approved this PR.

jomarko (approved):
Thank you for detailed explanation @porcelli

PR #6202, opened 2024-12-17
[incubator-kie-drools-6199] Clean up quartz related docs

tkobayas:
**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6199


tkobayas:
There is no longer quartz dependency in drools. Just the calendar cron expression is quartz style (e.g. having a "seconds" field) rather than standard cron.

mariofusco (approved):

lucamolteni (approved):

PR #6210, opened 2025-01-02
[KIE-DROOLS-6190] avoid reattaching a peer of a removed node during incremental compilation

mariofusco:


pibizza (approved):
Thanks! Looks good to me.

PR #6213, opened 2025-01-08
[incubator-kie-issues#1742] DMN: B-FEEL implementation

### LINKED ISSUE #1742: DMN: B-FEEL implementation
Verify/implement the new B-FEEL version inside DMN.
B-FEEL should be optionally enabled (i.e., disabled by default) setting expression language to

`“https://www.omg.org/spec/DMN/20240513/B-FEEL/”`

Verify the specs

[B-FEEL - Proposal.docx](https://github.com/bamoe/private-bamoe-issues/files/15232729/B-FEEL.-.Proposal.docx)

Requires 

- [x] https://github.com/apache/incubator-kie-issues/issues/73
- [x] https://github.com/apache/incubator-kie-issues/issues/1187
- [x] https://github.com/apache/incubator-kie-issues/issues/1206
- [x] https://github.com/apache/incubator-kie-issues/issues/1344

to be fixed, before

@baldimir @yesamer

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/1742

The B-FEEL proposal simply define a list of "default" values to return when the standard FEEL would return `null`

This PR:

1. create the `FEELDialect` enum to define different feel dialects, for the moment being the "standard" one and the B-FEEL one
2. populate the evaluation context with the current feel dialect
3. define `Object defaultValue() ` and  `List emendedList(List toEmend)` inside `FEELFunction`, to be override on specifc needs by concrete implementation
4. returns the result of the above methods in when the original result is null and the dialect is B-FEEL
5. add a specific BFEELTest to verify the cases listed in the B-FEEL proposal

The PR touched lot of files due to the pervasiveness of the logic, but the design itself seems clear, at least to me.

@AthiraHari77  @bncriju 

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


baldimir (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELBooleanFunction.java
```diff
@@ -0,0 +1,27 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
yesamer: @gitgabrio ouch

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELCollectionFunction.java
```diff
@@ -0,0 +1,29 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
yesamer: @gitgabrio ouch

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELDurationFunction.java
```diff
@@ -0,0 +1,29 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
yesamer: @gitgabrio ouch

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELNumberFunction.java
```diff
@@ -0,0 +1,35 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
yesamer: @gitgabrio ouch

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELStringFunction.java
```diff
@@ -0,0 +1,27 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
yesamer: @gitgabrio ouch

gitgabrio: ...

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/IsFunction.java
```diff
@@ -45,7 +46,7 @@ public FEELFnResult<Boolean> invoke(@ParameterName("value1") Object value1, @Par
                 return FEELFnResult.ofResult(BooleanEvalHelper.isEqualDateTimeInSemanticD(left, right));
             } // fallback; continue:
         }
-        Boolean fallback = BooleanEvalHelper.isEqual(value1, value2); // if null implying they are not the same semantic domain value.
+        Boolean fallback = BooleanEvalHelper.isEqual(value1, value2, FEELDialect.FEEL); // if null implying they are not the same semantic domain value.
```
jomarko, line 50: Does have `is` same behavior for `FEEL` and `BFEEL`? In other places I saw ve have on similar place code like `ctx.getDialect()`, not needed here?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/ListContainsFunction.java
```diff
@@ -56,7 +57,8 @@ public static boolean itemEqualsSC(Object value, Object itemFromList) {
         if (value instanceof String) {
             return value.equals(itemFromList);
         } else {
-            Boolean dmnEqual = BooleanEvalHelper.isEqual(value, itemFromList);
+            // Defaulting FEELDialect to FEEL
+            Boolean dmnEqual = BooleanEvalHelper.isEqual(value, itemFromList, FEELDialect.FEEL);
```
jomarko: `list contains` can not be used with `BFEEL`? I do not know `drools` codebase os much but in `ListContainsFunction.java` it seems no mention of `BFEEL`.

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/BFeelChecks.dmn
```diff
@@ -0,0 +1,74 @@
...
+      </dmn:literalExpression>
+   </dmn:decision>
+   <dmn:decision id="_F8CA8F98-EE6D-4D56-BCA0-1C716FAC2057" name="Decision2">
+      <dmn:extensionElements/>
+      <dmn:variable id="_C932A158-30E0-438B-8930-5704002CEA97" name="Decision2" typeRef="boolean"/>
+      <dmn:informationRequirement id="_B3961DA9-EB27-41E0-AAF2-386084E9E600">
+         <dmn:requiredInput href="#_DBC698E3-45FC-4D92-A1EC-B8C34629C259"/>
+      </dmn:informationRequirement>
+      <dmn:literalExpression id="_4B499A2E-513E-4938-8C74-C06E7DF47C73">
+         <dmn:text>user = 1</dmn:text>
```
jomarko: both `Decision1` and `Decision2` uses the same expression `user = 1` what does not bring much from test point of view:
```
        final DMNContext ctx1 = runtime.newContext();
        ctx1.set("user", "a");
        final DMNResult dmnResult1 = runtime.evaluateAll(dmnModel, ctx1);
        assertThat(dmnResult1.getDecisionResultByName("Decision1").getResult()).isEqualTo(false);
        assertThat(dmnResult1.getDecisionResultByName("Decision2").getResult()).isEqualTo(false);
        assertThat(dmnResult1.getDecisionResultByName("Decision3").getResult()).isEqualTo(Collections.emptyList());
```

I think we should either change one expression to like:
- `user = [true|false]`
- `user = [date expression]`
- `user = [collection]`
- ....

Or we should remove one of the nodes `Decision1`, `Decision2`

jomarko (commented):
Thank you for the PR @gitgabrio , especially for massive test coverage. I put some comments inline, some of them just for my better understanding. 

### Formatting
One general observation is about formatting changes, that I ma not sure are expected as part of this PR:
- on some places we add `<p> tag into Apache license header, while some licenses do not contain it
- on some places we split long string into concatenation, however not sure why it seems in codebase we have lines that are longer than places I have in mind, so not sure, is there some line length limit for the drools repository?

### FEELDialect.java
Do I understand correctly that B-FEEL is always represented by single value "https://www.omg.org/spec/DMN/20240513/B-FEEL/" while FEEL can be represented by multiple values, depending on the DMN version? My guess would be also for B-FEEL may appear more values/version in future? If that is true maybe the `FEELDialect` could be somehow refactored? Not saying this is somehow blocking from merge, just wanted to hear your opinion.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/ListContainsFunction.java
gitgabrio: Thanks for spotting that! Double checking....

gitgabrio:
@jomarko 
thanks for the carefull check of this PR!
About formatting, my bad, sorry.

About BFEEL multiple definitions, I also share some doubts, TBH.
The proposal doc does not mention any other possible identifier and syasy that the scope of BFEEL is only to manage the `null` values (without any reference to a specific FEEL version)

> B-FEEL shares the same grammar as FEEL but alters the semantics to be friendlier and more intuitive toward non-IT users.
In FEEL, the null value is used to both represent missing data or an execution error. In B-FEEL, null is used only to represent missing data. All operations and built-in functions that returns null in FEEL when an error occurs have their semantics modified in B-FEEL to return a non-null value. A warning message should still be produced when an error occurs.

So, my interpretation is that, regardless of the FEEL expression language used, whenever the "standard" version would return `null`, having defined the BFEEL dialect should replace it with "default" values.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/ListContainsFunction.java
baldimir: The logic should apply to all functions returning a boolean. So if it is missing on list contains, it should be added there.

gitgabrio: Thanks @baldimir - so it should also be applied to every function returning a number, a list, or whatever, IIUC

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/IsFunction.java
gitgabrio: I've addressed it. Anyway, I also kept the old "workaround" (the invocation with the FEEL dialect) to avoid potentially breaking changes

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/BFEELValueFunctionTest.java
```diff
@@ -0,0 +1,232 @@
...
+        FEEL_STRING_FUNCTIONS.add(SubstringFunction.INSTANCE);
+
+        FEEL_TIME_FUNCTIONS = new HashSet<>();
+        FEEL_TIME_FUNCTIONS.add(TimeFunction.INSTANCE);
+    }
+
+    @Test
+    void testBooleanFunctions() {
+        commonTest(FEEL_BOOLEAN_FUNCTIONS, false);
+    }
```
pibizza: You can redo all these tests as parametric. It will be clearer.

gitgabrio: @pibizza 
Done 😄

baldimir (approved):

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNInputRuntimeBFEELTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
jomarko: it is minor thing, however I think we should not include similar code as part of this PR and try to use the same withour `<p>`

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/InNode.java
```diff
@@ -26,6 +26,7 @@
 import org.kie.dmn.feel.runtime.Range;
 import org.kie.dmn.feel.runtime.UnaryTest;
 import org.kie.dmn.feel.runtime.functions.ListContainsFunction;
```
jomarko: I think this is unused now

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/BooleanEvalHelper.java
```diff
@@ -179,6 +180,70 @@ public static Boolean isEqualTimeInSemanticD(TemporalAccessor left, TemporalAcce
...
 
+    /**
+     * This method consider if the <code>value</code> object is a <code>String</code>
+     * In that case, return the <code>String.equal</code> result
+     * Otherwise, default to the <code>isEqual</code>
+     * @param value
+     * @param itemFromList
+     * @return
+     */
+    public static boolean isEqualsStringCompare(Object value, Object itemFromList) {
```
jomarko, line 191: Could we please improve this method name? I am not native english speaker, sorry if the problem is on my side only, however, from my point of view, the current name `isEqualsStringCompare` do not give enough information that:
- we compare strings for equality
- or we use FEELDialect logic for comparing objects

jomarko (commented):
Thank you for the PR @gitgabrio. I have one more question about the `package org.kie.dmn.feel.runtime.functions` tests. There we basically have tests for evaluation of expressions for each DMN built in function - using FEEL dialect. We do not use B-FEEL dialect there is that correct? And we are safe to keep it as it, because B-FEEL affect evaluation only for small amount of cases? For example when function is invoked with some invalid argument, then returned default value is different in FEEL vs B-FEEL and these default values we test in `BFEELValueFunctionTest` ?  Is my understanding correct?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/BooleanEvalHelper.java
gitgabrio: HI @jomarko 
I see your point and in general I agree.
Alas, in this very case, the method does both you mention, depending on the parameters.
I could not find a better way to describe (`isEqualWtihStringCompareOrByDefaultIsEqual` ? )
and that's why I also add javadoc.
Moreover, I'm not even 100% sure why this method is needed (from functional POV), even if I see it is used in two places. 
Anyway, as small "plus", keep in mind that this method previously was named `isEqualSC` and it was in a completely unrelated class, so, even if not "perfect", it seems a small improvement 😊

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNInputRuntimeBFEELTest.java
gitgabrio: Thanks for spotting that. I fixed.
Anyway, the root issue is that some license headers are written as _**javadoc**_, starting with
`/**`
for which a new line has to have the `<p>`, that my IDE keep introduce whenever I reformat the code.

The license header should be written as _**comment**_, starting with
`/*`

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/BooleanEvalHelper.java
```diff
@@ -30,6 +30,7 @@
 import java.util.Optional;
 import java.util.function.BiPredicate;
 
+import org.kie.api.io.Resource;
```
jomarko: is this import needed?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/BooleanEvalHelper.java
jomarko: ok, thank you

jomarko (commented):
Thank you for updates

jomarko (approved):

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/BFeelChecks.dmn
```diff
@@ -0,0 +1,64 @@
...
+  specific language governing permissions and limitations
+  under the License.
+  -->
+
+<dmn:definitions xmlns="http://www.trisotech.com/definitions/_238bd96d-47cd-4746-831b-504f3e77b442"
+                      xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/"
+                      xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
+                      xmlns:dmndi="https://www.omg.org/spec/DMN/20230324/DMNDI/"
+                      xmlns:feel="https://www.omg.org/spec/DMN/20240513/B-FEEL/"
+                      xmlns:dmn="https://www.omg.org/spec/DMN/20230324/MODEL/"
```
yesamer, line 26: @gitgabrio As B-FEEL is a DMN 1_6 feature, I guess we should move this file into `DMNv1_6` directory, and we should update the xmlns:dmn to 1.6, I guess.

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/BFeelOverrideChecks.dmn
```diff
@@ -0,0 +1,73 @@
...
+  specific language governing permissions and limitations
+  under the License.
+  -->
+
+<dmn:definitions xmlns="http://www.trisotech.com/definitions/_238bd96d-47cd-4746-831b-504f3e77b442"
+                      xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/"
+                      xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
+                      xmlns:dmndi="https://www.omg.org/spec/DMN/20230324/DMNDI/"
+                      xmlns:feel="https://www.omg.org/spec/DMN/20240513/B-FEEL/"
+                      xmlns:dmn="https://www.omg.org/spec/DMN/20230324/MODEL/"
```
yesamer, line 26: @gitgabrio As B-FEEL is a DMN 1_6 feature, I guess we should move this file into `DMNv1_6` directory, and we should update the xmlns:dmn to 1.6, I guess.

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/ConstraintsChecksBFeel.dmn
```diff
@@ -0,0 +1,94 @@
...
+  specific language governing permissions and limitations
+  under the License.
+  -->
+
+<dmn:definitions xmlns="http://www.trisotech.com/definitions/_238bd96d-47cd-4746-831b-504f3e77b442"
+                      xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/"
+                      xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
+                      xmlns:dmndi="https://www.omg.org/spec/DMN/20230324/DMNDI/"
+                      xmlns:feel="https://www.omg.org/spec/DMN/20240513/B-FEEL/"
+                      xmlns:dmn="https://www.omg.org/spec/DMN/20230324/MODEL/"
```
yesamer, line 26: @gitgabrio As B-FEEL is a DMN 1_6 feature, I guess we should move this file into `DMNv1_6` directory, and we should update the xmlns:dmn to 1.6, I guess.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/DateFunction.java
```diff
@@ -51,6 +52,7 @@ public class DateFunction
                                                   .withResolverStyle(ResolverStyle.STRICT);
     }
 
+
```
yesamer: @gitgabrio Please save github HD space removing this unnecessary line :asd:

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/MatchesFunction.java
```diff
@@ -57,4 +58,9 @@ public FEELFnResult<Boolean> invoke(@ParameterName("input") String input, @Param
             return FEELFnResult.ofError(new InvalidParametersEvent(Severity.ERROR, errorMessage, e));
         }
     }
+
+    @Override
+    public Object defaultValue() {
+        return false;
+    }
```
yesamer: @gitgabrio I guess this is override is not needed, as it duplicates the same behavior of the superclass's logic.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/impl/RangeImpl.java
```diff
@@ -97,32 +98,33 @@ public boolean isWithUndefined() {
         return withUndefined;
     }
 
-    private Boolean finiteRangeIncludes(Object param) {
+    private Boolean finiteRangeIncludes(FEELDialect feelDialect, Object param) {
+        // Defaulting FEELDialect to FEEL
```
yesamer: @gitgabrio This comment it's a bit misleading to me, where FEELDialect is set to FEEL as default?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/BaseFEELFunction.java
```diff
@@ -80,6 +81,14 @@ public Object invokeReflectively(EvaluationContext ctx, Object[] params) {
                 CandidateMethod cm = getCandidateMethod(ctx, params, isNamedParams);
 
                 if (cm != null) {
+                    if (cm.actualParams != null) {
+                        for (int i = 0; i < cm.actualParams.length; i++) {
+                            if (cm.actualParams[i] instanceof List list) {
+                                cm.actualParams[i] =  getFEELDialectAdaptedList(ctx, list);
+                            }
+                        }
+                    }
```
yesamer, line 90: @gitgabrio IIUC, the function parameters are coerced based on the FEEL dialect, right?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELFunction.java
```diff
@@ -70,6 +70,24 @@ public interface FEELFunction {
...
     Object invokeReflectively(EvaluationContext ctx, Object[] params);
 
+    /**
+     * The default value to return instead of <code>null</code>, to be used with the B-FEEL (and other different dialects) syntax
+     * @return
+     */
+    default Object defaultValue() {
+        // To be overridden by specific classes for B-FEEL compliance
+        return null;
+    }
```
yesamer, line 80: @gitgabrio Considering the vaule returned by `defaultValue` is used in a very specific scope, that is when B-FEEL is enabled (according to the usage in the BaseFEELFunction class), I would consider removing the default behavior of this method in FeelFunction interface. In this way, we will have the opportunity to discover at compile-time if we forgot to correctly manage a specific Function with B-FEEL and we'll remove any chance of returning null in such a case. Eg. if with the current behavior, we forget to manage a specific function overriding that method, with this default logic we will return null as a default value, that is conflicting with the B-FEEL specs that should never return null. If I wasn't clear enough, please reach me in private

yesamer (commented):
Thank you @gitgabrio, please consider my comments:

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/ConstraintsChecksBFeel.dmn
gitgabrio: For the moment being, I think it is better to leave it as it is, and then we'll review later. WDYT ?

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/BFeelOverrideChecks.dmn
gitgabrio: For the moment being, I think it is better to leave it as it is, and then we'll review later. WDYT ?

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/B-FEEL/BFeelChecks.dmn
gitgabrio: For the moment being, I think it is better to leave it as it is, and then we'll review later. WDYT ?

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/MatchesFunction.java
gitgabrio: 🤔 
Point is that there are lot of other Functions that does not covered in the boundaries of the BFEEL, so their original behavior is kept with this default method

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/BaseFEELFunction.java
gitgabrio: Yup. I used the term "adapted" because with "coerce" we usually mean type casting  (e.g. string to bigdecimal) while the BFEEL semantic is different (null to default value)

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/FEELFunction.java
gitgabrio: 🤔
Point is that there are lot (?) of other Functions that are not covered in the boundaries of the BFEEL, so their original behavior is kept with this default method

yesamer: Ok thank you, I assumed ALL the functions were involved -
That means the NOT involved functions should continue to act like regular FEEL (returning null is possible), is that correct?

yesamer: The following sections present the semantics of B-FEEL and compare it to the semantics of FEEL. **Anything not covered in this chapter has the behavior described for FEEL** in Chapter 10.

yesamer (approved):

PR #6219, opened 2025-01-16
[DO-NOT-MERGE] Experiment a special prefix for Custom Operator

tkobayas:
Experiment to use `##` as a prefix for Custom Operator.

This PR sets profile `enableNewParser` as activeByDefault to test with the new parser. We will not make it default yet. We will follow a gradual approach as discussed in the dev ML.

### DIFF
--- a/.github/workflows/pr-downstream.yml
+++ b/.github/workflows/pr-downstream.yml
@@ -33,6 +33,7 @@ on:
 
 jobs:
   kogito-downstream-build:
+    if: false
     concurrency:
       group: pr-${{ matrix.job_name }}_${{ matrix.os }}_${{ matrix.java-version }}_${{ matrix.maven-version }}_${{ github.head_ref }}
       cancel-in-progress: true
--- a/.github/workflows/pr-drools.yml
+++ b/.github/workflows/pr-drools.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Build Chain
         uses: apache/incubator-kie-kogito-pipelines/.ci/actions/build-chain@main
         env:
-          BUILD_MVN_OPTS_CURRENT: '-Dfull -Dreproducible'
+          BUILD_MVN_OPTS_CURRENT: '-Dfull -Dreproducible -DenableNewParser'
           MAVEN_OPTS: "-Dfile.encoding=UTF-8"
         with:
           definition-file: https://raw.githubusercontent.com/${GROUP:apache}/incubator-kie-kogito-pipelines/${BRANCH:main}/.ci/pull-request-config.yaml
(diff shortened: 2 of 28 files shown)


tkobayas:
Hi @mariofusco @lucamolteni @baldimir @gitgabrio @mdproctor @porcelli 

This PR is a draft implementation of "Custom Operator with a special prefix" mentioned in "Slim down DRL syntax with New Antlr4 Parser".

https://docs.google.com/document/d/1P6DuBGRNzz5JCOQvPOngyvrqzat2F0QOt0s1VNwFVE0/edit?tab=t.0

Here I use `##` as a prefix for Custom Operator.

Some points:

- I cannot find a single character which is suitable for this purpose (for example, `#` is used for subtype cast), so I chose `##`. Do you think of any other good candidates?
- The fixes for `isA` in `drools-traits` test cases a little annoying. If I add `isA` to DRLLexer.g4 as builtInOperator, we don't need to fix the test cases. Shall I do it?

porcelli:
@tkobayas though call on the symbol choices 😬 

+1 for isA for trait become built-in, so it won't be much annoying with a `custom` but `built-in` capability.

now in terms of symbol... a. few ideas (brain dump):
- ##customOp (current)
- @@customOp
- <customOp>

Honestly didn't really liked any.. so I'd be good with any - including the proposed.

mariofusco:
> I cannot find a single character which is suitable for this purpose (for example, `#` is used for subtype cast), so I chose `##`. Do you think of any other good candidates?

I cannot think to any better alternative too. I just wonder if reusing an existing symbol could create any ambiguity. For instance I cannot think to a situation when using `:` or even better `%` may not work, but it's impossible to foresee every possible scenarios without giving it a try. If there is any possibility to create any potential issue, even in future, it's ok for me to play safer and keep the `##` operator prefix.
 
> The fixes for `isA` in `drools-traits` test cases a little annoying. If I add `isA` to DRLLexer.g4 as builtInOperator, we don't need to fix the test cases. Shall I do it?

Sorry, but I disagree with @porcelli on this and I'd prefer to keep `isA` a custom operator, because

1. Traits are no more a core feature of Drools. They are implemented in a separated module and actually it was quite hard to perform this separation at time, so I would like to keep it separated, even at the parser level.
2. Traits are an almost unused feature, or at least I'm not aware of any user or customer using it. Do you?
3. Traits have a complicated (highly dynamic) implementation, and because of it they are also the only feature that we weren't able to provide also in the executable model. Also for these reason we repeatedly discussed if we want to mark them for deprecation (or completely remove).
4. Implementing `isA` as a custom operator will give us a chance to also do some dogfeeding with this new syntax and have a meaningful way to test it.

tkobayas:
Thanks @mariofusco , I agree with your points.

@porcelli , if you don't have a strong objection, I would go with the fixes for `isA` test cases.

porcelli:
I’m good with it; feel free to proceed @tkobayas !

tkobayas:
Thanks, I'll continue the work on https://github.com/apache/incubator-kie-drools/issues/6220

PR #6251, opened 2025-02-12
[incubator-kie-drools-6243] Property reactivity doesn't react to supe…

tkobayas:
…r class properties overridden by sub class getter

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6246

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/PhreakPropagationContext.java
+++ b/drools-core/src/main/java/org/drools/core/common/PhreakPropagationContext.java
@@ -233,7 +233,7 @@ public class PhreakPropagationContext
         Class<?> classType = classObjectType.getClassType();
         String pkgName = classType.getPackage().getName();
 
-        if (classType == modifiedClass || "java.lang".equals(pkgName) || !(classType.isInterface() || modifiedClass.isInterface())) {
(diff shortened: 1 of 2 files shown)


drools-core/src/main/java/org/drools/core/common/PhreakPropagationContext.java
```diff
@@ -233,7 +233,7 @@ public PropagationContext adaptModificationMaskForObjectType(ObjectType type, Re
         Class<?> classType = classObjectType.getClassType();
         String pkgName = classType.getPackage().getName();
 
-        if (classType == modifiedClass || "java.lang".equals(pkgName) || !(classType.isInterface() || modifiedClass.isInterface())) {
+        if (classType == modifiedClass || "java.lang".equals(pkgName)) {
```
tkobayas, line 236: The condition `!(classType.isInterface() || modifiedClass.isInterface())` allows the mask adaptation only for cases where Interface is involved. This was introduced by https://github.com/apache/incubator-kie-drools/commit/6d2840d9ac5e322b9df8b29355d5e5f5d1786004

But I think it's fine to allow for cases where both are classes. So this is the only change for this fix.

WDYT? @mariofusco

mariofusco (approved):

PR #6275, opened 2025-03-05
[Fix apache/incubator-kie-issues#1857] Adding triggerCount

### LINKED ISSUE #1857: Retrigger operation and data index feedback
Kogito currently support [retrigger](https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/addons/common/process-management/src/main/java/org/kie/kogito/process/management/BaseProcessInstanceManagementResource.java#L204-L215) operation. 
This operation allows an user to resume execution of a process instance that failed in a particular node (either after updating the model or after fixing the external service that failed)

On the other hand, Data index stores a list of executed node instances associated to a particular process instance. There is also an error property associated to the process instance, which contains the error message, but not a pointer to the node instance that failed

When the node is re-triggered, a new list of nodes, consequence of the dispatching of the re trigger request, is added to the existing list.

There is a problem for users that want to know  what are the nodes that are part of the re-trigger execution.  To cope with it, two alternatives come to my mind:
1) add a new `resume` property to data index process instance object, which will contain a list of items ( a workflow instance can be resumed many times), each item containing the nodeinstanceid that was retriggered and the date of the retrigger attemp. 
2) add a new `retriggered` flag to the node instance that was retriggered. 

Related with that, Data index is fed by events published by runtimes. Runtimes only publishes the events that fed data index when an unit of work is completed. A unit of work is completed when the process instance state is completed, error or waiting (if the workflow is waiting for event)

For workflows that basically invoke synchronous operations, this means that once the retrigger is executed, the state that the user will see in data index will be either completed or error. If the retrigger was executed when the process instance was in failed state (most typical situation) and the retrigger failed again (which can happen) the user wont see any change in process state (he will see more node instances added to the process instance, though). Therefore, it will be convenient to add error message to the node instance that failed. This way, a user can query by node instances which error messages are not null to find out how many times the process instance has failed. 

Also, it is probably sound (although not sure it is really required) to send a State Process instance event before dispatching the retrigger operation to set the status to active in data-index. This might be useful in case the workflow contains long operations and the event with the completed/error/waiting state result of the retrigger operation takes a while to be published.  

**Implemented solution**:

- `retrigger` and `errorMessage` properties has been added to `NodeInstance` GraphQL interface
- `nodeInstanceId` has been added to `ProcessInstance` Graphql `error` property
- A new event is fired when a Process instance retrigger operation is performed. This event set `ProcessInstance` Grapqhl `state` property to ACTIVE and clear the `error` property

fjtirado:
Fix https://github.com/apache/incubator-kie-issues/issues/1857

Add triggerCount method to ProcessNodeTriggeredEvent interface

To be merged with https://github.com/apache/incubator-kie-kogito-runtimes/pull/3858 and https://github.com/apache/incubator-kie-kogito-apps/pull/2204

### DIFF
--- a/kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
+++ b/kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
@@ -1,0 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.kie.api.event.process;
+
+/**
+ * A runtime event that is generated when a ProcessInstance is completed.
+ */
+public interface ProcessRetriggeredEvent
+    extends
+    ProcessEvent {
+
+}
--- a/kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
+++ b/kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
@@ -19,7 +19,7 @@
 package org.kie.api.event.process;
 
 /**
- * A runtime event that is generated when a ProcessInstance is completed.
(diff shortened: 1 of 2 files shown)


kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
```diff
@@ -0,0 +1,28 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.kie.api.event.process;
+
+/**
+ * A runtime event that is generated when a ProcessInstance is completed.
```
gmunozfe: Not sure if this description is correct as the name of the interface is related to "retriggered"

gmunozfe (commented):
Check the description comment of the new interface

kie-api/src/main/java/org/kie/api/event/process/ProcessRetriggeredEvent.java
fjtirado: its wrong, its a copy paste

wmedvede (approved):

pefernan (approved):

PR #6278, opened 2025-03-13
[incubator-kie-drools-6274] Support multi module and indexed dependen…

tkobayas:
…cies

**Issue**: 

* https://github.com/apache/incubator-kie-drools/issues/6274

### DIFF
--- a/drools-quarkus-extension/drools-quarkus-deployment/src/main/java/org/drools/quarkus/deployment/DroolsAssetsProcessor.java
+++ b/drools-quarkus-extension/drools-quarkus-deployment/src/main/java/org/drools/quarkus/deployment/DroolsAssetsProcessor.java
@@ -18,11 +18,17 @@
  */
 package org.drools.quarkus.deployment;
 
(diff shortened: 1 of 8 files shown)


tkobayas:
Hi @mariofusco  , I have created a draft PR for https://github.com/apache/incubator-kie-drools/pull/6278

Please check if it is an acceptable feature and share any concerns.

- Add `drools.asset.dependencies=groupId:artifactId,groupId:artifactId,...` in `application.properties` as depedencies to be scanned for rule assets. (If we scan all dependencies, it will take time).
- Such dependencies should be direct dependencies of the quarkus application project. (If we check names of all transitive dependencies, it will take time. But I'm not strongly opinionated)
- Of course, such external DRL files themselves are not subject to hot-reload, but `mvn install` for the dependency project triggers the hot-reload in quarkus dev-mode, so I think it's good enough.

Thanks!

mariofusco (approved):

lucamolteni (approved):

PR #6287, opened 2025-03-25
removal drools-docs from the repository as it contains invalid artifa…

porcelli:


porcelli:
although we don't have PR checks working, I'd argue that we don't need to block this change (as it blocks the 10.1.0 release).

thoughts?

tkobayas (approved):
I agree that it's safe to remove without the PR checks. Thanks.

mariofusco (approved):

PR #6308, opened 2025-04-14
Fix split package + add script to detect them

### LINKED ISSUE #6306: Conflicting packages in different modules with version 10.0.0
Hi

I tried to update from version `9.44.0.Final` to `10.0.0`. After changing the version I got `error: module xyz reads package org.drools.core.reteoo from both org.drools.core and org.drools.base` during compiling with my own multi module project. As on your main branch there are still the same packages in different modules (and I couldn't find an existing issue), I created this issue.

Is there an alternative ways of using Drools 10 or was this simply not taken into account during changing the package names?

If you need more information from my site, please let me know

mariofusco:
Fixes https://github.com/apache/incubator-kie-drools/issues/6306 and a github action launching a script to detect this problem and avoid it in future.


gitgabrio (approved):
Hi @mariofusco 
Thanks for PR. Fine for me, when license check is fixed

tkobayas (approved):

PR #6311, opened 2025-04-16
[incubator-kie-issues#1927] Removing DMN-related xsd from codebase

### LINKED ISSUE #1927: Remove DMN xsd files from codebase
Relates to https://github.com/apache/incubator-kie-issues/issues/1926

Scope of this ticket is to:

1.  create a module that automatically download them from reference source, and package them inside a jar
2. make all the modules that need those files depending on the above.

Possible approach:

Download files:
1.  [download-maven-plugin](https://github.com/download-maven-plugin/download-maven-plugin)
2. maven-ant-run **
3. other ?

Refer xsd-bundle artifact:

1. maven resources (as for kie-dmn-test-resources)
2. [maven remote resources plugin](https://maven.apache.org/plugins/maven-remote-resources-plugin/examples/sharing-resources.html)
3. other ?

**
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-antrun-plugin</artifactId>
    <executions>
        <execution>
            <id>download-files</id>
            <phase>prepare-package</phase>
            <goals>
                <goal>run</goal>
            </goals>
            <configuration>
                <target>
                    <!-- download file -->
                    <get src="http://url/to/some/file"
                         dest="${project.build.directory}/downloads/"
                         verbose="false"
                         usetimestamp="true"/>
                 </target>
             </configuration>
         </execution>
     </executions>
 </plugin>
```

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/1927

This PR:

1.  create a org.kie:kie-dmn-xsd-resources that, at compile time, download DMN-related xsds and bundle them in the generated jar
2. remove DMN-related xsd from all kie-dmn modules
3. modify modules that requires the above xsd, making them depend on the kie-dmn-xsd-resources and unpacking required xsd at expected name/position during the generate-resources phase
4. add a Readme.dmn file inside kie-dmn-xsd-resources to provide bare-minimum explanation

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


kie-dmn/kie-dmn-backend/pom.xml
```diff
@@ -36,8 +36,19 @@
   
   <properties>
     <java.module.name>org.kie.dmn.backend</java.module.name>
+    <dependency-plugin.version>3.6.1</dependency-plugin.version>
```
yesamer: @gitgabrio Is it possible to declare it globally?

porcelli:
really well done @gitgabrio - thank you for the effort!

however I'm afraid that the whole software might be broken if omg - for whatever reason - don't have that files available.

what if we moved the files to our kie website? https://kie.apache.org 

website is not part of the software release, so we have no legal restrictions for it.

thoughts?

kie-dmn/kie-dmn-backend/pom.xml
gitgabrio: @yesamer I removed it completely - at the end, it is not needed at all

gitgabrio:
> really well done @gitgabrio - thank you for the effort!
> 
> however I'm afraid that the whole software might be broken if omg - for whatever reason - don't have that files available.
> 
> what if we moved the files to our kie website? https://kie.apache.org
> 
> website is not part of the software release, so we have no legal restrictions for it.
> 
> thoughts?

@porcelli  I've no _strong_ opinion, TBH. 
Either case, there could be an outage, or files could be moved elsewhere (for whatever reason).

Some details make me prefer the OMG URL:
1. it is the "official" reference, and whenever updated/new models are published, those are the one to consider
2. the "kie-apache" approach would create a new burden to maintain a new "data storage"
3. the "kie-apache" approach would lead to two sources-of-truth/two databases-in-sync pattern, that generally is better to avoid.

Anyway, if someone else is willing to maintain it - and that means always, at every modification of DMN xsd, fine for me.

@baldimir  @yesamer : your thoughts ?

kie-dmn/kie-dmn-backend/pom.xml
```diff
@@ -38,6 +38,16 @@
...
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-dmn-xsd-resources</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
tiagobento: Shouldn't we be adding this new module to the BOM so that we don't need to declare it in multiple places?

kie-dmn/kie-dmn-xsd-resources/Readme.md
```diff
@@ -0,0 +1,35 @@
...
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+  -->
+
+KIE :: Decision Model Notation :: XSD-Bundle
```
tiagobento, line 20:
```suggestion
KIE :: Decision Model and Notation :: XSD-Bundle
```

kie-dmn/kie-dmn-xsd-resources/Readme.md
```diff
@@ -0,0 +1,35 @@
...
+============================================
+
+This modules is used to bundle all the DMN-related xsd in one single jar, fetching them from [official repository](https://www.omg.org/spec/DMN/).
```
tiagobento:
```suggestion
This module is used to bundle all the DMN-related XSD files in one single jar, fetching them from the [official repository](https://www.omg.org/spec/DMN/).
```

kie-dmn/kie-dmn-xsd-resources/pom.xml
```diff
@@ -0,0 +1,297 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie</groupId>
+    <artifactId>kie-dmn</artifactId>
+    <version>999-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>kie-dmn-xsd-resources</artifactId>
+  <name>KIE :: Decision Model Notation :: XSD-Bundle</name>
```
tiagobento, line 32:
```suggestion
  <name>KIE :: Decision Model and Notation :: XSD-Bundle</name>
```

kie-dmn/kie-dmn-xsd-resources/pom.xml
```diff
@@ -0,0 +1,297 @@
...
+  <description>Module meant to bundle all DMN-related xsd files</description>
+
+  <properties>
+    <java.module.name>org.kie.dmn.xsd.resources</java.module.name>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
tiagobento: Do we need those repeated here?

kie-dmn/kie-dmn-xsd-resources/pom.xml
```diff
@@ -0,0 +1,297 @@
...
+    <download-maven-plugin.version>2.0.0</download-maven-plugin.version>
```
tiagobento: Maybe this should be managed in the build-parent POM?

kie-dmn/kie-dmn-xsd-resources/pom.xml
```diff
@@ -0,0 +1,297 @@
...
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>io.github.download-maven-plugin</groupId>
+          <artifactId>download-maven-plugin</artifactId>
+          <version>${download-maven-plugin.version}</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
```
tiagobento: Same question here... shouldn't we be managing this in the build parent POM?

tiagobento (commented):
Thanks @gitgabrio. Left some questions/suggestions inline.

A general question I have is whether or not we need to have the `wget` tool available in our PATHs for building this.

Thanks!

yesamer:
> > really well done @gitgabrio - thank you for the effort!
> > however I'm afraid that the whole software might be broken if omg - for whatever reason - don't have that files available.
> > what if we moved the files to our kie website? https://kie.apache.org
> > website is not part of the software release, so we have no legal restrictions for it.
> > thoughts?
> 
> @porcelli I've no _strong_ opinion, TBH. Either case, there could be an outage, or files could be moved elsewhere (for whatever reason).
> 
> Some details make me prefer the OMG URL:
> 
> 1. it is the "official" reference, and whenever updated/new models are published, those are the one to consider
> 2. the "kie-apache" approach would create a new burden to maintain a new "data storage"
> 3. the "kie-apache" approach would lead to two sources-of-truth/two databases-in-sync pattern, that generally is better to avoid.
> 
> Anyway, if someone else is willing to maintain it - and that means always, at every modification of DMN xsd, fine for me.
> 
> @baldimir @yesamer : your thoughts ?

@gitgabrio I would consider that as a mirror, in case the official reference is not available.

tkobayas (approved):
Thank you, @gitgabrio !

jomarko (commented):
Hi @gitgabrio , we should update also:
- https://github.com/apache/incubator-kie-drools/blob/main/.rat-excludes
- https://github.com/apache/incubator-kie-drools/blob/main/LICENSE

@tkobayas do we need to update just the locations, or should we remove XSD entries from there completely?

tkobayas:
Good catch, @jomarko 

> do we need to update just the locations, or should we remove XSD entries from there completely?

As we remove xsd files from the code base, we should remove the entries completely. (note that this PR is for DMN xsd files)

gitgabrio:
> Hi @gitgabrio , we should update also:
> 
>     * https://github.com/apache/incubator-kie-drools/blob/main/.rat-excludes
> 
>     * https://github.com/apache/incubator-kie-drools/blob/main/LICENSE
> 
> 
> @tkobayas do we need to update just the locations, or should we remove XSD entries from there completely?

Thanks @jomarko for spotting that! 👍

gitgabrio:
> > > really well done @gitgabrio - thank you for the effort!
> > > however I'm afraid that the whole software might be broken if omg - for whatever reason - don't have that files available.
> > > what if we moved the files to our kie website? https://kie.apache.org
> > > website is not part of the software release, so we have no legal restrictions for it.
> > > thoughts?
> > 
> > 
> > @porcelli I've no _strong_ opinion, TBH. Either case, there could be an outage, or files could be moved elsewhere (for whatever reason).
> > Some details make me prefer the OMG URL:
> > 
> > 1. it is the "official" reference, and whenever updated/new models are published, those are the one to consider
> > 2. the "kie-apache" approach would create a new burden to maintain a new "data storage"
> > 3. the "kie-apache" approach would lead to two sources-of-truth/two databases-in-sync pattern, that generally is better to avoid.
> > 
> > Anyway, if someone else is willing to maintain it - and that means always, at every modification of DMN xsd, fine for me.
> > @baldimir @yesamer : your thoughts ?
> 
> @gitgabrio I would consider that as a mirror, in case the official reference is not available.

@porcelli @yesamer @baldimir @LightGuard 

Duplicating xsd at a different position respect of referral one seems just an anti-pattern that:
1. brings in more complexity
2. increase the workload
3. aim to solve a problem that currently does not exists, and, if ever it will happen, it will be a consequence of much bigger problems to solve

So, for the moment being, I think it does not make sense to spend time to that. 
If, ever, there will be a real need for that, of if someone else is willing to dedicate time to that, I think it could be done on a later PR.

kie-dmn/kie-dmn-xsd-resources/pom.xml
gitgabrio: `KIE :: Decision Model Notation :: `  is the definition used for all kie-dmn modules - so I would prefer to keep consistency

kie-dmn/kie-dmn-xsd-resources/Readme.md
gitgabrio: KIE :: Decision Model Notation :: is the definition used for all kie-dmn modules - so I would prefer to keep consistency

gitgabrio:
> Thanks @gitgabrio. Left some questions/suggestions inline.
> 
> A general question I have is whether or not we need to have the `wget` tool available in our PATHs for building this.
> 
> Thanks!

It uses Apache HttpClient, see [HttpFileRequester](https://github.com/download-maven-plugin/download-maven-plugin/blob/630ef4a5399bbbdd934f108e69ef62430ee90d48/src/main/java/io/github/download/maven/plugin/internal/HttpFileRequester.java) and its usage

kie-dmn/kie-dmn-backend/pom.xml
```diff
@@ -87,9 +91,59 @@
...
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.kie</groupId>
+                  <artifactId>kie-dmn-xsd-resources</artifactId>
+                  <version>${project.version}</version>
+                  <type>jar</type>
+                  <overWrite>true</overWrite>
+                  <outputDirectory>${project.build.directory}/test-classes/</outputDirectory>
```
yesamer, line 117: @gitgabrio Shouldn't be `${project.build.directory}/test-classes/resources` ?

yesamer (approved):
Thank you @gitgabrio, I have just a comment to raise, please see below. 

About the mirror proposal, I agree with you that can be considered in a future PR.  The ideal solution should be to have a mirror directly provided by the OMG group, but unfortunately this seems not the case.

kie-dmn/kie-dmn-backend/pom.xml
gitgabrio: @yesamer 
Nope. `src/` and ` resources/` folders are "_**sources**_" folder. When the module is compiled, their content ends in _target/classes_ or _target/test-classes_.
I usually do not like to dinamically create/write files inside "sources" folder, and that's why I put them directly under target.
Does this make sense ?

yesamer: It does. Thank you @gitgabrio

baldimir (approved):

porcelli:
OMG is returning 503 on all wget today June 6th

PR #6319, opened 2025-04-22
[incubator-kie-drools-6314] Ability to use pseudo clock with rule uni…

tkobayas:
…ts in tests only

**Issue**:

* https://github.com/apache/incubator-kie-drools/issues/6314

### DIFF
--- a/drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RuleUnitQuarkusTemplate.java
+++ b/drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RuleUnitQuarkusTemplate.java
@@ -23,6 +23,7 @@ import org.drools.modelcompiler.KieBaseBuilder;
 import org.drools.ruleunits.api.RuleUnit;
 import org.drools.ruleunits.api.conf.RuleConfig;
 import org.drools.ruleunits.api.RuleUnits;
+import org.drools.ruleunits.impl.conf.RuleConfigImpl;
 import org.drools.ruleunits.impl.factory.AbstractRuleUnit;
 import org.drools.ruleunits.impl.factory.AbstractRuleUnits;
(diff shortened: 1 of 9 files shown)


drools-ruleunits/drools-ruleunits-api/src/main/java/org/drools/ruleunits/api/conf/RuleConfig.java
```diff
@@ -36,4 +37,7 @@ public interface RuleConfig {
 
     List<RuleEventListener> getRuleEventListeners();
 
+    ClockType getClockType();
+
+    void setClockType(ClockType clockType);
```
tkobayas, line 42: I enhanced `RuleConfig` to override SessionConfiguration at runtime (e.g. for unit test). For now, only ClockType.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RuleUnitQuarkusTemplate.java
```diff
@@ -44,7 +45,7 @@ public CLASS_NAME() {
     }
 
     @jakarta.inject.Inject
-    public CLASS_NA(jakarta.enterprise.inject.Instance<RuleUnits> ruleUnits) {
+    public CLASS_NAME(jakarta.enterprise.inject.Instance<RuleUnits> ruleUnits) {
```
tkobayas, line 48: I happened to find this typo. This was luckily working because RuleUnitWriter processes all constructors, not checking the `CLASS_NA`.

baldimir: Ouch :)

baldimir (approved):

Bonehead5338:
awesome

mariofusco (approved):

PR #6343, opened 2025-05-13
[#1946]Rule engine half constraint or condition is not working as expected

### LINKED ISSUE #1946: Rule Engine: half constraint OR condition is not working as expected
The use of half constraint OR is not working as expected in the DRL file
Example
when $exec : RuleExecution( $req.id=="H1" || == "A4")

this rule is not getting fired even if we set the value of id as H1/A4.

 
Issue:

The generated Java file uses == operator for comparing the string values
return $req.getId() == "H1" || $req.getId() == "A4" 

This doesnt fire the rule, because  comparing in Java with == for Strings doesn't compare value.
 
Solution:
when $exec : RuleExecution( $req.id=="H1" || $req.id== "A4")
the generated java file should be like 

org.drools.modelcompiler.util.EvaluationUtil.areNullSafeEquals(_this.getId(), "H1") ||
        org.drools.modelcompiler.util.EvaluationUtil.areNullSafeEquals(_this.getId(), "A4"));

This works as its using .equal method for String literal comparison.

SoniyaAbraham89:
Fixes apache/incubator-kie-issues/issues/1946

Problem: The use of half constraint OR is not working as expected in the DRL file because it uses == operator for comparing String

Fix: the generated java file uses  org.drools.modelcompiler.util.EvaluationUtil.areNullSafeEquals() method for String comparison


drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/ConstraintParser.java
```diff
@@ -804,17 +816,17 @@ public static TypedExpression getCoercedRightExpression( PackageModel packageMod
...
+                .map(context::getTypedDeclarationById)
+                .anyMatch(optDecl -> {
+                    if (optDecl.isPresent()) {
+                        TypedDeclarationSpec decl = optDecl.get();
+                        if (!decl.isGlobal() && decl.getBelongingPatternDescr() != context.getCurrentPatternDescr()) {
+                            return true;
+                        }
+                    }
+                    return false;
+                });
```
yesamer: @SoniyaAbraham89 Thank you for your PR! May I ask you to revert all the formatting-only changes, just to focus on the real code changes only? Thanks!

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/ConstraintParser.java
```diff
@@ -62,8 +62,6 @@
 import org.drools.model.codegen.execmodel.generator.expressiontyper.ExpressionTyper;
 import org.drools.model.codegen.execmodel.generator.expressiontyper.ExpressionTyperContext;
 import org.drools.model.codegen.execmodel.generator.expressiontyper.TypedExpressionResult;
-import org.drools.mvel.parser.ast.expr.BigDecimalLiteralExpr;
```
SoniyaAbraham89, line 65: Removing the imports as its not used

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/drlxparse/ConstraintParser.java
```diff
@@ -412,7 +410,7 @@ private Expression combineExpressions( TypedExpressionResult leftTypedExpression
         return combo;
     }
 
-    private Expression combineExpressions(List<Expression> leftPrefixExpresssions, List<Expression> rightPrefixExpresssions, Expression combo) {
+    private Expression combineExpressions(List<Expression> leftPrefixExpressions, List<Expression> rightPrefixExpressions, Expression combo) {
```
SoniyaAbraham89, line 413: Handled the spelling mistake

tkobayas:
@SoniyaAbraham89 Thank you very much for the PR. I'm considering a different way of fixing the issue. Please allow me a little time. I'll get back to you soon.

(thinking if we can complement HalfBinaryExpr at the beginning of `ConstraintParser.parseBinaryExpr`, so we can reuse the same logic as the usual case)

tkobayas:
Hi @mariofusco , I have filed https://github.com/apache/incubator-kie-drools/pull/6345 as an alternative fix. Which one is preferred?

mariofusco:
Replaced by https://github.com/apache/incubator-kie-drools/pull/6345

PR #6352, opened 2025-05-20
[NO ISSUE] Quarkus ruleunit extension

pibizza:
This is a patch to create a separate ruleunits extension for quarkus. 

The idea is to simplify the usage of ruleunits in quarkus project, so that it is not necessary to explicitely import the quarkus ruleunit engine.

It also include a codestart specific for ruleunit, which can be useful to jump start the development.

There is no changes to the API for the old quarkus extension, so there is nothing to migrate in that case.

There is a small inconsistency between the names of the package - ruleunit - and the name of the various projects -all using ruleunitS. Since this was already present in the existing drools code, I kept it for consistency.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


porcelli:
@pibizza I'm not entirely sure what's the main difference between current drools extension and this new proposed extension.

It's my understanding that RuleUnits is the way to go for future development, so I'd argue this should be part of the default drools quarkus extension.

gitgabrio:
Hi @pibizza - I'm really doubtful about the idea behind that PR. 
IMO we should strive to remove all possible duplications/multiplications of "common" behaviors (like RuleUnit usage) between the different environments we support (Springboot, Quarkus, plain Java), but this seems to go in other direction, IIUC it
🤔

mariofusco:
If I understand correctly the two main objections to this PR are 

1. ruleunits are the favored and adviced way of doing stuff in drools
2. drools should be neutral between spring and quarkus integration

Regarding 1. I'm strongly against this idea, because ruleunits are useful and better than good-and-old KieBase/KieSession only in specific use cases while they miss some advanced features available in traditional

As for 2. I don't understand how adding one more quarkus extension could bring any damage to spring support.

For these reasons I'm going to reopen this PR, please let's keep this discussion here.

mariofusco (approved):

baldimir:
Hi, I don't want to raise any opinions here myself currently. I just want to ask - if there will be a decision to merge this, could you please wait, till the Quarkus 3.20 upgrade is done? There is already an opened PR chain, so hopefully it will not take a long time. https://github.com/apache/incubator-kie-drools/pull/6356

tkobayas (approved):

yesamer:
Hi all! I have the same Tibor's opinion, can you please wait to merge it until we merge the quarks update? (it should be completed this week)? Thanks!

pibizza:
@baldimir @yesamer I plan to wait at least unti next week. I understand there is some contention on this PR.

gitgabrio:
@mariofusco 
Point is not about 
"adding one more quarkus extension could bring any damage to spring support."
Point is about maintainability and architectural cleanness.

1.  in recent past we had harsh discussion, in community, about repository refactoring; root issue resolving about the inconsistency of our repositories in terms of scope and architectural layer; I repeat, drools repo should be the library framework-agnostic level (as it was enforced years ago exactly by you, by the way 😄  )
2. having quarkus dependency inside drools has, as consequence, the fact that whatever change we need to make related to quarkus, it involves also this repositories, increasing uselessy the effort and time required (see the long and sad history of this [PR](https://github.com/apache/incubator-kie-drools/pull/6356), as example )
3. another consequence is that, having quarkus inside drools, all the versions of the libraries inherited by it, have to be declared in the same common bom; and that lead to a clash whenever the same libraries are inherited also from springboot; e.g.

quarkus X requires some:library:10, so some:library:10 is declared in the common bom
springboot Y requires some:library:11, so, downstream, an exclude workaround has to be implemented for all the springboot cases; and, often, it is not a matter of a single library but most of a network of libraries.
And, again, this eventually resolve in increased cost to maintain and also brittleness of our codebase.

This, ultimately, is the consequence of having built our codebase around quarkus as primary target, and then managed springboot in sort of "workaround mode".


Last, but not least, being all our codebase a de-facto monolith (in terms of release), for users it does not make any difference in which repository a given module is written

porcelli (changes requested):
I still -1 for the approach 

It's very confusing to promote - at same level of support - 2 different models. 

I believe we should pick one - and I argue the one should be Rule Units and promote it.

For non-RuleUnit based - I'm OK with the idea to have a quarkus extension - but would name it something like Legacy or Classic.

mariofusco:
> I still -1 for the approach
> 
> It's very confusing to promote - at same level of support - 2 different models.
> 
> I believe we should pick one - and I argue the one should be Rule Units and promote it.
> 
> For non-RuleUnit based - I'm OK with the idea to have a quarkus extension - but would name it something like Legacy or Classic.

Ruleunits provide only a subset of the feature provided by KieSession and in many cases the syntax is more ambiguous, verbose and confusing. I already had this conversation with @gitgabrio and we seem to agree on this point: if you want to promote one single programming model, please NOT ruleunits.

baldimir:
I agree with Mario and Gabriele, we should not promote only rule units. Many users use the classic APIs and it will be too enforcing, when the classic APIs work well and are very commonly known by people using Drools. I will personally -1 any effort of deprecating classic APIs or promote only rule units, currently, because it doesn't make sense.

baldimir:
Quarkus upgrade PRs are merged.

mariofusco:
> Quarkus upgrade PRs are merged.

Ok, so can we please recap what are the objections in merging this PR?

@gitgabrio I understand all your doubts on having the quarkus extension inside the drools codebase, and I'm open to discuss this, but again it seems to me that this discussion is out of the scope of this PR and this PR itself doesn't change the current situation in any way. Do you agree?

porcelli:
My objection is the unclear what approach users should take when using Drools.

The overall feedback that I hear from users in different levels is that Drools is overwhelming and complex, now if we promote 2 very distinct ways to use Drools.. things became more complex.

I'd compromise if we rename the existing drools to something like Classic (to avoid use of Legacy).

@mariofusco can you provide a high level list of items missing in RuleUnits? It sounds to me that addressing those gaps could be turned a strong roadmap for Drools.

gitgabrio:
@mariofusco 

>  it seems to me that this discussion is out of the scope of this PR and this PR itself doesn't change the current situation in any way. Do you agree?

Nope, I disagree. 😄  , meaning that whenever there are bigger implications (i.e. architectural considerations) behind a PR, we can't simply ignore them because the classes modified are a lower arch level; sorry, this is a bad approach that has been taken a lot of time, result being a sort of messy overall situation (my POV - IMHO). 
Anyway, I did not put any review on this PR, to not block it.

@porcelli 
Of course @mariofusco will provide better inside, but there are some elephants in the room:

1. RuleUnit is not battle-tested as the classic approach - and reliability is THE critical factor for users that need to deliver stable products
2.  moving RuleUnit to the same level of test covering and battle-field verification of the classic approach would require an effort (in terms of time and resources) that probably right now we could not afford
3. side-note: it is not 100% clear, to me, what cases and actual improvements the "RuleUnit" approach is bringing, and if those are still relevant in the "kogito" era 

m2c

mariofusco:
> @mariofusco can you provide a high level list of items missing in RuleUnits? It sounds to me that addressing those gaps could be turned a strong roadmap for Drools.

I agree with everything @gitgabrio wrote, finding especially relevant his point 3. 

Moreover:

1. As a replacement for agenda-groups, ruleunits are supposed to encapsulate groups of related rules. However it has been never made clear how to pass the execution flow from a ruleunit(instance) to another, especially within the drl. The main problem there is that you can create multiple instances from the same ruleunit, but it isn't clear how to identify a specific instance. I made many experiments on this in the past, but without reaching a solution that made me happy.
2. Differently from agenda-groups each ruleunit is backed by its own session. In theory this design promotes better encapsulation. In practice this creates the problem of how to share facts and events across different ruleunits. This should be done by feeding multiple ruleunits with the same datasource, but I always found this idea difficult to understand for users.
3. The use of ruleunits also implies the use of the oopath notation in the drl. You could use the pattern matching notation referring to datasource as entry-points, but this would be extremely verbose and makes useless the fact that datasources are strongly typed. I must admit that I have never been satisfied with oopath: try to translate a drl of medium to high complexity in oopath and you will understand why. Some key features, like accumulate, are particularly inconvenient to be written in oopath. Finally oopath also has some unresolvable ambiguities, for instance if you write something like
```
b: /a/b
/c/d
```
it's impossible for the parser to decide if these are 2 different oopath or simply the second line is the continuation of the oopath started in the first. The parser always assumes the second option, concatenating the 2 lines in a single oopath, but this is often very confusing and surprising for users.

In essence for sure we are not in the position to say that ruleunits+oopath are battle-tested and fully functional equivalent to the traditional KieSession approach and that's why it's difficult to me to consider them the preferred way to write rules in drools.

gitgabrio (approved):
Thanks @pibizza 
I think we may work on further improvements, especially overall architectural, later on!

PR #6354, opened 2025-05-22
[incubator-kie-drools-6350] Custom Operator issue with executable-model

tkobayas:
- Enhanced mvel.jj and PointFreeExpr to accept custom operator with brackets
- Fixed CustomOperatorSpec to be able to generate codes for custom operator with brackets
- Fixed CustomOperatorWrapper and unit test custom operators to handle left/right input vs left/right operand
---
* https://github.com/apache/incubator-kie-drools/issues/6350


drools-test-coverage/test-compiler-integration/src/test/resources/logback-test.xml
```diff
@@ -32,7 +32,7 @@
     <logger name="org.drools" level="warn"/>
 
 <!--     <logger name="org.drools.core.management" level="debug"/> -->
-<!--    <logger name="org.drools.modelcompiler.builder" level="debug"/>-->
+<!--    <logger name="org.drools.model" level="debug"/>-->
```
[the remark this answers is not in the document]
tkobayas: This is intentional

drools-model/drools-mvel-parser/src/main/javacc/mvel.jj
```diff
@@ -6459,11 +6471,12 @@ Expression PointFreeExprMvel():
                           { negated = true; }
             )*
             (
-                (mvelOperator = <MVEL_STARTS_WITH> | mvelOperator = <MVEL_ENDS_WITH> | mvelOperator = <MVEL_LENGTH>)
+                operator = SimpleName()
+                subOperator = StringInBrackets()
```
tkobayas: Rather than hard-coded tokens (e.g. `str[startsWith]`), take an operator with brackets in general. (Temporal operators are handled in PointFreeExpr()).

drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/ast/expr/PointFreeExpr.java
```diff
@@ -33,17 +33,23 @@ public class PointFreeExpr extends Expression {
...
     public PointFreeExpr( TokenRange tokenRange, Expression left, NodeList<Expression> right, SimpleName operator, Boolean negated, Expression arg1, Expression arg2, Expression arg3, Expression arg4 ) {
+        this(tokenRange, left, right, operator, null, negated, arg1, arg2, arg3, arg4 );
+    }
+
+    public PointFreeExpr( TokenRange tokenRange, Expression left, NodeList<Expression> right, SimpleName operator, SimpleName subOperator, Boolean negated, Expression arg1, Expression arg2, Expression arg3, Expression arg4 ) {
         super(tokenRange);
         this.left = left;
         this.right = right;
         this.operator = operator;
+        this.subOperator = subOperator;
```
tkobayas: Now PointFreeExpr has `operator` and `subOperator` (e.g. operator = `str`, subOperator = `startsWith`)

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/expressiontyper/ExpressionTyper.java
```diff
@@ -278,7 +278,7 @@ private Optional<TypedExpression> toTypedExpressionRec(Expression drlxExpr) {
         if (drlxExpr instanceof PointFreeExpr pointFreeExpr) {
             Optional<TypedExpression> optLeft = toTypedExpressionRec(pointFreeExpr.getLeft());
             Optional<TypedExpression> optRight = pointFreeExpr.getRight().size() == 1 ? toTypedExpressionRec(pointFreeExpr.getRight().get( 0 )) : Optional.empty();
-            OperatorSpec opSpec = getOperatorSpec(pointFreeExpr.getRight(), pointFreeExpr.getOperator());
+            OperatorSpec opSpec = getOperatorSpec(pointFreeExpr.getRight(), pointFreeExpr.getFullOperatorString());
```
tkobayas: OperatorSpec is taken by fullOperatorString as a key (e.g. `str[startsWith`), because built-in operators are registered with the fullOperatorString.

Note: `str[xxx]` operators are managed as built-in operators (`NativeOperatorSpec`), not as custom operators (`CustomOperatorSpec`)

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/operatorspec/CustomOperatorWrapper.java
```diff
@@ -35,7 +35,8 @@ public CustomOperatorWrapper( Evaluator evaluator, String name ) {
 
     @Override
     public boolean eval( Object o1, Object o2 ) {
-        return evaluator.evaluate(null, null, dummyFactHandleOf(o2), null, dummyFactHandleOf(o1));
+        // o1 and o2 are the left and right operands of the expression, so this order. (See Evaluator.evaluate)
+        return evaluator.evaluate(null, null, dummyFactHandleOf(o1), null, dummyFactHandleOf(o2));
```
tkobayas: This is an important change. I had to fix this order to pass both exec-model and non-exec-model custom operator tests.

The tricky part is that we often use the words `left` and `right` for "left input" and "right input" to a JoinNode, where "left input" becomes "right operand" of an expression and "right input" becomes "left operand" of the expression. However, in this case, `o1` and `o2` are "left operand" and "right operand" of the expression. Then, the method `Evaluator.evaluate(ValueResolver, ReadAccessor, FactHandle, ReadAccessor, FactHandle)` expects such "operand" order. Other impl classes follow the order (e.g. `ObjectContainsEvaluator`). So I changed this `CustomOperatorWrapper` accordingly. Now the javadoc for the `Evaluator.evaluate` method looks inconsistent (explains left/right "input"), so I changed the javadoc, too (changed to left/right "operand"). I fixed the unit tests' custom operators in the same way.

@mariofusco Does this change make sense? Please share your thoughts. Thanks!

drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/ast/expr/PointFreeExpr.java
mariofusco: Ok, now I see why you introduced the `subOperator`, but I'm still doubtful if this is necessary and we want it. What I mean is that the syntax op[subOp] has always been very weird to me and we kept there only for legacy/historical reasons. In other words there's no reason why we have `str[startsWith]` and `str[endsWith]` instead of `stringStartsWith` and `stringEndsWith`, they are 2 distinct operators anyway and doesn't have anything in common. What I mean is that I would prefer to avoid the additional `subOperator` and in this case just having operator=`str[startsWith]`. Anyway I'm not sure if this is feasible or if I'm missing some impilcations, so feel free to ignore this comment if this is not possible.

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/operatorspec/CustomOperatorWrapper.java
mariofusco: Yes it makes sense. Maybe just change the variable names from `o1` and `o2` to `leftOperand` and `rightOperand`.

drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/ast/expr/PointFreeExpr.java
tkobayas: Thank you, @mariofusco . Your point makes a lot of sense. I'll follow the approach (so this PR will become simpler) and will raise a docs PR to explain that.

tkobayas:
PR updated:
- `CustomOperatorOnlyDrlTest` is removed as a custom operator with brackets is not supported.

FYI)
Filed a document PR to explain Custom Operators : https://github.com/apache/incubator-kie-docs/pull/4565

mariofusco (approved):

yesamer (approved):

drools-model/drools-mvel-parser/src/main/java/org/drools/mvel/parser/ast/expr/PointFreeExpr.java
porcelli: @mariofusco I hear you, but this is an interesting use case that I've seen:

```
$listOfBigIntegerType contains 123 //expected, but doesn't work as 123 needs to be BigDecimal
$listOfBigDecimal contains new BigDecimal(123) // works, but not much legible
$listOfBigDecimal has[bigdec] 123 //expected use of pluggable operators
```

if not having the parameter... 

```
$listOfBigDecimal hasBigDec 123 //no parameters
```

I see that users could live with hasBigDec, but I also see that has[bigdec] with has[other_type] feels more legible.

tkobayas: Thank you for the input, @porcelli . IMHO, `has[bigdec]` may be slightly better than `hasBigDec`, but it isn’t worth introducing that extra complexity into the signature...

Let me merge this PR for now and we may revisit the need of brackets in an operator.

PR #6371, opened 2025-06-09
[incubator-kie-issues#1997] Implementing fallback downloads for DMN XSDs.

baldimir:
Issue: https://github.com/apache/incubator-kie-issues/issues/1997

This is an alternative approach of setting fallback URLs for DMN XSDs. Original approach is here: https://github.com/apache/incubator-kie-drools/pull/6370. 

By setting failOnError for the original downloads, if some download doesn't work, it doesn't fail the build. With overwrite set to false for the fallback ones, if the original download works, the fallback is skipped.

### DIFF
--- a/kie-dmn/kie-dmn-xsd-resources/pom.xml
+++ b/kie-dmn/kie-dmn-xsd-resources/pom.xml
@@ -62,6 +62,7 @@
               <unpack>false</unpack>
               <outputDirectory>${output.spec.directory}/20151101/</outputDirectory>
               <md5>ff4a266fa7a91b51b5c7ffde0e19aa5f</md5>
+              <failOnError>false</failOnError>
             </configuration>
           </execution>
           <!-- DMN 1.2 -->
@@ -76,6 +77,7 @@
               <unpack>false</unpack>
               <outputDirectory>${output.spec.directory}/20180521/</outputDirectory>
               <md5>89afb8f008550f81cc3989c5d0b9e4ea</md5>
+              <failOnError>false</failOnError>
             </configuration>
           </execution>
           <execution>
@@ -89,6 +91,7 @@
               <unpack>false</unpack>
               <outputDirectory>${output.spec.directory}/20180521/</outputDirectory>
               <md5>635f0e1e62ba09c22ca302c6103738ba</md5>
+              <failOnError>false</failOnError>
             </configuration>
           </execution>
           <execution>
@@ -102,6 +105,7 @@
               <unpack>false</unpack>
               <outputDirectory>${output.spec.directory}/20180521/</outputDirectory>
               <md5>f2fdcef4f5a1a32473e0ddca2dd72592</md5>
+              <failOnError>false</failOnError>
             </configuration>
           </execution>
           <execution>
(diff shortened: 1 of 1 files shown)


gitgabrio:
I personally think this configuration is  contrived/cumbersome to maintain, being a big workaround of the limitations of the used plugin.
I personally fail to see concrete motivation (i.e. not abstract/ideological) against the implementation/fix of specific code/plugin, as done in original [PR](https://github.com/apache/incubator-kie-drools/pull/6370)
That PR may be modified to make a smarter usage of programmatic maven invocation, eventually, and that would make it even slimmer.
Anyway, being involved, I won't give any approval/reject on this PR

baldimir:
> I personally think this configuration is contrived/cumbersome to maintain, being a big workaround of the limitations of the used plugin. I personally fail to see concrete motivation (i.e. not abstract/ideological) against the implementation/fix of specific code/plugin, as done in original [PR](https://github.com/apache/incubator-kie-drools/pull/6370) That PR may be modified to make a smarter usage of programmatic maven invocation, eventually, and that would make it even slimmer. Anyway, being involved, I won't give any approval/reject on this PR

I personally see the custom Maven plugin in the original PR as not necessary as the original approach can be used to also specify fallback downloads without any custom code. The configuration stays in the pom.xml and we don't need to maintain the custom Maven plugin that is used only for this one specific thing.

baldimir:
There may be also licensing problem with the copy-pasted code from the original plugin. I am not sure what are the rules for such cases in Apache, so I personally would rather avoid copying code from other projects. Especially when the problem is solvable just by extending the configuration in the pom.xml.

gitgabrio:
As I wrote, the implementation of the plugin could be changed - I think it would be cleaner to programmatically invoke the original one, and that would also remove the copy'n'paste issue.
From my POV, the problem is not if, with some kind of configuration, it is possible to have a plugin working as we need, but how contrived this configuration is, compared to cost of build/maintain custom one. And, using invocation programmatically, code would be just a bunch of lines.
But, again, it seems to me more an "abstract" objection then a concrete pros/cons evaluation

tiagobento:
How much do we trust the `omg-dmn-taskforce` organization on GitHub? If we trust it as much as `https://omg.org` or more, I think switching to it by default could produce more reliable results, as I think `github.com` is much more prepared to be used as a "build resources registry" than `https://omg.org`, no? 

Maybe the middle-ground here between your proposed solution @baldimir and @gitgabrio's suggestions is to change the source of download so we evaluate how it behaves over a few months.

baldimir:
@tiagobento omg-dmn-taskforce is GitHub group for the OMG DMN RTF group. I personally think the primary source for download should be the original source and that is the OMG links we already have. GitHub should be just a fallback.  

To any solution, I am ok with any solution. I am just trying us to not end with an overengineered one.

baldimir:
As we cannot agree on the solution, I posted this to the mailing list, so we get feedback.

yesamer (approved):

tkobayas (approved):
Thanks!

porcelli (approved):

PR #6389, opened 2025-07-02
[kie-issues#2017] `range()` function shouldn't allow descendant range endpoints

### LINKED ISSUE #2017: `range()` function shouldn't allow descendant range endpoints
According to the specs:

> Ranges
>   FEEL supports a compact syntax for a range of values, useful in decision table test cells and elsewhere. Ranges can
>   be syntactically represented either:
>   a) as a comparison operator and a single endpoint (grammar rule 7.a.)
>   b) or a pair of endpoints and endpoint inclusivity flags that indicate whether one or both endpoints are
>   included in the range (grammar rule 7.b.); on this case, endpoints must be of equivalent types (see section
>   10.3.2.9.1for the definition of type equivalence) **and the endpoints must be ordered such that range start
>   <= range end**

Ranges endpoints should be ascendant (i.e range start <= range end)

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/2017

According to the specs:

> Ranges
>   FEEL supports a compact syntax for a range of values, useful in decision table test cells and elsewhere. Ranges can
>   be syntactically represented either:
>   a) as a comparison operator and a single endpoint (grammar rule 7.a.)
>   b) or a pair of endpoints and endpoint inclusivity flags that indicate whether one or both endpoints are
>   included in the range (grammar rule 7.b.); on this case, endpoints must be of equivalent types (see section
>   10.3.2.9.1for the definition of type equivalence) and **the endpoints must be ordered such that range start
>   <= range end**

Ranges endpoints should be ascendant (i.e range start <= range end)

TCK Tests status after applying the change.
`[ERROR] Tests run: 6765, Failures: 0, Errors: 15, Skipped: 0`

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
@@ -189,7 +189,7 @@ public class RangeFunction extends BaseFEELFunction {
     /**
      * @param leftObject
      * @param rightObject
-     * @return It checks if the leftObject is a greater than rightObject, false otherwise. If one of the endpoints is null,
+     * @return It checks if the leftObject is lower or equals to rightObject, false otherwise. If one of the endpoints is null,
      * the endpoint range is considered to be in ascending order.
      */
     @SuppressWarnings("unchecked")
--- a/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
+++ b/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
@@ -137,7 +137,7 @@ public class RangeFunction extends BaseFEELFunction {
             return FEELFnResult.ofError(new InvalidParametersEvent(FEELEvent.Severity.ERROR, "from", "endpoints must be of equivalent types"));
         }
 
-        if (!nodesValueRangeAreAscending(left, right)) {
+        if (!nodesValuesRangeAreAscending(left, right)) {
             return FEELFnResult.ofError(new InvalidParametersEvent(FEELEvent.Severity.ERROR, "from", "range endpoints must be in ascending order"));
         }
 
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
```diff
@@ -181,6 +186,18 @@ protected boolean nodesReturnsSameType(Object leftObject, Object rightObject) {
...
     }
 
+    /**
+     * @param leftObject
+     * @param rightObject
+     * @return It checks if the leftObject is lower or equals to rightObject, false otherwise. If one of the endpoints is null,
+     * the endpoint range is considered to be in ascending order.
+     */
+    @SuppressWarnings("unchecked")
+    protected boolean nodesValueRangeAreAscending(Object leftObject, Object rightObject) {
```
gitgabrio: Thanks @yesamer 
Could you please check/consider also `NullNode` and what should be the behavior in that case ?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/RangeFunctionTest.java
```diff
@@ -329,6 +329,59 @@ void nodesReturnsSameType_False() {
                 .isFalse();
     }
 
+    @Test
```
gitgabrio, line 332: Please look and consider also `NullNode` in tests, thanks!

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/RangeFunction.java
yesamer: @gitgabrio The null check in place doesn't cover that case?

gitgabrio: Nope, IINW. `NullNode` are concrete instances, not `null`. And I'm not sure/remember of what would be the behavior when one, or both, of the objects are `NullNode` instances

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/runtime/functions/RangeFunctionTest.java
```diff
@@ -356,6 +356,26 @@ void nodesEndpointsAscendant_True() {
         assertThat(rangeFunction.nodesValueRangeAreAscending(left, right))
                 .withFailMessage("P2DT20H14M - P2DT20H15M")
                 .isTrue();
+        left = new NullNode("null");
```
gitgabrio: @yesamer @baldimir 
here's the issue I wanted to highlight:
IIRC, UndefinedValueComparable are semantically valid value for one - or both - of the range ends.
Am I misreading/misunderstanding something here ?

yesamer: @gitgabrio It's the opposite, if there is at least one null endpoint, that method returns true. That method's scope is to return false when the two endpoints defines a descendant range, and with at least one null node that is NOT the case.

gitgabrio (approved):

baldimir (approved):

PR #6395, opened 2025-07-09
[kie-issues#1841] Date and time related functions as non literal not permitted

### LINKED ISSUE #1841: Date and time related functions as non literal not permitted
"compliance-level-3/1156-range-function","1156-range-function-test-01","007_b","ERROR","FAILURE: 'decision007_b' expected='null' but found='[ 1970-01-01 .. 1970-01-02 ]'"
"compliance-level-3/1156-range-function","1156-range-function-test-01","008_b","ERROR","FAILURE: 'decision008_b' expected='null' but found='[ 1970-01-01T00:00 .. 1970-01-02T00:00 ]'"
"compliance-level-3/1156-range-function","1156-range-function-test-01","009_b","ERROR","FAILURE: 'decision009_b' expected='null' but found='[ 00:00 .. 00:00 ]'"
"compliance-level-3/1156-range-function","1156-range-function-test-01","010_b","ERROR","FAILURE: 'decision010_b' expected='null' but found='[ PT24H .. PT48H ]'"

Core issue is that the following expression

date(string("1970-01-1")) 
seems invalid, because it uses a function (StringFunction) as parameter for Date function (the specs does not include it as valid parameter).
On the other side, our engine recursively evaluate all parameters, innermost to outermost), so the stringfunction is evaluated to string before invoking the date function.

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1841

According to the DMN 1.6 Specs:

Screenshot 2025-07-09 at 13 46 44

In this PR:
- In case of Endpoints of FunctionInvocation type, only the functions type are allowed
- The above function type are allowed as literal only (e.g they can't have functions as parameters)
- Updated Tests accordingly
- Allowed node type, values and function are now stored as Set, that is a better fit for this use case.


gitgabrio (approved):
Thx - LGTM!

baldimir (approved):

PR #6408, opened 2025-07-29
[kie-issues#1831] Return coerced null from decision with knowledge requirements

### LINKED ISSUE #1831: DMN - TCK failure: Return coerced null from decision with knowledge requirements
TCK Failures

"compliance-level-3/0082-feel-coercion","0082-feel-coercion-test-01","decision_bkm_002","ERROR","FAILURE: 'decision_bkm_002' expected='null' but found='false'"
"compliance-level-3/0082-feel-coercion","0082-feel-coercion-test-01","invoke_001","ERROR","FAILURE: 'invoke_001' expected='null' but found='false'"
"compliance-level-3/0082-feel-coercion","0082-feel-coercion-test-01","invoke_005","ERROR","FAILURE: 'invoke_005' expected='null' but found='10'"

Model https://github.com/dmn-tck/tck/blob/master/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
Test https://github.com/dmn-tck/tck/blob/master/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion-test-01.xml

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1831

**Case 1: `decision_bkm_002` and `invoke_001 tests`**
Not found a clear statement in the specs, but according to this [DMN TCK PR that modified those two tests ](https://github.com/dmn-tck/tck/pull/495):

`if types don't match each other, the function is not invoked.`

This leads to the 2 line changes applied in `DMNFunctionDefinitionEvaluator.java`. If the types doesn't match each other, an error is thrown and the function is not invoked - null is returned.

**Case 2: `invoke_005`test**
In this case, an array with single element (namely `[10]`) is passed to a BKM node. The BKM takes that input and put it inside another array. Then, we have [[10]] as a processed result. This should fail, because the BKM node should return a numer and not an array. The problem here is that, in our code, the coercion of an array of single element to a plain element is applied twice, so [[10]] is coerced to 10 that is indeed not correct.

First coercion is applied [here](https://github.com/apache/incubator-kie-drools/pull/6408/files#diff-9a500edade4dcb8870c963537818452a826d55dd2802cc34bc353c6395c7dc59R451) and the second is applied inside the [isAssignableValue()](https://github.com/apache/incubator-kie-drools/pull/6408/files#diff-9a500edade4dcb8870c963537818452a826d55dd2802cc34bc353c6395c7dc59R458) method. Passing the original value instead of the already coerced element makes the trick. 

Merging this will solve 3 TCK failures - 8 are still failing

Note: Unit Testing those changes requires a relevant effort, the involved classes requires a lot of mocking to be tested, and that means they are not in the best possible shape to be properly tested - i.e I guess we need to refactor them to properly test them.


gitgabrio (approved):

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
```diff
@@ -455,7 +455,7 @@ public static Object coerceUsingType(Object value, DMNType type, boolean typeChe
         if (!typeCheck) {
             return result;
         }
-        if (type.isAssignableValue(result)) {
+        if (type.isAssignableValue(value)) {
```
baldimir: I think this whole method looks a bit cumbersome (not because of the change, it just builds on top of what is already there). It returns on various places and it is not visible on first sight what it is doing. Maybe we could replace it with something like this: 

```
public static Object coerceUsingType(Object value, DMNType type, boolean typeCheck, BiConsumer<Object, DMNType> nullCallback) {
        if (typeCheck) {
            if (type.isAssignableValue(value)) {
                return coerceSingleItemCollectionToValue(value, type);
            } else {
                nullCallback.accept(value, type);
                return null;
            }
        } else {
            return coerceSingleItemCollectionToValue(value, type);
        }
    }

    /**
     * Checks a type and if it is not a collection type, checks if the specified value is a collection 
     * that contains only a single value and if yes, coerces the collection to the single item itself.  
     * E.g. [1] becomes 1. Basically it unwraps the single item from a collection, if it is required. 
     * 
     * @param value Value that is checked and potentially coerced to a single item.  
     * @param type Required type. Based on this type, it is determined, if the coercion happens. 
     *             If the requirement is for a non-collection type and the value is a single item collection, 
     *             the coercion happens. 
     * @return If all requirements are met, returns coerced value. Otherwise returns the original value. 
     */
    private static Object coerceSingleItemCollectionToValue(Object value, DMNType type) {
        if (!type.isCollection() && value instanceof Collection && ((Collection<?>) value).size() == 1) {
            // as per Decision evaluation result.
            return ((Collection<?>) value).toArray()[0];
        } else {
            return value;
        }
    }
```
What do you think, please?

baldimir (approved):
Thanks, approving. I made one suggestion, however it is not a blocker. It can be merged also as it is. Just an idea for further improvement.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNRuntimeImpl.java
yesamer: @baldimir +1, changes applied, thanks!

PR #6416, opened 2025-08-12
[incubator-kie-issue#2052]Bug fix - Decision Services doesn't work with unnamed imported Elements

### LINKED ISSUE #2052: DMN Decision Services doesn't work with unnamed imported Elements
It seems a Decision Service doesn't correctly manage an imported element from an unnamed model.

Image

In this scenario, an error is thrown:

`IMPORT_NOT_FOUND: Required import not found: {[https://kie.org/dmn/_420326F7-3ECC-40FA-9E5B-64186F028144}DMN_928E1EA0-C443-4A77-BAB7-1850538DF803](https://kie.org/dmn/_420326F7-3ECC-40FA-9E5B-64186F028144%7DDMN_928E1EA0-C443-4A77-BAB7-1850538DF803) for node 'org.kie.dmn.model.v1_5.TInputData@1fd95e25'; missing DMN Import name alias.`

After a quick check, it seems the current code doesn't handle the case of an element imported by an unnamed model

https://github.com/apache/incubator-kie-drools/blob/main/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java#L106

[decision-service.zip](https://github.com/user-attachments/files/21598212/decision-service.zip)

ChinchuAjith:
Fixes https://github.com/apache/incubator-kie-issues/issues/2052

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
@@ -111,7 +111,7 @@ public class DecisionServiceCompiler implements DRGElementCompiler {
         } else {
             try {
                 return getInputNamePrefix(input, model);
-            } catch(IllegalStateException e) {
+            } catch (IllegalStateException e) {
                 MsgUtil.reportMessage(LOG,
                                       DMNMessage.Severity.ERROR,
                                       ((DMNBaseNode)input).getSource(),
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
@@ -108,19 +108,7 @@ public class DecisionServiceCompiler implements DRGElementCompiler {
             return null;
         } else {
             Optional<String> importAlias = model.getImportAliasFor(input.getModelNamespace(), input.getModelName());
-            if (importAlias.isEmpty()) {
-                MsgUtil.reportMessage(LOG,
-                                      DMNMessage.Severity.ERROR,
-                                      ((DMNBaseNode)input).getSource(),
-                                      model,
-                                      null,
-                                      null,
-                                      Msg.IMPORT_NOT_FOUND_FOR_NODE_MISSING_ALIAS,
-                                      new QName(input.getModelNamespace(), input.getModelName()),
-                                      ((DMNBaseNode)input).getSource());
-                return null;
-            }
-            return importAlias.get();
+            return importAlias.orElse(null);
         }
     }
 
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
@@ -110,7 +110,7 @@ public class DecisionServiceCompiler implements DRGElementCompiler {
             return null;
         } else {
             try {
-                return resolveImportAlias(input, model);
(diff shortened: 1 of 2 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
```diff
@@ -103,12 +105,15 @@ public boolean accept(DMNNode node) {
      * The qualified name of an element named E that is defined in the same decision model as S is simply E.
      * Otherwise, the qualified name is I.E, where I is the name of the import element that refers to the model where E is defined.
      */
-    private static String inputQualifiedNamePrefix(DMNNode input, DMNModelImpl model) {
+     static String inputQualifiedNamePrefix(DMNNode input, DMNModelImpl model) {
         if (input.getModelNamespace().equals(model.getNamespace())) {
             return null;
+        } else if (isInUnnamedImport(input, model)) {
+            return input.getName();
```
yesamer: Shouldn't this method return the Import prefix (as the method names states)?. Can you please clarify why in this case the input name is returned, while in other cases the prefix is returned?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
```diff
@@ -120,8 +125,16 @@ private static String inputQualifiedNamePrefix(DMNNode input, DMNModelImpl model
                                       ((DMNBaseNode)input).getSource());
                 return null;
             }
-            return importAlias.get();
         }
+     }
+
+    private static String resolveImportAlias(DMNNode input, DMNModelImpl model) {
```
yesamer: Please be consistent: We refer to the prefix as:
- Name Prefix
- Import Alias

Please use the same name for that.

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.kie.dmn.model.api.Definitions;
+import org.kie.dmn.model.api.Import;
+import org.kie.dmn.model.api.NamedElement;
+import org.mockito.Mockito;
+
+import java.util.List;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
```
yesamer: Please do not use package imports

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
ChinchuAjith: Corrected this. For unnamed imports, since no alias is defined, no prefix should be used when referencing the imported element. So updated the logic to return null in that case.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
ChinchuAjith: Done

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
ChinchuAjith: Corrected

yesamer (approved):
Thank you @ChinchuAjith, copilot's suggestion could make sense, even if minors,  please address them.

gitgabrio (approved):

yesamer:
Closed as I accepted a Suggestion by Copilot by mistake

PR #6419, opened 2025-08-14
[incubator-kie-drools-6410] Multiple rules in an activation group can…

tkobayas:
- https://github.com/apache/incubator-kie-drools/issues/6410

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -101,6 +101,7 @@ public class PhreakRuleTerminalNode {
             return;
         }
 
+        // most recent PropagationContext is required to maintain the right recency which triggers the match
         PropagationContext pctx = leftTuple.findMostRecentPropagationContext();
         if ( rtnNode.getRule().isNoLoop() && sameRules(rtnNode, pctx.getTerminalNodeOrigin()) ) {
             return;
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -101,14 +101,9 @@ public class PhreakRuleTerminalNode {
             return;
         }
 
-        PropagationContext pctx;
-        if ( rtnNode.getRule().isNoLoop() ) {
-            pctx = leftTuple.findMostRecentPropagationContext();
-            if ( sameRules(rtnNode, pctx.getTerminalNodeOrigin()) ) {
-                return;
-            }
-        } else {
-            pctx = leftTuple.getPropagationContext();
+        PropagationContext pctx = leftTuple.findMostRecentPropagationContext();
+        if ( rtnNode.getRule().isNoLoop() && sameRules(rtnNode, pctx.getTerminalNodeOrigin()) ) {
+            return;
         }
 
         int salienceInt = getSalienceValue(rtnNode, ruleAgendaItem, leftTuple, reteEvaluator);
@@ -117,8 +112,8 @@ public class PhreakRuleTerminalNode {
 
         activationsManager.getAgendaEventSupport().fireActivationCreated(leftTuple, activationsManager.getReteEvaluator());
 
-        if ( rtnNode.getRule().isLockOnActive() && pctx.getType() != PropagationContext.Type.RULE_ADDITION ) {
-            pctx = leftTuple.findMostRecentPropagationContext();
+        if (  rtnNode.getRule().isLockOnActive() &&
+                leftTuple.getPropagationContext().getType() != PropagationContext.Type.RULE_ADDITION ) {
             InternalAgendaGroup agendaGroup = executor.getRuleAgendaItem().getAgendaGroup();
             if (blockedByLockOnActive(rtnNode.getRule(), pctx, agendaGroup)) {
                 activationsManager.getAgendaEventSupport().fireActivationCancelled(leftTuple, reteEvaluator, MatchCancelledCause.FILTER );
--- a/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
+++ b/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
@@ -1,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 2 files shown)


tkobayas:
https://github.com/apache/incubator-kie-drools/commit/3de05e07086ae761fb742f8bb0ebd95f38559650 contains several changes. With regard to the change from `findMostRecentPropagationContext` to `getPropagationContext`, 3 parts were changed.

- `PhreakRuleTerminalNode.doLeftTupleInsert` : This is the cause of the issue illustrated by `ActivationGroupDualFactMatchTest.testRuleAMatchingTwoFactsButNeverFires`. `findMostRecentPropagationContext` is required to maintain the right recency which triggers the match. So reverted in this PR.
- `PhreakRuleTerminalNode.doLeftTupleUpdate` : I was not able to write a test case which causes a problem. As long as `PhreakRuleTerminalNode.doLeftTupleInsert` is fixed, the tuple's context's recency is valid even if the tuple is updated, I think. So not reverted in this PR.
- `QueryElementNode.UnificationNodeViewChangedEventListener.rowAdded` : In this case, the context is only used as `context.getReaderContext()`. ReaderContext is shared across contexts in the same session, so replacing from `findMostRecentPropagationContext` to `getPropagationContext` is fine. Not reverted in this PR.

@mariofusco  Could you share your thoughts on this?

drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
```diff
@@ -117,8 +112,8 @@ public static void doLeftTupleInsert(TerminalNode rtnNode, RuleExecutor executor
 
         activationsManager.getAgendaEventSupport().fireActivationCreated(leftTuple, activationsManager.getReteEvaluator());
 
-        if ( rtnNode.getRule().isLockOnActive() && pctx.getType() != PropagationContext.Type.RULE_ADDITION ) {
-            pctx = leftTuple.findMostRecentPropagationContext();
+        if (  rtnNode.getRule().isLockOnActive() &&
+                leftTuple.getPropagationContext().getType() != PropagationContext.Type.RULE_ADDITION ) {
```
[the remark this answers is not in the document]
tkobayas: This is an intentional change to revert https://github.com/apache/incubator-kie-drools/commit/3de05e07086ae761fb742f8bb0ebd95f38559650 . For `getType()`, "getPropagationContext()" (not most recent one) is consistently used across commits.

mariofusco (approved):

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
```diff
@@ -0,0 +1,205 @@
...
+        System.out.println(">>> Total rules fired: " + fired);
+
+        // === PHASE 5: Validation - Verify Expected Behavior ===
+
+        final boolean ruleAFired = firedRules.contains("Rule A - GroupA");
+        final long ruleBFires = firedRules.stream().filter(r -> r.equals("Rule B - GroupB")).count();
+        final long setupFires = firedRules.stream().filter(r -> r.equals("Setup Rule - GroupA")).count();
+
+        // Validate the expected execution pattern
+        assertEquals(3, setupFires, "Setup rule should fire 3 times (once per dynamic fact).");
```
pibizza: Please use assertThat from assertJ instead of assertEquals and assertFalse

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
```diff
@@ -0,0 +1,205 @@
...
+        System.out.println(">>> Firing all rules...");
+
+        // Execute all rules - this is where the complex interaction happens:
+        // 1. Setup Rule fires for each dynamic fact (3 times)
+        // 2. Each Setup Rule execution modifies a dynamic fact and switches to GroupB
+        // 3. Rule B fires in GroupB for each modified dynamic fact (3 times)
+        // 4. Rule A never fires despite having matching conditions (activation group prevents it)
+        final int fired = kSession.fireAllRules();
+
+        System.out.println(">>> Total rules fired: " + fired);
```
pibizza: I suggest not using println in tests. they are useful for debugging, but they pollute the execution logs with tons of confusing output. if you want to have some traces maybe log.debug is a better choice?

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
```diff
@@ -0,0 +1,205 @@
...
+            // Each dynamic fact starts with flag=false, matching Setup Rule conditions
+            kSession.insert(new FactWrapper("dynamic-%d".formatted(i)));
+        }
+
+        // === PHASE 4: Rule Execution - The Core Test Logic ===
+
+        // Set initial agenda focus to GroupA where Setup Rule and Rule A reside
+        kSession.getAgenda().getAgendaGroup("GroupA").setFocus();
+
+        System.out.println(">>> Firing all rules...");
```
pibizza: See below about using println

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
```diff
@@ -0,0 +1,205 @@
...
+            import %s.FactWrapper;
+            
+            global org.kie.api.runtime.KieSession ksession;
+            
+            rule "Setup Rule - GroupA"
+                agenda-group "GroupA"
+                when
+                    $d: FactWrapper(id matches "dynamic-\\\\d+", flag == false)
+                then
+                    System.out.println(">>> Fired Setup Rule with fact: " + $d);
```
pibizza: Using a println inside a consequence makes it harder to test if the rule was correctly fired or not. maybe you can remove it and use instead a simple listener to check which rule has fired?

pibizza (changes requested):
Some changes - relatively minor - to make the test a bit clearer and easier to read.  Thanks!

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
tkobayas: @pibizza Thanks! Fixed.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
tkobayas: @pibizza Thanks! Fixed.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
tkobayas: @pibizza Thanks! Fixed.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java
tkobayas: @pibizza Thanks! Fixed. Yeah, the test uses a listener to assert the fired rules.

tkobayas:
@pibizza Applied the suggestion. Please review again, thanks!

pibizza (approved):

PR #6429, opened 2025-08-20
[incubator-kie-issues#2069]Resolve ArchUnit Stacktrace Flooding in Drools Build on Semeru JDK 21

### LINKED ISSUE #2069: Resolve ArchUnit Stacktrace Flooding in Drools Build on Semeru JDK 21
The build of drools project floods output logs with similar stacktraces coming from following maven goal : arch-unit:2.9.1:arch-test (default) 
java.lang.IllegalArgumentException: Unsupported class file major version 65
    at com.tngtech.archunit.thirdparty.org.objectweb.asm.ClassReader.<init> (ClassReader.java:199)

ChinchuAjith:
Fixes https://github.com/apache/incubator-kie-issues/issues/2069

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -221,7 +221,9 @@
     <version.download-maven-plugin>2.0.0</version.download-maven-plugin>
     <version.property-maven-plugin>1.2.1</version.property-maven-plugin>
     <!-- plugin used to enforce architectural constraints -->
-    <archunit.maven.plugin.version>4.0.2</archunit.maven.plugin.version>
+    <!--Added JDK-specific ArchUnit plugin profiles for Java 17 and Java 21.
+    Commented out the global plugin declaration to prevent version mismatch.-->
+    <!--<archunit.maven.plugin.version>3.0.1</archunit.maven.plugin.version>-->
     <version.archunit.junit5>1.4.0</version.archunit.junit5>
   </properties>
 
@@ -1900,11 +1902,6 @@
           <version>${version.com.github.eirslett}</version>
         </plugin>
         <plugin>
-          <groupId>com.societegenerale.commons</groupId>
-          <artifactId>arch-unit-maven-plugin</artifactId>
-          <version>${archunit.maven.plugin.version}</version>
-        </plugin>
-        <plugin>
           <groupId>io.github.download-maven-plugin</groupId>
           <artifactId>download-maven-plugin</artifactId>
           <version>${version.download-maven-plugin}</version>
@@ -2397,6 +2394,54 @@
                 <phase>compile</phase>
                 <goals>
                   <goal>unpack</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
+      <id>archunit-java17</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+        <jdk>17</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.societegenerale.commons</groupId>
+            <artifactId>arch-unit-maven-plugin</artifactId>
+            <version>2.9.1</version> <!-- Compatible with Java 17 -->
+            <executions>
+              <execution>
+                <goals>
+                  <goal>arch-test</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <!-- Profile for Java 21 -->
+    <profile>
+      <id>archunit-java21</id>
+      <activation>
+        <jdk>21</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.societegenerale.commons</groupId>
+            <artifactId>arch-unit-maven-plugin</artifactId>
+            <version>4.0.2</version> <!-- Compatible with Java 21 -->
+            <executions>
+              <execution>
+                <goals>
+                  <goal>arch-test</goal>
                 </goals>
               </execution>
             </executions>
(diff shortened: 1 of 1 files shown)


gitgabrio (approved):
Thanks @ChinchuAjith !

build-parent/pom.xml
```diff
@@ -221,7 +221,9 @@
     <version.download-maven-plugin>2.0.0</version.download-maven-plugin>
     <version.property-maven-plugin>1.2.1</version.property-maven-plugin>
     <!-- plugin used to enforce architectural constraints -->
-    <archunit.maven.plugin.version>2.9.1</archunit.maven.plugin.version>
+    <!--Added JDK-specific ArchUnit plugin profiles for Java 17 and Java 21.
+    Commented out the global plugin declaration to prevent version mismatch.-->
+    <!--<archunit.maven.plugin.version>3.0.1</archunit.maven.plugin.version>-->
```
yesamer: @ChinchuAjith Please restore the version here, the one compatible with JDK 17 (3.0.1?)

build-parent/pom.xml
```diff
@@ -1899,11 +1901,6 @@
           <artifactId>frontend-maven-plugin</artifactId>
           <version>${version.com.github.eirslett}</version>
         </plugin>
-        <plugin>
-          <groupId>com.societegenerale.commons</groupId>
-          <artifactId>arch-unit-maven-plugin</artifactId>
-          <version>${archunit.maven.plugin.version}</version>
-        </plugin>
         <plugin>
```
yesamer, line 1902: @ChinchuAjith Please restore this

build-parent/pom.xml
```diff
@@ -2404,6 +2401,54 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
```
yesamer, line 2415: @ChinchuAjith Please remove the jdk 17 profile

build-parent/pom.xml
```diff
@@ -2404,6 +2401,54 @@
...
+    <!-- Profile for Java 21 -->
+    <profile>
+      <id>archunit-java21</id>
+      <activation>
+        <jdk>21</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.societegenerale.commons</groupId>
```
yesamer: @ChinchuAjith You don't need to declare the plugin, it's enough to override the version property:



```
<id>jdk-21</id>
        <activation>
            <jdk>[21,)</jdk>
        </activation>
        <properties>
            <archunit.maven.plugin.version>4.0.2</archunit.maven.plugin.version>
        </properties>
```

yesamer (commented):
@ChinchuAjith Thank you, I guess this can be simplified. The idea is that JDK 17 should be considered as the default JDK

build-parent/pom.xml
baldimir: We shouldn't have this commented out here. Please delete or use as a default version. I think here it could be some default version and later in the profile there could be an override for a specific Java version.

ChinchuAjith: Done

build-parent/pom.xml
ChinchuAjith: Done

build-parent/pom.xml
ChinchuAjith: keeping jdk 21 specific version as default and added profile for 17

build-parent/pom.xml
ChinchuAjith: Corrected

build-parent/pom.xml
```diff
@@ -221,7 +221,7 @@
     <version.download-maven-plugin>2.0.0</version.download-maven-plugin>
     <version.property-maven-plugin>1.2.1</version.property-maven-plugin>
     <!-- plugin used to enforce architectural constraints -->
-    <archunit.maven.plugin.version>2.9.1</archunit.maven.plugin.version>
+    <archunit.maven.plugin.version>4.0.2</archunit.maven.plugin.version>
```
yesamer: @ChinchuAjith  Not part of your change, but can you please rename this version using the same convention currently in place for all other dependencies ? eg. `archunit.maven.plugin.version` to `version.archunit-maven-plugin`

build-parent/pom.xml
```diff
@@ -1900,9 +1900,16 @@
...
+        <groupId>com.societegenerale.commons</groupId>
+        <artifactId>arch-unit-maven-plugin</artifactId>
+        <version>${archunit.maven.plugin.version}</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>arch-test</goal>
+            </goals>
+          </execution>
+        </executions>
```
yesamer, line 1912: @ChinchuAjith Why is this executions required?

build-parent/pom.xml
```diff
@@ -2404,6 +2411,18 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>archunit-java17</id>
+      <activation>
+        <jdk>[17,18)</jdk>
```
yesamer: @ChinchuAjith This should be activated to any version in the 17 - 20 range

build-parent/pom.xml
ChinchuAjith: I'm not entirely sure whether we are currently running ArchUnit rules separately via the Maven plugin or just through JUnit tests. However, if we are using the Maven plugin to run architectural rules, then the <executions> block is required. This block binds the arch-test goal to the Maven build lifecycle, ensuring that architectural rules are executed automatically during the build process. 
So, if we’re relying on the Maven plugin (not just JUnit), this configuration is necessary. If not, and we’re only using archunit-junit5 in test scope, we might be able to remove it.

build-parent/pom.xml
ChinchuAjith: Done

build-parent/pom.xml
ChinchuAjith: Changed to [17, 21) Ranging from 17...20

yesamer (approved):
Thank you @ChinchuAjith

PR #6440, opened 2025-09-02
[NO ISSUE] Remove getknowledgebase usages

pibizza:
This is another patch related to cleaning and simplifying drools core.

The strategy is always the same - remove excessive reliance on ReteEvaluator as a mother object.  Where possible, split dependency and provide reference directly to the underlying objects.

This patch focuses mostly removing usages of getKnowledgeBase and passing directly the rule base in the constructor.

All tests pass and PR should be straightforward.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/AgendaGroupsManager.java
+++ b/drools-core/src/main/java/org/drools/core/common/AgendaGroupsManager.java
@@ -256,21 +256,15 @@ public interface AgendaGroupsManager extends Externalizable {
             this.kieBase = kieBase;
             // stacked agenda groups are supported only for InternalWorkingMemory
             this.workingMemory = workingMemory;
-            if (this.mainAgendaGroup == null) {
-                initMainAgendaGroup(kieBase);
-            }
+            this.mainAgendaGroup = agendaGroupFactory.createAgendaGroup( InternalAgendaGroup.MAIN, kieBase);
+			this.agendaGroups.put( InternalAgendaGroup.MAIN, this.mainAgendaGroup );
+			this.focusStack.add( this.mainAgendaGroup );
             this.mainAgendaGroup.setReteEvaluator( workingMemory );
         }
 
         @Override
         public InternalAgendaGroup getMainAgendaGroup() {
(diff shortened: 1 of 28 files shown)


tkobayas (approved):

drools-core/src/main/java/org/drools/core/common/AgendaGroupsManager.java
```diff
@@ -248,15 +247,17 @@ class StackedAgendaGroupsManager implements AgendaGroupsManager {
...
 
-        public StackedAgendaGroupsManager(InternalWorkingMemory workingMemory) {
+        public StackedAgendaGroupsManager(InternalRuleBase kieBase, InternalWorkingMemory workingMemory) {
             this.agendaGroupFactory = RuntimeComponentFactory.get().getAgendaGroupFactory();
+            this.kieBase = kieBase;
             // stacked agenda groups are supported only for InternalWorkingMemory
             this.workingMemory = workingMemory;
             if (this.mainAgendaGroup == null) {
-                initMainAgendaGroup(workingMemory.getKnowledgeBase());
+                initMainAgendaGroup(kieBase);
```
Rikkola: If initMainAgendeGroup always takes in this.kieBase would it be better to not pass it in as a parameter? Just a note.

Rikkola (approved):
Approving this while admitting I am not an expert on the area, but I know where you are trying to get us and the direction is good.

mariofusco (approved):

drools-core/src/main/java/org/drools/core/common/AgendaGroupsManager.java
mariofusco: @pibizza please give a look at @Rikkola's comment and then I will merge this PR.

pibizza: @Rikkola fixed, thanks.

pibizza: @mariofusco this is resolved

PR #6447, opened 2025-09-09
[incubator-kie-drools-6421] Rules in agenda group with auto-focus may…

tkobayas:
- https://github.com/apache/incubator-kie-drools/issues/6421

Firstly, fixing the Insert use case.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/AlphaTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/AlphaTerminalNode.java
@@ -51,7 +51,9 @@ public class AlphaTerminalNode extends LeftInputAdapterNode {
             leftTuple.setPropagationContext( propagationContext );
 
             if ( rtn.getRule().getAutoFocus() && !agendaItem.getAgendaGroup().isActive() ) {
-                activationsManager.getAgendaGroupsManager().setFocus( agendaItem.getAgendaGroup() );
+                if (activationsManager.getAgendaGroupsManager().setFocus(agendaItem.getAgendaGroup())) {
+                    activationsManager.haltGroupEvaluation();
+                }
             }
 
             PhreakRuleTerminalNode.doLeftTupleInsert( rtn, agendaItem.getRuleExecutor(), activationsManager, agendaItem, leftTuple );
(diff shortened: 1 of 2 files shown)


drools-core/src/main/java/org/drools/core/reteoo/AlphaTerminalNode.java
```diff
@@ -51,7 +51,9 @@ public void assertObject( InternalFactHandle factHandle, PropagationContext prop
             leftTuple.setPropagationContext( propagationContext );
 
             if ( rtn.getRule().getAutoFocus() && !agendaItem.getAgendaGroup().isActive() ) {
-                activationsManager.getAgendaGroupsManager().setFocus( agendaItem.getAgendaGroup() );
+                if (activationsManager.getAgendaGroupsManager().setFocus(agendaItem.getAgendaGroup())) {
+                    activationsManager.haltGroupEvaluation();
+                }
```
tkobayas, line 56: If agenda group has changed, group evaluation should be halted.

Following the same approach as `AgendaGroupQueueImpl.SetFocusAction`.
https://github.com/apache/incubator-kie-drools/blob/10.1.0/drools-core/src/main/java/org/drools/core/common/AgendaGroupQueueImpl.java#L142-L146

mariofusco (approved):

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/FiringOrderTest.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            final List<String> firingOrder = new ArrayList<>();
+
+            // Listener to capture firing order for B and C
+            final AgendaEventListener listener = new DefaultAgendaEventListener() {
+                @Override
+                public void beforeMatchFired(BeforeMatchFiredEvent event) {
+                    firingOrder.add(event.getMatch().getRule().getName());
+                }
+            };
```
pibizza, line 117: I think we can make a general test oriented listener for collecting these events and perform these assertions. I will open a ticket for this.

pibizza (approved):
Looks good to me. I suggest we create a series of listeners for testing purposes.

PR #6449, opened 2025-09-10
[incubator-kie-drools-6421] Rules in agenda group with auto-focus may…

tkobayas:
- https://github.com/apache/incubator-kie-drools/issues/6421

Update use case.


drools-core/src/main/java/org/drools/core/reteoo/AlphaTerminalNode.java
```diff
@@ -87,12 +83,23 @@ public void modifyObject(InternalFactHandle factHandle, ModifyPreviousTuples mod
                 if ( context.getModificationMask().intersects( rtn.getLeftInferredMask() ) ) {
                     leftTuple = TupleFactory.createLeftTuple( rtn, factHandle, true );
                     leftTuple.setPropagationContext( context );
+
+                    autoFocusIfNeeded(rtn, agendaItem, activationsManager);
```
tkobayas, line 87: Indeed, it was not implemented in the past versions. I think it's natural to expect that `auto-focus` works in case of update, too.

mariofusco (approved):

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/FiringOrderTest.java
```diff
@@ -100,10 +98,56 @@ public void setAFired(boolean aFired) {
                                       end
                                       """.formatted(FiringOrderTest.class.getName());
 
+    private static final String DRL_UPDATE_FACT = """
+                                      package com.example.drools
+                                      
+                                      import %s.State;
+                                      
+                                      global java.util.List firingOrder;
```
pibizza: This seems useless in the test, am I wrong?

pibizza (commented):
Just left a comment in the test about something I am not fully convinced. If this has a good reason we can move it forward and merge it.

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/FiringOrderTest.java
tkobayas: Thank you for spotting! Fixed.

mariofusco:
@tkobayas Please resolve the conflict and I will merge this.

tkobayas:
@mariofusco Resolved and pushed. Thanks.

PR #6454, opened 2025-09-11
[incubator-kie-drools 6453] Unifying listeners for testing

pibizza:
This provides an implementation for the requirements described in [incubator-kie drools 6453](https://github.com/apache/incubator-kie-drools/issues/6453).

The listener is placed for the moment inside Drools core, since it is not part of the API and has to be visibile in several places in the tests.

Besides that the PR is pretty straightforward.


mariofusco (approved):
Ok for me, but if the `TrackingAgendaEventListener` class is only used for tests, shouldn't it also be in a test folder?

pibizza:
@mariofusco you may be right. I was thinking that this is kind of a generic feature that can be useful outside of tests, but I am ok moving this into test folder. I will push it there tomorrow.

Rikkola (approved):

tkobayas (approved):
Thank you, @pibizza . It's great that you slimmed down those many similar listener usage :)

PR #6455, opened 2025-09-15
[incubator-kie-issues#2099] Fixed BFEEL execution

### LINKED ISSUE #2099: DMN Runner returns wrong values for B-FEEL expressions
using the DMN Runner and these B-FEEL expressions user is getting wrong results.

- day of week("a") : throws an error about non existing function instead of ""
- split("abc", 22) : throws an error 'FEEL_EVALUATION_ERROR: FEEL ERROR while evaluating literal expression 'split("abc", 22)': Unable to find function 'split( lass org.kie.dmn.feel.runtime.functions.SplitFunctio )'' instead of []
- "Today is " + today() : returns null instead of "Today is 202x-xx-xx"
- "The result is: " + 1 : returns null instead of "The result is: 1"
- 5 + " minutes" : returns null instead of "5 minutes"
- "This is " + null : returns null instead of "This is "
- 1 + null : returns null instead of 1
- null - 6 : returns null instead of -6
- date("2021-01-01") + 7 : throws error 'FEEL_EVALUATION_ERROR: FEEL ERROR while evaluating literal expression 'date("2021-01-01") + 7': Based on the specification, the operation is undefined for the specified parameter set.' instead of 7
- "abc" - 2 : returns null instead of ""
- 22 * "a" : returns null instead of 0
- null / 22 : returns null instead of  0
- duration("P1Y") * null : returns null instead of duration("P0M")

For trying those expressions, this model can be used:

[bfeel-tests.dmn.txt](https://github.com/user-attachments/files/22299604/bfeel-tests.dmn.txt)

One of reported cases demonstrated also as screenshot

Image

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/2099

DO-NOT-MERGE

This PR
1. conditionally manage invocation of functions with unexpected parameters
2. adapt specific functions to comply with BFEEL specification

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNInputRuntimeTest.java
+++ b/kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/DMNInputRuntimeTest.java
@@ -67,7 +67,7 @@ public class DMNInputRuntimeTest extends BaseInterpretedVsCompiledTest {
     void sFeelInvalidBFeelValidExpressions(boolean useExecModelCompiler) {
         init(useExecModelCompiler);
         final DMNRuntime runtime = DMNRuntimeUtil
-                .createRuntime("invalid_models/DMNv1_6/B-FEEL/S-FEELInvalidExpressions" + ".dmn", this.getClass());
+                .createRuntime("invalid_models/DMNv1_6/B-FEEL/S-FEELInvalidExpressions.dmn", this.getClass());
         final DMNModel dmnModel = runtime.getModel("https://kie.org/dmn/_71957705-0381-4AEB-9BFD-59059D89A18C",
                 "DMN_E73F88CE-4E67-4B67-BFB2-7977441F2343");
         assertThat(dmnModel).isNotNull();
@@ -78,7 +78,6 @@ public class DMNInputRuntimeTest extends BaseInterpretedVsCompiledTest {
         final DMNResult dmnResult = runtime.evaluateAll(dmnModel, context);
         assertThat(dmnResult.hasErrors()).isFalse();
         List<DMNMessage> warnings = dmnResult.getMessages(DMNMessage.Severity.WARN);
-        System.out.println("warnings: " + warnings.size());
         assertThat(warnings).hasSize(3);
         assertThat(dmnResult.getDecisionResults()).hasSize(14);
         assertThat(dmnResult.getDecisionResultByName("DayOfWeek").getResult()).isEqualTo("");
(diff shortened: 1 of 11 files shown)


gitgabrio:
All three kogito examples build failed for docker registry access error: 

> https://registry-1.docker.io/v2....
> unexpected HTTP status: 500 Internal Server Error

Restarted

jomarko (commented):
Thank you @gitgabrio for this fix, I found two cases that still do not work:

### split("abc", 22)
Does not return empty array [] but:
image


### date("2021-01-01") + 7
Does not return number **7** but date **2021-01-01**

Could you please double check if that is issue in specification we used to implement the feature or is an issue in implementation? Thank you.

gitgabrio:
@jomarko 
As per` date("2021-01-01") + 7`, I could not find the reference of this particular case in the specs, but I refer to this:

image

and, the invalid `7`, is converted to `0`

image

Does this make sense ?
@baldimir  @yesamer

gitgabrio:
@jomarko 
As per `split("abc", 22)`, I've implemented test of it as

1. unit test [here](https://github.com/apache/incubator-kie-drools/pull/6455/files#diff-503dc8564595623bb7a68d9676eb15e2c2ec6966380223bf047374c8f2f6b1ceR86)
2. inside the testing [model](https://github.com/apache/incubator-kie-drools/pull/6455/files#diff-288b26bf7ae432b994c3c8e3119fbafd9585636e06634eed56f032ec3b3dbcefR44)
3. inside the integration [test](https://github.com/apache/incubator-kie-drools/pull/6455/files#diff-e37ffabf528fd419f19b46b4f870b2a6352d429a104078bbbf200aec59c1706eR85) 

It seems to me it correctly returns an empty list (with a warning message)

Am I missing something ?

@baldimir @yesamer

yesamer (approved):

jomarko (approved):
Situation clarified in the private conversation. I can confirm the backed changes are correct.

ljmotta:
@gitgabrio Thank you for addressing this. I have just one question. The current default value of ther `time` function is 00:00Z. Shouldn't be 00:00:00+00:00?

gitgabrio:
> @gitgabrio Thank you for addressing this. I have just one question. The current default value of ther `time` function is 00:00Z. Shouldn't be 00:00:00+00:00?

`00:00:00+00:00` should mean `00:00:00` with offset `0`
`00:00Z` is its string representation from java
This is how it instantiated: 
`OffsetTime.of(0, 0, 0, 0, ZoneOffset.ofHoursMinutes(0, 0))`


Does this make sense ?
@yesamer @baldimir

ljmotta:
@gitgabrio Thanks for the explanation. I think we need to be consistent between how we display the default time. The `date and time` function is returning `1970-01-01T00:00:00` instead of `1970-01-01T00:00:00+00:00`. I'm fine with `time` returning `00:00Z`, but I believe both should have the same pattern. Maybe we could change the `date and time` to be `1970-01-01T00:00Z` as well?

gitgabrio:
@ljmotta 
AFAIK, the specs dictate the object returned, not its String representation. Anyway, we'll give it a look 👍

ljmotta:
@gitgabrio Taking a better look, I believe the `date and time` default value is not correct, as it doesn't have the timezone. We need to add the +00:00 or the Z (based on what you said of the short representation) at the end of it. 😅

gitgabrio:
Replaced by github.com/apache/incubator-kie-drools/pull/6519

PR #6456, opened 2025-09-16
[incubator-kie-drools#6448] Decision Engine: Fixing NPE in DMNEventUtils

### LINKED ISSUE #6448: NPE in DMNEventUtils.extractDSOutputDecisionsValues when used in DMN runtime listener with included models
## Issue Details

**Affected Version:** 10.1.0

**Component:** kie-dmn-core

## Description

When implementing a custom DMN runtime event listener (extending `DefaultDMNRuntimeEventListener`) to log decision service execution, the utility method `DMNEventUtils.extractDSOutputDecisionsValues()` throws a NullPointerException when processing decision services from models that include other models with different namespaces.

## Use Case

We use a custom logging listener to capture decision service inputs/outputs for audit and debugging purposes:

```java
public class DecisionLoggingListener extends DefaultDMNRuntimeEventListener {
    @Override
    public void afterEvaluateDecisionService(AfterEvaluateDecisionServiceEvent event) {
        var outputs = DMNEventUtils.extractDSOutputDecisionsValues(event);
        logger.debug("Decision Service outputs: " + outputs);
        // ... additional logging logic
    }
}
```

## Root Cause

The method at line 70 of `org.kie.dmn.core.impl.DMNEventUtils` calls `getDecisionById(id).getName()` without checking if `getDecisionById(id)` returns null. When decision IDs contain namespace information from included models, `getDecisionById()` returns null because it cannot find decisions with namespace-prefixed IDs, causing the NPE.

## Stack Trace

```
java.lang.NullPointerException: Cannot invoke "org.kie.dmn.api.core.ast.DecisionNode.getName()" because the return value of "org.kie.dmn.api.core.DMNModel.getDecisionById(String)" is null
    at org.kie.dmn.core.impl.DMNEventUtils.extractDSOutputDecisionsValues(DMNEventUtils.java:70)
    at com.example.DecisionLoggingListener.afterEvaluateDecisionService(DecisionLoggingListener.java:xx)
    at org.kie.dmn.core.impl.DMNRuntimeEventManagerUtils.fireAfterEvaluateDecisionService(DMNRuntimeEventManagerUtils.java:81)
    at org.kie.dmn.core.ast.DMNDecisionServiceEvaluator.evaluate(DMNDecisionServiceEvaluator.java:111)
```

## Steps to Reproduce

1. Create a parent DMN model that includes a child model from a different namespace
2. Have the child model contain a decision service with output decisions
3. Register a custom DMN runtime listener that calls `DMNEventUtils.extractDSOutputDecisionsValues()` in `afterEvaluateDecisionService()`
4. Evaluate the decision service
5. The listener's event processing will fail with NPE

## Expected Behavior

The `DMNEventUtils.extractDSOutputDecisionsValues()` utility method should safely handle namespace-prefixed decision IDs and extract output values correctly for use in logging listeners.

## Actual Behavior

NPE is thrown when attempting to call getName() on a null DecisionNode, breaking the logging functionality.

## Impact

This prevents proper logging/auditing of decision service executions when using included models, which is a common pattern in complex DMN implementations.

## Workaround

Implement a custom extraction method in the logging listener that handles null returns from getDecisionById() with fallback strategies instead of using the utility method.

ChinchuAjith:
Fixes https://github.com/apache/incubator-kie-drools/issues/6448

### DIFF
--- a/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNEventUtils.java
+++ b/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNEventUtils.java
@@ -25,6 +25,7 @@ import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
 import org.kie.dmn.api.core.ast.DMNNode;
+import org.kie.dmn.api.core.ast.DecisionNode;
 import org.kie.dmn.api.core.event.AfterEvaluateDecisionServiceEvent;
 import org.kie.dmn.api.core.event.BeforeEvaluateDecisionServiceEvent;
 import org.kie.dmn.api.core.event.BeforeInvokeBKMEvent;
@@ -67,9 +68,12 @@ public final class DMNEventUtils {
         Map<String, Object> results = new LinkedHashMap<>();
         List<String> decisionIDs = event.getDecisionService().getDecisionService().getOutputDecision().stream().map(er -> DMNCompilerImpl.getId(er)).collect(Collectors.toList());
         for (String id : decisionIDs) {
-            String decisionName = ((DMNResultImpl) event.getResult()).getModel().getDecisionById(id).getName();
-            Object decisionResult = event.getResult().getContext().get(decisionName);
-            results.put(decisionName, decisionResult);
+            DecisionNode decisionNode = ((DMNResultImpl) event.getResult()).getModel().getDecisionById(id);
+            if(decisionNode != null) {
+                String decisionName = decisionNode.getName();
+                Object decisionResult = event.getResult().getContext().get(decisionName);
+                results.put(decisionName, decisionResult);
+            }
         }
         return results;
     }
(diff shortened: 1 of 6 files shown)


kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/impl/DMNEventUtils.java
```diff
@@ -67,9 +68,12 @@ public static Map<String, Object> extractDSOutputDecisionsValues(AfterEvaluateDe
         Map<String, Object> results = new LinkedHashMap<>();
         List<String> decisionIDs = event.getDecisionService().getDecisionService().getOutputDecision().stream().map(er -> DMNCompilerImpl.getId(er)).collect(Collectors.toList());
         for (String id : decisionIDs) {
-            String decisionName = ((DMNResultImpl) event.getResult()).getModel().getDecisionById(id).getName();
-            Object decisionResult = event.getResult().getContext().get(decisionName);
-            results.put(decisionName, decisionResult);
+            DecisionNode decisionNode = ((DMNResultImpl) event.getResult()).getModel().getDecisionById(id);
+            if (decisionNode != null) {
```
yesamer, line 72: If I understood correctly, the passed id is not a real decision id. Is it a valid case? If yes, should we add an else branche to manage that case too? @gitgabrio @baldimir

gitgabrio: @yesamer I do not understand what you mean with:

> passed id is not a real decision id. 


Decisions may be referred to by name or by id.

gitgabrio (approved):
Thanks @ChinchuAjith LGTM!

yesamer (approved):

PR #6471, opened 2025-10-01
[NO-ISSUE] Spring Boot 3.4.10 and other libraries update to fix vulnerabilities.

baldimir:
This PR updates libraries with Spring Boot upgrade to 3.4.10 done in kogito-runtimes. It also introduces forcing version for netty libraries as those are often hit with vulnerabilities, so to not need to declare the version overrides any time in the future anymore, I added the forced version for all netty libraries in this PR. So in the future, it should be enough to just update the netty version property and it shouldn't be needed to do any dependency exclusions or similar. 

Related PRs: 
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4074
https://github.com/apache/incubator-kie-tools/pull/3295


gitgabrio (approved):
Thanks @baldimir !

yesamer (approved):

PR #6473, opened 2025-10-02
[incubator-kie-issues-6472] Make exec‑model consequence argument order deterministic

### LINKED ISSUE #6472: Exec‑model/codegen generates nondeterministic consequence argument order
### Summary
In the exec-model code generator, the set of “used declarations” for a rule consequence is collected and then iterated to (1) build `D.on(...)` and (2) build parameters for the generated `LambdaConsequence…` (`BlockN<…>`). Because this was done via an unordered collection, iteration order could vary across environments. When the two iterations disagree, the generated `BlockN<…>` signature no longer matches the `D.on(...)` argument order, causing a compile-time error in generated sources.

**Modules:** `drools-model/drools-model-codegen`  
**Area:** Exec-model code generation (consequences / `D.on(...)` and `LambdaConsequence...`)  
**Relates to:** Nondeterministic order exposed by `TypeCoercionTest#testComparisonCoercion`  

### Steps to Reproduce 
Running the test normally:

```bash
mvn -pl drools-model/drools-model-codegen \
    test -Dtest=org.drools.model.codegen.execmodel.TypeCoercionTest#testComparisonCoercion 
```
passes. However, the same test fails when executed under the NonDex plugin, which perturbs the iteration order of non-deterministic Java collections:
```bash
mvn -pl drools-model/drools-model-codegen \
    edu.illinois:nondex-maven-plugin:2.1.7:nondex \
    -Dtest=org.drools.model.codegen.execmodel.TypeCoercionTest#testComparisonCoercion
```
The failure is a compile-time error in generated sources:
```
The method execute(Block2<String,List<String>>) in the type ConsequenceBuilder._2<String,List<String>>
is not applicable for the arguments (LambdaConsequence…)
```
It indicates a mismatch between the order expected by `ConsequenceBuilder._N<...>.execute(...)` (as implied by `D.on(var_A, var_B, ...)`) and the functional type implemented by the generated lambda (`LambdaConsequence…`). 

### Root Cause

Use of unordered `Set` collection(s) (e.g., via `Collectors.toSet()`) during consequence generation. Java’s `HashSet` does not guarantee iteration order; `Collectors.toSet()` is unordered.

### Fixes

I am currently working on the fixes and will later provide a PR related to this issue.

hwu2024:

**NOTE!:** Double-check the target branch for this PR.
The default is `main` so it will target Drools 8 / Kogito.

**Ports** If a forward-port or a backport is needed, paste the forward port PR here

* N/A

**Issue**: _(please edit the GitHub Issues link if it exists)_

* [Link](https://github.com/apache/incubator-kie-drools/issues/6472)

* N/A

## Summary

**Modules:** `drools-model/drools-model-codegen`  
**Area:** Exec‑model consequence generation (`D.on(...)` & `LambdaConsequence…`)

This PR makes exec‑model consequence argument order deterministic so the list passed to `D.on(...)` always matches the parameter order of the generated `LambdaConsequence…` (i.e., the `BlockN<…>` functional type). The change prevents compile‑time mismatches that sometimes occurred in generated sources when iteration order differs across environments.

## Root cause

* Unordered iteration of a `Set` (e.g., from `Collectors.toSet()`) when determining the argument list for `D.on(...)` and the parameter list for the corresponding lambda.  
* Because the two lists are computed from a container with unspecified iteration order, the two orders can differ—leading to compile‑time errors like:
```
The method execute(Block2<String,List<String>>) in the type ConsequenceBuilder._2<String,List<String>>
is not applicable for the arguments (LambdaConsequence…)
```

## Related Tests

`TypeCoercionTest#testComparisonCoercion` 
`TypeCoercionTest#testComparisonCoercion2` 
`TypeCoercionTest#testFloatOperation` 
`TypeCoercionTest#testPrimitiveCoercion` 
`TypeObjectCoercionTest#testCoercionStringVsExplicitIntegerWithMap` 
`RuleAttributesTest#testDynamicSalience`

## Fixes

1) Preserve encounter order when collecting used declarations, and reuse that single ordered sequence in both places:
   * Replace unordered `toSet()` collectors with `toCollection(LinkedHashSet::new)` (or collect to a `List`), ensuring a stable encounter order.  
   * Thread the same ordered sequence through the construction of `D.on(...)` and the `LambdaConsequence…` parameter boxing.

2) Define ordering policy:
   * Java dialect (AST): order by first appearance in `ruleConsequence.findAll(NameExpr)` (reading order).
   * MVEL dialect (string): order by first index of occurrence in the consequence string.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


yesamer:
Hi @tkobayas, can you please review this PR?

tkobayas:
Thank you for reporting. @hwu2024 

So there are 2 places which iterate the HashSet. Practically, the same JVM results in the same order, so we haven't seen the compilation error so far, but it's not guaranteed in the future. So this PR.

The order in `D.on()`
```java
    private MethodCallExpr onCall(Collection<String> usedArguments) {
        MethodCallExpr onCall = null;

        if (!usedArguments.isEmpty()) {
            onCall = createDslTopLevelMethod(ON_CALL);
            usedArguments.stream().map(context::getVar).forEach(onCall::addArgument);
        }
        return onCall;
    }
```

The order in `LambdaConsequence.execute()`
```java
    public NodeList<Parameter> getParametersForPrototype(Collection<String> declarationUsedInRHS, BlockStmt methodBody) {
        NodeList<Parameter> parameters = NodeList.nodeList();
        for (String parameterName : declarationUsedInRHS) {
            DeclarationSpec declaration = context.getDeclarationByIdWithException(parameterName);
            Parameter boxedParameter = declaration instanceof TypedDeclarationSpec tSpec ?
                    getTypedParameter(methodBody, parameterName, tSpec) :
                    new Parameter(toClassOrInterfaceType(PrototypeFactInstance.class), parameterName);
            parameters.add(boxedParameter);
        }
        return parameters;
    }
```

@mariofusco  If I correctly remember, you had a similar PR in the past and you rejected it (I cannot find it though). WDYT this time? (The actual PR implementation aside ... I guess no need to sort, just using List would be enough)

mariofusco:
> @mariofusco If I correctly remember, you had a similar PR in the past and you rejected it (I cannot find it though). WDYT this time? (The actual PR implementation aside ... I guess no need to sort, just using List would be enough)

Sorry but I don't remember a similar PR form you (but very likely my memory is not working very well). Anyway, I believe that the iteration order of an `HashSet` is always stable, even if not predetermined, so I think there isn't any real problem. That said I'm not against this PR, but if it is fixing an actual issue it would be great to have a test case reproducing it.

tkobayas:
@mariofusco 

> it would be great to have a test case reproducing it.

https://github.com/apache/incubator-kie-drools/issues/6472 has the reproduce steps using `edu.illinois:nondex-maven-plugin` to forcibly cause the issue.

https://github.com/TestingResearchIllinois/NonDex 
```
NonDex is a tool for detecting and debugging wrong assumptions on under-determined Java APIs. An example of such an assumption is when code assumes the order of iteration through the entries in a java.util.HashMap is deterministic 
```

tkobayas:
Hi @hwu2024 ,

I wonder if sort is required. I guess just using `LinkedHashSet` is enough. Am I missing something?
```
        if (context.getRuleDialect() == RuleContext.RuleDialect.MVEL) {
            return new LinkedHashSet<>(existingDecls.stream().filter(d -> containsWord(d, consequenceString)).collect(toSet()));
        } else if (ruleConsequence != null) {
            Set<String> declUsedInRHS = ruleConsequence.findAll(NameExpr.class).stream().map(NameExpr::getNameAsString).collect(toSet());
            return new LinkedHashSet<>(existingDecls.stream().filter(declUsedInRHS::contains).collect(toSet()));
        }
```

hwu2024:
Hi @tkobayas,

Thanks for the updates and suggestions. 

My original thought is that since the method `toSet()` is unordered and would discard the stream's encounter order, I would instead sort the upstream and put it in `LinkedHashSet` to ensure both steps preserve the order. 

I have done some additional testing and found out that just using `toSet()` with `LinkedHashSet` will no longer produce any related error. In this case, would it be safer using `toCollection(LinkedHashSet::new)`, instead of using `toSet()` and wrapping it in `LinkedHashSet`, to ensure the deterministic order?

And for the AST section using RHS order, I was thinking that streaming `existingDecls` directly may lead to the encounter order driven by `existingDecls` and thus lose the RHS order. I am wondering whether we do need that order in this case. 

Thanks again for your time to follow up.

tkobayas:
> In this case, would it be safer using toCollection(LinkedHashSet::new), instead of using toSet() and wrapping it in LinkedHashSet, to ensure the deterministic order?

Thanks, `toCollection(LinkedHashSet::new)` is better :)

> And for the AST section using RHS order, I was thinking that streaming existingDecls directly may lead to the encounter order driven by existingDecls and thus lose the RHS order. I am wondering whether we do need that order in this case.

The RHS order is not important. As long as `extractUsedDeclarations` returns a `LinkedHashSet`, the compilation issue of `D.on(var_A, var_B, ...)` vs `LambdaConsequence` is solved.

hwu2024:
Thanks, @tkobayas — I have updated the patch to remove the sorting and to collect directly into a `LinkedHashSet` via `toCollection(LinkedHashSet::new)`. The current change follows your guidance that the RSH order isn’t important as long as it’s stable.

tkobayas (approved):
Thank you!

tkobayas:
Please review @mariofusco  @Rikkola 

Note that this PR doesn't contain a test case. The issue can be reproduced with `nondex-maven-plugin` (Explained in https://github.com/apache/incubator-kie-drools/issues/6472). I locally confirmed that this PR solved the issue.
```
mvn -pl drools-model/drools-model-codegen \
    edu.illinois:nondex-maven-plugin:2.1.7:nondex \
    -Dtest=org.drools.model.codegen.execmodel.TypeCoercionTest#testComparisonCoercion
```

Rikkola (approved):

mariofusco (approved):

PR #6466, opened 2025-09-26
Splitbetas

Rikkola:
https://github.com/apache/incubator-kie/issues/7047

Clean up and rebase for https://github.com/apache/incubator-kie-drools/issues/5787

PR #6481, opened 2025-10-10
Fix ClassCastException in TupleToObjectNode after beta node split

Rikkola:
https://github.com/apache/incubator-kie-issues/issues/2140

Fixes ClassCastException: AccumulateRight cannot be cast to BetaNode in TupleToObjectNode.SubnetworkPathMemory.updateRuleTerminalNodes().

After the beta node split (#6466), AccumulateRight extends RightInputAdapterNode instead of BetaNode. The updateRuleTerminalNodes method was still casting all object sinks to BetaNode, causing failures when accumulate operations invoke queries.

Changes:
- Update TupleToObjectNode.updateRuleTerminalNodes() to handle both BetaNode and RightInputAdapterNode types
- Add comprehensive unit tests for the fix

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/RuleNameExtractor.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/RuleNameExtractor.java
@@ -1,0 +1,72 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


Rikkola:
This change is tricky to unit test since it is hard to reproduce without an internal toString().

mariofusco (changes requested):
The fix looks ok to me, I'm just surprised that this `updateRuleTerminalNodes` method is only called in a `toString`. Please double check this and if so rename the method accordingly (or maybe inline its body directly in the `toString` method?).

tkobayas (approved):

mariofusco (approved):

PR #6488, opened 2025-10-15
[incubator-kie-issues #203] Cleanup Phreak code - next step 

pibizza:
More progress in the cleanup of the drl engine code. This is mostly related to phreak code.

- I have moved a number of statich methods to SegmentMemory and SegmentPrototype. This has made the code cleaner
- I have reordered the parameters in a number of methods following a more sensible approach - namely first global objects like RuleBase, ReteEvaluator and the likes, then Path/Segment memories, then nodes, then tuples, and finally flags. It was not always clean and easy to do, so you will find possibly still something not perfect, but still.
- I have reformetted the code with the drools formatter. Code is now a bit cleaner and more uniform as you look at it.

@mariofusco please can you look at org.drools.core.reteoo.SegmentMemory.mergeBitMasks(SegmentMemory). The code is copy pasted from the corresponding code in LazyPhreakBuilder, but it seems wrong.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
@@ -34,6 +34,7 @@ import org.drools.core.phreak.BuildtimeSegmentUtilities;
 import org.drools.core.phreak.RuntimeSegmentUtilities;
 import org.drools.core.reteoo.AsyncReceiveNode.AsyncReceiveMemory;
 import org.drools.core.reteoo.QueryElementNode.QueryElementNodeMemory;
+import org.drools.core.reteoo.SegmentMemory.SegmentPrototype;
 import org.drools.core.reteoo.TupleToObjectNode.SubnetworkPathMemory;
 import org.drools.core.reteoo.TimerNode.TimerNodeMemory;
 import org.drools.core.util.LinkedList;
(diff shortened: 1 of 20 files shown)


drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
```diff
@@ -282,32 +297,77 @@ public void addPathMemory(PathMemory pathMemory) {
...
+
+        sm.linkedNodeMask = currentLinkedNodeMask >> splitPos;
+        sm.dirtyNodeMask = currentDirtyNodeMask >> splitPos;
+    }
+
+    public void mergeBitMasks(SegmentMemory sm) {
+        long mask = sm.allLinkedMaskTest << sm.nodeMemories.length;
+        this.allLinkedMaskTest = mask & allLinkedMaskTest;
+
+        mask = sm.allLinkedMaskTest << sm.nodeMemories.length;
```
mariofusco, line 331: I believe this should be 

mask = sm.**linkedNodeMask** << sm.nodeMemories.length;

correct?

mariofusco: If so it's weird that we this has never caused a problem? Are you sure that this method is used at all?

pibizza: @mariofusco yes, if you check the original method it was called. Also the & should be a |, correct?

mariofusco: Yes , you're right. Not sure what are the implications of changing this, but you could try to fix this according to your intuition and rerun the test suite.

mariofusco: Thinking twice, since you didn't change the logic, and indeed all tests are green, I would rather merge this pull request as it and attempt to rewrite this method according to our intuition in a separated one, do you agree?

mariofusco (approved):

Rikkola (approved):

tkobayas (approved):

PR #6508, opened 2025-10-30
[incubator-kie-issues #203] Cleanup Phreak code - Minor refactoring in SegmentMemory handling

pibizza:
This PR introduces two improvements to segment memory handling:

  1. Added isOnlyLiaSegment() helper method to SegmentMemory
  2. Refactored createChildSegments() → initializeChildSegmentsIfNeeded() with simplified signature

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/SegmentMemorySupport.java
+++ b/drools-core/src/main/java/org/drools/core/common/SegmentMemorySupport.java
@@ -19,7 +19,6 @@
 package org.drools.core.common;
 
 import org.drools.core.reteoo.LeftTupleNode;
-import org.drools.core.reteoo.LeftTupleSinkPropagator;
 import org.drools.core.reteoo.LeftTupleSource;
 import org.drools.core.reteoo.PathEndNode;
 import org.drools.core.reteoo.PathMemory;
@@ -36,7 +35,7 @@ public interface SegmentMemorySupport {
     
     public SegmentMemory getQuerySegmentMemory(QueryElementNode queryNode);
     
-    public void createChildSegments(LeftTupleSinkPropagator sinkProp, SegmentMemory smem);
+    public void initializeChildSegmentsIfNeeded(SegmentMemory smem);
       
     public SegmentMemory createChildSegment(LeftTupleNode node);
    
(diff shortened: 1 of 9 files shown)


drools-core/src/main/java/org/drools/core/common/SegmentMemorySupport.java
```diff
@@ -36,7 +35,7 @@ public interface SegmentMemorySupport {
     
     public SegmentMemory getQuerySegmentMemory(QueryElementNode queryNode);
     
-    public void createChildSegments(LeftTupleSinkPropagator sinkProp, SegmentMemory smem);
+    public void initializeChildSegmentsIfNeeded(SegmentMemory smem);
```
mariofusco, line 38: Nitpic: I would remove the `IfNeeded` bit from the name of this method (unless you have a good reason to keep it).

drools-core/src/main/java/org/drools/core/phreak/EagerPhreakBuilder.java
```diff
@@ -1121,8 +1121,9 @@ private static TupleImpl insertPeerLeftTuple(InternalWorkingMemory wm,
             memory.getSegmentMemory().getStagedLeftTuples().addInsert(peer);
         } else {
             // If parent is Lian, then this must be called, so that any linking or unlinking can be done.
-            LeftInputAdapterNode.doInsertSegmentMemoryWithFlush(wm, true, liaMem, memory.getSegmentMemory(), peer, node
-                    .getLeftTupleSource().isStreamMode());
+            List<PathMemory> pathsToFlush = LeftInputAdapterNode.doInsertSegmentMemory(wm, true, liaMem, memory.getSegmentMemory(), peer, node
+                                .getLeftTupleSource().isStreamMode() );
+            wm.getRuleNetworkEvaluator().forceFlushPaths(pathsToFlush);
```
mariofusco, line 1126: Why is this force flush necessary now? Did it do this also before in a different way?

drools-core/src/main/java/org/drools/core/common/SegmentMemorySupport.java
pibizza: @mariofusco It is to clarify that the method is idempotent - you can call it twice and it won't get extra initailization.

drools-core/src/main/java/org/drools/core/phreak/EagerPhreakBuilder.java
pibizza: @mariofusco yes. If you see the method called originally was doInsertSegmentMemoryWithFlush. Now it is doInsertSegmentMemory.  See changes in LeftInputAdapterNode.

tkobayas (approved):

mariofusco (approved):

PR #6510, opened 2025-11-05
[incubator-kie-issues #203] Cleanup Phreak code - Extract SegmentCursor

pibizza:
Refactors PHREAK rule network evaluation to encapsulate traversal state into a dedicated SegmentCursor class.

  - Extracted SegmentCursor.java  - Encapsulates segment traversal state and navigation logic
  - Refactored RuleNetworkEvaluatorImpl.java - Simplified to focus on evaluation logic
  - Updated method signatures foro the new API

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
@@ -40,8 +40,9 @@ import static org.drools.core.phreak.BuildtimeSegmentUtilities.nextNodePosMask;
 public class SegmentCursor {
     private static final Logger log = LoggerFactory.getLogger(SegmentCursor.class);
     
-    private PathMemory pmem;
-    private SegmentMemory[] smems;
+    private final PathMemory pmem;
+    private final SegmentMemory[] smems;
+
     private int smemIndex;
     private long bit;
     private Memory nodeMem;
@@ -359,7 +360,6 @@ public class SegmentCursor {
                     sm.getPos(),
                     1L,
                     sm.getNodeMemories()[0],
-                    // The segment is the first and it has the lian shared with other nodes, the lian must be skipped, so adjust the bit and sink
                     sm.getRootNode(), 
                     tupleSets);
         }
--- a/drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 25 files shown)


tkobayas (approved):
It looks great.

tkobayas:
Btw, Drools GHA (ubuntu-latest) fails these days. Not related to this PR.

I filed a GH issue, and am looking. (Hmm, it doesn't reproduce now. Just I keep this in mind)
https://github.com/apache/incubator-kie-drools/issues/6511

Rikkola (approved):
Looks like it isolates and clarifies the SegmentCursor functionality nicely.

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
```diff
@@ -0,0 +1,475 @@
...
+import org.drools.core.reteoo.SegmentMemory;
+import org.drools.core.reteoo.SegmentNodeMemory;
+import org.drools.core.reteoo.TerminalNode;
+import org.drools.core.reteoo.TupleToObjectNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.drools.core.phreak.BuildtimeSegmentUtilities.nextNodePosMask;
+
+public class SegmentCursor {
```
mariofusco, line 40: Can we make this whole class package private?

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
```diff
@@ -0,0 +1,475 @@
...
+    private static final Logger log = LoggerFactory.getLogger(SegmentCursor.class);
+    
+    private PathMemory pmem;
```
mariofusco: I believe that some of these fields could be flagged as `final`?

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
```diff
@@ -0,0 +1,475 @@
...
+        this.pmem = pmem;
+        this.smems = smems;
+        this.smemIndex = smemIndex;
+        this.bit = bit;
+        this.nodeMem = nodeMem;
+        this.node = node;
+        this.srcTuples = srcTuples;
+    }
+    
+    public PathMemory getPathMemory() {
```
mariofusco, line 69: Are all these getters necessary? Maybe if this class is package private and used only by the `RuleNetworkEvaluator` we could actually get rid of all of them, also make the fields package private and access directly them?

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
pibizza: Not really, it is used in the metric project by a phreak metric node. I plan to extend to other phreak nodes also.

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
pibizza: Done

drools-core/src/main/java/org/drools/core/phreak/SegmentCursor.java
pibizza: See comment above. I plan to extend the use of the SegmentCursor to other prheak nodes, so probably it's better to keep the getters. Eventually we can decide to throw them away - I really hope we can encapsulate most of their usage anyway.

mariofusco (approved):

PR #6516, opened 2025-11-10
[incubator-kie-issues #203] Cleanup Phreak code - Remove getActivationsManager from ReteEvaluator

pibizza:
Another relatively simple PR, moving the code into the direction of breaking the dependencies for ReteEvaluator. 

In this PR I have removed the getReteEvaluator from the ActivationsManager interface. 

This is because the implementations of the  ActivationsManager are created inside the implementations of the ReteEvaluator.

Sso it should not be their responsibility to provide ReteEvaluator to the other objects.

The objects that need a reference to the ReteEvaluator now have it directly at creation time.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/EagerPhreakBuilder.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/EagerPhreakBuilder.java
@@ -1098,8 +1098,8 @@ public class EagerPhreakBuilder implements PhreakBuilder {
             if (insert) {
                 TerminalNode rtn = (TerminalNode) node;
                 InternalAgenda agenda = wm.getAgenda();
-                RuleAgendaItem agendaItem = AlphaTerminalNode.getRuleAgendaItem(wm, agenda, rtn, insert);
-                PhreakRuleTerminalNode.doLeftTupleInsert(rtn, agendaItem.getRuleExecutor(), agenda, agendaItem,
+                RuleAgendaItem agendaItem = AlphaTerminalNode.getRuleAgendaItem(wm, rtn, insert);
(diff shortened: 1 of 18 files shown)


drools-core/src/main/java/org/drools/core/phreak/EagerPhreakBuilder.java
```diff
@@ -1098,8 +1098,8 @@ private static TupleImpl insertPeerLeftTuple(InternalWorkingMemory wm,
             if (insert) {
                 TerminalNode rtn = (TerminalNode) node;
                 InternalAgenda agenda = wm.getAgenda();
-                RuleAgendaItem agendaItem = AlphaTerminalNode.getRuleAgendaItem(wm, agenda, rtn, insert);
-                PhreakRuleTerminalNode.doLeftTupleInsert(rtn, agendaItem.getRuleExecutor(), agenda, agendaItem,
+                RuleAgendaItem agendaItem = AlphaTerminalNode.getRuleAgendaItem(wm, rtn, insert);
+                PhreakRuleTerminalNode.doLeftTupleInsert(wm, rtn, agendaItem.getRuleExecutor(), agenda, agendaItem,
```
Rikkola, line 1102: Not related to this change, but I noticed this passes in agendaItem and rule exectutor from agendaItem. Could it just pass agendaItem and get rule executor once in?

Rikkola: Ok looking down on the changes, in other locations it would not work.

Rikkola (approved):

mariofusco (approved):

tkobayas (approved):

PR #6517, opened 2025-11-11
Fix issue where activation-group is not applied after rule fires initially. fixes #6509

### LINKED ISSUE #6509: Activation Groups are only applied for initial rule activation
When multiple rules belonging to the same activation group are activated repeatedly by fact changes, only the first activation fires the cancellation of other activations in the activation-group.  Subsequent activations result in the firing of ALL activated rules in the activation group.  None are cancelled.

A simple ruleset that demonstrates this behavior is as follows.

```
package com.example
            
import StringFact;
            
rule "First rule in first activation group"
    activation-group "first-group"
    salience 10
when
    $strf : StringFact(value == "force", count < 5)
then
end

rule "Second rule in first activation group"
    activation-group "first-group"
    salience 5
when
    $strf : StringFact(value == "force", count < 5)
then
    modify($strf) { setValue("fail") }
end

rule "Rule without activation group"
    salience 2
when
    $strf : StringFact(value == "force", count < 1)
then
    modify($strf) { incrementCount() }
end
```
The second rule should never fire as long as the first rule has also been activated.

This test fails in the 7.x branch as well as the main branch.

I tracked down the issue to differences in PhreakRuleTerminalNode.doLeftTupleInsert(...) vs PhreakRuleTerminalNode.doLeftTupleUpdate(...).  The insert method calls         activationsManager.addItemToActivationGroup( leftTuple ) but the update method does not.  Adding this call to the update methods allows the test to succeed.

See attached test class for the Drools::Model::Codegen module and PhreakRuleTerminalNode.java with suggested fix.  This is my first time looking at  this code base so "fix" needs to be evaluated for correctness.

[ActivationGroupReactivateTest.java](https://github.com/user-attachments/files/23243570/ActivationGroupReactivateTest.java)

[PhreakRuleTerminalNode.java](https://github.com/user-attachments/files/23243581/PhreakRuleTerminalNode.java)

aner-perez:
If multiple rules in the same activation-group are re-triggered by a fact change, any subsequent firing of these rules do not cancel other activations for the activation-group.

A unit test is included that reproduces the problem.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -228,6 +228,7 @@ public class PhreakRuleTerminalNode {
 
                     leftTuple.update( salienceInt, pctx );
                     executor.modifyActiveTuple(leftTuple );
+                    activationsManager.addItemToActivationGroup( leftTuple );
                     reteEvaluator.getRuleEventSupport().onUpdateMatch( leftTuple );
                 }
             }
--- a/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupReactivateTest.java
+++ b/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupReactivateTest.java
@@ -24,7 +24,7 @@ import java.util.List;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.kie.api.event.rule.AfterMatchFiredEvent;
-import org.kie.api.event.rule.DefaultAgendaEventListener;
+import org.drools.core.event.TrackingAgendaEventListener;
 import org.kie.api.runtime.KieSession;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


tkobayas:
looking...

tkobayas:
Thank you @aner-perez , the fix looks pretty valid.

Just one point, recently we have introduced `TrackingAgendaEventListener` which is exactly useful for test cases like yours.

https://github.com/apache/incubator-kie-drools/blob/main/drools-core/src/main/java/org/drools/core/event/TrackingAgendaEventListener.java

So your custom eventListener can be replaced like this:

https://github.com/apache/incubator-kie-drools/blob/main/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java#L140-L141
https://github.com/apache/incubator-kie-drools/blob/main/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/ActivationGroupDualFactMatchTest.java#L175-L177

Please use TrackingAgendaEventListener. Thanks!

tkobayas (approved):

aner-perez:
Thanks for reviewing this @tkobayas , I changed the unit test to use TrackingAgendaEventListener as suggested

mariofusco (approved):

PR #6520, opened 2025-11-20
[incubator-kie-drools-6445] Problem in calling RuleUnitProviderImpl l…

### LINKED ISSUE #6445: Problem in calling RuleUnitProviderImpl loadRuleUnits from a multithreaded context
While calling RuleUnitProviderImpl - loadRuleUnits() i.e. from the contructor, we have encountered a problem while doing so in a multithreaded context. The ruleUnitMap is a normal HashMap instead of a ConcurrentHashMap. This is causing a NPE for a brief moment before a next retry.

tkobayas:
- closes https://github.com/apache/incubator-kie-drools/issues/6445

### DIFF
--- a/drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
+++ b/drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
@@ -32,6 +32,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


tkobayas:
`ConcurrentHashMap` is enough to avoid NPE, but I also added `synchronized (generationLock)` to avoid duplicated generation.

drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
```diff
@@ -66,21 +68,29 @@ public class RuleUnitProviderImpl implements RuleUnitProvider {
...
     public RuleUnitProviderImpl() {
-        this.ruleUnitMap = loadRuleUnits(Thread.currentThread().getContextClassLoader());
+        this.ruleUnitMap = new ConcurrentHashMap<>(loadRuleUnits(Thread.currentThread().getContextClassLoader()));
     }
 
     @Override
     public <T extends RuleUnitData> RuleUnit<T> getRuleUnit(T ruleUnitData) {
         String ruleUnitName = getRuleUnitName(ruleUnitData);
-        RuleUnit<T> ruleUnit = ruleUnitMap.get(ruleUnitName);
+        RuleUnit<T> ruleUnit = (RuleUnit<T>) ruleUnitMap.get(ruleUnitName);
```
gitgabrio, line 84: Hi @tkobayas 
I saw that modification just by chance, but I'm not sure to understand why that -> 

```
ruleUnit = (RuleUnit<T>) ruleUnitMap.get(ruleUnitName);
        if (ruleUnit != null) {
            return ruleUnit;
        }
```

has to be repeated inside the `synchronized` block, if first retrieval is `null`...

drools-ruleunits/pom.xml
```diff
@@ -59,4 +59,32 @@
...
+                <systemProperties>
+                  <property>
+                    <name>runTurtleTests</name>
+                    <value>true</value>
+                  </property>
+                </systemProperties>
+              </configuration>
+            </plugin>
+          </plugins>
+        </pluginManagement>
```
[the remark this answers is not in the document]
tkobayas: This `<pluginManagement>` is actually effective, because `maven-surefire-plugin` is implicitly effective by maven default. But probably it's worth reviewing the configurations. Filed a GH issue:
https://github.com/apache/incubator-kie-drools/issues/6522

drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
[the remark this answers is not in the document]
tkobayas: Added `@SuppressWarnings("unchecked")` with a comment accordingly.

tkobayas: This prevents redundant generation when concurrent accesses both pass the initial null check. I added a comment to the source.

drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
```diff
@@ -66,21 +68,33 @@ public class RuleUnitProviderImpl implements RuleUnitProvider {
 
     private static final boolean USE_EXEC_MODEL = true;
 
-    private final Map<String, RuleUnit> ruleUnitMap;
+    private final ConcurrentMap<String, RuleUnit> ruleUnitMap;
+    private final Object generationLock = new Object();
```
mariofusco: Nitpick: using a `synchronized` block is mostly considered an antipattern nowadays and we should actually get rid of all of them throughout our codebase. Please consider replacing it with a proper `ReentrantLock`.

mariofusco (approved):

Rikkola (approved):

drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/RuleUnitProviderImpl.java
tkobayas: Sure. Updated.

PR #6523, opened 2025-11-21
Add Definition component for DRL type declaration verification in Drool Verifier

Rikkola:
https://github.com/apache/incubator-kie-issues/issues/2176

With AI generated DRL the AI tries to cheat by using facts and fields of facts that it did not actually define in the DRL Definitions. These changes make it possible to verify that the DRL is complete.


Rikkola:
I will squash the commits, but after pushing this my local git installation fails to force push.

Rikkola:
Had to reopen this in a new PR here https://github.com/apache/incubator-kie-drools/pull/6524

PR #6527, opened 2025-11-21
Fix for optaplanner test failures

pibizza:
This fixes the issue with optaplanner tests. 

The root cause is an error in migrating static methods from RuleNetworkEvaluator to PhreakNodeOperations.


mariofusco (approved):

gitgabrio:
THanks @pibizza !
I can't evaluate the modification, but if optaplanner successfully build, it is fine for me!

pibizza:
The failure in kogito runtimes seems unrelated.

pibizza:
There are two failures in kogito apps that seem unrelated also. I am reexecuting just to check if they were transient failures.

yesamer (approved):

pibizza:
@gitgabrio optaplanner tests are fixed, however I see a kogito runtime test failing. I have tried reexecuting it and it fails again. Do you know if it is an unstable test or it is something we should investigate?

yesamer:
All checks are now green. Merging it.

PR #6555, opened 2026-01-16
Refactor Equality Operator handling code in FEEL simple positive unary test Parsing

### LINKED ISSUE #2213: Refactor Equality Operator handling code in FEEL Parsing
Previously, equality (=) and not-equality (!=) operators were handled together in a dedicated method, while all other operators were processed via a separate switch-case structure in another method. This split led to inconsistent operator handling and unnecessary fragmentation in the decision table expression parsing flow.

As part of this refactoring:

- The equality operator handling has been moved into the same method used by other relational operators, ensuring a single, consistent entry point for operator parsing.
 - This aligns equality handling with the existing switch-case logic, since all operators now invoke the same method.
- The not-equality (!=) operator remains in a separate method, as it follows a distinct parsing path and behavior that differs from other operators.

This change improves code readability, reduces duplication, and makes the operator parsing logic easier to maintain and extend, while preserving existing behavior.

ChinchuAjith:
Fixes https://github.com/apache/incubator-kie-issues/issues/2213


yesamer (approved):

gitgabrio (approved):
Great, thanks @ChinchuAjith

PR #6557, opened 2026-01-20
[Incubator kie issues#2028] Comparing lists using in does not work

### LINKED ISSUE #2028: Comparing lists using `in` does not work
The following case are supposed to work:

`[1,2,3] in =[1,2,3]` should return `true` (currently it throws an error)
`[1,2,3] in (=[1,2,3])` should return `true` (currently it throws an error)
`[1,2,3] in =[1,2,3,4]`should return `false` (currently it throws an error)

In those cases, the right operand is converted to Range (because of this change https://github.com/apache/incubator-kie-drools/pull/6231), and I guess this is not the correct logic as a List is not necessary a Range (reference ASTBuilderVisitor:visitPositiveUnaryTestIneq:188)

`[1,2,3] in !=[1,2,3]`should return `false` (currently it returns true)
`[1,2,3] in (!=[1,2,3])`should return `false` (currently it returns true)

In those cases, the right operand is converted to an UnaryTest, that is not detected as a Collection. Then, the comparison fails because a comparison between 2 collections is not performed (reference InNode:evaluate:68)

Reference are TCK Test 0072-feel-in.dmn

AthiraHari77:
Fixes : https://github.com/apache/incubator-kie-issues/issues/2028

</details>


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/UnaryTestNode.java
```diff
@@ -158,15 +159,66 @@ public UnaryTest getUnaryTest() {
...
+
+    /**
+     * Checks if a collection contains a specific element.
+     *
+     * @param collection the collection to search in
+     * @param element the element to search for
+     * @return true if collection contains the element, false otherwise
+     */
+    static Boolean isElementInCollection(Collection<?> collection, Object element) {
+        return collection.contains(element);
```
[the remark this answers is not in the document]
gitgabrio: @yesamer @AthiraHari77 @ChinchuAjith 
That comment seems reasonable (not the suggested code).
But, the original implementation was the Collection.contains: wdyt ?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/UnaryTestNodeTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class UnaryTestNodeTest {
+
+    @Test
+    void testAreCollectionsEqualInOrder_withEqualCollections() {
+        List<Integer> left = Arrays.asList(1, 2, 3);
```
gitgabrio: Thanks @AthiraHari77 
All the methods invoking ""UnaryTest.areCollectionsEqual") could be replaced with a single method and using the `Parameterized` semantic: this would make test reading/writing easier (this comment has been done multiple times, in the past, by some very expert member in that area)

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/UnaryTestNodeTest.java
```diff
@@ -0,0 +1,223 @@
...
+        List<Integer> element = Arrays.asList(1, 2, 3);
+        List<List<Integer>> collection = Arrays.asList(
+            Arrays.asList(1, 2, 3, 4), Arrays.asList(1, 2, 3));
+
+        Boolean result = UnaryTestNode.isElementInCollection(collection, element);
+        assertThat(result).isTrue();
+    }
+
+    @Test
+    void testListInCollection_NoMatch() {
```
gitgabrio: same comment as before

gitgabrio (commented):
Thanks @AthiraHari77 !
From my pov, we should investigate a bit the suggesiont from copilot and, maybe, the tests could be refactored as "parameterized" for better readability/maintainability

gitgabrio (approved):
Thanks @AthiraHari77 well done!

yesamer (approved):
Well Done!

PR #6562, opened 2026-01-23
[incubator-kie-drools-6549] Drop unused drools soft keywords

### LINKED ISSUE #6549: Drop unused drools soft keywords
Confirm these keywords are not actually used. Consider dropping. 
- `refract` : Not used. Consider dropping. A test case (to parse `refract` as a rule attribute) exists, but actually `refract` has no effect.
- `direct` : Not used. Consider dropping. (`@activationListener('direct')` is not related)
- `template` : not used. Consider dropping. `drools-template` is not related

tkobayas:
- Fixes https://github.com/apache/incubator-kie-drools/issues/6549

    - `refract` : Not used. There is a test case to parse `refract` as a rule attribute, but actually `refract` has no effect.
    - `direct` : Not used. (`@activationListener('direct')` is not related)
    - `template` : Not used. `drools-template` uses a word `template`, but it is not related to the `DroolsSoftKeyword.TEMPLATE`

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -1828,23 +1828,6 @@ class MiscDRLParserTest {
     }
 
     @Test
-    void attributeRefract() {
-        final String source = "rule Test refract when Person() then end";
-
-        PackageDescr pkg = parseAndGetPackageDescr(
-                source);
-
-        RuleDescr rule = (RuleDescr) pkg.getRules().get(0);
-
-        assertThat(rule.getName()).isEqualTo("Test");
-        Map<String, AttributeDescr> attributes = rule.getAttributes();
-        assertThat(attributes).hasSize(1);
-        AttributeDescr refract = attributes.get("refract");
-        assertThat(refract).isNotNull();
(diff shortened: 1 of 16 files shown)


mariofusco (approved):

yesamer:
Hi @tkobayas! Thank you for your PR and your commitment to keep the codebase clean. However, I have some concerns about this change. 
When you said NOT used: that means that using those keywords in a DRL have no effect, right? 
So I wonder that endusers could still using them (maybe with very old DRL?) unaware those keywords doesn't produce any effect, but removing them could break the DRL compilation, right?
If the described scenario is possibile, can you please:
- Double check if there is any reference in the Apache KIE documentation? And in case remove that?
- In the upcoming 10.2.0 release we should mention this change in the Release note.
WDYT?
Thanks! Yeser

tkobayas:
@yesamer Thank you for pointing this out.

> When you said NOT used: that means that using those keywords in a DRL have no effect, right?
So I wonder that endusers could still using them (maybe with very old DRL?) unaware those keywords doesn't produce any effect, but removing them could break the DRL compilation, right?

Yes, that's true.

> Double check if there is any reference in the Apache KIE documentation? And in case remove that?

Confirmed that there is no reference about these keywords.

> In the upcoming 10.2.0 release we should mention this change in the Release note.

Sure. I have filed https://github.com/apache/incubator-kie-docs/pull/4578

Thanks!

yesamer (approved):
Thank you @tkobayas

PR #6569, opened 2026-01-28
[incubator-kie-drools-6561] Update supported java versions > 19 for DRL compilation

tkobayas:
- https://github.com/apache/incubator-kie-drools/issues/6561
  - Added java versions up to `21` for DRL compilation.
  - This PR is about DRL compilation, not about building drools project itself
  - Upgraded eclipse-ecj from `3.33.0` to `3.44.0`
  - Added Java 21 syntax test (unit test, drl test, kjar test)
  - Supports only dialect "java" (default)
  - For dialect "mvel"
    - non-exec-mode) `drools-mvel` is deprecated, so we will not likely upgrade this part
    - exec-model) `drools-model`/`mvel.jj` is based on Java 16 syntax now. Enhancement would be a separate issue/PR -> https://github.com/apache/incubator-kie-drools/issues/6571
  - Refactored `JavaConfiguration` to remove duplicate java version definitions as @gitgabrio suggested
  - We should clean up old versions usage / tests (`1.5` - `16`) , because the current drools requires at least version `17` to run. It will be a separated issue/PR -> https://github.com/apache/incubator-kie-drools/issues/6572

Currently PR check GHA is running with Java17. We plan to change it to Java21 after `10.2.0` release : https://lists.apache.org/thread/o38t7bswk1w7nzd78ftsld2t1psbspdr

So at the moment, I'm testing this PR code with Java21 GHA in https://github.com/apache/incubator-kie-drools/pull/6570

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -94,7 +94,7 @@
     <version.org.apache.poi>5.4.1</version.org.apache.poi>
     <version.org.apache.tomcat.tomcat-dbcp>10.1.48</version.org.apache.tomcat.tomcat-dbcp>
     <version.org.assertj>3.27.3</version.org.assertj>
-    <version.org.eclipse.jdt>3.33.0</version.org.eclipse.jdt>
+    <version.org.eclipse.jdt>3.44.0</version.org.eclipse.jdt>
     <version.org.freemarker>2.3.32</version.org.freemarker>
     <version.org.glassfish.jaxb>4.0.6</version.org.glassfish.jaxb>
     <!--This needs to be in sync with JUnit-->
--- a/drools-ecj/src/main/java/org/drools/ecj/EclipseJavaCompiler.java
+++ b/drools-ecj/src/main/java/org/drools/ecj/EclipseJavaCompiler.java
@@ -45,6 +45,8 @@ import org.eclipse.jdt.internal.compiler.env.ICompilationUnit;
 import org.eclipse.jdt.internal.compiler.env.INameEnvironment;
 import org.eclipse.jdt.internal.compiler.env.NameEnvironmentAnswer;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment;
+import org.eclipse.jdt.internal.compiler.lookup.ModuleBinding;
 import org.eclipse.jdt.internal.compiler.problem.DefaultProblemFactory;
 import org.kie.memorycompiler.AbstractJavaCompiler;
 import org.kie.memorycompiler.CompilationProblem;
@@ -155,6 +157,16 @@ public final class EclipseJavaCompiler extends AbstractJavaCompiler {
 
         public boolean ignoreOptionalProblems() {
             return true;
+        }
+
+        @Override
+        public char[] getModuleName() {
+            return ModuleBinding.UNNAMED;
+        }
+
+        @Override
+        public ModuleBinding module(LookupEnvironment environment) {
+            return environment.getModule(ModuleBinding.UNNAMED);
         }
     }
 
--- a/drools-ecj/src/test/java/org/drools/ecj/JavaCompilerI18NTest.java
+++ b/drools-ecj/src/test/java/org/drools/ecj/JavaCompilerI18NTest.java
@@ -43,8 +43,8 @@ public class JavaCompilerI18NTest {
         reader.add(fileStr, fileContents.getBytes());
 
         EclipseJavaCompilerSettings settings = new EclipseJavaCompilerSettings();
-        settings.setSourceVersion( "1.5" );
-        settings.setTargetVersion( "1.5" );
+        settings.setSourceVersion( "17" );
+        settings.setTargetVersion( "17" );
         EclipseJavaCompiler compiler = new EclipseJavaCompiler( settings, "" );
         CompilationResult res = compiler.compile( classes.toArray( new String[classes.size()] ), reader, store );
         assertThat(0).isEqualTo(res.getErrors().length);
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
@@ -124,6 +124,10 @@ import static org.drools.util.MethodUtils.findMethod;
 
 public class DrlxParseUtil {
 
+    // Regardless of the configured language level, always parse with the highest level at this point,
+    // so compiler will report any incompatible syntax at the compilation phase.
+    public static final ParserConfiguration.LanguageLevel PARSABLE_LANGUAGE_LEVEL = ParserConfiguration.LanguageLevel.JAVA_21;
+
     public static final String THIS_PLACEHOLDER = "_this";
 
     private static final ConcurrentMap<String, Method> ACCESSOR_CACHE = new ConcurrentHashMap<>();
@@ -497,7 +501,7 @@ public class DrlxParseUtil {
 
     public static BlockStmt parseBlock(String ruleConsequenceAsBlock) {
         ParserConfiguration parserConfiguration = new ParserConfiguration();
-        parserConfiguration.setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_21);
+        parserConfiguration.setLanguageLevel(PARSABLE_LANGUAGE_LEVEL);
         com.github.javaparser.JavaParser javaParser = new com.github.javaparser.JavaParser(parserConfiguration);
 
         ParseResult<BlockStmt> blockStmtParseResult = javaParser.parseBlock(String.format("{%n%s%n}", ruleConsequenceAsBlock));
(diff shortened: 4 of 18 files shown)


build-parent/pom.xml
```diff
@@ -94,7 +94,7 @@
     <version.org.apache.poi>5.4.1</version.org.apache.poi>
     <version.org.apache.tomcat.tomcat-dbcp>10.1.48</version.org.apache.tomcat.tomcat-dbcp>
     <version.org.assertj>3.27.3</version.org.assertj>
-    <version.org.eclipse.jdt>3.33.0</version.org.eclipse.jdt>
+    <version.org.eclipse.jdt>3.44.0</version.org.eclipse.jdt>
```
tkobayas, line 97: Updated eclipse ecj, because 3.33.0 doesn't support Java21.

drools-ecj/src/main/java/org/drools/ecj/EclipseJavaCompiler.java
```diff
@@ -156,6 +158,16 @@ public char[][] getPackageName() {
...
+
+        @Override
+        public char[] getModuleName() {
+            return ModuleBinding.UNNAMED;
+        }
+
+        @Override
+        public ModuleBinding module(LookupEnvironment environment) {
+            return environment.getModule(ModuleBinding.UNNAMED);
+        }
```
tkobayas, line 170: These methods are required (= provide a module name) to compile Java codes with eclipse ecj 3.44.0.

drools-ecj/src/test/java/org/drools/ecj/JavaCompilerI18NTest.java
```diff
@@ -43,8 +43,8 @@ public void testi18NFile () throws Exception {
         reader.add(fileStr, fileContents.getBytes());
 
         EclipseJavaCompilerSettings settings = new EclipseJavaCompilerSettings();
-        settings.setSourceVersion( "1.5" );
-        settings.setTargetVersion( "1.5" );
+        settings.setSourceVersion( "17" );
+        settings.setTargetVersion( "17" );
```
tkobayas, line 47: eclipse ecj 3.44.0 doesn't support language level lower than `1.8`.

drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/DrlxParseUtil.java
```diff
@@ -497,7 +497,7 @@ public int hashCode() {
 
     public static BlockStmt parseBlock(String ruleConsequenceAsBlock) {
         ParserConfiguration parserConfiguration = new ParserConfiguration();
-        parserConfiguration.setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_15);
+        parserConfiguration.setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_21);
```
tkobayas: Using `ParserConfiguration.LanguageLevel.JAVA_21` is fine even if the runtime java is 17, because this is used for parsing and java code generation. Of course, if you use java21 syntax with java17 runtime, it will result in a compilation error, which is expected.

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/integrationtests/SwitchOverStringTest.java
```diff
@@ -70,18 +69,4 @@ public void testCompileSwitchOverStringWithLngLevel17(KieBaseTestConfiguration k
...
-        System.setProperty("drools.dialect.java.compiler.lnglevel", "1.6");
-        try {
-            KieBuilder kieBuilder = KieUtil.getKieBuilderFromDrls(kieBaseTestConfiguration, false, FUNCTION_WITH_SWITCH_OVER_STRING);
-            List<Message> errors = kieBuilder.getResults().getMessages(Message.Level.ERROR);
-            assertThat(errors.isEmpty()).as("Should have an error").isFalse();
-            
-        } finally {
-            System.clearProperty("drools.dialect.java.compiler.lnglevel");
-        }
-    }
```
tkobayas, line 86: removed the test because drools runs on java 17 or higher, so java version 1.6 test is no longer required,

kie-memory-compiler/src/main/java/org/kie/memorycompiler/JavaConfiguration.java
```diff
@@ -86,31 +86,19 @@ public static String findJavaVersion() {
...
-            return "17";
-        } else if (level.startsWith("18")) {
-            return "18";
-        } else if (level.startsWith("19")) {
-            return "19";
+        String normalized = normalizeVersion(level);
+        if (Arrays.binarySearch(LANGUAGE_LEVELS, normalized) >= 0) {
+            return normalized;
         }
+        return "11"; // default
```
yesamer: @tkobayas Shouldn't the default be 17?

tkobayas: @yesamer You are right. Thank you!

gitgabrio (commented):
Hi @tkobayas , many thanks for the PR and the improvements you made.
Anyway, IINW, there are still different places where the same "data" are somehow duplicated:
1. EclipseJavaCompilerSettings
2. ParserConfiguration.LanguageLevel
3. JavaConfiguration


Do you think it would be possible to centralize all of them in one single place, maybe also a new class in some topmost module... dunno....

tkobayas:
@gitgabrio Thank you for the suggestion. I made further refactoring to remove the duplication in `EclipseJavaCompilerSettings`, so the java version is centralized to `JavaConfiguration`. `ParserConfiguration.LanguageLevel` is an external class (javaparser), so we cannot refactor it, but I clarified the purpose with a comment.

build-parent/pom.xml
mariofusco: I believe that at some point we could get rid of the Ecplise compiler and rely only on the native Java one, but that's ok for now.

kie-memory-compiler/src/main/java/org/kie/memorycompiler/JavaConfiguration.java
```diff
@@ -31,16 +31,16 @@
...
- * The default compiler is Eclipse and the default lngLevel is 1.5.
+ * The default compiler is Eclipse and the default lngLevel is 17.
  * The lngLevel will attempt to autodiscover your system using the 
  * system property "java.version"
  */
 public class JavaConfiguration {
 
     // This should be in alphabetic order to search with BinarySearch
-    protected static final String[]  LANGUAGE_LEVELS = new String[]{"1.5", "1.6", "1.7", "1.8", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "9"};
+    protected static final String[]  LANGUAGE_LEVELS = new String[]{"1.5", "1.6", "1.7", "1.8", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "9"};
```
mariofusco, line 43: I believe we could remove everything below `17`?

mariofusco (approved):

kie-memory-compiler/src/main/java/org/kie/memorycompiler/JavaConfiguration.java
tkobayas: @mariofusco Yes, I will work on it next. https://github.com/apache/incubator-kie-drools/issues/6572

build-parent/pom.xml
tkobayas: @mariofusco I can file a GH issue to remove the Eclipse compiler. Is there any blocker that prevents us from removing the Eclipse compiler?

For now, if Eclipse compiler exists in classpath, Eclipse compiler is used by default. https://github.com/apache/incubator-kie-drools/blob/10.1.0/drools-compiler/src/main/java/org/drools/compiler/compiler/JavaDialectConfiguration.java#L99

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/compiler/commons/jci/compilers/NativeJavaCompilerSettingsTest.java
```diff
@@ -47,17 +47,17 @@ public void allSettings() {
         settings.setWarnings(true);
         settings.setDeprecations(true);
         settings.setSourceEncoding("My-Custom-Encoding");
-        settings.setSourceVersion("1.9");
-        settings.setTargetVersion("1.9");
+        settings.setSourceVersion("9");
+        settings.setTargetVersion("9");
```
yesamer, line 51: @tkobayas Why version 9 here?

yesamer (approved):

drools-test-coverage/test-compiler-integration/src/test/java/org/drools/mvel/compiler/commons/jci/compilers/NativeJavaCompilerSettingsTest.java
tkobayas: @yesamer  `1.9` was simply wrong and the version has to be `9` (`1.x` format was used until `1.8`). I will drop versions lower than `17` in https://github.com/apache/incubator-kie-drools/issues/6572 , but at the moment, testing with `9` is valid.

build-parent/pom.xml
tkobayas: Filed https://github.com/apache/incubator-kie-drools/issues/6576

PR #6592, opened 2026-02-20
[NO-ISSUE] max-parallel for GHA

tkobayas:
ASF Infrastructure notified us of the following issue:
```
Our analysis has found that the following GitHub Actions workflows need remediation:
        Kogito Downstream: `max-parallel: 20` is required for job matrices. see https://s.apache.org/max-parallel for more details
```
In our case, the number of jobs is currently 7, which is below the limit of 20. However, this appears to be a precautionary requirement, so I’ve applied the fix.

### DIFF
--- a/.github/workflows/pr-downstream.yml
+++ b/.github/workflows/pr-downstream.yml
@@ -38,6 +38,7 @@ jobs:
       cancel-in-progress: true
     timeout-minutes: 180
     strategy:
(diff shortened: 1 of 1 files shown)


gitgabrio (approved):

yesamer (approved):

yesamer:
Thank you @tkobayas !!

PR #6618, opened 2026-03-18
Add configurable compile-time warning for self-referencing constraints

### LINKED ISSUE #2272: Self-referencing constraint (e.g. fieldA == fieldA) compiles silently and produces "incorrect runtime behavior"
A community member reported that a rule containing a self-referencing constraint compiles without any warning or error but produces incorrect runtime behavior. The constraint `fieldA == fieldA` becomes a self-comparison that always evaluates to `true`, causing rules to fire incorrectly for all facts. The correct intent was to reference a bound variable from another pattern (e.g., `fieldA == $fieldA`).

This is particularly dangerous in large-scale rule systems and during migrations, where such issues are difficult to detect and can lead to incorrect business outcomes without obvious failures.

### Reproducer

**Incorrect (compiles but behaves "incorrectly"):**
```drl
rule "SampleRule"
when
  temp : TempObject(fieldA == "X", fieldB : fieldB)
  AnotherObject(fieldA == fieldA)
then
  temp.setFlag("Y");
end
```

- `fieldA == fieldA` becomes a self-comparison, always true
- The rule fires incorrectly for all facts

**Correct (expected behavior):**
```drl
rule "SampleRule"
when
  $temp : TempObject(fieldA == "X", $fieldB : fieldB)
  AnotherObject(fieldA == $fieldB)
then
  $temp.setFlag("Y");
end
```

### Request

Detect self-comparisons like `fieldA == fieldA` at compile time and provide at minimum a warning-level diagnostic. 

### Original report

From community member Geeta:

> We observed a scenario where a rule compiles successfully but produces incorrect runtime behavior due to missing `$` in variable bindings. Ideally, such cases should fail during compilation rather than silently executing with unintended logic.
>
> In large-scale rule systems, especially during migrations or automated transformations, such issues are hard to catch and can lead to incorrect business outcomes without obvious failures.
>
> Even a warning-level validation would significantly help prevent subtle production issues.

porcelli:
Resolves https://github.com/apache/incubator-kie-issues/issues/2272

Self-comparisons like fieldA == fieldA (missing $ prefix) now emit a WARNING at compile time instead of silently producing always-true constraints. 

Severity is configurable via drools.kbuilder.severity.selfReferencingConstraint (WARNING default, upgradable to ERROR or suppressible with INFO).

### DIFF
--- a/drools-compiler/src/main/java/org/drools/compiler/compiler/SelfReferencingConstraint.java
+++ b/drools-compiler/src/main/java/org/drools/compiler/compiler/SelfReferencingConstraint.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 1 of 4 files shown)


porcelli:
The failing test seems unrelated to the changes in this PR

tkobayas (approved):

mariofusco (approved):

PR #6619, opened 2026-03-18
[incubator-kie-issues#2261] Group all version declarations in single pom

### LINKED ISSUE #2261: Group all version declarations in single pom
As demontrated by PRs on quarkus upgrade, version of the same libraries are declared/duplicated in multiple boms/poms, making harder to maintain/keep in sync them all over the repositories.

Scope of this ticket is to:

1. identify all of them
2. group all of them in one single pom.xml inside the "drools" repository and have all of them inherited throughout the codebase, so that such version would be declared only in one single place
3. one of the observed pattern is that the version is declared in some pom, and then the library itself is declared in some dependencyManagement tag in children pom - which some how break the concept of dependencyManagement in two different places (the pom where the version is declared as property and the one where it is used in the dependencyManagement) and lead to the duplication - ideally, we should need one single dependencyManagement declaration all over the stack

gitgabrio:
Fixes: https://github.com/apache/incubator-kie-issues/issues/2261

Needed by:
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4220
https://github.com/apache/incubator-kie-kogito-apps/pull/2308
https://github.com/apache/incubator-kie-kogito-examples/pull/2196

** Please read ticket description to understand the motivation behind this PR **

Please consider that the motivation of this PR rely on the unitary view of all our codestack (from drools to examples to kie tools) and not on specific considerations of repositories/modules considered by themselves in isolation.

This PR:

1. move all the `dependencyManagement`(s) tags to the `drools-build-parent` pom
2. remove all the `dependencyManagement`(s) tags from all other poms
3. create a `kie-quarkus-build-parent`specific bom that extend `drools-build-parent` one
4. modify the `drools-quarkus-extension` pom to inherit from `kie-quarkus-build-parent`
5. implement a custom enforcer rule `noDependencyManagementRule` to "forbid" declaration of `dependencyManagement` tag in module, not even inside `profile`(s) (it could be overridden, though: see later)

The `kie-quarkus-build-parent` _extends_ the `drools-build-parent` one, instead of importing it, to also inherit the `plugingManagement`. 
It declares the 

`<allowedPomsList>org.kie:kie-quarkus-build-parent</allowedPomsList>`
property to not fail on the `noDependencyManagementRule` check.

The reason for the `noDependencyManagementRule` is to avoid/limit the habitual approach of adding `dependencyManagement` in downstream/leaf modules.

The problem to solve is that, in our code base, we have multiple versions inconsistency due to the fact that:
1. the same dependency is declared, with different versions, in "parent" and "children" modules
2. the same dependency is inherited transitively on some module but explicit declared in other ones
3. at the very bottom of the stack (i.e. final applications/examples) there are multiple convergence issues due to the above
4. the trickiest part is that in most cases the wrong behavior is not at compile time, where it would be easily spotted, but at runtime, maybe in some combination of artifacts/examples that are not even tested in community

Those points makes our code base frail and hard to maintain (e.g. long time required for frameworks migrations, with overall delay in community major versions releases) for two reasons:

1.  there are situations where it is not clear why there is the version clash
2. whenever there is a request for massive versions upgrade (i.e. quarkus/springboot migration) the modifications have to be done in multiple places and possible failures (when not at compile level) could be hidden

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


kie-pmml-trusty/pom.xml
yesamer: @gitgabrio Do you think we can move those in the main bom?

kie-dmn/kie-dmn-feel/pom.xml
```diff
@@ -41,16 +41,6 @@
     <version.net.sf.saxon.Saxon-HE>12.7</version.net.sf.saxon.Saxon-HE>
```
yesamer: @gitgabrio Is this still required?

drools-test-coverage/drools-test-coverage-jars/drools-test-coverage-jars-with-invoker/pom.xml
```diff
@@ -38,16 +38,6 @@
     <org.kie.version>${project.version}</org.kie.version>
```
yesamer, line 38: @gitgabrio Is this still required?

drools-reliability/drools-reliability-tests/pom.xml
```diff
@@ -38,32 +38,7 @@
     <version.testcontainers>2.0.3</version.testcontainers>
```
yesamer: @gitgabrio Is this still required?

PR #6631, opened 2026-03-31
[incubator-kie-drools-6422] Fix NPE in RuleExecutor.removeDormantTuple

Rikkola:
https://github.com/apache/incubator-kie-drools/issues/6422#issuecomment-4157177207

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/RuleExecutor.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/RuleExecutor.java
@@ -311,7 +311,9 @@ public class RuleExecutor {
                 throw new IllegalStateException();
             }
         }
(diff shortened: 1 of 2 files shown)


mariofusco (approved):

yesamer (approved):

pibizza:
I would suggest reviewing the tests. they seem quite verbose and it really unclear how they relate to the NPE at all. This does not mean that they don't, it means that htey have a lot of details that seem unrelated to the specifics. Why we need for example two different objects in this case? why each object needs three properties? and so on. I am not a superexpert of the code, but I think it would be better to try to simply them and put together the bare minimum that showcases the issue. In which case we can have this NPE? I would suggest reviewing and simplifying the tests before moving on.

PR #6632, opened 2026-04-01
[incubator-kie-issues#2174] Unit test cases for Dialect Handler classes

### LINKED ISSUE #2174: Unit testing for DialectHandlers
Create unit test cases for all the dialect handlers.

ChinchuAjith:
Fixes https://github.com/apache/incubator-kie-issues/issues/2174 

This PR is for adding unit test cases for below dialect handler classes

FEELDialectHandler
BFEELDialectHandler
DefaultDialectHandler
DialectHandlerFactory

### DIFF
--- a/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
+++ b/kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
@@ -1,0 +1,151 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.kie.dmn.feel.lang.ast.dialectHandlers;
+
+import java.math.BigDecimal;
+import java.time.Duration;
+import java.time.LocalDate;
+import java.time.Period;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.kie.dmn.feel.lang.EvaluationContext;
+import org.kie.dmn.feel.lang.types.impl.ComparablePeriod;
+import org.kie.dmn.feel.util.EvaluationContextTestUtil;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for BFEELDialectHandler - BFEEL uses type coercion and returns default values
+ */
+class BFEELDialectHandlerTest {
+
+    private BFEELDialectHandler handler;
+    private EvaluationContext ctx;
+
+    @BeforeEach
+    void setUp() {
+        handler = new BFEELDialectHandler();
+        ctx = EvaluationContextTestUtil.newEmptyEvaluationContext();
+    }
+
+    @Test
+    void testAddOperations() {
+        // String concatenation with type coercion
+        assertThat(handler.executeAdd("Hello", "World", ctx)).isEqualTo("HelloWorld");
+        assertThat(handler.executeAdd("Hello", 123, ctx)).isEqualTo("Hello123");
+        assertThat(handler.executeAdd("Hello", null, ctx)).isEqualTo("Hello");
+        
+        // Numbers
+        assertThat(handler.executeAdd(10, 20, ctx)).isEqualTo(new BigDecimal("30"));
+        assertThat(handler.executeAdd(10, null, ctx)).isEqualTo(10);
+        
+        // Date operations - returns number instead of null
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), 5, ctx)).isEqualTo(5);
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), Duration.ofDays(5), ctx))
+                .isEqualTo(LocalDate.of(2024, 1, 6));
+    }
+
+    @Test
+    void testLogicalOperations() {
+        // AND with boolean coercion
+        assertThat(handler.executeAnd(Boolean.TRUE, Boolean.TRUE, ctx)).isEqualTo(Boolean.TRUE);
+        assertThat(handler.executeAnd(Boolean.TRUE, Boolean.FALSE, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeAnd("test", Boolean.TRUE, ctx)).isEqualTo(Boolean.FALSE); // Non-boolean coerced to false
+        
+        // OR operations
+        assertThat(handler.executeOr(Boolean.TRUE, Boolean.FALSE, ctx)).isEqualTo(Boolean.TRUE);
+        assertThat(handler.executeOr(Boolean.FALSE, Boolean.FALSE, ctx)).isEqualTo(Boolean.FALSE);
+    }
+
+    @Test
+    void testEqualityOperations() {
+        // Equal - returns false for different types (not null)
+        assertThat(handler.executeEqual(null, null, ctx)).isEqualTo(Boolean.TRUE);
+        assertThat(handler.executeEqual(10, 10, ctx)).isEqualTo(Boolean.TRUE);
+        assertThat(handler.executeEqual(10, 20, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeEqual("10", 10, ctx)).isEqualTo(Boolean.FALSE); // Different types
+        assertThat(handler.executeEqual(null, 10, ctx)).isEqualTo(Boolean.FALSE);
+        
+        // Not Equal
+        assertThat(handler.executeNotEqual(null, null, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeNotEqual(10, 20, ctx)).isEqualTo(Boolean.TRUE);
+    }
+
+    @Test
+    void testComparisonOperations() {
+        // BFEEL returns false for invalid comparisons (not null)
+        assertThat(handler.executeGte(Boolean.FALSE, Boolean.FALSE, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeGte("test", Boolean.TRUE, ctx)).isEqualTo(Boolean.TRUE);
+        assertThat(handler.executeGte(20, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeGte("test", 10, ctx)).isEqualTo(Boolean.FALSE); // Incomparable
+        
+        assertThat(handler.executeGt(20, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeLte(5, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeLt(5, 10, ctx)).isEqualTo(Boolean.FALSE);
+    }
+
+    @Test
+    void testArithmeticOperations() {
+        // Power - returns false for invalid
+        assertThat(handler.executePow("2", 3, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executePow(2, 3, ctx)).isEqualTo(new BigDecimal("8"));
+        
+        // Subtraction - returns empty string for invalid
+        assertThat(handler.executeSub("10", 5, ctx)).isEqualTo("");
+        assertThat(handler.executeSub(20, 10, ctx)).isEqualTo(new BigDecimal("10"));
+        assertThat(handler.executeSub(LocalDate.of(2024, 1, 10), Duration.ofDays(5), ctx))
+                .isEqualTo(LocalDate.of(2024, 1, 5));
+        
+        // Multiplication - returns zero for invalid
+        assertThat(handler.executeMult("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeMult(5, 10, ctx)).isEqualTo(new BigDecimal("50"));
+        assertThat(handler.executeMult(Duration.ofDays(5), null, ctx)).isEqualTo(Duration.ZERO);
+        assertThat(handler.executeMult(Period.ofMonths(3), null, ctx)).isEqualTo(ComparablePeriod.ofMonths(0));
+        assertThat(handler.executeMult(Duration.ofDays(5), Duration.ofDays(3), ctx)).isNull();
+        
+        // Division - returns zero for invalid, null for division by zero
+        assertThat(handler.executeDivision("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeDivision(20, 4, ctx)).isEqualTo(new BigDecimal("5"));
+        assertThat(handler.executeDivision(10, Duration.ofDays(5), ctx)).isNull();
+        assertThat(handler.executeDivision(Duration.ofDays(5), null, ctx)).isEqualTo(Duration.ZERO);
+        assertThat(handler.executeDivision(Period.ofMonths(3), null, ctx)).isEqualTo(ComparablePeriod.ofMonths(0));
+        assertThat(handler.executeDivision(null, 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeDivision(10, 0, ctx)).isNull(); // Division by zero
+    }
+
+    @Test
+    void testCompareMethod() {
+        // Returns false for null or incomparable types
+        assertThat(handler.compare(null, null, (l, r) -> l.compareTo(r) > 0)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.compare(20, 10, (l, r) -> l.compareTo(r) > 0)).isTrue();
+        assertThat(handler.compare("test", "test", (l, r) -> l.compareTo(r) == 0)).isTrue();
+        assertThat(handler.compare("test", 10, (l, r) -> l.compareTo(r) > 0)).isEqualTo(Boolean.FALSE);
+    }
+
+    @Test
+    void testBFEELSpecificBehavior() {
+        // Type coercion in logical operations
+        assertThat(handler.executeAnd("notBoolean", "alsoNotBoolean", ctx)).isEqualTo(Boolean.FALSE);
+        
+        // Comparison with non-comparable types
+        assertThat(handler.executeGt(new Object(), new Object(), ctx)).isEqualTo(Boolean.FALSE);
+    }
+}
+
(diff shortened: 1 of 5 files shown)


kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertThat(handler.executeMult("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeMult(5, 10, ctx)).isEqualTo(new BigDecimal("50"));
+        assertThat(handler.executeMult(Duration.ofDays(5), null, ctx)).isEqualTo(Duration.ZERO);
+        assertThat(handler.executeMult(Period.ofMonths(3), null, ctx)).isEqualTo(ComparablePeriod.ofMonths(0));
+        assertThat(handler.executeMult(Duration.ofDays(5), Duration.ofDays(3), ctx)).isNull();
+        
+        // Division - returns zero for invalid, null for division by zero
+        assertThat(handler.executeDivision("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeDivision(20, 4, ctx)).isEqualTo(new BigDecimal("5"));
+        assertThat(handler.executeDivision(10, Duration.ofDays(5), ctx)).isNull();
```
gitgabrio: HI @ChinchuAjith : are you sure of the returned `null`? IIRC, BFeel should always return a default value - I may be wrong in those cases, of course

ChinchuAjith: The operations number ÷ duration and number ÷ period are not defined because they have no semantic meaning and are mathematically undefined. BFEEL returns default values for type coercion failures, but returns null for mathematically/semantically undefined operations.That’s why we explicitly map these cases to null in the implementation. Does this make sense?

gitgabrio: @ChinchuAjith 
I see this in specs:


>  In B-FEEL, null is used only to
represent missing data. All operations and built-in functions that returns null in FEEL when an error occurs have their
semantics modified in B-FEEL to return a non-null value. A warning message should still be produced when an error
occurs.

And also:

image


in which operations with invalid parameters returns anyway a number...

ChinchuAjith: Got it @gitgabrio So based on further analysis, arrived at the following steps
**Step 1: Identify Incompatible Types**
	The engine identifies that the operands for the division operation are a Number and a Duration (either years and months or days and time)
**Step 2: Apply B-FEEL Implicit Coercion**
	In B-FEEL, when performing division and one operand is a number while the other is not, the non-number type is implicitly converted to a number using the B-FEEL number() function logic before the operation proceeds
**Step 3: Translate Duration to Numeric Value**
	The duration is translated into its numeric semantic value as defined in the semantic domain: (The spec defines ( Section: 10.3.2 Semantics defines : 10.3.2.3.8 years and months duration))

Years and Months Duration: The value is expressed as the total number of months it contains (Years×12+Months). (Implemented already in drools -ComparablePeriod.java )
Days and Time Duration: The value is expressed as the total number of seconds it contain (duration.getSeconds())
**Step 4: Execute Number ÷ Number Calculation**
	The operation is then performed as a standard numeric division using the value derived in Step 3.
		Example: If you divide 26 by duration("P2Y"), the duration becomes 24 (months). The calculation is 26 / 24, resulting in approximately 1.0833.
Step 5: Return Default Value for Errors
	If the operation encounters a semantic error—such as division by zero (e.g., dividing by duration("P0M"))—B-FEEL semantics are modified to return 0 everywhere standard FEEL would return null. (Spec reference 11.3 )

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertThat(handler.executeAdd("Hello", "World", ctx)).isEqualTo("HelloWorld");
+        assertThat(handler.executeAdd("Hello", 123, ctx)).isEqualTo("Hello123");
+        assertThat(handler.executeAdd("Hello", null, ctx)).isEqualTo("Hello");
+        
+        // Numbers
+        assertThat(handler.executeAdd(10, 20, ctx)).isEqualTo(new BigDecimal("30"));
+        assertThat(handler.executeAdd(10, null, ctx)).isEqualTo(10);
+        
+        // Date operations - returns number instead of null
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), 5, ctx)).isEqualTo(5);
```
gitgabrio, line 70: @yesamer @ChinchuAjith 

This is very tricky. I think you are right with that, but we need to double check everyone.
My interpretation is:
1. normal FEEL defines valid addiction/subtraction with Table 58
2. for BFEEL, we have to "extend" this table for invalid cases, with the Table at 11.9
3. the order of the rows in that latter take precedence, so if there is a number, everything has to be considered a number.

But, please, double check yourself.

@ChinchuAjith would it be possible to create a test for each row in that latter table , with parameters at both left and right side ?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertThat(handler.executePow(2, 3, ctx)).isEqualTo(new BigDecimal("8"));
+        assertThat(handler.executePow(2, "3", ctx)).isEqualTo(new BigDecimal("8")); // "3" coerced to 3
+        assertThat(handler.executePow(2, "invalid", ctx)).isEqualTo(BigDecimal.ZERO); // Invalid string returns 0
+        assertThat(handler.executePow(null, 3, ctx)).isEqualTo(BigDecimal.ZERO); // Null returns 0
+        assertThat(handler.executePow(2, null, ctx)).isEqualTo(BigDecimal.ZERO); // Null returns 0
+        
+        // Subtraction - returns empty string for invalid
+        assertThat(handler.executeSub("10", 5, ctx)).isEqualTo("");
+        assertThat(handler.executeSub(20, 10, ctx)).isEqualTo(new BigDecimal("10"));
+        assertThat(handler.executeSub(LocalDate.of(2024, 1, 10), Duration.ofDays(5), ctx))
```
gitgabrio, line 120: @ChinchuAjith would it be possible to create a test for each row in the BFEEL table , with parameters at both left and right side ?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertThat(handler.executeGte(20, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeGte("test", 10, ctx)).isEqualTo(Boolean.FALSE); // Incomparable
+        
+        assertThat(handler.executeGt(20, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeLte(5, 10, ctx)).isEqualTo(Boolean.FALSE);
+        assertThat(handler.executeLt(5, 10, ctx)).isEqualTo(Boolean.FALSE);
+    }
+
+    @Test
+    void testArithmeticOperations() {
```
gitgabrio: @ChinchuAjith 
Could you split those methods by actual operation executed, i.e. one for "Pow", one for "Sub", etc ?
It would simplify reading and check about the different cases... Thanks 🙏

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: I understand the concern @gitgabrio Based on the further analysis, this is the implementation we are doing currently

Reordered and corrected the addition operations in getAddOperations() method to comply with BFEEL Section 11.9 specification:
**Key Fix: Number operations now take precedence over string concatenation, following the BFEEL Table 11.9 precedence rules.**
**Precedence:**
1. Date + Number operations → Returns the number
    * BFEEL Section 11.9, when one operand is a number, the non-number value (date) is converted to a number using the number() function
    * In BFEEL, number(date) returns 0 (per Section 11.3)
    * Therefore: 0 + 5 = 5, so we return the number operand
2. Number + Date operations → Returns the number
    * Same logic: 5 + 0 = 5, so we return the number operand
3. Number + null operations → Returns the number
4. null + Number operations → Returns the number
5. String concatenation → Only applies when no number operands are present

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: Done. please check

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: Done . Please review

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: i have arranged the tests in the order of the rows specified in the table. Please check

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -44,192 +44,240 @@ void setUp() {
...
+
+        // multiplication - invalid types
+        assertThat(handler.executeMult("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
         assertThat(handler.executeMult(Duration.ofDays(5), null, ctx)).isEqualTo(Duration.ZERO);
         assertThat(handler.executeMult(Period.ofMonths(3), null, ctx)).isEqualTo(ComparablePeriod.ofMonths(0));
-        
-        // Case A: Same Duration Types - Duration × Duration (same type) → zero duration (B-FEEL default)
+
+        // Duration × Duration (same type) → zero duration (B-FEEL default)
         assertThat(handler.executeMult(Duration.ofDays(5), Duration.ofDays(3), ctx)).isEqualTo(Duration.ZERO);
```
gitgabrio, line 141: Hi @ChinchuAjith 
How do you get to this, please ? Mind you, I'm not saying it is wrong, but I would like to understand where that approach is rooted in specs.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -44,192 +44,240 @@ void setUp() {
...
+                .isEqualTo(ComparablePeriod.ofMonths(12));
         
-        // Duration × Period: Period converted to months (P1Y = 12 months), result is Duration
+        // Duration × Period: Period converted to months, result is Duration
         // PT1S (1 second) × P1Y (12 months) = PT12S
-        assertThat(handler.executeMult(Duration.ofSeconds(1), Period.ofYears(1), ctx)).isEqualTo(Duration.ofSeconds(12));
-        
-        // Division - B-FEEL implicit coercion: converts duration to numeric value before division
-        assertThat(handler.executeDivision("test", 5, ctx)).isEqualTo(BigDecimal.ZERO);
+        assertThat(handler.executeMult(Duration.ofSeconds(1), Period.ofYears(1), ctx))
```
gitgabrio, line 169: @ChinchuAjith 
That seems strange, but specs are not very friendly.
@yesamer

gitgabrio (commented):
Thanks @ChinchuAjith , great stuff.
I still have some doubts, but mostly because of the unfriendliness of specs themselves.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: Hi @gitgabrio Reached in that assumption based on a chain of semantic rules defined across different sections of the DMN 1.6 specification.
1. FEEL Table 60: Duration * Duration is "Not Allowed".
2. FEEL Section 7.3.8: "Not Allowed" operations return null.
3. B-FEEL Section 11.1: Any operation returning null in FEEL due to error must return a non-null default.
4. B-FEEL Section 11.6: The default for durations is 0 months or 0 seconds ( duration(“P0M”) )
Thus, in B-FEEL, multiplying Duration.ofDays(5) by Duration.ofDays(3) results in a duration of 0 seconds (represented in Java as Duration.ZERO), because this operation is "Not allowed" in standard FEEL and is modified in the business dialect to return a non-null default value. WDYT?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: This was wrong. actually i was reversing the precedence. It is treating the Duration (DTD) as the anchor and coercing the Period (YMD) into a number.  But Section 11.10 explicitly places YMD in Row 2 and DTD in Row 3.
So correct Result according to Spec: Duration.ofSeconds(1) * Period.ofYears(1) = Period.ofMonths(12) (or Period.ofYears(1)). Corrected this and committed

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,285 @@
...
+        assertThat(handler.executeAdd("Period:", Period.ofMonths(3), ctx)).isEqualTo("Period:P3M");
+        
+        // Right: other + string
+        assertThat(handler.executeAdd(123, "World", ctx)).isEqualTo("123World");
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), " is the date", ctx)).isEqualTo("2024-01-01 is the date");
+        assertThat(handler.executeAdd(Duration.ofDays(5), " duration", ctx)).isEqualTo("PT120H duration");
+        assertThat(handler.executeAdd(Period.ofMonths(3), " period", ctx)).isEqualTo("P3M period");
+        
+        // Case 2 : NUMBER - If either operand is a number (and neither is string), convert non-number to number
+        // Per BFEEL Section 11.3: number(date) = 0, number(duration) = seconds, number(period) = months
```
gitgabrio, line 67: @ChinchuAjith 
Are you sure this comes from section 11.3 ?

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
```diff
@@ -0,0 +1,285 @@
...
+        
+        // Left: date + number → 0 + number = number
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), 5, ctx)).isEqualTo(5);
+        assertThat(handler.executeAdd(LocalDate.of(2024, 1, 1), new BigDecimal("10.5"), ctx))
+                .isEqualTo(new BigDecimal("10.5"));
+        
+        // Right: number + date → number + 0 = number
+        assertThat(handler.executeAdd(5, LocalDate.of(2024, 1, 1), ctx)).isEqualTo(5);
+        assertThat(handler.executeAdd(new BigDecimal("10.5"), LocalDate.of(2024, 1, 1), ctx))
+                .isEqualTo(new BigDecimal("10.5"));
```
gitgabrio, line 77: IINW there is a missing case of Number + Duration (Maybe I do not see it 😊  )

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
gitgabrio: I see your point, @ChinchuAjith . 
On the other side, IINW, there is a default Duration -> Number conversion

>  // Per BFEEL Section 11.3: number(date) = 0, number(duration) = seconds, number(period) = months

so, I'm still a bit confused, TBH.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: @gitgabrio The DMN specification does not define a signature for the number() function that accepts a date type. Standard FEEL treats it as a semantic error and returns null (find the points below), 
while B-FEEL modifies this result to 0. 

In standard FEEL, the number() function has a restricted domain for its parameters. 
According to Table 73 (Semantics of conversion functions), the valid signatures for number() are:
* number(from, grouping separator, decimal separator) where from must be a string.
* number(from) where from must be a string.
* number(from) where from must be a number.
There is no signature provided in the specification that allows a date type as the from parameter.

Section 10.3.4 (Built-in functions) defines ”Whenever a parameter is outside its domain, the result of the built-in is null".
This is reiterated in Section 10.3.2.13.1, which states that actual parameters must conform to the parameter domains or the result is null.

According to Section 11.3 (Built-in functions returning a number): "Several FEEL built-in functions return a numeric result. In B-FEEL, those functions’ semantics are modified to return 0 everywhere FEEL would return null for them".

This is the maximum we will get from spec :)

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
ChinchuAjith: You were absolutely correct! Added the missing cases : Number + Duration and Number + Period.

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
yesamer: @ChinchuAjith very good explanation!

kie-dmn/kie-dmn-feel/src/test/java/org/kie/dmn/feel/lang/ast/dialectHandlers/BFEELDialectHandlerTest.java
yesamer: @gitgabrio @ChinchuAjith My interpretation:

```
If type(e1) or type(e2) is … e1 * e2
The non-duration type is converted to a number using the number B-FEEL function 
```

In this case, both operands are duration types, so the conversion to number() should not occur.

yesamer (approved):

gitgabrio (approved):
Thanks @ChinchuAjith !

PR #6634, opened 2026-04-03
Fix `spotbug-maven-plugin` usage

### LINKED ISSUE #2279: Fix `spotbug-maven-plugin` usage
Remove the obsolete kie-build-tools dependency from SpotBugs configuration.

Currently, drools defines spotbugs-maven-plugin in 4 submodules: drools-base, drools-compiler, drools-core, and drools-drl-parser.

In build-parent/pom.xml, the plugin used to depend on external org.kie:kie-build-tools:${project.version}. That artifact points to an old non-Apache module and 999-SNAPSHOT is not available anymore.

```
   <groupId>com.github.spotbugs</groupId>
          <artifactId>spotbugs-maven-plugin</artifactId>
          <version>${version.com.github.spotbugs-maven-plugin}</version>
          <dependencies>
            <dependency>
              <groupId>org.kie</groupId>
              <artifactId>kie-build-tools</artifactId>
              <version>${project.version}</version>
            </dependency>
          </dependencies>
```

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/2279

Summary of Changes in This PR

- Removed the kie-build-tools dependency from the SpotBugs configuration. It points to an old external module, and version 999-SNAPSHOT is no longer available, which breaks dependency resolution.
- Removed references to missing SpotBugs exclude filter files. These references pointed to files that no longer exist and to the kie-build-tools JAR.
- Upgraded the spotbugs-maven-plugin.
- Fixed a potential NullPointerException detected by the `spotbugs-maven-plugin`.

As a follow-up, we could agree to enable the `spotbugs-maven-plugin` in this project’s submodules (namely `drools-base`, `drools-compiler`, `drools-core`, and `drools-drl-parser` that already declare it) so that the build fails when SpotBugs detects any issues at compile time.
It appears there was already an intention to enable this behavior in the past, since the following configuration is present in the Drools build-parent POM:
```
    <!-- Set to "true" on every project that has no violations. -->
    <spotbugs.failOnViolation>false</spotbugs.failOnViolation>
```

To enable this in specific submodules, we should:

1. In the target submodule’s pom.xml, set spotbugs.failOnViolation to true under <properties>.
2. Keep the spotbugs-maven-plugin execution bound to the verify phase, with both the spotbugs and check goals enabled:

```
<plugin>
  <groupId>com.github.spotbugs</groupId>
  <artifactId>spotbugs-maven-plugin</artifactId>
  <executions>
    <execution>
      <id>spotbugs</id>
      <phase>verify</phase>
      <goals>
        <goal>spotbugs</goal>
        <goal>check</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -177,8 +177,7 @@
     <enforcer.failOnBannedDependencies>true</enforcer.failOnBannedDependencies>
     <!-- Set to "true" on every project that has no violations. -->
     <spotbugs.failOnViolation>false</spotbugs.failOnViolation>
-    <!-- replaces findbugs-maven-plugin -->
-    <version.com.github.spotbugs-maven-plugin>3.1.8</version.com.github.spotbugs-maven-plugin>
+    <version.com.github.spotbugs-maven-plugin>4.9.8.3</version.com.github.spotbugs-maven-plugin>
 
     <version.maven.min>3.8.1</version.maven.min>
     <!-- property for productisation to know the last released version -->
@@ -1935,19 +1934,11 @@
           <groupId>com.github.spotbugs</groupId>
           <artifactId>spotbugs-maven-plugin</artifactId>
           <version>${version.com.github.spotbugs-maven-plugin}</version>
-          <dependencies>
-            <dependency>
-              <groupId>org.kie</groupId>
-              <artifactId>kie-build-tools</artifactId>
-              <version>${project.version}</version>
-            </dependency>
-          </dependencies>
           <configuration>
             <maxRank>6</maxRank>
             <effort>Max</effort>
             <xmlOutput>true</xmlOutput>
             <failOnError>${spotbugs.failOnViolation}</failOnError>
-            <excludeFilterFile>spotbugs-excludes.xml</excludeFilterFile>
           </configuration>
         </plugin>
         <plugin>
--- a/drools-base/pom.xml
+++ b/drools-base/pom.xml
@@ -214,9 +214,6 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <configuration>
-          <excludeFilterFile>${project.basedir}/src/main/spotbugs/spotbugs-exclude.xml</excludeFilterFile>
-        </configuration>
       </plugin>
 
     </plugins>
--- a/drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
+++ b/drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
@@ -614,7 +614,7 @@ public class KnowledgePackageImpl
             Class<?> typeClass = null;
             try {
                 typeClass = typeDeclaration.getTypeClass();
-                if (typeClass != null || !typeClass.isPrimitive()) {
+                if (typeClass != null && !typeClass.isPrimitive()) {
                     Class<?> cls = getPackageClassLoader().loadClass(typeClass.getName());
                     typeDeclaration.setTypeClass(cls);
                 }
--- a/drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
+++ b/drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
@@ -833,4 +833,3 @@ public class KnowledgePackageImpl
         this.cloningResources.put(key, resource);
     }
 }
-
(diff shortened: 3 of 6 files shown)


tkobayas (approved):

Rikkola (approved):
Worth using.

drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
```diff
@@ -614,7 +614,7 @@ public void wireTypeDeclarations() {
             Class<?> typeClass = null;
             try {
                 typeClass = typeDeclaration.getTypeClass();
-                if (typeClass != null || !typeClass.isPrimitive()) {
+                if (typeClass != null && !typeClass.isPrimitive()) {
```
gitgabrio, line 617: 🤔

build-parent/pom.xml
```diff
@@ -1935,19 +1934,11 @@
...
-              <artifactId>kie-build-tools</artifactId>
-              <version>${project.version}</version>
-            </dependency>
-          </dependencies>
           <configuration>
             <maxRank>6</maxRank>
             <effort>Max</effort>
             <xmlOutput>true</xmlOutput>
             <failOnError>${spotbugs.failOnViolation}</failOnError>
-            <excludeFilterFile>spotbugs-excludes.xml</excludeFilterFile>
```
gitgabrio, line 1950: out of curiosity: do you know why that exclude was there and how it behaved ?

drools-base/pom.xml
```diff
@@ -214,9 +214,6 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <configuration>
```
gitgabrio, line 217: out of curiosity: do you know why that exclude was there and how it behaved ?

drools-compiler/pom.xml
```diff
@@ -185,7 +185,6 @@
           <groupId>com.github.spotbugs</groupId>
           <artifactId>spotbugs-maven-plugin</artifactId>
           <configuration>
-            <excludeFilterFile>${project.basedir}/src/main/spotbugs/spotbugs-exclude.xml</excludeFilterFile>
```
gitgabrio, line 188: Same as above

drools-core/pom.xml
```diff
@@ -225,9 +225,6 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <configuration>
```
gitgabrio, line 228: Same as above

drools-drl/drools-drl-parser/pom.xml
```diff
@@ -105,7 +105,6 @@
                     <groupId>com.github.spotbugs</groupId>
                     <artifactId>spotbugs-maven-plugin</artifactId>
                     <configuration>
-                        <excludeFilterFile>${project.basedir}/src/main/spotbugs/spotbugs-exclude.xml</excludeFilterFile>
```
gitgabrio, line 108: Same as above

build-parent/pom.xml
yesamer: @gitgabrio Yes, you can find that file here --> https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/main/kie-build-tools/src/main/resources, imported through kie-build-tools dependency
It contains a single rule. 
I didn't find any other reference.

drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
yesamer: That is a bug spotted by spotbugs-maven-plugin

gitgabrio: Yes, clearly is a bug.. My doubt is that "in theory" it should throw a NPE (the right-side should be executed only when typeClass is `null`, IINW), but it no-one noticed that until now... anyway, good catch 😄

gitgabrio (approved):

drools-base/src/main/java/org/drools/base/definitions/impl/KnowledgePackageImpl.java
yesamer: @gitgabrio Yes! I think that plugin is quite useful, we should evaluate to enable it some modules

PR #6660, opened 2026-04-10
[incubator-kie-drools-6659] PseudoClockScheduler may retain 1000 canc…

### LINKED ISSUE #6659: PseudoClockScheduler may retain 1000 canceled jobs at most
When an event is deleted, ExpireJob is marked as canceled, but not yet removed from the queue. Canceled jobs are explicitly removed from the queue only when cancelledJob > 1000.

https://github.com/apache/incubator-kie-drools/blob/10.1.x/drools-core/src/main/java/org/drools/core/time/impl/PseudoClockScheduler.java#L117-L123
```
    public synchronized void removeJob(JobHandle jobHandle) {
        jobHandle.cancel();
        jobFactoryManager.removeTimerJobInstance(jobHandle);
        if ( ++cancelledJob > 1000 ) {
            purgeCancelledJob();
        }
    }
...
    private void purgeCancelledJob() {
        Iterator<TimerJobInstance> i = queue.iterator();
        while (i.hasNext()) {
            if (i.next().isCanceled()) {
                i.remove();
            }
        }
        cancelledJob = 0;
    }
```

So the engine may retain 1000 jobs at most. When the event size is large, it could be an issue.
```
  Class Name                                                                            | Objects | Shallow Heap |  Retained Heap
  --------------------------------------------------------------------------------------------------------------------------------
  org.drools.core.time.impl.DefaultTimerJobInstance                                     |   1,000 |       40,000 |  >= 93,025,040
  org.drools.core.common.DefaultEventHandle                                             |   1,000 |      104,000 |  >= 92,909,992
  org.drools.core.time.SelfRemovalJobContext                                            |   1,000 |       24,000 |  >= 92,889,048
  org.drools.core.reteoo.ObjectTypeNode$ExpireJobContext                                |   1,000 |       24,000 |  >= 92,865,040
  --------------------------------------------------------------------------------------------------------------------------------
```

Note: `PseudoClockScheduler.runCallBacksAndIncreaseTimer` may also remove canceled jobs when `peek`, but it doesn't guarantee clean up all canceled jobs, because when it peeks an "active, but not yet firing job", it stops examining.

I think `1000` should be configurable.

tkobayas:
…eled jobs at most

- This PR makes hard-coded `1000` configurable. See the GH issue https://github.com/apache/incubator-kie-drools/issues/6659

Closes https://github.com/apache/incubator-kie-drools/issues/6659

### DIFF
--- a/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/PseudoClockSchedulerRetentionTest.java
+++ b/drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/PseudoClockSchedulerRetentionTest.java
@@ -1,0 +1,163 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.drools.model.codegen.execmodel;
+
+import java.lang.reflect.Field;
+import java.util.PriorityQueue;
+
+import org.drools.core.time.impl.PseudoClockScheduler;
+import org.drools.core.time.impl.TimerJobInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.kie.api.definition.type.Expires;
+import org.kie.api.definition.type.Role;
+import org.kie.api.runtime.KieSession;
+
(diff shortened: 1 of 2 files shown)


drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/PseudoClockSchedulerRetentionTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PseudoClockSchedulerRetentionTest extends BaseModelTest {
+
+    private static final String CANCELLED_JOB_PURGE_THRESHOLD_PROPERTY = "drools.pseudoclock.cancelledJobPurgeThreshold";
+    private static final int CANCELLED_JOB_PURGE_THRESHOLD = 1000;
+
+    @ParameterizedTest
+    @MethodSource("parameters")
```
Rikkola, line 40: Just a question about a problem I have. I am failing to run these parameterizedTests directly from IntelliJ, is there a trick to it?

Rikkola (approved):

mariofusco (approved):

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/PseudoClockSchedulerRetentionTest.java
tkobayas: I can run the test with Intellij without an issue. Btw, when I hit an issue with IntelliJ (which is seemingly project resolution related), I usually run `mvn clean install -Dquickly` in terminal, then "Reload All Maven Projects" in IntelliJ.

PR #6672, opened 2026-04-21
[NO-ISSUE] Mitigate CVE-2025-67721 by update Jackson core to 2.21.1

athirakm94:
Mitigate CVE-2025-67721 by update Jackson core to 2.21.1

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -57,9 +57,9 @@
     <version.commons-logging>1.1.1</version.commons-logging>
     <version.commons-io>2.20.0</version.commons-io>
     <version.common-text>1.14.0</version.common-text>
-    <version.com.fasterxml.jackson>2.19.2</version.com.fasterxml.jackson>
-    <version.com.fasterxml.jackson.databind>2.19.2</version.com.fasterxml.jackson.databind>
-    <version.com.fasterxml.jackson.annotations>2.19.2</version.com.fasterxml.jackson.annotations>
+    <version.com.fasterxml.jackson>2.21.1</version.com.fasterxml.jackson>
+    <version.com.fasterxml.jackson.databind>2.21.1</version.com.fasterxml.jackson.databind>
+    <version.com.fasterxml.jackson.annotations>2.21</version.com.fasterxml.jackson.annotations>
     <version.com.github.victools>4.37.0</version.com.github.victools> <!-- victools should align with Jackson if possible -->
     <version.com.miglayout>3.7.4</version.com.miglayout>
     <version.domino-slf4j-logger>1.0.1</version.domino-slf4j-logger>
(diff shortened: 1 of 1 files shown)


yesamer (approved):

build-parent/pom.xml
```diff
@@ -57,9 +57,9 @@
     <version.commons-logging>1.1.1</version.commons-logging>
     <version.commons-io>2.20.0</version.commons-io>
     <version.common-text>1.14.0</version.common-text>
-    <version.com.fasterxml.jackson>2.19.2</version.com.fasterxml.jackson>
-    <version.com.fasterxml.jackson.databind>2.19.2</version.com.fasterxml.jackson.databind>
-    <version.com.fasterxml.jackson.annotations>2.19.2</version.com.fasterxml.jackson.annotations>
+    <version.com.fasterxml.jackson>2.21.1</version.com.fasterxml.jackson>
```
gitgabrio, line 60: @yesamer @athirakm94 
I was curious to understand the reason of those duplication, and I went through history.
[Originally](https://github.com/apache/incubator-kie-drools/commit/608cdde9a8d9bf0c42551bbcd2201d534fd3ce27), this is how the pom was written:

```xml
<version.com.fasterxml.jackson>2.12.5</version.com.fasterxml.jackson>
    <version.com.fasterxml.jackson.databind>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.databind>
    <version.com.fasterxml.jackson.annotations>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.annotations>
```
so, basically:
1. the actual version should be the same for all
2. jackson specific libraries should be aligned with root one

Later on, a kie-ci commit replaced the variable usage with actual value, and that created this situation.
So, I suggest to come back to original setup, i.e.


```xml
<version.com.fasterxml.jackson>2.21.1</version.com.fasterxml.jackson>
 <version.com.fasterxml.jackson.databind>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.databind>
<version.com.fasterxml.jackson.annotations>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.annotations>
```
at least:
1. we won't have to duplicate the modification
2. we'll be sure that such dependencies are always kept aligned

athirakm94: @gitgabrio @yesamer 
My change was limited to version.com.fasterxml.jackson.annotations because jackson-annotations:2.21.1 is not available, while jackson-databind:2.21.1 is. So I used the closest available version (2.21) as a direct value.

Also, the CVE fix (per Mend) targets Jackson core with version 2.21.1. Since jackson-annotations doesn’t have that version, 2.21 is the nearest compatible option while keeping alignment with the rest.

If we revert to the property-based approach, jackson-annotations may need to stay as a separate property unless all artifacts share the exact same version.

gitgabrio: ok, thanks, misread that

gitgabrio (approved):

PR #6675, opened 2026-04-22
[NO-ISSUE] Fix flaky PersistenceTest by disabling parallel test execution

yesamer:
### Fix flaky PersistenceTest by disabling parallel test execution

**Problem:**
`PersistenceTest.testTraitsSerialization` fails intermittently with:
```
org.apache.openjpa.persistence.ArgumentException: This configuration disallows runtime optimization, 
but the following listed types were not enhanced at build time or at class load time with a javaagent:
org.drools.persistence.info.WorkItemInfo
org.drools.persistence.info.SessionInfo
```

**Root Cause:**
The test uses OpenJPA with `-javaagent` for runtime entity enhancement. With `surefire.forkCount=2`, two test instances run in parallel, creating a race condition:
- The javaagent enhancement is asynchronous
- When tests run concurrently, EntityManagerFactory creation sometimes happens before enhancement completes
- This triggers the "unenhanced classes" error

**Solution:**
Removed `surefire.forkCount=2` property to use Maven's default sequential execution (`forkCount=1`). This ensures the javaagent completes enhancement before any EntityManagerFactory is created.

**Testing:**
The fix can be verified by running the test multiple times:
```bash
mvn clean test -pl drools-traits -Dtest=PersistenceTest -Dsurefire.rerunFailingTestsCount=10
```

### DIFF
--- a/drools-traits/pom.xml
+++ b/drools-traits/pom.xml
@@ -37,6 +37,7 @@
 
   <properties>
     <java.module.name>org.drools.traits</java.module.name>
+
     <!-- These are properties used in the database profiles. Some of them must be initialized
      to be empty so that Maven applies their values via filtering to the resources. -->
     <maven.datasource.classname>org.h2.jdbcx.JdbcDataSource</maven.datasource.classname>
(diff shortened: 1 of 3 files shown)


tiagobento (approved):

gitgabrio (approved):

tkobayas (approved):
Thank you!

PR #4270, opened 2022-03-18
Drools 6875 - Removed use of deprecated methods from assertions

pibizza:
The patch removes a number of methods that are deprecated in assertions and starts to move tests towards assertj.

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>

* for a <b>full downstream build</b> 
  * for <b>jenkins</b> job: please add comment: <b>Jenkins run fdb</b>
  * for <b>github actions</b> job: add the label `run_fdb`

* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #6686, opened 2026-04-28
[incubator-kie-issues#2288] - Springboot 4.0.x upgrade 

### LINKED ISSUE #2288: Springboot 4.0.x upgrade
This issue is to upgrade Spring Boot to a newer supported version, as the current Spring Boot version(3.5.x) used in incubator KIE repositories is nearing its end of support (EoS/EOL) by 30 Jun 2026. The upgrade ensures continued security fixes, stability, and long‑term maintainability.

nrknithin:
Closes: https://github.com/apache/incubator-kie-issues/issues/2288

Related PRs
- Optaplanner: https://github.com/apache/incubator-kie-optaplanner/pull/3221
- Kogito Runtimes: https://github.com/apache/incubator-kie-kogito-runtimes/pull/4270
- Kogito Apps: https://github.com/apache/incubator-kie-kogito-apps/pull/2331
- Kogito Examples: https://github.com/apache/incubator-kie-kogito-examples/pull/2210

### Why
Drools does not depend on Spring Boot, but every downstream KIE repository (kogito-runtimes, kogito-apps, kogito-examples) imports both `drools-build-parent` and `spring-boot-dependencies`. 

This PR is the first step of the Spring Boot 3.5 → 4.0 upgrade for the KIE stack. It updates the versions in `drools-build-parent` so they match what `spring-boot-dependencies:4.0.5` already pins. Once this is in, downstream repositories can move to Spring Boot 4.0.5 without adding per-library overrides.

### Things to call out

- **JUnit `5.13.4` → `6.0.3` (not `5.x` or `1.14.x`).** JUnit 6 changed how the version numbers work. `junit-bom:6.0.3` now pins both `junit-jupiter` and `junit-platform` to `6.0.3`. If `junit-jupiter` and `junit-platform` end up on different major versions (for example Jupiter 6 + Platform 1.x), tests fail at discovery time with `NoSuchMethodError`. Spring Boot 4.0.5 ships JUnit 6, so drools needs to be on JUnit 6 too.
- **Infinispan `15.0` → `15.2`.** Required so the downstream `infinispan-spring-boot-starter` works with Spring Boot 4. Infinispan `15.2.6` ships byte-identical copies of two classes (`OffHeapMemory`, `ThreadCreator`) in both `infinispan-commons` and `infinispan-commons-spi`. This breaks the duplicate-class check in `drools-reliability`. Fixed with two `<ignoreClass>` entries scoped only to `infinispan-commons-spi`. The same file already has a similar entry for `infinispan-objectfilter`, so this follows an existing pattern.
- **`RestObjectMapperSpringTemplate` codegen template.** Spring Boot 4 removed `Jackson2ObjectMapperBuilderCustomizer` and stopped automatically registering `MappingJackson2HttpMessageConverter` when only a Jackson 2 `ObjectMapper` bean is present. The template is updated to register both directly. There is one subtle override on `canWrite(String.class) → false`: kogito's DMN-generated REST resources return an already-serialized JSON `String` body. Without the override, Jackson 2 would wrap that string in quotes (`"\"{…}\""`) and break response parsers. The override lets the built-in `StringHttpMessageConverter` write those `String` bodies as-is.

### How to review

- `build-parent/pom.xml`: each property change is a simple version bump. The JUnit version is the only one that needs careful reading — see the note above.
- `drools-reliability/drools-reliability-{infinispan,tests}/pom.xml`: two new `<ignoreClass>` entries. Check the artifact scope is `infinispan-commons-spi` and that the class names match the duplicate classes Infinispan 15.2 actually ships.
- `drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java`: this template is copied into every kogito Spring Boot application that uses rule units, so the diff has a large blast radius. Read the new `@Bean MappingJackson2HttpMessageConverter` method carefully; the `canWrite(String) → false` override is the part most likely to surprise a reviewer.
- The best end-to-end check is to build the downstream repositories (kogito-runtimes, kogito-apps, kogito-examples) against this branch installed locally. Version-conflict issues only show up when the downstream classpath is actually assembled.

### Verification
`mvn clean install` on this branch passes locally.


build-parent/pom.xml
```diff
@@ -121,15 +122,15 @@
     <version.jakarta.json-api>2.1.3</version.jakarta.json-api>
     <version.org.apache.openjpa>4.0.0</version.org.apache.openjpa>
     <version.org.jpmml.model>1.6.4</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.5.1 intentional, because 1.5.1 evaluators works with 1.5.1 -->
-    <version.org.junit.jupiter>5.13.4</version.org.junit.jupiter>
-    <version.org.junit.platform>1.13.4</version.org.junit.platform> <!-- Keep synchronized with junit-jupiter (middle and minor should be the same) -->
+    <version.org.junit.jupiter>6.0.3</version.org.junit.jupiter>
+    <version.org.junit.platform>6.0.3</version.org.junit.platform> <!-- JUnit 6 unified versioning: platform shares the jupiter version, managed by junit-bom 6.0.3 -->
```
yesamer: @nrknithin In this case, I guess we can use a single entry for both, WDYT?

build-parent/pom.xml
```diff
@@ -1342,6 +1343,13 @@
           </exclusion>
         </exclusions>
       </dependency>
+
+      <!-- CVE fix: use at.yawk.lz4:lz4-java instead of org.lz4:lz4-java -->
+      <dependency>
+        <groupId>at.yawk.lz4</groupId>
+        <artifactId>lz4-java</artifactId>
+        <version>${version.at.yawk.lz4.java}</version>
```
yesamer: @nrknithin IS this required for SPringBoot4 or for a CVE?

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
```diff
@@ -18,43 +18,56 @@
...
 
+    public RestObjectMapper(ObjectMapper objectMapper) {
+        SimpleModule module = new SimpleModule();
+        module.addDeserializer(DataStream.class, new DataStreamDeserializer());
+        module.addDeserializer(DataStore.class, new DataStoreDeserializer());
+        module.addDeserializer(SingletonStore.class, new SingletonStoreDeserializer());
+        objectMapper.registerModule(module);
+    }
+
+    // TODO Jackson 3 migration: SB 4 no longer auto-registers a Jackson 2 HTTP converter; register one
```
yesamer: @nrknithin Is this TODO still needed?

yesamer (commented):
@nrknithin Thank you.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
```diff
@@ -18,43 +18,56 @@
...
 
-import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer;
 import org.springframework.boot.SpringBootConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
-import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
 
+// TODO Jackson 3 migration: drop this shim when the Spring add-ons move to tools.jackson.databind.*.
```
yesamer: @nrknithin Is this TODO still needed?

build-parent/pom.xml
nrknithin: @yesamer  Good catch. Consolidated to a single version.org.junit property — both jupiter and platform reference it now.

build-parent/pom.xml
nrknithin: @yesamer  Good question. On closer look this was a CVE fix that became dead code with the Kafka 4.1.2 bump in this PR — Kafka 4.x's pom already brings at.yawk.lz4:lz4-java 1.10.1 directly (verified org.lz4 is not resolved anywhere in ~/.m2). Removed the exclusion, the standalone dep, and the version.at.yawk.lz4.java property.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
nrknithin: @yesamer Yes — the class is a Jackson 2 shim required while kogito-events-core autowires the Jackson 2 ObjectMapper on both the Quarkus and Spring Boot sides. Quarkus 3.33 still ships Jackson 2 (jackson-bom 2.21.1), so the Jackson 3 migration won't happen this cycle and likely not before the LTS after 3.33. The TODO is the removal trigger for that future cleanup.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
nrknithin: @yesamer Yes — needed for the same reason as the TODO above. Quarkus 3.33 still ships Jackson 2 (jackson-bom 2.21.1), so the trigger won't fire this cycle.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
nrknithin: reframed as a plain comment since TODO reads as "incomplete work" — but this is a future removal trigger, not pending work.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
nrknithin: same as above reframed as a plain comment since TODO reads as "incomplete work" — but this is a future removal trigger, not pending work.

drools-model/drools-model-codegen/src/main/resources/class-templates/ruleunits/RestObjectMapperSpringTemplate.java
yesamer: @nrknithin Thank you, I guess we should keep trace of those. Can you please:
- Open a ticket in drools issues --> https://github.com/apache/incubator-kie-drools/issues (Jackson 3 migration)
- This should contains all the info that will help future maintainers to migrate to Jackson 3
- In the code, please link that issue directly
- Repeat for all the impacted points.

nrknithin: @yesamer Done — opened https://github.com/apache/incubator-kie-drools/issues/6702 with the migration context and the list of affected sites. Each Jackson 2 helper now links the issue directly. Same update is applied in kogito-runtimes [#4270](https://github.com/apache/incubator-kie-kogito-runtimes/pull/4270), kogito-apps [#2331](https://github.com/apache/incubator-kie-kogito-apps/pull/2331), and kogito-examples [#2210](https://github.com/apache/incubator-kie-kogito-examples/pull/2210) so every helper across the stack points at the same tracker.

yesamer (approved):
Thank you @nrknithin

drools-traits/src/main/java/org/drools/traits/core/factmodel/AbstractTraitFactory.java
```diff
@@ -194,6 +198,10 @@ protected static String getKey(Class core, Class trait) {
     }
 
     protected Class<T> buildProxyClass(K core, Class<?> trait) {
+        // The trait class builder factory is static and may have been left in a different
+        // VirtualPropertyMode by an earlier call. Re-sync it to this factory's mode so the
+        // generated proxy class has constructors that match the lookup in cacheConstructor.
+        syncBuilderFactoryWithMode(mode);
```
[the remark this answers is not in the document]
nrknithin: This race is pre-existing — the original setMode had no synchronization at all. This PR added synchronized to the write path to fix a sequential state-pollution case JUnit 6 ordering exposed: TraitTest (parameterized MAP/TRIPLES tests) left the static factory in TRIPLES mode, then TraitTypeGenerationTest ran in MAP mode and got the wrong proxy constructor (the "konst is null" failure). The fix re-syncs the factory to the instance's mode at the top of buildProxyClass, which solves the sequential case. Fully fixing the unsynchronized reads (the concurrent case you flagged) needs a redesign — per-instance builders or full-method locking — out of scope for the SB 4 upgrade. 

@yesamer @gitgabrio do u think we need to create a follow up issue for this?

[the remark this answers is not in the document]
yesamer: @nrknithin I don't know that module, but it looks weird that a SB update impacts it. I mean, that should be a framework agnostic module.

build-parent/pom.xml
```diff
@@ -121,15 +123,15 @@
     <version.jakarta.json-api>2.1.3</version.jakarta.json-api>
     <version.org.apache.openjpa>4.0.0</version.org.apache.openjpa>
     <version.org.jpmml.model>1.6.4</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.5.1 intentional, because 1.5.1 evaluators works with 1.5.1 -->
-    <version.org.junit.jupiter>5.13.4</version.org.junit.jupiter>
-    <version.org.junit.platform>1.13.4</version.org.junit.platform> <!-- Keep synchronized with junit-jupiter (middle and minor should be the same) -->
+    <!-- JUnit 6: single pin for jupiter + platform + vintage via junit-bom. See version.junit above for the legacy JUnit 4 API. -->
```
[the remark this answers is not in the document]
nrknithin: updated the comment

pefernan (approved):

gitgabrio (approved):

PR #6697, opened 2026-05-07
[incubator-kie-drools-6696] DRL10 rejects diamond operators in functi…

### LINKED ISSUE #6696: DRL10 rejects diamond operators in function body
DRL10 currently rejects Java diamond operators in function bodies due to a parser grammar gap; 

```java
   function List<String> testOfGenerics(String a) {
       List<String> results = new ArrayList<>();
       return results;
   }
```
```
Unable to create KieModule, Errors Existed: [Message [id=1, kieBase=KBase.xxxx, level=ERROR, path=snip.dslr, line=11, column=0
   text=no viable alternative at input 'List<String> results = new ArrayList<>'], Message [id=2, kieBase=KBase.xxx, level=ERROR, path=snip, line=11, column=0
   text=mismatched input '=' expecting '('], Message [id=3, kieBase=KBase.xxxx, level=ERROR, path=snip, line=11, column=0
   text=mismatched input '>' expecting {'unit', 'function', 'global', 'declare', 'trait', 'type', 'rule', 'query', 'when', 'then', 'end', 'and', 'or', 'exists', 'not', 'in', 'from', 'collect', 'accumulate', 'acc', 'init', 'action', 'reverse', 'result', 'entry-point', 'eval', 'forall', 'over', 'matches', 'memberOf', 'contains', 'excludes', 'soundslike', 'str', 'after', 'before', 'coincides', 'during', 'includes', 'finishes', 'finishedby', 'meets', 'metby', 'overlaps', 'overlappedby', 'starts', 'startedby', 'window', 'attributes', 'salience', 'enabled', 'no-loop', 'auto-focus', 'lock-on-active', 'activation-group', 'ruleflow-group', 'date-effective', 'date-expires', 'dialect', 'calendars', 'timer', 'duration', 'abstract', 'assert', 'boolean', 'break', 'byte', 'case', 'catch', 'char', 'class', 'const', 'continue', 'default', 'do', 'double', 'else', 'enum', 'extends', 'final', 'finally', 'float', 'for', 'if', 'goto', 'implements', 'import', 'instanceof', 'int', 'interface', 'long', 'native', 'package', 'private', 'protected', 'public', 'return', 'short', 'static', 'strictfp', 'super', 'switch', 'synchronized', 'this', 'throw', 'throws', 'transient', 'try', 'void', 'volatile', 'while', 'module', 'open', 'requires', 'exports', 'opens', 'to', 'uses', 'provides', 'with', 'transitive', 'var', 'yield', 'record', 'sealed', 'permits', 'non-sealed', IDENTIFIER}], Message [id=4, kieBase=KBase.xxx, level=ERROR, path=snip.dslr, line=11, column=0
   text=mismatched input '(' expecting {'unit', 'function', 'global', 'declare', 'trait', 'type', 'rule', 'query', 'when', 'then', 'end', 'and', 'or', 'exists', 'not', 'in', 'from', 'collect', 'accumulate', 'acc', 'init', 'action', 'reverse', 'result', 'entry-point', 'eval', 'forall', 'over', 'matches', 'memberOf', 'contains', 'excludes', 'soundslike', 'str', 'after', 'before', 'coincides', 'during', 'includes', 'finishes', 'finishedby', 'meets', 'metby', 'overlaps', 'overlappedby', 'starts', 'startedby', 'window', 'attributes', 'salience', 'enabled', 'no-loop', 'auto-focus', 'lock-on-active', 'activation-group', 'ruleflow-group', 'date-effective', 'date-expires', 'dialect', 'calendars', 'timer', 'duration', 'abstract', 'assert', 'boolean', 'break', 'byte', 'case', 'catch', 'char', 'class', 'const', 'continue', 'default', 'do', 'double', 'else', 'enum', 'extends', 'final', 'finally', 'float', 'for', 'if', 'goto', 'implements', 'import', 'instanceof', 'int', 'interface', 'long', 'native', 'package', 'private', 'protected', 'public', 'return', 'short', 'static', 'strictfp', 'super', 'switch', 'synchronized', 'this', 'throw', 'throws', 'transient', 'try', 'void', 'volatile', 'while', 'module', 'open', 'requires', 'exports', 'opens', 'to', 'uses', 'provides', 'with', 'transitive', 'var', 'yield', 'record', 'sealed', 'permits', 'non-sealed', IDENTIFIER}], Message [id=5, kieBase=KBase.xxxx, level=ERROR, path=snip.dslr, line=11, column=0
   text=mismatched input ';' expecting '->']]
```

explicit type arguments are a valid workaround until the fix is released.
```java
   function List<String> testOfGenerics(String a) {
       List<String> results = new ArrayList<String>();
       return results;
   }
```

tkobayas:
…on body

Fixes https://github.com/apache/incubator-kie-drools/issues/6696

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -4038,6 +4038,23 @@ class MiscDRLParserTest {
     }
 
     @Test
+    void functionWithDiamondOperatorInLocalVariableDeclaration() {
+        final String text = "function List<String> testOfGenerics(String a) {\n" +
+                "    List<String> results = new ArrayList<>();\n" +
+                "    return results;\n" +
+                "}";
+        PackageDescr packageDescr = parseAndGetPackageDescr(text);
+
+        FunctionDescr function = packageDescr.getFunctions().get(0);
+
+        assertThat(function.getName()).isEqualTo("testOfGenerics");
+        assertThat(function.getReturnType()).isEqualToIgnoringWhitespace("List<String>");
+        assertThat(function.getParameterTypes().get(0)).isEqualTo("String");
+        assertThat(function.getParameterNames().get(0)).isEqualTo("a");
+        assertThat(function.getBody()).isEqualToIgnoringWhitespace("List<String> results = new ArrayList<>(); return results;");
+    }
+
+    @Test
     void lhsPatternAnnotation() {
         final String text = "package org.drools\n" +
                 "rule R1\n" +
--- a/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL10Expressions.g4
+++ b/drools-drl/drools-drl-parser/src/main/antlr4/org/drools/drl/parser/antlr4/DRL10Expressions.g4
@@ -801,14 +801,14 @@ creator
     ;
 
 createdName
-    :	drlIdentifier typeArguments?
-        ( DOT drlIdentifier typeArguments?)*
+    :	drlIdentifier typeArgumentsOrDiamond?
+        ( DOT drlIdentifier typeArgumentsOrDiamond? )*
         |	primitiveType
     ;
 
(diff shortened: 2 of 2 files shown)


yesamer (approved):

daniel-clark-mint:
This was annoying, tested locally and it looks good, thank you!

Rikkola (approved):

PR #6708, opened 2026-05-19
[incubator-kie-drools-6706] Parallel Build of Inline Accumulate rules…

### LINKED ISSUE #6706: Parallel Build of Inline Accumulate rules cause a build error
If the rules have Inline Accumulate and rules are more than 10 rules (= parallel build), it causes a build error
```
[ERROR] Failures: 
[ERROR]   ParallelBuildInlineAccumulateTest.buildExecutableModelWithParallelInlineAccumulates:44->assertBuildDoesNotThrow:58 Build attempt 1 should not throw an exception ==> Unexpected exception thrown: java.lang.RuntimeException: [Message [id=1, level=ERROR, path=src/main/java/org/drools/model/codegen/execmodel/RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97RuleMethods0.java, line=79, column=8716
   text=accumulate1Accumulate20 cannot be resolved to a type], Message [id=2, level=ERROR, path=src/main/java/org/drools/model/codegen/execmodel/RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97RuleMethods0.java, line=0, column=0
   text=Java source of src/main/java/org/drools/model/codegen/execmodel/RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97RuleMethods0.java in error:
package org.drools.model.codegen.execmodel;

import org.drools.modelcompiler.dsl.pattern.D;
import org.drools.model.Index.ConstraintType;
import org.drools.model.codegen.execmodel.domain.Person;
import static org.drools.model.codegen.execmodel.RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97.*;
import static org.drools.model.codegen.execmodel.RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97.*;

public class RulesDE9C0B44B9BC8EB52ECA1B01C4B7DB97RuleMethods0 {
...
```

Workaround: Disable parallel build. `-Ddrools.parallelRulesBuildThreshold=-1`

tkobayas:
… cause a build error

- Fixes https://github.com/apache/incubator-kie-drools/issues/6706

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
 
 import com.github.javaparser.StaticJavaParser;
@@ -154,7 +155,7 @@ public class PackageModel {
 
     private final List<MethodDeclaration> functions = new ArrayList<>();
 
-    private final List<TypeDeclaration> generatedPOJOs = new ArrayList<>();
+    private final List<TypeDeclaration> generatedPOJOs = new CopyOnWriteArrayList<>();
     private final List<GeneratedClassWithPackage> generatedAccumulateClasses = new ArrayList<>();
 
     private final Set<Class<?>> domainClasses = new LinkedHashSet<>();
--- a/drools-model/drools-model-codegen/pom.xml
+++ b/drools-model/drools-model-codegen/pom.xml
@@ -150,4 +150,32 @@
         </dependency>
     </dependencies>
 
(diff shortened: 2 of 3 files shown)


drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/PackageModel.java
```diff
@@ -154,7 +155,7 @@ public class PackageModel {
 
     private final List<MethodDeclaration> functions = new ArrayList<>();
 
-    private final List<TypeDeclaration> generatedPOJOs = new ArrayList<>();
+    private final List<TypeDeclaration> generatedPOJOs = new CopyOnWriteArrayList<>();
```
tkobayas, line 158: During build, helper classes like `accumulate5Accumulate68` are added to this list. With a plain `ArrayList`, the added class may be lost under concurrency and result in a build error.

We may have to fix other lists in `PackageModel`, but so far I can only reproduce this issue, so the fix is only this one.

Rikkola (approved):

yesamer (approved):

PR #6719, opened 2026-05-22
NO-ISSUE Update xercesImpl to 2.12.2

yesamer:
It solves CVE-2022-23437.

Not sure it's still used somewhere.


tkobayas (approved):
Thanks. Yeah, at least, there is no reference to xerces in pom files in drools repo.

PR #6734, opened 2026-05-28
[incubator-kie-drools-6733] Preserve insertion order in ListDataStore

### LINKED ISSUE #6733: ListDataStore loses insertion order due to IdentityHashMap, breaking sliding window replay
**Description:**

`ListDataStore` uses `IdentityHashMap<T, DataHandle>` for internal storage (line 40). `IdentityHashMap` does not guarantee iteration order, so when `subscribe()` replays existing events to a new `DataProcessor` via
`store.values().forEach(...)` (line 97), events are delivered in arbitrary order.

This causes incorrect behavior for CEP sliding windows (`SlidingLengthWindow`, `SlidingTimeWindow`) when events are added to a `DataStore` before a subscriber is attached. The window keeps the "last N" events, but
which N survive depends on the non-deterministic iteration order — producing flaky results.

**Reproduction:**

1. Create a `DataStore<T>` via `DataSource.createStore()`
2. Add N events (N > window size) to the store
3. Build a `RuleUnit` with a sliding length window rule
4. Create a `RuleUnitInstance` and bind the `DataStore` (triggers subscribe() → replay)
5. `fireAllRules()` returns different counts across runs

**Expected:** Events replayed in insertion order, consistent with how they'd behave if inserted after subscription.

**Root cause:** `IdentityHashMap` provides identity-based keying (== instead of .equals()) — which is correct — but sacrifices insertion order. Java has no LinkedIdentityHashMap.

**Possible fix:** Maintain a separate `ArrayList<DataHandle>` for ordered iteration alongside the `IdentityHashMap` for identity-based lookup, or use a `LinkedHashMap` with identity-based key wrappers.

tkobayas:
ListDataStore used IdentityHashMap which does not preserve insertion order, causing subscribe() to replay events in arbitrary order and breaking sliding window CEP behavior. Added an ArrayList<DataHandle> alongside the IdentityHashMap to maintain insertion order for iteration and replay while preserving identity-based lookup semantics.

Fixes: https://github.com/apache/incubator-kie-drools/issues/6733

### DIFF
--- a/drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/datasources/ListDataStore.java
+++ b/drools-ruleunits/drools-ruleunits-impl/src/main/java/org/drools/ruleunits/impl/datasources/ListDataStore.java
@@ -18,8 +18,10 @@
  */
 package org.drools.ruleunits.impl.datasources;
 
+import java.util.ArrayList;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 import org.drools.base.definitions.rule.impl.RuleImpl;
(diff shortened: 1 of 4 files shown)


tkobayas:
> ListDataStore used IdentityHashMap which does not preserve insertion order, causing subscribe() to replay events in arbitrary order and breaking sliding window CEP behavior. 

Note that the existing `CepTest` doesn't hit this issue, because it uses `DirectDataStream` which doesn't store data, so doesn't have this problem.

CEP use cases suppose to use `DataStream` rather than `DataStore`, so this issue may not be a real problem. However, users would still expect "List" named class to preserve insertion order. Right?

Rikkola (approved):

Rikkola:
I would myself assume that the default List keeps the order.

mariofusco (approved):

yesamer:
@tkobayas Can we merge this please?

tkobayas:
yes, merged.

PR #6744, opened 2026-06-03
[CVE][Medium] CVE-2026-45292 opentelemetry-api-1.44.1.jar

### LINKED ISSUE #2335: [CVE][Medium] CVE-2026-45292 opentelemetry-api-1.44.1.jar

josedee:
Closes https://github.com/apache/incubator-kie-issues/issues/2335

### DIFF
--- a/kie-parent/pom.xml
+++ b/kie-parent/pom.xml
@@ -928,12 +928,6 @@
         <artifactId>netty-transport-udt</artifactId>
         <version>${version.io.netty}</version>
       </dependency>
-      <!-- Override to fix vulnerability: https://github.com/open-telemetry/opentelemetry-java/security/advisories/GHSA-rcgg-9c38-7xpx -->
-      <dependency>
-        <groupId>io.opentelemetry</groupId>
-        <artifactId>opentelemetry-api</artifactId>
-        <version>${version.io.opentelemetry-api}</version>
-      </dependency>
       <dependency>
         <groupId>io.opentelemetry.proto</groupId>
         <artifactId>opentelemetry-proto</artifactId>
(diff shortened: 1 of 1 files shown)


yesamer:
Hi @josedee, which dependency is transitively importing that?

josedee:
Hi @yesamer, it is coming from `smallrye-reactive-messaging-api`

gitgabrio (changes requested):
Hi @josedee 
I think we should first upgrade opentelemetry.proto, i.e.

`<version.io.opentelemetry>1.0.0-alpha</version.io.opentelemetry>` -> 

`<version.io.opentelemetry>1.10.0-alpha</version.io.opentelemetry>`

since it is affected by known vulnerabilities -> https://mvnrepository.com/artifact/io.opentelemetry.proto/opentelemetry-proto/1.0.0-alpha

I'm not sure if the newer version also include the fix you are proposing here.

yesamer:
@josedee Which version of smallrye-reactive-messaging-api are we importing please? Quarkus is importing that, right?

josedee:
@yesamer `smallrye-reactive-messaging` version that is currently used is `4.28.0` which mainly comes through Quarkus

josedee:
@gitgabrio Yes we can update `opentelemetry.proto` which has a different vulnerability. But I don't think it fixes the current one. Would you prefer to do it in this PR

gitgabrio:
thanks @josedee 
If the two are completely unrelated, fine for me to tackle in a different PR: as you prefer!

yesamer:
@josedee I would suggest using the same approach we used in the past to manage new declared dependencies for CVE only ---> https://github.com/apache/incubator-kie-drools/pull/6500/changes
I would add a comment saying that Quarkus 3.27.3 is  transitively importing that.

kie-parent/pom.xml
```diff
@@ -1102,6 +1096,13 @@
         <artifactId>vertx-web-client</artifactId>
         <version>${version.io.vertx}</version>
       </dependency>
+      <!-- Quarkus 3.27.3 transitively imports io.opentelemetry:opentelemetry-api:1.44.1 -->
```
gitgabrio: Sorry @josedee , could you pls move that snippet before 
`<groupId>io.opentelemetry.proto</groupId>` ?

@yesamer To simplify management, the pom should be sorted alphabetically, both in the `properties` and in the `dependencyManagment` sections.

Sorting by a different logic makes it extremely hard to manage in long term.

yesamer: @gitgabrio I’m fine with the order; my point is to keep track of the reason we introduced that dependency.

gitgabrio (approved):

yesamer (approved):

PR #6748, opened 2026-06-05
Fix CVE-2024-6763, CVE-2025-11143, CVE-2026-2332, CVE-2025-5115: Upgade to Jetty 12.0.33

ChinchuAjith:
This PR resolves four critical and high-severity CVEs in Jetty by upgrading from version 11.0.24 to 12.0.33.

WireMock 3.13.2 ships a separate artifact — `wiremock-jetty12` — which pulls Jetty 12.0.33 instead of 11.x. Switching to this artifact resolves all four CVEs.

**Changes Done**
   - Added version.org.eclipse.jetty=12.0.33`
   - wiremock : 3.13.0 -> 3.13.2`
   - Added `jetty-bom:12.0.33` in dependencyManagement
   - Added explicit dependency management other Jetty artifacts:**
           
 This explicit management was necessary because wiremock-jetty12:3.13.2 brings in mixed Jetty versions (12.1.7 for core, 12.0.30 for EE10), causing NoSuchMethodError at runtime.

**Replaced wiremock with wiremock-jetty12:**
   - Changed from `<artifactId>wiremock</artifactId>` to `<artifactId>wiremock-jetty12</artifactId>`
   - Added exclusion for `jetty-jakarta-servlet-api` to avoid conflicts

**Impact**
- All Jetty artifacts now resolve to version 12.0.33
- All four CVEs are resolved

### DIFF
--- a/kie-parent/pom.xml
+++ b/kie-parent/pom.xml
@@ -329,7 +329,6 @@
   </properties>
   <dependencyManagement>
     <dependencies>
-      <!-- Import Jetty BOM to ensure consistent Jetty 12.0.33 versions -->
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-bom</artifactId>
@@ -1599,7 +1598,7 @@
         <artifactId>ecj</artifactId>
         <version>${version.org.eclipse.jdt}</version>
       </dependency>
-      <!-- Explicit Jetty 12.0.33 dependency management to prevent version conflicts -->
+      <!-- Explicit set to Jetty 12.0.33 version, overriding version transitively imported by wiremock-jetty12 3.12.2  -->
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-http</artifactId>
--- a/kie-parent/pom.xml
+++ b/kie-parent/pom.xml
@@ -1598,7 +1598,22 @@
         <artifactId>ecj</artifactId>
         <version>${version.org.eclipse.jdt}</version>
       </dependency>
-      <!-- Explicit set to Jetty 12.0.33 version, overriding version transitively imported by wiremock-jetty12 3.12.2  -->
+      <!-- Explicit set to Jetty 12.0.33 version, overriding version transitively imported by wiremock-jetty12 3.13.2  -->
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-alpn-client</artifactId>
+        <version>${version.org.eclipse.jetty}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-alpn-server</artifactId>
+        <version>${version.org.eclipse.jetty}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-client</artifactId>
+        <version>${version.org.eclipse.jetty}</version>
+      </dependency>
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-http</artifactId>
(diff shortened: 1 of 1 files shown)


kie-parent/pom.xml
```diff
@@ -1596,6 +1604,77 @@
         <artifactId>jetty-jakarta-servlet-api</artifactId>
         <version>${version.org.eclipse.jetty.jakarta.servlet.api}</version>
       </dependency>
+      <!-- Explicit Jetty 12.0.33 dependency management to prevent version conflicts -->
```
yesamer: @ChinchuAjith consider you're importing the bom, it's not necessary to declare the below dependencies here, those should be already imported by the above bom. Please check here if all dependencies are declared in the bom:  https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-bom/12.1.10/jetty-bom-12.1.10.pom, but I assume the bom already contains all of them, so this block is not necessary

yesamer (changes requested):

kie-parent/pom.xml
ChinchuAjith: @yesamer The BOM itself is fine, but wiremock-jetty12:3.13.2 is pulling in conflicting Jetty versions (12.1.7 core and 12.0.30 EE10). This mismatch caused a NoSuchMethodError at runtime, so explicit dependency management was needed to force everything to align consistently to 12.0.33. I’m also double‑checking by removing the explicit versions to see if the BOM alone resolves it

yesamer: > I’m also double‑checking by removing the explicit versions to see if the BOM alone resolves it

That is my expectation.

ChinchuAjith: WireMock 3.13.2 ships with Jetty 12.0.30, but transitive dependencies were pulling in incompatible 12.1.8 versions. The explicit entries override WireMock's transitive dependencies to ensure consistent 12.0.33 across all Jetty modules (core, HTTP/2, EE10), maintaining API compatibility while getting latest 12.0.x patches. So these 8 dependencies need to define explicitely.

yesamer: Thank you for your the explanation, @ChinchuAjith. Now I have the full picture of the changes.

kie-parent/pom.xml
```diff
@@ -296,6 +295,7 @@
     <version.org.w3c.dom>2.3.0-jaxb-1.0.6</version.org.w3c.dom>
     <version.org.webjars.bootstrap>5.1.3</version.org.webjars.bootstrap>
     <version.org.webjars.jquery>3.6.0</version.org.webjars.jquery>
+    <version.org.eclipse.jetty>12.0.33</version.org.eclipse.jetty>
```
gitgabrio: hi @ChinchuAjith 
Could you please preserve the alphabetic order of both `properties` and `dependency declarations` ? THanks!

kie-parent/pom.xml
```diff
@@ -329,6 +329,14 @@
   </properties>
   <dependencyManagement>
     <dependencies>
+      <!-- Import Jetty BOM to ensure consistent Jetty 12.0.33 versions -->
```
gitgabrio: hi @ChinchuAjith
Could you please preserve the alphabetic order of both properties and dependency declarations ? THanks!

kie-parent/pom.xml
```diff
@@ -1596,6 +1604,47 @@
         <artifactId>jetty-jakarta-servlet-api</artifactId>
         <version>${version.org.eclipse.jetty.jakarta.servlet.api}</version>
       </dependency>
+      <!-- Explicit Jetty 12.0.33 dependency management to prevent version conflicts -->
```
gitgabrio: hi @ChinchuAjith
Could you please preserve the alphabetic order of both properties and dependency declarations ? THanks!

gitgabrio (changes requested):

kie-parent/pom.xml
ChinchuAjith: @gitgabrio The Jetty version property declaration and its related dependencies have been relocated to the correct section, following alphabetical order. The Jetty BOM import remains the first entry in dependency management, since no other imports are present.

kie-parent/pom.xml
ChinchuAjith: The Jetty version property declaration and its related dependencies have been relocated to the correct section, following alphabetical order. The Jetty BOM import remains the first entry in dependency management, since no other imports are present.

kie-parent/pom.xml
ChinchuAjith: same as above

gitgabrio (approved):

yesamer (approved):

PR #6756, opened 2026-06-10
fix(drools-core): guard subnetwork-not right-delete against phantom-delete NPE

### LINKED ISSUE #2338: NPE in PhreakSubnetworkNotExistsNode.deleteRight for two subnetwork nots in STREAM mode
Evaluating a rule with two subnetwork not nodes — not(and(...)), each with its own correlation memory — in an event-processing (STREAM) session throws a NullPointerException on the right-tuple delete path. Pure mainline: plain PatternDSL, no sequencing or temporal constructs.

Stack trace
```
  java.lang.NullPointerException: Cannot invoke "...TupleList.remove(...)"
      because the return value of "...TupleImpl.getMemory()" is null
    at PhreakSubnetworkNotExistsNode.deleteRight(PhreakSubnetworkNotExistsNode.
  java:92)
    at PhreakSubnetworkNotExistsNode.doSubNetworkNode(...)
    at PhreakNotNode.doNode(...)
```
Root cause

In the right-delete loop, the code dereferences the right tuple's memoryunconditionally:
```
  // don't use matches here, as it may be null, if the LT was also being
  removed.
  rightTuple.getMemory().remove(rightTuple);
```
Under this topology a subnetwork right tuple whose insert was staged then unstaged in the same evaluation cycle reaches the delete having never entered the matches list, so getMemory() is null and .remove(...) NPEs.

The sibling PhreakNotNode.doRightDeletes already guards this exact case ("it may have been staged and never actually added"); the subnetwork variant (PhreakSubnetworkNotExistsNode / RuleNetworkEvaluatorImpl.doSubnetwork2) does not — its delete loop cancels the pending right INSERT (removeInsert) but then falls through and addDeletes it anyway, producing a phantom delete against null memory. This diverges from TupleSetsImpl.addDelete, which returns after an INSERT/DELETE clash.

Conditions (all load-bearing)

Dropping any of these stops the reproduction:

- STREAM (EventProcessingOption.STREAM) — CLOUD mode does not reproduce.
- Two subnetwork not(and(...)) nodes, each with its own correlation memory.
- Event-role facts.

Reproducer

PatternDSL rule, two event types, STREAM mode:
```
  Rule rule = rule("twoSubnetworkNots").build(
          not(and(pattern(eventA1).expr(x -> true), pattern(eventB1).expr(x ->
  true))),
          not(and(pattern(eventA2).expr(x -> true), pattern(eventB2).expr(x ->
  true))),
          execute(() -> { }));

  KieBase kieBase = KieBaseBuilder.createKieBaseFromModel(
          new ModelImpl().addRule(rule), EventProcessingOption.STREAM);

  KieSession ksession = kieBase.newKieSession();
  ksession.insert(new EventA(0)); ksession.fireAllRules();
  ksession.insert(new EventB(1)); ksession.fireAllRules();  // NPE here
```
(Distilled from a randomized sequencing stress harness — MODEL_AFTER reference path, shape 46/seed 42.)

Environment

- Affects drools-core Phreak runtime.

Rikkola:
`PhreakSubnetworkNotExistsNode.deleteRight` removed each delete-staged right
tuple from its memory unconditionally. A subnetwork right tuple whose insert
was staged then unstaged in the same cycle never enters a memory list, so
`rightTuple.getMemory()` is null and the remove NPEs:

```
  java.lang.NullPointerException: Cannot invoke "TupleList.remove(...)"
    because the return value of "TupleImpl.getMemory()" is null
      at PhreakSubnetworkNotExistsNode.deleteRight(PhreakSubnetworkNotExistsNode.java:92)
```

Guard the removal exactly as the sibling beta nodes already do — `PhreakNotNode`, 
`PhreakExistsNode`, `PhreakJoinNode` and `PhreakAccumulateNode` all skip the remove
when `getMemory()` is null ("it may have been staged and never actually added").
This node was the lone right-delete path missing that guard.

Reproduces on mainline with plain PatternDSL — two subnetwork nots in STREAM
mode, no sequencing involved. Adds `SubnetworkNotDeleteRightNpeTest`.

Fixes https://github.com/apache/incubator-kie-issues/issues/2338

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakSubnetworkNotExistsNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakSubnetworkNotExistsNode.java
@@ -89,7 +89,12 @@ public class PhreakSubnetworkNotExistsNode {
 
                 TupleImpl leftTuple = node.getStartTuple(rightTuple);
                 // don't use matches here, as it may be null, if the LT was also being removed.
-                rightTuple.getMemory().remove(rightTuple);
+                // memory can likewise be null: the right tuple may have been staged then unstaged in
+                // the same cycle, never actually added to a memory list.
+                TupleList rightMemory = rightTuple.getMemory();
+                if (rightMemory != null) {
+                    rightMemory.remove(rightTuple);
+                }
 
                 TupleList matches = (TupleList) leftTuple.getContextObject();
                 if (matches != null && matches.isEmpty()) { // matches is null, if LT was deleted too
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/SubnetworkNotDeleteRightNpeTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/SubnetworkNotDeleteRightNpeTest.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


Rikkola:
A bug I ran into while working on sequential.

tkobayas (approved):

yesamer (approved):

PR #6760, opened 2026-06-15
[incubator-kie-issues-2344] Fix NPE evaluating after() under an OR-of-negated-join

### LINKED ISSUE #2344: Model compiler: NPE evaluating temporal after() when an OR branch is a negated join
When an executable-model rule contains a top-level or(...) whose branch is a
  join containing a negation (and(not(x), y)), and a later pattern carries a
  temporal after constraint, the rule throws an NPE during fact evaluation. The
  DRL path is unaffected (its parser expands disjunctions into separate rules
  before compilation); the bug is specific to the model→KiePackage compilation
  path.

  Symptom
```
  org.drools.modelcompiler.constraints.ConstraintEvaluationException:
      Error evaluating constraint '' in  and in more rules
    at LambdaConstraint.isAllowedCachedLeft(LambdaConstraint.java:222)
    at SingleBetaConstraints.isAllowedCachedLeft(...)
    at PhreakJoinNode.doLeftInserts(PhreakJoinNode.java:118)
    ...
  Caused by: java.lang.NullPointerException:
      Cannot invoke "org.drools.core.reteoo.TupleImpl.getIndex()" because
  "entry" is null
    at org.drools.core.reteoo.TupleImpl.get(TupleImpl.java:274)
    at TemporalConstraintEvaluator.getBetaInvocationFactHandles(TemporalConstra
  intEvaluator.java:70)
    at
  TemporalConstraintEvaluator.evaluate(TemporalConstraintEvaluator.java:50)
```
  Minimal reproducer (executable model, STREAM)
```
  // Event types with @Role(EVENT): EvE, EvC, EvD
  Variable<EvE> e1 = declarationOf(EvE.class);
  Variable<EvE> e2 = declarationOf(EvE.class);
  Variable<EvC> c  = declarationOf(EvC.class);
  Variable<EvE> et = declarationOf(EvE.class);
  Variable<EvD> d  = declarationOf(EvD.class);

  Rule r = rule("min-or-negjoin-after").build(
      or(
          pattern(e1).expr("e1", v -> true),
          and(not(pattern(e2).expr("e2", v -> true)),   // OR branch is a
  negated join
              pattern(c).expr("c", v -> true))
      ),
      pattern(et).expr("et", v -> true),
      pattern(d).expr("d", v -> true).expr("d_after_et", et, after()),  //
  downstream temporal beta
      execute(() -> {})
  );
```
  // build STREAM KieBase, insert a few events of each type, fireAllRules() ->
  NPE

  Both ingredients are required (each removed individually makes it pass):
  - a bare or(pattern, pattern) does not trigger it,
  - a lone not(...) branch does not,
  - the negated join without the OR does not,
  - the OR + negated join without the downstream after does not.

  Root cause

  When the LHS or is expanded into per-branch subrules (LogicTransformer), the
  shared trailing elements (here ET and D after ET) are cloned into each
  subrule, and each clone's declarations are independently rebound to its own
  subrule's patterns via Constraint.replaceDeclaration (which reassigns a slot
  in the evaluator's declarations[] array).

  TemporalConstraintEvaluator.clone() passed the original declarations array by
  reference (getDeclarations()) instead of copying it, so all clones shared
  one array. The last subrule's rebinding therefore clobbered the others,
  leaving the narrow branch's D after ET pointing at the wide branch's ET
  pattern (tuple index 3) — an index absent from the narrow branch's 2-deep
  tuple, so TupleImpl.get walks off the parent chain and dereferences a null
  entry.

  (Equal-width OR branches mask the bug, since the shared index stays valid —
  which is why a plain or(pattern, pattern) doesn't reproduce.)

Rikkola:
An executable-model rule with a top-level or(...) whose branch is a negated join (and(not(x), y)) and a later pattern carrying a temporal after() constraint threw an NPE during evaluation:

  ConstraintEvaluationException: Error evaluating constraint '' ... in more rules
  Caused by NullPointerException:
    Cannot invoke "TupleImpl.getIndex()" because "entry" is null

When the LHS OR is expanded into per-branch subrules, the shared trailing elements are cloned and each clone's declarations are rebound to its own subrule's patterns (replaceDeclaration reassigns an array slot). TemporalConstraintEvaluator.clone() shared the declarations array by reference, so one subrule's rebinding clobbered the others, leaving a temporal beta pointing at a sibling branch's pattern tuple index -- an index absent from this branch's shorter tuple, causing the NPE in TupleImpl.get. (DRL is unaffected: its parser expands disjunctions before compilation.)

Copy the array (shallow) so each clone rebinds independently. A deep clone via getClonedDeclarations() regresses the NamedConsequence path, whose clones are not re-bound and rely on the original Declaration objects.

Adds OrNegatedJoinAfterReproTest (minimal repro + two controls). 

Closes apache/incubator-kie-issues#2344

### DIFF
--- a/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/OrNegatedJoinAfterReproTest.java
+++ b/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/OrNegatedJoinAfterReproTest.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.drools.modelcompiler;
+
+import org.drools.model.Model;
+import org.drools.model.Rule;
+import org.drools.model.Variable;
+import org.drools.model.impl.ModelImpl;
+import org.drools.modelcompiler.KieBaseBuilder;
+import org.junit.jupiter.api.Test;
+import org.kie.api.KieBase;
+import org.kie.api.KieServices;
+import org.kie.api.conf.EventProcessingOption;
+import org.kie.api.definition.type.Role;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.KieSessionConfiguration;
+import org.kie.api.runtime.conf.ClockTypeOption;
+import org.kie.api.time.SessionPseudoClock;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.drools.model.DSL.after;
+import static org.drools.model.DSL.declarationOf;
+import static org.drools.model.DSL.execute;
+import static org.drools.model.DSL.not;
+import static org.drools.model.PatternDSL.and;
+import static org.drools.model.PatternDSL.or;
+import static org.drools.model.PatternDSL.pattern;
+import static org.drools.model.PatternDSL.rule;
+
+public class OrNegatedJoinAfterReproTest {
+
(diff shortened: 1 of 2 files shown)


drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/OrNegatedJoinAfterReproTest.java
```diff
@@ -0,0 +1,126 @@
+package org.drools.modelcompiler;
```
yesamer, line 19: @Rikkola Apache header is required.

tkobayas:
Thanks! Copilot suggestions look minor, but the `drools-model/drools-codegen-common/build/**` files seem to be a mistake :)

Rikkola:
This is what happens when you do not let AI do your PRs :---)

tkobayas (approved):

yesamer (approved):

PR #6768, opened 2026-06-19
[incubator-kie-drools-6765] Park no-loop/lock-on-active blocked tuples in dormantMatches

tkobayas:
Minimal fix: Make sure that when a tuple is suppressed by no-loop/lock-on-active, set it to dormantMatches.

### DIFF
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -280,6 +280,7 @@ public class PhreakRuleTerminalNode {
             // Expiration propagations should not be removed from the list, as they still need to fire
             executor.removeActiveTuple( leftTuple );
         } else if ( leftTuple.getStagedType() == Tuple.DELETE && !leftTuple.isQueued() ) {
+            // if Tuple.DELETE, then the tuple should be removed/dereferenced entirely, so not in activeMatches nor dormantMatches
             executor.removeDormantTuple( leftTuple );
         }
 
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -280,7 +280,8 @@ public class PhreakRuleTerminalNode {
             // Expiration propagations should not be removed from the list, as they still need to fire
             executor.removeActiveTuple( leftTuple );
         } else if ( leftTuple.getStagedType() == Tuple.DELETE && !leftTuple.isQueued() ) {
-            // if Tuple.DELETE, then the tuple should be removed/dereferenced entirely, so not in activeMatches nor dormantMatches
+            // if Tuple.DELETE, it's not in activeMatches nor dormantMatches. Basically, unlinked from parent nodes, so will be GC'ed.
+            // Note: there could be a case where such a tuple will be retained and activated again. (e.g., accPropCtx.getResultLeftTuple())
             executor.removeDormantTuple( leftTuple );
         }
 
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
@@ -113,6 +113,7 @@ public class PhreakRuleTerminalNode {
         // most recent PropagationContext is required to maintain the right recency which triggers the match
         PropagationContext pctx = leftTuple.findMostRecentPropagationContext();
         if ( rtnNode.getRule().isNoLoop() && sameRules(rtnNode, pctx.getTerminalNodeOrigin()) ) {
+            executor.addDormantTuple(leftTuple);
             return;
         }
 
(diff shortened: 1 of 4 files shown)


drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
```diff
@@ -113,6 +113,7 @@ public static void doLeftTupleInsert(ReteEvaluator reteEvaluator,
         // most recent PropagationContext is required to maintain the right recency which triggers the match
         PropagationContext pctx = leftTuple.findMostRecentPropagationContext();
         if ( rtnNode.getRule().isNoLoop() && sameRules(rtnNode, pctx.getTerminalNodeOrigin()) ) {
+            executor.addDormantTuple(leftTuple);
```
tkobayas, line 116: When the tuple is suppressed because of `no-loop`, the tuple was not belong to activeMatches nor dormantMatches. This is how it became orphan in the NPE case.

Explicitly add it to dormantMatches.

drools-core/src/main/java/org/drools/core/phreak/PhreakRuleTerminalNode.java
```diff
@@ -127,6 +128,7 @@ public static void doLeftTupleInsert(ReteEvaluator reteEvaluator,
             InternalAgendaGroup agendaGroup = executor.getRuleAgendaItem().getAgendaGroup();
             if (blockedByLockOnActive(rtnNode.getRule(), pctx, agendaGroup)) {
                 activationsManager.getAgendaEventSupport().fireActivationCancelled(leftTuple, reteEvaluator, MatchCancelledCause.FILTER );
+                executor.addDormantTuple(leftTuple);
```
tkobayas, line 131: The same for `lock-on-active`.

drools-core/src/main/java/org/drools/core/phreak/RuleExecutor.java
```diff
@@ -314,7 +314,7 @@ public void removeDormantTuple(RuleTerminalNodeLeftTuple tuple) {
         if (tuple.getPrevious() != null || dormantMatches.getFirst() == tuple) {
             dormantMatches.remove(tuple);
         } else {
-            log.debug("Skipping removeDormantTuple for orphan tuple {} on rule \"{}\" — tuple is not present in dormantMatches",
+            log.warn("Skipping removeDormantTuple for orphan tuple {} on rule \"{}\" — tuple is not present in dormantMatches",
```
tkobayas, line 317: Raised the log level because we should have fixed all known paths that could lead to orphaned tuples.

drools-core/src/main/java/org/drools/core/phreak/RuleExecutor.java
```diff
@@ -345,7 +345,8 @@ public void removeActiveTuple(RuleTerminalNodeLeftTuple tuple) {
         activeMatches.remove(tuple);
         if (tuple.getStagedType() != Tuple.DELETE) {
             addDormantTuple(tuple);
-        }
+        } // if Tuple.DELETE, it's not in activeMatches nor dormantMatches. Basically, unlinked from parent nodes, so will be GC'ed.
+          // Note: there could be a case where such a tuple will be retained and activated again. (e.g., accPropCtx.getResultLeftTuple())
```
tkobayas, line 349: Previously we were looking at this path as the "orphan path", but actually, this is intentional -- the tuple is unlinked, so no longer referenced by parent node's memory. So the tuple should not belong to activeMatches nor dormantMatches. Finally, will be GC'ed. So we could say it's "REMOVED" state, rather than "ORPHANED".

However, as noted, there is a case where such a tuple is retained and activated again. So, it's not easy to frame the lifecycle.

mariofusco (approved):
I prefer this solution, it is much less invasive, it is easier to validate and seems to address the problem correctly.

tkobayas:
Note that this fix is to make the tuple in the correct state.

Actually, even without this fix, the rule execution works correctly by the NPE guard implemented by https://github.com/apache/incubator-kie-drools/pull/6707.

---

UPDATE path (orphan receives UPDATE from a different rule):

With the guard:
1. modifyActiveTuple → guarded removeDormantTuple detects orphan → skips remove
2. addActiveTuple → tuple enters active → fires

With this PR:
1. modifyActiveTuple → removeDormantTuple succeeds (tuple is in dormant, not orphan)
2. addActiveTuple → tuple enters active → fires

Same result. The tuple correctly fires because the no-loop condition no longer applies (different rule triggered the UPDATE).

DELETE path (orphan receives DELETE):

With the guard:
1. doLeftDelete → guarded removeDormantTuple detects orphan → skips remove
2. setContextObject(null) → tuple cleaned up

With this PR:
1. doLeftDelete → removeDormantTuple succeeds (tuple is in dormant, not orphan)
2. setContextObject(null) → tuple cleaned up

Same result.

So functionally the NPE guard is sufficient. The advantage of this fix is:
- No log noise on every orphan encounter
- The tuple's list membership is always consistent — no "not in any list but still alive" state
- Doesn't mask a genuine bug where a tuple was incorrectly removed from dormant

yesamer (approved):
Thank you @tkobayas

PR #6769, opened 2026-06-22
[DO-NOT-MERGE] [incubator-kie-drools-6765] Fix removeDormantTuple NPE with explicit MatchState

tkobayas:
Enhanced version of https://github.com/apache/incubator-kie-drools/pull/6767
---

The root cause of the NPE in RuleExecutor.removeDormantTuple is that a
tuple's list membership (active, dormant, or neither) was inferred from
indirect signals (memory, isQueued, previous) rather than tracked
explicitly. DORMANT and orphan (never added to any list) were
indistinguishable, so removeDormantTuple was called on orphan tuples
that were never in dormantMatches.

Add MatchState enum (NONE, ACTIVE, DORMANT, REMOVED) to
RuleTerminalNodeLeftTuple and consolidate all state transitions into
three self-contained methods on RuleExecutor:

- transitionToActive: NONE/DORMANT -> ACTIVE (replaces addActiveTuple
  and modifyActiveTuple). Validates from-state, removes from dormant
  list if needed, adds to active list and salience queue.
- transitionToDormant: NONE/ACTIVE -> DORMANT (replaces addDormantTuple).
  Validates from-state, removes from active list if needed, adds to
  dormant list.
- deactivateTuple: any -> DORMANT or REMOVED (replaces removeActiveTuple
  and removeTuple). Handles all states: ACTIVE uses stagedType to decide
  DORMANT (park for expiration) vs REMOVED; DORMANT/NONE -> REMOVED.
  Fixes the NPE by safely handling orphan tuples (NONE state).

Each method encapsulates state validation, list manipulation, and
MatchState update in one place. Callers no longer need to check
MatchState before calling — eliminating the repeated if/else pattern
in PhreakRuleTerminalNode.doLeftDelete and the redundant pre-checks
in PhreakBranchNode.

Also fix RuleTerminalNode.cancelMatch which bypassed RuleExecutor by
directly removing from activeMatches via getMemory().remove().

### DIFF
--- a/drools-core/src/main/java/org/drools/core/common/ActivationsManager.java
+++ b/drools-core/src/main/java/org/drools/core/common/ActivationsManager.java
@@ -93,7 +93,7 @@ public interface ActivationsManager {
         if (!ruleAgendaItem.isQueued()) {
             ruleAgendaItem.getRuleExecutor().getPathMemory().queueRuleAgendaItem();
         }
-        ruleAgendaItem.getRuleExecutor().modifyActiveTuple((RuleTerminalNodeLeftTuple) justified.getTuple() );
+        ruleAgendaItem.getRuleExecutor().transitionToActive((RuleTerminalNodeLeftTuple) justified.getTuple() );
     }
 
     default ActivationsManager getPartitionedAgenda(int partitionNr) {
--- a/drools-core/src/main/java/org/drools/core/phreak/PhreakBranchNode.java
+++ b/drools-core/src/main/java/org/drools/core/phreak/PhreakBranchNode.java
@@ -142,9 +142,6 @@ public class PhreakBranchNode {
             if (oldRtn != null) {
                 if (newRtn == null) {
                     // old exits, new does not, so delete
-                    if ( branchTuples.rtnLeftTuple.getMemory() != null ) {
-                        executor.removeActiveTuple(branchTuples.rtnLeftTuple);
-                    }
                     PhreakRuleTerminalNode.doLeftDelete(activationsManager, executor, branchTuples.rtnLeftTuple);
 
                 } else if (newRtn == oldRtn) {
@@ -153,9 +150,6 @@ public class PhreakBranchNode {
 
                 } else {
                     // old and new on different branches, delete one and insert the other
-                    if ( branchTuples.rtnLeftTuple.getMemory() != null ) {
-                        executor.removeActiveTuple(branchTuples.rtnLeftTuple);
-                    }
                     PhreakRuleTerminalNode.doLeftDelete(activationsManager, executor, branchTuples.rtnLeftTuple);
 
                     branchTuples.rtnLeftTuple = (RuleTerminalNodeLeftTuple) TupleFactory.createLeftTuple(leftTuple,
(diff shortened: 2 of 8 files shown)


tkobayas:
```
[ERROR]   AccumulateTest.testAccumulateMultipleFunctionsConstraint:1073 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   AccumulateTest.testAccumulateMultipleFunctionsConstraint:1073 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   AccumulateTest.testAccumulateWithFromChaining:784 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   AccumulateTest.testAccumulateWithFromChaining:784 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testSimpleLengthWindow:1029 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testSimpleLengthWindow:1029 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testSimpleTimeWindow:973 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testSimpleTimeWindow:973 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testTimeAndLengthWindowConflict:2171 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   CepEspTest.testTimeAndLengthWindowConflict:2171 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   IncrementalCompilationTest.testUnchangedAccumulate:3723 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   NamedConsequencesTest.testDynamicSalience:788 » IllegalState Cannot transition to ACTIVE from REMOVED
[ERROR]   NamedConsequencesTest.testDynamicSalience:788 » IllegalState Cannot transition to ACTIVE from REMOVED
```
  Unlinking happens UPSTREAM in the beta node, BEFORE the terminal node processes the DELETE. The beta node (PhreakJoinNode, PhreakNotNode, etc.) calls unlinkAndDeleteChildLeftTuple() which does
  unlinkFromRightParent() + unlinkFromLeftParent() + stages as DELETE. By the time PhreakRuleTerminalNode.doLeftDeletes runs, the tuple is already disconnected from its parents.

  So in the normal DELETE flow, the REMOVED tuple IS dereferenced — the parent no longer holds it. This means the REMOVED→ACTIVE transition should NOT happen through the normal parent-child path.

  But there's an important exception: doLeftDelete can be called WITHOUT the upstream unlink. This happens in:
  - RuleTerminalNode.cancelMatch — calls doLeftDelete directly, no upstream beta node involved
  - PhreakBranchNode — calls doLeftDelete directly on branch tuples
  - ActivationsManager.stageLeftTuple → transitionToActive via TMS unblocking — the blocked tuple was never unlinked from its parent

  In these paths, the tuple goes to REMOVED (or DORMANT) but stays linked to its parent. A subsequent network re-evaluation at the parent node can find it and propagate an UPDATE, leading to REMOVED→ACTIVE.

  So the answer is: REMOVED tuples from the normal DELETE propagation are fully unlinked and will be GC'd. REMOVED tuples from cancelMatch, branch node, or TMS paths remain linked to their parents and can 
  legitimately be re-activated. That's why the failing tests (accumulate, CEP windows, dynamic salience) hit REMOVED→ACTIVE — they use these non-standard deletion paths.

---
Not yet clarifying the state transition use cases...

mariofusco (changes requested):
Under a pure correctness point of view this solution is probably better than the other, or at least it is more "sounding" to me, but I agree that this is more invasive, and I'm afraid that the introduction of the additional `MatchState` could also have a non-trivial performance impact, so I would vote for the other one.

tkobayas:
Closing in favor of https://github.com/apache/incubator-kie-drools/pull/6768

PR #6776, opened 2026-06-25
[Incubator kie issues#2188] Fix the compilation failure on DecisionService–Decision output type mismatch

### LINKED ISSUE #2188: Fix the compilation failure on DecisionService–Decision output type mismatch
Currently, compilation fails if the output type of a DecisionService and its contained Decision differ, particularly when the DecisionService is defined as a FunctionItem. This needs to be addressed. 

Example:
DecisionService typeRef: dateList
Decision output typeRef: date
Result: Compilation error due to mismatch.
**Return type typeRef mismatch while compiling node 'To Singleton List DS'; FunctionItem defines typeRef 'dateList', the defined decision logic defines typeRef 'date'. (DMN id: _BF3F2933-A588-41F1-AC98-492A79088A53, The typeRef does not match**

This is not a fundamental type mismatch but a collection difference (single value vs list). 
If the DecisionService expects a collection and the Decision provides a single item (or vice versa), coercion should be applied where possible.
The coercion logic should be applied during evaluation to handle such cases consistently.

[1157-implicit-conversions.txt](https://github.com/user-attachments/files/24082576/1157-implicit-conversions.txt)

AthiraHari77:
Closes : https://github.com/apache/incubator-kie-issues/issues/2188


yesamer:
@AthiraHari77 Is this ready to be reviewed? Can you please provide Unit Tests?

jomarko:
Does this PR handle also a situation when the decision service output data type is structure? Like **Person(name, age)**, then decision service upper part may contain two decision nodes: **name** and **age** and it should be valid scenario.

AthiraHari77:
> Does this PR handle also a situation when the decision service output data type is structure? Like **Person(name, age)**, then decision service upper part may contain two decision nodes: **name** and **age** and it should be valid scenario.

@jomarko The DMN v1.6 specification explicitly defines implicit conversions only for primitive and list types (singleton list wrapping/unwrapping, date → dateTime). It does not explicitly specify any additional implicit conversion semantics for custom or structured types such as Person(name, age).

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDecisionServiceEvaluator.java
```diff
@@ -73,7 +75,14 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult r
         for (String id : decisionIDs) {
             DMNDecisionResult decisionResultById = evaluateById.getDecisionResultById(id);
             String decisionName = dmnModel.getDecisionById(id).getName();
-            ctx.put(decisionName, decisionResultById.getResult());
+
+            DMNType expectedType = dsNode.getResultType();
+            Object originalResult = decisionResultById.getResult();
+            Object coercedResult = CoerceUtil.coerceValue(expectedType, originalResult);
+            if (coercedResult != originalResult) {
```
gitgabrio, line 81: HI @AthiraHari77 
Please check for `null` here. both `originalResult`and `courcedResult` could be `null`, IINW

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDecisionServiceEvaluator.java
```diff
@@ -84,8 +93,9 @@ public EvaluatorResult evaluate(DMNRuntimeEventManager eventManager, DMNResult r
             }
         }
         boolean typeCheck = ((DMNRuntimeImpl) eventManager.getRuntime()).performRuntimeTypeCheck(result.getModel());
+        Object finalResult = decisionIDs.size() == 1 ? ctx.values().iterator().next() : ctx;
         if (typeCheck) {
-            Object c = DMNRuntimeUtils.coerceUsingType(decisionIDs.size() == 1 ? ctx.values().iterator().next() : ctx,
+            Object c = DMNRuntimeUtils.coerceUsingType(finalResult,
```
yesamer: @AthiraHari77 Why this change, please?

AthiraHari77: The extraction was kept for better readability.

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDecisionServiceEvaluator.java
```diff
@@ -22,14 +22,15 @@
...
-import java.util.stream.Collectors;
 
 import org.kie.dmn.api.core.DMNDecisionResult;
 import org.kie.dmn.api.core.DMNMessage;
 import org.kie.dmn.api.core.DMNMessage.Severity;
 import org.kie.dmn.api.core.DMNModel;
 import org.kie.dmn.api.core.DMNResult;
 import org.kie.dmn.api.core.DMNRuntime;
+import org.kie.dmn.api.core.DMNType;
+import org.kie.dmn.api.core.ast.DecisionNode;
```
gitgabrio: HI @AthiraHari77 IINW, this import is unneeded

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDecisionServiceEvaluator.java
gitgabrio: @AthiraHari77 
I know that code was already there, but the name of this method:
`DMNRuntimeUtils.coerceUsingType` is unclear and, keeping in mind we also have `CoerceUtil.coerceValue` , the overall result is pretty confusing.
`DMNRuntimeUtils.coerceUsingType` actually deals only with `Collections`: could you please verify that and, eventually, rename it to make that clearer ? Just to reduce a bit tech debts while we found them. Thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
```diff
@@ -216,115 +219,166 @@ private void reportReferenceError(DecisionServiceNodeImpl ni, DMNModelImpl model
...
-                                      fiComposite,
-                                      fdComposite);
+                                      decisionServiceNode.getName(),
+                                      functionCompositeType,
+                                      decisionServiceCompositeType);
             }
         }
     }
 
+    static boolean isReturnTypeCollectionCompatible(QName functionReturnTypeRef,  QName outputDecisionReturnTypeRef, DMNModelImpl model) {
```
gitgabrio: Hi @AthiraHari77 
Those two static methods IMO should go to some common utility class, since there behavior is very "generic" and probably already needed elsewhere.
Maybe inside `DMNRuntimeUtils` ? Or maybe there is already some "util" class that deals with similar implementations ?

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
```diff
@@ -216,115 +219,166 @@ private void reportReferenceError(DecisionServiceNodeImpl ni, DMNModelImpl model
...
+                        null,
+                        null,
+                        builtInType
+                );
+            }
+        }
+        return type;
+    }
+
+    static BuiltInType resolveBuiltInType(String name) {
```
gitgabrio: See comment above.  I've the impression that this functionality is already implemented somewhere, but I can't recall where 🤔  (I may be wrong, of course)

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DecisionServiceCompiler.java
```diff
@@ -216,115 +219,166 @@ private void reportReferenceError(DecisionServiceNodeImpl ni, DMNModelImpl model
                 ni.getName());
     }
 
-    private void checkFnConsistency(DMNModelImpl model, DecisionServiceNodeImpl ni, DMNType type, List<DecisionNode> outputDecisions) {
-        SimpleFnTypeImpl fnType = ((SimpleFnTypeImpl) type);
-        FunctionItem fi = fnType.getFunctionItem();
-        if (fi.getParameters().size() != ni.getInputParameters().size()) {
+    private void validateDecisionServiceFunctionSignature(DMNModelImpl model, DecisionServiceNodeImpl decisionServiceNode, DMNType decisionServiceFunctionType, List<DecisionNode> outputDecisions) {
```
gitgabrio, line 220: 👍  🙏 ❤️

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+    @Test
+    void resolveBuiltInType_boolean() {
+        assertThat(DecisionServiceCompiler.resolveBuiltInType("boolean")).isEqualTo(BuiltInType.BOOLEAN);
+    }
+
+
+    @Test
+    void resolveDMNTypeString() {
+
+        DMNModelImpl model = mock(DMNModelImpl.class);
```
gitgabrio: @AthiraHari77 
The need of those mocks (that actually creates a "fake" implementation) is related to that implementation -> 


```java
    static DMNType resolveDMNType(QName typeRef, DMNModelImpl model) {
        DMNType type = model.getTypeRegistry().resolveType(model.getNamespace(), typeRef.getLocalPart());
        if (type == null) {
            BuiltInType builtInType = resolveBuiltInType(typeRef.getLocalPart());
            if (builtInType != null) {
                boolean isCollection = (builtInType == BuiltInType.LIST);
                type = new SimpleTypeImpl(
                        model.getNamespace(),
                        builtInType.getName(),
                        null,
                        isCollection,
                        null,
                        null,
                        null,
                        builtInType
                );
            }
        }
        return type;
    }
 ```
What this test is about is the `type==null` branch of execution but, since it is embedded inside the `DMNType resolveDMNTYpe`, hence the need of (slightly contrived) mocking.
Extracting the content of `type==null`  branch to a specific method, would reduce the "cyclomatic complexity" of the method ( 😄  ) and simplify testing it, removing the need of "mocks".
(In general terms, lot of time mocks are always a smell of something not perfectly encapsulated).
Am I clear ? Does this make sense ?

File: kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java [hunk 8]
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+        QName qName = new QName("", "string");
+
+        DMNType result = DecisionServiceCompiler.resolveDMNType(qName, model);
+        assertThat(result).isInstanceOf(SimpleTypeImpl.class);
+        assertThat(((SimpleTypeImpl) result).getFeelType()).isEqualTo(BuiltInType.STRING);
+    }
+
+    @Test
+    void resolveDMNTypeDateAndTime() {
+        DMNTypeRegistry registry = mock(DMNTypeRegistry.class);
```
gitgabrio: Same comment as above

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+
+        DMNType result = DecisionServiceCompiler.resolveDMNType(new QName("", "date and time"), model);
+        assertThat(result).isInstanceOf(SimpleTypeImpl.class);
+        assertThat(((SimpleTypeImpl) result).getFeelType()).isEqualTo(BuiltInType.DATE_TIME);
+        assertThat(result.isCollection()).isFalse();
+    }
+
+    @Test
+    void resolveDMNType_listBuiltInIsMarkedAsCollection() {
+        DMNTypeRegistry registry = mock(DMNTypeRegistry.class);
```
gitgabrio: Same comment as above

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+
+        DMNType result = DecisionServiceCompiler.resolveDMNType(new QName("", "list"), model);
+        assertThat(result).isInstanceOf(SimpleTypeImpl.class);
+        assertThat(((SimpleTypeImpl) result).getFeelType()).isEqualTo(BuiltInType.LIST);
+        assertThat(result.isCollection()).isTrue();
+    }
+
+    @Test
+    void resolveDMNType_unknownNameReturnsNull() {
+        DMNTypeRegistry registry = mock(DMNTypeRegistry.class);
```
gitgabrio: Same comment as above

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+
+        assertThat(DecisionServiceCompiler.isReturnTypeCollectionCompatible(
+                new QName(ns, "tDateList"), new QName(ns, "date"), model)).isTrue();
+    }
+
+
+    // fi="date and time", fd="date" — date-to-dateTime coercion pattern → true
+    @Test
+    void isReturnTypeCollectionCompatible_dateTimeFi_dateFd_compatible() {
+        DMNTypeRegistry registry = mock(DMNTypeRegistry.class);
```
gitgabrio: Same comment as above

kie-dmn/kie-dmn-core/src/test/java/org/kie/dmn/core/compiler/DecisionServiceCompilerTest.java
```diff
@@ -144,4 +148,179 @@ void inputQualifiedNamePrefixWithEmptyImportAlias() {
...
+        when(model.getTypeRegistry()).thenReturn(registry);
+
+        assertThat(DecisionServiceCompiler.isReturnTypeCollectionCompatible(
+                new QName("", "date and time"), new QName("", "date"), model)).isTrue();
+    }
+
+    // fi="date", fd="date and time" — order matters → false
+    @Test
+    void isReturnTypeCollectionCompatible_dateFi_dateTimeFd_incompatible() {
+        DMNTypeRegistry registry = mock(DMNTypeRegistry.class);
```
gitgabrio: Same comment as above

kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_6/DecisionService/DecisionServiceImplicitConversions.dmn
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
```
gitgabrio, line 3: HI @AthiraHari77 
Plkease, do not add new dmn models outside `kie-dmn-test-resources`.
Our approach is to have all testing DMN inside `kie-dmn-test-resources` and then refer to them properly.
Thanks!

gitgabrio (changes requested):
HI @AthiraHari77 
Thanks for that PR.
I see only simple refactoring/moving needed, and then it would be perfect: thanks!

kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/ast/DMNDecisionServiceEvaluator.java
yesamer: ok, but you changed the scope of the finalResult variable, previously it was managed inside the if, now is outside, is that necessary?

yesamer (approved):

gitgabrio (approved):
Good stuff, @AthiraHari77 , thanks!

PR #6778, opened 2026-06-26
[DO-NOT-MERGE][incubator-kie-drools-6775]CI build macos java17 failure in MavenDeployTest 

tkobayas:
- https://github.com/apache/incubator-kie-drools/issues/6775

### DIFF
--- a/drools-core/src/main/java/org/drools/core/time/impl/KieCronExpression.java
+++ b/drools-core/src/main/java/org/drools/core/time/impl/KieCronExpression.java
@@ -189,6 +189,12 @@ import java.util.TreeSet;
  */
 public class KieCronExpression implements Serializable {
 
+    static {
+        System.err.println("[DIAG] KieCronExpression loaded: java.version=" + System.getProperty("java.version")
+                + ", java.home=" + System.getProperty("java.home")
+                + ", drools.dialect.java.compiler.lnglevel=" + System.getProperty("drools.dialect.java.compiler.lnglevel"));
+    }
+
     private static final long serialVersionUID = 510L;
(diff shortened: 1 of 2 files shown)


tkobayas:
Note:
---
● Yes, look at line 21 of CiComputeBuildScopes.java:
     
  //JAVA 21
     
  This JBang directive forces JDK 21. Here's the chain:
     
  1. JBang sees //JAVA 21 → downloads and installs JDK 21 → runs the script with JDK 21
  2. The script spawns mvn validate via ProcessBuilder (line 202) which inherits JBang's environment — including JAVA_HOME pointing to JDK 21
  3. So mvn validate runs across the entire project with JDK 21, before the real JDK 17 build starts

  The ProcessBuilder at line 205 only explicitly sets MAVEN_OPTS — it does NOT pin JAVA_HOME back to the CI's Java 17. So the spawned mvn process inherits JBang's JDK 21 environment.

  The non-reproduction after your changes could be because:
  - The issue is intermittent (timing/cache-dependent)
  - Or the different set of changed files changes the affected module set enough to avoid the trigger

  Quickest way to confirm: change //JAVA 21 to //JAVA 17 in CiComputeBuildScopes.java and push with only the original 2-file change. If the failure disappears, that's the root cause. The permanent fix would
  either be:
  - Downgrade the JBang directive to //JAVA 17 (the script uses nothing Java 21-specific)
  - Or have the ProcessBuilder explicitly set JAVA_HOME to the CI's original value before spawning mvn validate

tkobayas:
Hmm, it doesn't reproduce...

PR #6787, opened 2026-06-29
[incubator-kie-drools-6783] Type inline-accumulate result binding used in a result constraint

### LINKED ISSUE #6783: Executable model: a result constraint referencing an inline `accumulate` binding fails to compile (binding typed as `Object`)
**Summary:** For an inline-binding accumulate `accumulate( …; $b : fn(…); <constraint using $b> )`, a result constraint that uses the binding (e.g. `$b.size > 0`) fails to compile because `$b` is typed as `Object` in the generated constraint lambda, so member access doesn't resolve. The classic compiler types the binding as the accumulate function's result type.

**Reproducer** (`Person` with `int getAge()`):
```drl
import org.test.Person;
rule R when
  accumulate(
    $p : Person( $a : age );
    $ages : collectList( $a );
    $ages.size > 0
  )
then end
```
- **Expected:** compiles (`$ages` typed as `List`).
- **Actual:** compile error — `size` cannot be resolved; generated lambda is `(Object $ages) -> $ages.size > 0`.

**Root cause:** the result pattern correctly keeps its `Object` object-type node (for RETE matching), but the constraint lambdas built for the result constraints are not retyped to the function's result type. *(Naively changing the pattern's object type instead breaks RETE matching — it caused ~28 AccumulateTest/GroupByTest failures — so only the constraint lambdas must be retyped, not the OTN.)*

**Suggested fix:** after building the result pattern, retype the binding references in the already-built constraint expressions to the accumulate function's result type, leaving the pattern object type unchanged (in `ModelGeneratorVisitor`, inline-binding accumulate branch):
```java
context.getTypedDeclarationById(function.getBind())
    .map(DeclarationSpec.class::cast)
    .ifPresent(decl -> {
        Class<?> resultType = decl.getDeclarationClass();
        if (resultType != null && resultType != Object.class)
            context.getExpressions().forEach(e ->
                replaceTypeInExprLambdaAndIndex(decl.getBindingId(), resultType, e));
    });
```

drccrd:
Fixes: #6783 

A result constraint that references an inline-binding accumulate result (e.g. "$ages : collectList($a); $ages.size > 0") failed to compile because the binding was left typed as Object. After building the accumulate pattern, retype the binding references in the result-constraint lambdas to the accumulate function's result type, without altering the pattern object type used for matching

### DIFF
--- a/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/ModelGeneratorVisitor.java
+++ b/drools-model/drools-model-codegen/src/main/java/org/drools/model/codegen/execmodel/generator/visitor/ModelGeneratorVisitor.java
@@ -35,6 +35,7 @@ import org.drools.drl.ast.descr.OrDescr;
 import org.drools.drl.ast.descr.PatternDescr;
 import org.drools.drl.ast.descr.PatternSourceDescr;
 import org.drools.model.codegen.execmodel.PackageModel;
+import org.drools.model.codegen.execmodel.generator.DeclarationSpec;
 import org.drools.model.codegen.execmodel.generator.RuleContext;
 import org.drools.model.codegen.execmodel.generator.visitor.accumulate.AccumulateVisitor;
 import org.drools.model.codegen.execmodel.generator.visitor.accumulate.GroupByVisitor;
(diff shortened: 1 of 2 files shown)


tkobayas:
@drccrd Thank you very much for the PR. I see that the fix rewrites the lambda after `buildPattern`. While the approach is clean enough, I think we can fix the root problem earlier --- in the pattern building. I filed a PR with the alternative approach (https://github.com/apache/incubator-kie-drools/pull/6791).

@mariofusco  , could review the both PRs and which one is preferred?

mariofusco (changes requested):
I believe this other one https://github.com/apache/incubator-kie-drools/pull/6791 is a cleaner solution.

tkobayas:
Closing this PR because it has been superseded by https://github.com/apache/incubator-kie-drools/pull/6791. Thanks!

drccrd:
Much cleaner, thank you!

PR #2652, opened 2019-11-22
decouple ruleunit + pluggable classloader

mariofusco:

PR #1874, opened 2018-04-24
RHPAM-795 RHPAM-55 Register project classloaders as parallel capable

baldimir:
- This avoids the deadlock mentioned in https://issues.jboss.org/browse/RHPAM-795 and probably also the deadlock from RHPAM-55. 

@mariofusco @DuncanDoyle could you please review?

PR #6809, opened 2026-07-15
[incubator-kie-drools-6758] Register concrete project classloaders as parallel capable

### LINKED ISSUE #6758: DynamicProjectClassLoader does not call registerAsParallelCapable()
I am running unit tests and would like each scenario to have a clean load of a test-specific subset of rules.  This is triggering an issue where DynamicProjectClassLoader is effectively single-threaded.

This appears to be an oversight.  Is there some reason this classloader should not be parallel capable?

nikhiln64:
**Issue**:

* Closes https://github.com/apache/incubator-kie-drools/issues/6758

**Summary**

Parallel capability is per-class, not inherited: `ProjectClassLoader` calls `registerAsParallelCapable()`, but the JVM checks the loader's **concrete runtime class** (and its whole superclass chain), so the concrete loaders never benefited - `getClassLoadingLock` fell back to locking the loader instance, serializing every class load through one monitor.

Registered all five concrete classes: `DynamicProjectClassLoader`, `IBMDynamicClassLoader`, `StaticProjectClassLoader`, `IBMStaticClassLoader`, and `DummyInternalTypesClassLoader` (the dynamic `DefaultInternalTypesClassLoader` already had it).

Thread-safety audit of what parallel loading newly exposes: the `defineType`/`defineClass`/`undefineClass` paths are already `synchronized`, and the `loadedClasses`/`nonExistingClasses` caches are `ConcurrentHashMap`s - the base class was designed for this; the subclasses just never opted in.

**Testing**

New tests in both wiring modules use a probe subclass: `registerAsParallelCapable()` returns `false` when any ancestor class is not itself registered, so the assertion is deterministic (no timing games) and goes red if a registration is ever removed - verified red against the unfixed loaders, green with the fix. Full `drools-wiring-dynamic` + `drools-wiring-static` test suites pass on JDK 17.

Developed with AI assistance (Claude Code), reviewed and driven by a human.

### DIFF
--- a/drools-wiring/drools-wiring-dynamic/src/main/java/org/drools/wiring/dynamic/DynamicProjectClassLoader.java
+++ b/drools-wiring/drools-wiring-dynamic/src/main/java/org/drools/wiring/dynamic/DynamicProjectClassLoader.java
@@ -33,6 +33,13 @@ import org.drools.wiring.api.util.ClassUtils;
 
 public class DynamicProjectClassLoader extends ProjectClassLoader {
 
+    static {
+        // Parallel capability is per-class, not inherited: without this registration the JVM
+        // ignores the base class' one and getClassLoadingLock falls back to locking `this`,
+        // serializing all class loading through this loader (issue #6758).
+        registerAsParallelCapable();
+    }
+
     private static boolean isIBM_JVM = System.getProperty("java.vendor").toLowerCase().contains("ibm");
 
     protected DynamicProjectClassLoader(ClassLoader parent, ResourceProvider resourceProvider) {
@@ -50,6 +57,10 @@ public class DynamicProjectClassLoader extends ProjectClassLoader {
     }
 
     public static class IBMDynamicClassLoader extends DynamicProjectClassLoader {
+
+        static {
+            registerAsParallelCapable();
+        }
 
         private final boolean parentImplementsFindResources;
 
--- a/drools-wiring/drools-wiring-dynamic/src/test/java/org/drools/wiring/dynamic/ParallelCapableClassLoaderTest.java
+++ b/drools-wiring/drools-wiring-dynamic/src/test/java/org/drools/wiring/dynamic/ParallelCapableClassLoaderTest.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.drools.wiring.dynamic;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Parallel capability is per-class, not inherited: the JVM only enables per-name class loading
+ * locks when the loader's concrete runtime class and its whole superclass chain are registered.
+ * Without the subclass registration, getClassLoadingLock falls back to locking the loader itself
+ * and all class loading serializes on one monitor (issue #6758).
+ *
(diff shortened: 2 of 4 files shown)


mariofusco (changes requested):
I believe we attempted this in the past (quite a few years ago), we had a few problems and we have to revert it. I honestly don't if something relevant is changed in this area in the meantime.

I would vote against this change unless there isn't a very compelling reason to do this. If it is to improve performances I would like to have at least a benchmark clarifying the amount of the gain. 

I also gave a look to the original issue ( https://github.com/apache/incubator-kie-drools/issues/6758 ) and I honestly don't see how the things are related and which kind of advantages could bring this change for testing. Can you please clarify?

nikhiln64:
Thanks for the careful look @mariofusco - I dug into the history before this PR and again now, and I'd like to share what I found, because I think this PR completes an old fix rather than re-attempts a reverted one.

**The past attempt appears to have stuck, not reverted:** parallel-capable registration was introduced by 4381219ecf / #1874 (RHPAM-795/RHPAM-55, 2018) - and notably it was itself a **deadlock fix** (the commit adds "a test for a classloader deadlock"). That registration is still present today in `ProjectClassLoader` (drools-wiring-api, line 55), together with the descendant of that deadlock regression test (`ClassLoaderTest` in drools-wiring-dynamic). Searching `-S registerAsParallelCapable` across all branches I find no revert - only the 2018 introduction and the 2019 wiring-module split (#2652). If the revert you remember was in a different repo (jBPM/RHPAM product layer?), I'd genuinely appreciate the pointer.

**The gap this PR closes:** when the split created the concrete subclasses, the per-class registration wasn't carried over - and the JVM checks the loader's *runtime* class, so the 2018 registration is silently inert for every `DynamicProjectClassLoader`/`StaticProjectClassLoader` instance. `getClassLoadingLock` returns `this`, i.e. exactly the coarse-lock configuration the 2018 fix was eliminating.

**Relation to #6758:** the reporter runs unit tests where each scenario loads a test-specific subset of rules concurrently and hits the single-monitor serialization. So the benefit claimed here is not raw throughput - it's (a) restoring the deadlock-avoidance property of #1874 for the loaders actually instantiated at runtime, and (b) unblocking concurrent class loading for use cases like the reporter's.

That said, you have institutional memory I don't. I'm happy to (1) add a JMH or multi-threaded stress benchmark quantifying the concurrent-load behavior if that would help, or (2) park this if the team's risk assessment says the coarse lock is the safer default - your call either way.

mariofusco:
I'm afraid that the test failures in CI could be also related and caused by this change. Please check them and if they aren't it is ok for me to try to merge this PR. 

I just relaunched CI to see what happens.

nikhiln64:
@mariofusco I looked into the CI failures before asking you to consider merging anything. The failing tests are the `KJarLoadingTest` ones in kie ci, and they fail with "Cannot find KieModule" while creating the container. That happens during module resolution, which is a step before any class loading goes through the project classloaders touched here.

I also tried to reproduce it locally. The interesting part is that the test fails in exactly the same way whether my change is applied or not. I checked out the parent commit's wiring sources, reinstalled them and reran the test, and got the identical error. My local environment cannot resolve the test kjar in either state though, so I would not call that conclusive on its own.

One thing I could rule out cleanly is branch staleness. This branch was cut from a July 14 main while the recently green PRs run against the current one, so I rebased onto today's main and pushed. The new CI runs are currently waiting for approval to start. If they still show the same KJar failures on the fresh base I will treat that as real evidence of an interaction with this change and dig into the kie ci module discovery path myself first. I would not want this merged with an unexplained failure either.

nikhiln64:
@mariofusco I went deeper on this after the rerun on the rebased branch showed the same failures. The CI logs point somewhere unexpected. Right before the KJarLoadingTest errors there is a stream of warnings from MavenRepository.resolveArtifact saying it is unable to resolve org.drools.testcoverage:drools-kie-ci-with-domain-test-kjar:999-SNAPSHOT. So the tests are failing while resolving the fixture kjar through the embedded maven machinery, which happens before any class loading that this change could influence. The "Cannot find KieModule" message is what surfaces after the scanner swallows that resolution failure.

I also redid the local comparison more carefully this time. I installed the test kjar and domain modules into my local repo first, then ran the kieci suite with my change applied and again with the parent commit's wiring sources reinstalled. The outcome was identical in both states every time.

Given that main is green and the recently passing PRs only touched dependency versions, my best guess is that the PR build computes its module set from the diff, and when only drools-wiring changes the test kjar does not get built into the repository the kieci tests resolve from. Is that a known quirk of the kieci standalone suite? If there is a preferred way to debug or force the kjar build in the PR pipeline I am happy to do it, and equally happy to be proven wrong if you see a real interaction here.

tkobayas:
I confirmed that the `CI :: Build` issue is reproducible with a simple edit in a file in `drools-wiring`. I filed a GH issue and will work on a fix : https://github.com/apache/incubator-kie-drools/issues/6819

So the CI failure doesn't block this PR. `Drools` job is green, we may think that the tests passed.

tkobayas (approved):

mariofusco (approved):

PR #6817, opened 2026-07-20
NO-ISSUE:fix: upgrade org.eclipse.parsson:parsson from 1.1.7 to 1.1.8 to fix CVE

athirakm94:
fix: upgrade org.eclipse.parsson:parsson from 1.1.7 to 1.1.8 to fix CVE


yesamer (changes requested):
@athirakm94 Parsson is already declared (line 1690). The version is set with a misleading name (version.jakarta.json), so I guess we can upgrade that version and rename that variable.

yesamer (approved):

gitgabrio (approved):

PR #6821, opened 2026-07-22
[incubator-kie-issues#2377] Message start event without data mapping fails code generation

### LINKED ISSUE #2377: Message start event without data mapping fails code generation
A message start event with no data mapping fails to build. `MessageHandler` requires an itemDefinition for every `<message>`, so a message with no `itemRef` fails with `Could not find itemDefinition null`, and a message whose itemDefinition has an empty `structureRef` is never registered and fails with `Could not find message <id>`.

Both `itemRef` (on `<message>`) and `structureRef` (on `<itemDefinition>`) are optional in the BPMN2 spec, so a message with just an id and a name should be valid. The legacy jBPM parser defaulted an empty `structureRef` to `java.lang.Object` and these processes used to build.

Abhitocode:
- **MessageHandler**:  message with no `itemRef` has no data, so only look up the itemDefinition when an `itemRef` is actually there (looking it up when it's missing is what caused the error). Still register the message so the event can find it, just with no type.

- **ServiceTaskHandler**:  a Java service task uses a message to describe the method it calls (like the argument type). Now that a no-data message has no type, reading the type from it would leave the service task with nothing and break code generation. So when the message has no type, skip it and get the argument type from the service task's own data inputs instead.

- **ProcessHandler** and **TriggerMetaData**: for a message event, code generation needs a real type. A no-data message doesn't have one, so it defaults to `java.lang.Object` in these two spots.

**Tests**
- `CodegenMessageStartEventTest`:  a message with no `itemRef`, and an itemDefinition with an empty `structureRef`.
- `StartEventTest.testMessageStartWithoutDataMapping`:  sends the message with no payload and checks the process starts.

closes: https://github.com/apache/incubator-kie-issues/issues/2377

### DIFF
--- a/kogito-api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/AbstractCloudEventDataConverter.java
+++ b/kogito-api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/AbstractCloudEventDataConverter.java
@@ -56,7 +56,7 @@ public abstract class AbstractCloudEventDataConverter<O> implements Converter<Cl
             return Optional.of(targetClass.isAssignableFrom(pojo.getClass()) ? targetClass.cast(pojo) : PojoCloudEventDataMapper.from(objectMapper, targetClass).map(value).getValue());
         } else if (value instanceof JsonCloudEventData) {
             JsonNode node = ((JsonCloudEventData) value).getNode();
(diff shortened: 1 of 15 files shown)


martinweiler (approved):
Great work @Abhitocode

josedee (approved):

PR #6823, opened 2026-07-23
improve protobuf session validation with deserialization

tkobayas:
Improved the validation.


tkobayas:
`Drools` GHA job is marked as cancelled, but the result log was actually successful. Probably it's better to set a longer timeout value in a separated PR.

PR #6837, opened 2026-07-24
[NO-ISSUE] Increase job timeout

tkobayas:
Increased timeout because `optaplanner` and `kogito-runtimes` have been added.

### DIFF
--- a/.github/workflows/pr-drools.yml
+++ b/.github/workflows/pr-drools.yml
@@ -36,7 +36,7 @@ jobs:
     concurrency:
       group: pr-drools_${{ matrix.os }}_${{ matrix.java-version }}_${{ matrix.maven-version }}_${{ github.head_ref }}
       cancel-in-progress: true
-    timeout-minutes: 120
+    timeout-minutes: 240
     strategy:
       matrix:
         os: [ubuntu-latest]


yesamer (approved):

.github/workflows/pr-drools.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     concurrency:
       group: pr-drools_${{ matrix.os }}_${{ matrix.java-version }}_${{ matrix.maven-version }}_${{ github.head_ref }}
       cancel-in-progress: true
-    timeout-minutes: 120
+    timeout-minutes: 240
```
tkobayas, line 39: With this CI run, `Drools` job took 120 minutes. `timeout-minutes: 240` seems like a reasonable timeout value.

Rikkola: Yes the build time is huge now. I am trying to use scripts for my local build needs with only selected modules. Optaplanner has some long lasting tests on full build so even 240 might not be enough, but we will see.

Rikkola (approved):

.github/workflows/pr-drools.yml
Rikkola: I guess if you saw it take 120 minutes then it is good :)

PR #6839, opened 2026-07-24
incubator-kie-issues#2307: adjust root process references after migration

jstastny-cz:
To maintain correct data-isolation boundaries, we need to update root_process references for relevant entries in runtime persistence tables. Specifically, in a parent -> child relationship, once parent is migrated and its process id and/or process version are changed, such change needs to be reflected also on the side of given instanes' children.

The solution in schema connects each "child" instance with its root process instance also using `root_process_instance_id` column - which then allows adjustment of the migrate queries in the Repository class.


pefernan (approved):

martinweiler (approved):

pefernan:
Failures are not related to this change... @martinweiler are we good to go? @kbowers-ibm ^

klbowers:
@pefernan Full build is passing, so no problem with me!

PR #6850, opened 2026-07-29
[incubator-kie-issues#2382] Handle concurrent duplicate ProcessDefinition insert

### LINKED ISSUE #2382: Duplicate key violation on definitions table when multiple replicas start at the same time
When several replicas of the same application start at the same time, each one indexes its own process definitions. They all read an empty state before any of them commits, so they all insert the same `(id, version)` rows. All but the first fail with a primary key violation on the definitions` table, and the replica crashes on startup.

Quarkus:
`org.hibernate.exception.ConstraintViolationException: could not execute statement
[ERROR: duplicate key value violates unique constraint "definitions_pkey"
Detail: Key (id, version)=(myProcess, 1.0) already exists.]
at com.arjuna.ats.arjuna.coordinator.TwoPhaseCoordinator.beforeCompletion(TwoPhaseCoordinator.java:348)
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "definitions_pkey"`

Springboot:
On Spring Boot the same race gives `org.springframework.dao.DataIntegrityViolationException`,
thrown from `JpaTransactionManager.doCommit`.

**Expected behaviour**

A definition already registered by another replica is not an error. The replica should keep the
stored definition and continue starting up

Abhitocode:
When a few replicas of the same app start together, they all see an empty state and insert the same `(id, version)` rows into `definitions`. All but the first hit a primary key violation and the replica dies on startup. 

The reason is  `em.merge()` only queues the INSERT, and `put` joins the transaction that indexing already runs in, so the INSERT is only sent when that outer transaction commits - after any `try`/`catch` around `put` has returned.

**Changes**

- **`ProcessDefinitionEntityStorage`** - new `put` overload with the actual fix: flush the insert right away so the failure happens where we can catch it, run it in a separate transaction so the rollback doesn't kill the caller's work, and on a duplicate just return the definition that's already stored.

- **`QuarkusProcessDefinitionEntityStorage`** (jpa-quarkus *and* postgresql) - overrides `put` to run the insert in a `@Transactional(REQUIRES_NEW)` method. Two copies of the same class, both used by an embedded add-on, so both need it.

- **`SpringBootProcessDefinitionEntityStorage`** (new, springboot) - same thing with a`TransactionTemplate`. `@Transactional` doesn't work here: Spring ignores it when a class calls its own method, so the insert would just run in the caller's transaction.

This is also why the catch sits outside the transaction, not inside, here a failed flush marks the transaction as rollback only, so the commit still fails after the `catch` has run. The exception is different too, Quarkus throws `ConstraintViolationException`, Spring Boot throws `DataIntegrityViolationException`.

- **`DataIndexStorageProducer`** (springboot) - removed the `processDefinitionEntityStorage` bean, otherwise there'd be two beans of the same type now that the class above provides it.

**Tests**

`AbstractConcurrentProcessDefinitionStorageIT` plus a PostgreSQL subclass per runtime. A plain JDBC connection acts as the competing replica, it inserts the row and holds it uncommitted, so the storage still sees nothing and queues its own insert, which blocks on the primary key index. The commit is triggered from `pg_locks` once that insert is actually waiting, so there are no sleeps.

Closes: https://github.com/apache/incubator-kie-issues/issues/2382

### DIFF
--- a/kogito-data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/ProcessDefinitionEntityStorage.java
+++ b/kogito-data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/ProcessDefinitionEntityStorage.java
@@ -19,7 +19,6 @@
 package org.kie.kogito.index.jpa.storage;
 
 import java.util.Optional;
-import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.hibernate.exception.ConstraintViolationException;
@@ -37,7 +36,7 @@ import jakarta.persistence.EntityManager;
 
 import static org.kie.kogito.index.DependencyInjectionUtils.getInstance;
 
-public class ProcessDefinitionEntityStorage extends AbstractStorage<ProcessDefinitionKey, ProcessDefinitionEntity, ProcessDefinition> {
+public abstract class ProcessDefinitionEntityStorage extends AbstractStorage<ProcessDefinitionKey, ProcessDefinitionEntity, ProcessDefinition> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ProcessDefinitionEntityStorage.class);
 
@@ -53,10 +52,10 @@ public class ProcessDefinitionEntityStorage extends AbstractStorage<ProcessDefin
                 e.getVersion()), Optional.ofNullable(getInstance(predicateBuilder)), Optional.ofNullable(getInstance(processes)));
     }
 
-    protected ProcessDefinition put(ProcessDefinitionKey key, ProcessDefinition value,
-            Function<Supplier<ProcessDefinition>, ProcessDefinition> isolatedTransaction) {
+    @Override
+    public ProcessDefinition put(ProcessDefinitionKey key, ProcessDefinition value) {
         try {
-            return isolatedTransaction.apply(() -> {
+            return wrapInTransaction(() -> {
                 super.put(key, value);
                 em.flush();
                 return value;
@@ -67,8 +66,10 @@ public class ProcessDefinitionEntityStorage extends AbstractStorage<ProcessDefin
             }
             LOGGER.info("ProcessDefinition with id '{}' and version '{}' is already present, skipping insert.", key.getId(), key.getVersion());
             LOGGER.debug("Duplicate ProcessDefinition insert suppressed", e);
-            return get(key);
+            return wrapInTransaction(() -> get(key));
         }
     }
+
+    protected abstract ProcessDefinition wrapInTransaction(Supplier<ProcessDefinition> supplier);
 
 }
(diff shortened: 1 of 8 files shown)


kogito-data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/ProcessDefinitionEntityStorage.java
```diff
@@ -44,4 +53,22 @@ public ProcessDefinitionEntityStorage(EntityManager em, Iterable<JsonPredicateBu
...
+        } catch (RuntimeException e) {
+            if (Throwables.getCausalChain(e).stream().noneMatch(ConstraintViolationException.class::isInstance)) {
+                throw e;
+            }
+            LOGGER.info("ProcessDefinition with id '{}' and version '{}' is already present, skipping insert.", key.getId(), key.getVersion());
+            LOGGER.debug("Duplicate ProcessDefinition insert suppressed", e);
+            return get(key);
+        }
+    }
+
```
pefernan: I think this code could be simplified to something like:

```suggestion
    public ProcessDefinition put(ProcessDefinitionKey key, ProcessDefinition value) {
    try {
        return wrapInTransaction(() -> {
            super.put(key, value);
            em.flush();
            return value;
        });
    } catch (RuntimeException e) {
        if (Throwables.getCausalChain(e).stream().noneMatch(ConstraintViolationException.class::isInstance)) {
            throw e;
        }
        LOGGER.info("ProcessDefinition with id '{}' and version '{}' is already present, skipping insert.", key.getId(), key.getVersion());
        LOGGER.debug("Duplicate ProcessDefinition insert suppressed", e);
        return wrapInTransaction(() -> get(key));
    }
}

protected abstract ProcessDefinition wrapInTransaction(Supplier<ProcessDefinition> supplier);

```

pefernan: Then each concrete implementation should only implement thet insertInNewTransaction method with the @Transactional annotation like:
```java
@Override
@Transactional(TxType.REQUIRES_NEW)
protected ProcessDefinition wrapInTransaction(Supplier<ProcessDefinition> supplier) {
    return supplier.get();
}

// Spring Boot
@Override
@Transactional(Transactional.TxType.REQUIRES_NEW)
protected ProcessDefinition wrapInTransaction(Supplier<ProcessDefinition> supplier) {
    return supplier.get();
}
```

pefernan (changes requested):
@Abhitocode great work! The `flush` + `REQUIRES_NEW` + `ConstraintViolationException` catch pattern is exactly right. Just had a thought on simplifying the design a bit.

The current `put(key, value, Function<Supplier<…>, …>)` overload creates a bit of an awkward bounce: the base class builds a Supplier and hands it to the subclass, the subclass calls it and returns the result back to the base class. It works, but it means each subclass has to override `put()` just to inject the transaction boundary.

What if instead the base class owns `put()` entirely and subclasses only implement a single abstract method `wrapInTransaction(Supplier<ProcessDefinition>)`? The insert logic and the catch/fallback all live in one place, and each subclass becomes a one-liner that just delegates to its framework's transaction mechanism.

As a bonus, the `get(key)` fallback in the catch could also run in a new transaction ensuring we get the real value in db.

Also on the Spring Boot side — with `@Transactional(propagation = Propagation.REQUIRES_NEW)` on `wrapInTransaction`, the `TransactionTemplate` field and the `PlatformTransactionManager` constructor param are no longer needed, making `SpringBootProcessDefinitionEntityStorage` as lean as the Quarkus ones.

pefernan (approved):

martinweiler (approved):
Great work @Abhitocode

PR #6852, opened 2026-07-29
[incubator-kie-issues-2307]Isolate data retrieved by data-audit addons

### LINKED ISSUE #2307: Compact architecture persistence layer leaks data between apps connected to single relational database
In compact architecture setup, when multiple apps share the same relational database, various endpoints return data not belonging to particular service instance, but rather a superset of data coming from all the apps connected to the particular database instance.

To be able to isolate runtime data coming from database based on sense of "locality", existing process definition identifiers can be used without the need to bring in any notion of deployment topology information into the database schema. Such identifier should be process id and version composed key - but at given moment, throughout the engine and also DB schema, the uniqueness is judged based on process id itself.

To bear such simplification in mind, but also to allow for possible future refactoring, the persistence modules for all relevant components should implement filtering support. For the time being based on a single column, possibly extensible to multiple columns.

The relevant components:
* runtime persistence addon
* jbpm usertasks jpa addon
* Embedded data-index
* Embedded jobs addon
* Embedded data-audit

Goals:
* Make sure all data loading operations apply filtering based on the "locality" of queried items.
* Keep schema changes minimal, no artificial deployment topology information added, rely on existing values where possible. Possible schema denormalization to facilitate filtering (e.g. extract value from stored JSON column into newly added string column).
* Account for possible future change in filters from single value (id) to multiple (e.g. id + version).

Non goals:
* Any changes to components outside of the compact architecture setup.
* Any changes to storage options other than relational databases.

SoniyaAbraham89:
Closes https://github.com/apache/incubator-kie-issues/issues/2307

Included the code to achieve isolation in data-audit
The filtering is currently done based on process id and version pairs. considering root process id and root process version in case of sub-processes


jstastny-cz (approved):
Looks good, thank you @SoniyaAbraham89

kogito-data-audit/kogito-addons-data-audit-jpa/kogito-addons-data-audit-jpa-common/src/main/java/org/kie/kogito/app/audit/jpa/queries/JPAAbstractQuery.java
```diff
@@ -85,4 +93,124 @@ protected <T> List<T> executeWithQueryEntityManagerAndArguments(EntityManager en
...
+                processIdColumn, processVersionColumn, rootProcessIdColumn, rootProcessVersionColumn);
+
+        LOGGER.debug("About to execute isolated native query {} with arguments {}", finalQuery, params);
+        Query jpaQuery = entityManager.createNativeQuery(finalQuery);
+
+        params.forEach(jpaQuery::setParameter);
+
+        bindIsolationParameters(jpaQuery, allowedKeys);
+
+        if (pagination != null) {
```
martinweiler, line 128: This logic is missing the `offset` parameter. Please reuse the logic from `executeWithQueryEntityManagerAndArguments`, ideally extract into a common method to be used in both places.

martinweiler (changes requested):
Looks good overall @SoniyaAbraham89, thank you!
Requesting a change regarding the `offset` parameter.

kogito-data-audit/kogito-addons-data-audit-jpa/kogito-addons-data-audit-jpa-common/src/main/java/org/kie/kogito/app/audit/jpa/queries/JPAAbstractQuery.java
SoniyaAbraham89: Moved the pagination logic of executeWithQueryEntityManagerAndArguments () to applyPagination() and used the same in case of isolation too

martinweiler (approved):
Great work @SoniyaAbraham89

PR #6860, opened 2026-07-31
[DO-NOT-MERGE] [incubator-kie-6854] Parallel CI Jobs

tkobayas:
- Issue : https://github.com/apache/incubator-kie/issues/6854


tkobayas:
(deleted)

tkobayas:
Confirmed sequential CI (original ci.yaml) and parallel CI have the same test coverage.

tkobayas:
(deleted)

tkobayas:
# Split PR CI into 4 parallel test jobs

## Problem

The monorepo PR build ran all ~850 modules sequentially in a single job, taking ~2-3h.

## Solution

Split the PR test step into 4 concurrent jobs by project category: **drools**, **optaplanner**, **kogito-runtimes**, **kogito-apps**. Wall-clock time drops to ~1h15m (limited by the longest job, kogito-apps).

## Key changes

**`script/ci/CiComputeBuildScopes.java`** -- Added `parseModuleCategories()` which reads `<!-- BEGIN/END xxx modules (auto) -->` markers in the root `pom.xml` to categorize modules. Writes per-category `affected-*.txt` files (e.g., `affected-drools.txt`, `affected-kogito-runtimes.txt`) alongside the existing combined output.

**`.github/actions/ci-setup/action.yaml`** -- New composite action extracting shared CI setup: free disk space, checkout-pr, Java, JBang, Maven settings, changed-file collection, and build scope computation.

**`.github/workflows/ci.yaml`** -- Push-to-main remains a single full-build job. PRs fan out to 4 parallel test jobs. Each job:
1. Checks if its category has affected modules -- if not, skips entirely
2. Installs upstream dependencies without tests (`-T 1C -Dquarkus.build.skip=true` for speed)
3. Runs tests only for its own category's affected modules

**`.rat-excludes`** -- Replaced 48 individual golden file entries with a single glob `script/ci/tests/scenarios-compute-build-scopes/**`.

**Golden files** -- 48 new per-category expected files added across 12 test scenarios. Existing golden files updated for new modules on main.

## Cross-category dependency handling

Each test job installs all upstream categories' affected modules (without tests) before running its own:

| Job | Upstream install (no tests) |
|---|---|
| test-drools | upstream only |
| test-optaplanner | upstream + affected-drools |
| test-kogito-runtimes | upstream + affected-drools + affected-optaplanner |
| test-kogito-apps | upstream + affected-drools + affected-optaplanner + affected-kogito-runtimes |

## Build time comparison

Sequential CI (run `85920644738`):

| Job | Duration |
|---|---:|
| Java 17 | 3h 05m 04s |
| Java 21 | 2h 14m 10s |

Parallel CI (run `86233087403`):

| Job | Java 17 | Java 21 |
|---|---:|---:|
| Drools | 44m 01s | 41m 55s |
| OptaPlanner | 15m 27s | 19m 06s |
| Kogito runtimes | 1h 12m 45s | 1h 06m 30s |
| Kogito apps | 1h 16m 13s | 1h 05m 33s |

.github/workflows/ci.yaml
```diff
@@ -61,110 +63,82 @@ jobs:
...
-      # See https://github.com/apache/infrastructure-actions/pull/806
-      - name: 'SETUP :: Install JBang'
+      - name: "CI :: BUILD :: Full"
         shell: bash
-        run: |
-          curl -sL https://github.com/jbangdev/jbang/releases/download/v0.138.0/jbang-0.138.0.zip -o jbang.zip
-          echo "3c9fb9ac5823ae5ab9d136dec08b896e2a0f6f6b313689b21f43c511b7c34b85  jbang.zip" | sha256sum -c -
-          unzip -q jbang.zip
-          echo "${{ github.workspace }}/jbang-0.138.0/bin" >> $GITHUB_PATH
+        run: mvn --batch-mode --no-transfer-progress -fae -Dsurefire.redirectTestOutputToFile=true -Dfull -Dreproducible install
```
tkobayas: On push, we still use the non-parallel full build.

.github/workflows/ci.yaml
```diff
@@ -61,110 +63,82 @@ jobs:
...
+
+  test-drools:
+    if: github.event.action != 'closed' && github.event_name == 'pull_request'
+    name: 'drools :: ${{ matrix.os }}, Java ${{ matrix.java }}'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:   [ubuntu-latest]
+        java: [17, 21]
```
tkobayas: In each parallel job, I don't use `exclude`, but exact `ubuntu-latest` and java `17` `21` for conciseness. This should be enhanced when we accept mac and windows.

.github/workflows/ci.yaml
```diff
@@ -178,32 +152,90 @@ jobs:
...
-      - name: "PR CHECK :: BUILD :: Upstream modules${{ github.event_name == 'push' && ' (skipped)' || '' }}"
-        if: github.event_name == 'pull_request'
+      - name: "PR CHECK :: BUILD :: Upstream modules"
         shell: bash
         run: |
+          [ ! -s "$MAVEN_PL_AFFECTED_DROOLS_FILE" ] && echo "No affected drools modules. Skipping upstream build." && exit 0
           pl=$(paste -sd, "$MAVEN_PL_UPSTREAM_FILE")
           [ -z "$pl" ] && echo "No upstream modules. Skipping." && exit 0
-          mvn -T 1C --batch-mode --no-transfer-progress -fae -DskipTests -DskipITs -Denforcer.skip=true -Dcheckstyle.skip=true -Dformatter.skip=true -Darchunit.skip=true -Dsurefire.redirectTestOutputToFile=true -pl "$pl" install
+          mvn --batch-mode --no-transfer-progress -T 1C -fae -DskipTests -DskipITs -Dquarkus.build.skip=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dformatter.skip=true -Darchunit.skip=true -Dsurefire.redirectTestOutputToFile=true -pl "$pl" install
```
tkobayas: I added `-Dquarkus.build.skip=true` to solve the quarkus build ordering issue with `-T 1C`.

.github/workflows/ci.yaml
```diff
@@ -178,32 +152,90 @@ jobs:
...
-      - name: "PR CHECK :: BUILD :: Changed and affected modules${{ github.event_name == 'push' && ' (skipped)' || '' }}"
-        if: github.event_name == 'pull_request'
+      - name: "PR CHECK :: TEST :: drools"
         shell: bash
         run: |
-          pl=$(paste -sd, "$MAVEN_PL_AFFECTED_FILE")
-          [ -z "$pl" ] && echo "No affected modules. Skipping." && exit 0
+          pl=$(paste -sd, "$MAVEN_PL_AFFECTED_DROOLS_FILE")
+          [ -z "$pl" ] && echo "No affected drools modules. Skipping." && exit 0
           mvn --batch-mode --no-transfer-progress -fae -Dsurefire.redirectTestOutputToFile=true -Dfull -Dreproducible -pl "$pl" install
```
tkobayas, line 195: I don't use `-T 2` for the test phase because it will raise many issues, and they will likely require a long time to fix (See https://github.com/apache/incubator-kie/pull/6864).

pom.xml
```diff
@@ -152,6 +152,9 @@
   </build>
 
   <modules>
+    <!-- DO NOT remove the BEGIN/END comments. They are used by CiComputeBuildScopes for parallel CI job split
+         When you add a new module, put it in one of the 4 categories (drools, optaplanner, kogito-runtimes, kogito-apps) -->
+    <!-- BEGIN drools modules (auto) -->
```
tkobayas: `CiComputeBuildScopes` depends on these `BEGIN` `END` markers for 4 categories split. It's a little fragile, but when broken, `CiComputeBuildScopes` makes sure to exit(1) with an error, so we will be able to detect.

tkobayas:
@tiagobento  @yesamer @Rikkola @athirakm94 @kbowers-ibm 

Hi, this PR splits the PR check CI into four parallel jobs to reduce the overall CI time.

Please have a look at the above summary.

The Build time comparison is the "big impact" case where almost all modules are tested (e.g., root pom.xml change, drools-core change, etc.). But other use cases like only kogito-runtimes change would also benefit from the parallel CI.

The CI logic (`CiComputeBuildScopes`, `ci.yaml`) gets a little more complicated, but I think the gain is larger

Please review, thanks!

Rikkola:
Looks like this is also making the scripts a bit easier to read, but is there an easy way to run this locally? First to test this, but also to run my own nighties on my WIP topic branches.

tkobayas:
@Rikkola 
> is there an easy way to run this locally?

I haven't tried, but this seems to be for the purpose > https://github.com/nektos/act 

Btw, sometimes I use my fork repo to run GHA jobs (e.g., https://github.com/tkobayas/drools/pull/17) So it doesn't consume my local PC resource :)

Rikkola (approved):

script/ci/CiComputeBuildScopes.java
```diff
@@ -202,6 +220,102 @@ private static void writeLines(Path out, Collection<String> lines) throws IOExce
...
+        }
+
+        String currentCategory = "drools";
+        boolean inModules = false;
+        Set<String> seenBegins = new HashSet<>();
+        Set<String> seenEnds = new HashSet<>();
+
+        Pattern beginPattern = Pattern.compile("<!--\\s*BEGIN\\s+(\\S+)\\s+modules\\s+\\(auto\\)\\s*-->");
+        Pattern endPattern = Pattern.compile("<!--\\s*END\\s+(\\S+)\\s+modules\\s+\\(auto\\)\\s*-->");
+        Pattern modulePattern = Pattern.compile("<module>(.+)</module>");
```
tiagobento: I'm a concerned with this parsing strategy. As it relies on comments being maintained in the top-level pom.xml file. This simply mimics the old structure we had with the repos, and doesn't to separate things logically per domain. I don't think we should carry forward the old, artificial grouping we had. 

Can we try to make this partitioned by domain? We can have even more partitions, like Kogito Data-Index, Kogito Jobs Service, for example.

In `kie-tools`, when we needed to do a similar partitioning effort in order to keep PR checks build times in check, what I did was I created partitioning files which listed the leaf modules which represented good vertical partitions of the dependency DAG.

We don't have this anymore since the 10.3.x+ stream effort began, but you can see how it used to look like in 10.2.

https://github.com/apache/incubator-kie-tools/tree/10.2.0/.github/supporting-files/ci/partitions

Note that there are 2 partition files only, with an implicit 3rd partition which contained all the modules that were not listed explicitly in neither of those 2 partition files.

pom.xml
tiagobento: I see you're putting some guardrails in place, but I think we can do better. The "partitions" files listing leaf modules which bring a larger dependency hierarchy with them will give us a lot more flexibility, and will allow us to not couple the pom.xml with a partitioning solution for PR checks and CI.

.github/actions/ci-setup/action.yaml
```diff
@@ -0,0 +1,128 @@
...
+        echo "MAVEN_PL_AFFECTED_KOGITO_RUNTIMES_FILE=${{ runner.temp }}/maven-pl-affected-kogito-runtimes.txt" >> "$GITHUB_ENV"
+        echo "MAVEN_PL_AFFECTED_KOGITO_APPS_FILE=${{ runner.temp }}/maven-pl-affected-kogito-apps.txt" >> "$GITHUB_ENV"
+        touch "${{ runner.temp }}/changed-files.txt"
+        touch "${{ runner.temp }}/maven-pl-affected.txt"
+        touch "${{ runner.temp }}/maven-pl-upstream.txt"
+        touch "${{ runner.temp }}/maven-pl-changed.txt"
+        touch "${{ runner.temp }}/maven-pl-affected-drools.txt"
+        touch "${{ runner.temp }}/maven-pl-affected-optaplanner.txt"
+        touch "${{ runner.temp }}/maven-pl-affected-kogito-runtimes.txt"
+        touch "${{ runner.temp }}/maven-pl-affected-kogito-apps.txt"
```
tiagobento: Declaring 4 fixes partitions I believe will limit the flexibility of the whole mechanism, and further constrain us into fitting new modules into these 4 boxes. Can we try to make partitions more fluid? Adding/removing a new partition should be easy enough. We can use file names to name them, and list the leaf modules which belong in each partition inside the files. E.g.,

```
drools.txt
optaplanner.txt
kogito-runtimes.txt
kogito-data-index.txt
kogito-jobs-service.txt
```

and so on.

.github/workflows/ci.yaml
```diff
@@ -216,14 +248,145 @@ jobs:
...
+      - name: "PR CHECK :: BUILD :: Upstream + drools + optaplanner modules (tests skipped)"
+        shell: bash
+        run: |
+          [ ! -s "$MAVEN_PL_AFFECTED_KOGITO_RUNTIMES_FILE" ] && echo "No affected kogito-runtimes modules. Skipping upstream build." && exit 0
+          upstream=$(paste -sd, "$MAVEN_PL_UPSTREAM_FILE")
+          drools=$(paste -sd, "$MAVEN_PL_AFFECTED_DROOLS_FILE")
+          optaplanner=$(paste -sd, "$MAVEN_PL_AFFECTED_OPTAPLANNER_FILE")
+          pl=$(echo "$upstream,$drools,$optaplanner" | sed 's/^,//;s/,,*/,/g;s/,$//')
+          [ -z "$pl" ] && echo "No modules to build. Skipping." && exit 0
+          mvn --batch-mode --no-transfer-progress -T 1C -fae -DskipTests -DskipITs -Dquarkus.build.skip=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dformatter.skip=true -Darchunit.skip=true -Dsurefire.redirectTestOutputToFile=true -pl "$pl" install
```
tiagobento: Commenting on this one, but it applies to each of the fixed 4 partitions. There's a lot of duplication here. If we make partitions dynamic (and thus more flexible), we remove the duplication and keep things simpler to change as we evolve our codebase. We really need a CI system that's not rigid for us to have room to grow / adapt.

tiagobento (changes requested):
Hi @tkobayas. First of all I would like to thank you immensely for the effort and time you're putting into making our infrastructure better. It's great to see such great initiatives coming together recently.

I left a few comments which I believe will steer the direction of this partitioning mechanism you're introducing here for the better. And I'd also like to bring a thinking point for the total time spent on GitHub Actions that such parallelization introduces.

One additional, general topic:

**Total GitHub Actions resource usage**

Recently, we had our own PR checks and CI jobs taking a lot of time to start. We reached out to Apache Infra and were told some projects sometimes end up using too much of the GitHub Actions quota. Since we share that with all of Apache, we need to be mindful of our growth too. Sequential builds are slower from start to finish, but they take on less resources too, as there's less repeated effort compared to Parallel builds.

Image

Probably something for us to keep in mind is to have an easy-enough mechanism to revert PR checks to sequential builds if we ever reach a GitHub Actions usage scale that will be called out by Apache Infra.

---

Lastly, but not least important: maybe we should try to combine partitioning and a more aggressive -T configuration? Taking advantage of multi-core builds will probably help even more, now that we're aiming to have less scope to build on each partition.

---

Thanks @tkobayas!

tkobayas:
@tiagobento  Thank you for the feedback. I'll work on the suggestions.

One thing:

> Lastly, but not least important: maybe we should try to combine partitioning and a more aggressive -T configuration? Taking advantage of multi-core builds will probably help even more, now that we're aiming to have less scope to build on each partition.

I experimented with the `-T 4` approach in https://github.com/apache/incubator-kie/pull/6864 and achieved a 34% speedup, but at the cost of disabling 10 flaky tests, with one hang issue still unresolved. In theory, we should fix these issues anyway, but I'm afraid it could take a long time. Also, it would be hard to keep all tests reliable under parallel Maven builds.

tiagobento:
@tkobayas 34% is a HUGE gain! That combined with the parallelization this PR brings will put our PR checks into another level of speed.

tkobayas:
# Parallel CI Test Run: Partition Files Approach

Run `89495155157` — all modules affected (root pom change), 857 modules total.

## Module distribution

| Partition | Affected modules | Tests |
|---|---:|---:|
| partition0 (optaplanner) | 28 | 1,535 |
| partition1 (kogito-runtimes) | 179 | 2,260 |
| partition2 (kogito-apps) | 194 | 2,083 |
| default (drools + shared) | 452 | 29,432 |
| **Total** | **853** | **35,310** |

## Job elapsed time

| Job | Java 17 ||| Java 21 |||
|---|---:|---:|---:|---:|---:|---:|
| | upstream | test | **total** | upstream | test | **total** |
| compute-partitions | — | — | **2m 24s** | — | — | — |
| partition0 (optaplanner) | 3m 26s | 9m 17s | **15m 21s** | 3m 06s | 7m 52s | **13m 28s** |
| partition1 (kogito-runtimes) | 9m 32s | 40m 21s | **52m 57s** | 7m 26s | 39m 34s | **49m 28s** |
| partition2 (kogito-apps) | 14m 07s | 1h 03m | **1h 20m 31s** | 13m 12s | 59m 06s | **1h 14m 20s** |
| default (drools + shared) | 8m 41s | 53m 32s | **1h 05m 15s** | 8m 28s | 51m 53s | **1h 03m 29s** |

Wall-clock time = compute-partitions (2m 24s) + slowest test job (1h 20m 31s) = **~1h 23m**

Note:
- The upstream doesn't have `-T 1C` to avoid quarkus build ordering issue. (will investigate next)
- `default` partition contains "unspecified modules and shared modules between partitions" e.g., `kie-addons-springboot-persistence-jdbc`. Anyway, we can tailor the partition files. This is just an example.

tkobayas:
# Parallel CI Test Run: Parallel Upstream with Quarkus Application Rebuild

Run `89559962903` — all jobs completed successfully.

## Job elapsed time

| Job | Java 17 ||| Java 21 |||
|---|---:|---:|---:|---:|---:|---:|
| | upstream | test | **total** | upstream | test | **total** |
| compute-partitions | — | — | **2m 31s** | — | — | — |
| partition0 (optaplanner) | 4m 55s | 9m 52s | **16m 47s** | 4m 30s | 9m 06s | **16m 00s** |
| partition1 (kogito-runtimes) | 7m 06s | 43m 29s | **52m 37s** | 7m 29s | 46m 53s | **57m 15s** |
| partition2 (kogito-apps) | 10m 18s | 1h 04m | **1h 17m 16s** | 10m 05s | 1h 01m | **1h 14m 28s** |
| default (drools + shared) | 8m 09s | 56m 39s | **1h 07m 18s** | 7m 51s | 51m 20s | **1h 02m 12s** |

The partition2 upstream figures include both upstream phases:

- Java 17: parallel upstream 8m 39s + required Quarkus applications 1m 39s = 10m 18s
- Java 21: parallel upstream 8m 30s + required Quarkus applications 1m 35s = 10m 05s

The workflow wall-clock time, measured from the start of `compute-partitions` to the end of the slowest matrix job, was approximately **1h 24m 45s**. This includes an approximately five-minute scheduling gap between `compute-partitions` and the matrix jobs.

tkobayas:
# Parallel CI Test Run: Computed Image-Producer Rebuild

Run `89788266903` — all jobs completed successfully.

## Job elapsed time

| Job | Java 17 ||| Java 21 |||
|---|---:|---:|---:|---:|---:|---:|
| | upstream | test | **total** | upstream | test | **total** |
| compute-partitions | — | — | **3m 06s** | — | — | — |
| partition0 (optaplanner) | 1m 52s | 9m 42s | **14m 18s** | 1m 50s | 9m 13s | **13m 36s** |
| partition1 (kogito-runtimes) | 5m 09s | 42m 09s | **51m 56s** | 6m 21s | 49m 36s | **58m 50s** |
| partition2 (kogito-apps) | 8m 23s | 1h 03m | **1h 14m 30s** | 7m 04s | 1h 18m | **1h 29m 02s** |
| default (drools + shared) | 5m 29s | 52m 50s | **1h 00m 33s** | 5m 35s | 54m 54s | **1h 02m 50s** |

The partition2 upstream figures include both upstream phases:

- Java 17: parallel upstream 6m 36s + required image-producing modules 1m 47s = 8m 23s
- Java 21: parallel upstream 5m 33s + required image-producing modules 1m 31s = 7m 04s

The workflow wall-clock time, measured from the start of `compute-partitions` to the end of the slowest matrix job, was approximately **1h 32m 16s**. The matrix jobs started approximately five seconds after `compute-partitions` finished.

yesamer:
@tkobayas Very nice work!! It's curious that JDK21 is slower than JDK17 :o

tiagobento:
@tkobayas I wonder if those partitions are the most optimal. It seems we have a big opportunity to break down `kogito-apps` and the `default` partition as well. 

One thing that crossed my mind too: Are we sure that the sum of these partitions result in all the modules being built? In KIE Tools we added a guard clause for it that helped be confident that no module was being forgotten by bugs or misconfiguration.

If we can extract the time each test takes from the logs, we can heuristically determine what's the best partition allocation by number of partitions and members of each partition.

.github/workflows/ci.yaml
tkobayas: outdated

.github/workflows/ci.yaml
tkobayas: outdated

.github/workflows/ci.yaml
tkobayas: outdated

.github/workflows/ci.yaml
tkobayas: outdated

pom.xml
tkobayas: Applied the partition approach.

script/ci/CiComputeBuildScopes.java
```diff
@@ -202,6 +220,102 @@ private static void writeLines(Path out, Collection<String> lines) throws IOExce
...
+        dir = dir.toAbsolutePath().normalize();
+        for (var entry : categories.entrySet()) {
+            for (Path catPath : entry.getValue()) {
+                if (dir.equals(catPath) || dir.startsWith(catPath + "/")) {
+                    return entry.getKey();
+                }
+            }
+        }
+        return "drools";
+    }
```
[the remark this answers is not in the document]
tkobayas: outdated

script/ci/CiComputeBuildScopes.java
```diff
@@ -202,6 +220,102 @@ private static void writeLines(Path out, Collection<String> lines) throws IOExce
...
+                break;
+            }
+            if (!inModules) continue;
+
+            Matcher beginMatcher = beginPattern.matcher(trimmed);
+            if (beginMatcher.matches()) {
+                currentCategory = beginMatcher.group(1);
+                if (!categories.containsKey(currentCategory)) {
+                    System.err.println("ERROR: unknown module category '" + currentCategory
+                            + "' in pom.xml marker. Expected one of: " + EXPECTED_CATEGORIES);
```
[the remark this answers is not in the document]
tkobayas: outdated

script/ci/CiComputeBuildScopes.java
```diff
@@ -166,13 +167,30 @@ public static void main(String[] args) throws Exception {
...
+          .append(" changed=").append(changed.size())
+          .append(" affected=").append(affected.size())
+          .append(" upstream=").append(upstreamAll.size())
+          .append(" ignored=").append(ignored);
+        for (String cat : categories.keySet()) {
+            long catCount = affected.stream()
+                    .filter(ga -> categorizeGa(ga, gaToDir, categories).equals(cat))
+                    .count();
+            sb.append(" affected-").append(cat).append("=").append(catCount);
+        }
```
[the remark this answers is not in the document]
tkobayas: outdated

pom.xml
```diff
@@ -152,6 +152,9 @@
   </build>
 
   <modules>
+    <!-- DO NOT remove the BEGIN/END comments. They are used by CiComputeBuildScopes for parallel CI job split
+         When you add a new module, put it in one of the 4 categories (drools, optaplanner, kogito-runtimes, kogito-apps) -->
```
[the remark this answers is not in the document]
tkobayas: outdated

script/ci/CiComputeBuildScopes.java
tkobayas: Thanks. Applied the partition approach.

.github/actions/ci-setup/action.yaml
tkobayas: Thanks. Applied the partition approach.
Filenames are `partition0.txt`, `partition1.txt`... for easier processing. We can add a comment line like `# optaplanner` for explanation.

Also we have "implicit default" partition which catches:
- Modules in zero explicit partitions' closures (= not referenced by any leaves in the partition files)
- Modules in multiple explicit partitions' closures (shared modules like `kogito-api`)

.github/workflows/ci.yaml
tkobayas: Thanks, yes. I removed the duplication.

script/ci/CiComputeBuildScopes.java
```diff
@@ -166,13 +167,30 @@ public static void main(String[] args) throws Exception {
...
         writeLines(changedOut, changed);
 
+        Map<String, Set<Path>> categories = parseModuleCategories(cwd.resolve("pom.xml"), cwd);
+
+        for (String cat : categories.keySet()) {
+            Set<String> catAffected = affected.stream()
+                    .filter(ga -> categorizeGa(ga, gaToDir, categories).equals(cat))
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
+            writeLines(partitionedPath(affectedOut, cat), catAffected);
+        }
```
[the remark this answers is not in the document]
tkobayas: outdated

script/ci/CiComputeBuildScopes.java
```diff
@@ -241,6 +277,104 @@ private static void writeLines(Path out, Collection<String> lines) throws IOExce
...
+    }
+
+    // The upstream set intentionally includes the partition's own affected modules.
+    // Without them, an upstream module from another partition could fail to resolve
+    // dependencies on this partition's affected modules (e.g., shared module U depends
+    // on affected module V — if V is removed from upstream, building U fails).
+    private static void computePerPartitionUpstream(List<Partition> partitions, DepGraph graph) {
+        for (Partition p : partitions) {
+            if (p.assigned.isEmpty()) continue;
+            p.upstream = DepGraph.traverse(p.assigned, graph.upstreamOf);
```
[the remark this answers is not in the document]
tkobayas: The overlap is intentional: phase 1 compiles affected modules in parallel with tests and Quarkus skipped, so the serial test pass largely reuses those outputs. Removing them would shift compilation into the slower serial pass and could also break upstream or image-producer builds whose snapshot dependencies are assigned to the same partition. Since phase 1 takes only 2–7 minutes versus 9–79 minutes for testing, the duplicated overhead is small.

tkobayas:
Thanks. I'm still working out the details. I'll post an update with a summary of the new changes.

.github/workflows/ci-parallel.yaml
```diff
@@ -0,0 +1,178 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: 'CI :: Build - Parallel'
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, closed, labeled]
```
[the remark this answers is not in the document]
tkobayas: Rather I'm removing `labeled` from `types`, so `labeled`/`unlabeled` doesn't trigger CI. CI will be triggered by usual ways (opening a PR, push a new commit etc.) Label is just to decide sequential or parallel.

tkobayas:
## PR update: Partition files approach

Thanks for the review @tiagobento. I've reworked the PR based on your feedback. Here's a summary of the changes.

### Label-based switching

The parallel CI workflow (`ci-parallel.yaml`) is activated by attaching the **`parallel ci`** label to a PR. Without the label, the original sequential `ci.yaml` runs as before. This lets us opt in per-PR and makes it easy to revert if Apache Infra raises concerns about GHA runner quota — each partition job rebuilds upstream modules independently, so total runner-minutes increase even though wall-clock time decreases.

**Keeping both `ci.yaml` and `ci-parallel.yaml` would be a maintenance burden. We may finally combine them, but keep them separated for a while as `ci.yaml` is the stable CI and `ci-parallel.yaml` is the opt-in experimental CI (for only people who hit many affected modules and want a quick CI cycle). WDYT?**

### Partition files

Replaced the pom.xml `BEGIN/END` markers with partition files in `.github/supporting-files/ci/partitions/`:

- `partition0.txt` — optaplanner (14 leaf modules)
- `partition1.txt` — kogito-runtimes (87 leaf modules)
- `partition2.txt` — kogito-apps (39 leaf modules)
- Implicit default — drools + shared modules (no file needed)

Adding or removing a partition is just adding or removing a file. The GHA workflow uses a dynamic matrix (`compute-partitions` job discovers partition files and outputs a JSON array, `test` job uses `partition × java × os` cross-product), so no YAML changes are needed.

**Exclusive assignment**: each affected module goes to a partition only if it's in exactly one partition's transitive closure. Shared modules (like `kie-api`, `kogito-api`) go to the implicit default. This avoids redundant tests — each module is tested in exactly one job.

The current 4 partitions are just a starting point. We can split further (e.g., `kogito-data-index`, `kogito-jobs-service`) by adding more partition files.

### 2-phase upstream build (image-producer modules)

The upstream build uses `-T 1C` for parallelism, but Quarkus runtime extensions have an implicit dependency on their `-deployment` counterpart that Maven doesn't capture. To avoid race conditions:

**Phase 1** — Build all upstream modules in parallel with Quarkus disabled:
```
-T 1C -Dquarkus.build.skip=true -Ddisable.quarkus.plugin=true -Dskip.quarkus.image.assembly=true
```

**Phase 2** — Rebuild image-producing modules normally (without the skip flags). These modules need Quarkus augmentation to produce `target/quarkus-app` and container image assemblies required by integration tests. The module list is maintained in `image-producers.txt`.

This 2-phase approach would also benefit `ci.yaml`, which currently uses `-T 1C` in its upstream build and has the same latent Quarkus race condition.

The five image-producing modules that use `maven-assembly-plugin` to package `target/quarkus-app` require a small POM change: a `skip.quarkus.image.assembly` property to skip the assembly when Quarkus is disabled (otherwise the archive fails as empty).

### What this PR does NOT do

- **`-T 1C` for the test phase** — parallel test execution with surefire/failsafe is risky (port conflicts, flaky tests). This could be explored in a separate PR.

### CI results

All checks green (run `89788266903`). Wall-clock time comparison (all modules affected):

| Job | Java 17 ||| Java 21 |||
|---|---:|---:|---:|---:|---:|---:|
| | upstream | test | **total** | upstream | test | **total** |
| compute-partitions | — | — | **3m 06s** | — | — | — |
| partition0 (optaplanner) | 1m 52s | 9m 42s | **14m 18s** | 1m 50s | 9m 13s | **13m 36s** |
| partition1 (kogito-runtimes) | 5m 09s | 42m 09s | **51m 56s** | 6m 21s | 49m 36s | **58m 50s** |
| partition2 (kogito-apps) | 8m 23s | 1h 03m | **1h 14m 30s** | 7m 04s | 1h 18m | **1h 29m 02s** |
| default (drools + shared) | 5m 29s | 52m 50s | **1h 00m 33s** | 5m 35s | 54m 54s | **1h 02m 50s** |

Wall-clock: **~1h 32m** (vs ~3h sequential).

.github/workflows/ci-parallel.yaml
```diff
@@ -0,0 +1,178 @@
...
+        if: always()
+        shell: bash
+        env:
+          MERMAID_EXPANDED: ${{ matrix.java == 21 }}
+          MATRIX_OS: ${{ matrix.os }}
+          MATRIX_JAVA: ${{ matrix.java }}
+        run: |
+          MAVEN_PL_AFFECTED=$(paste -sd, "$MAVEN_PL_AFFECTED_FILE" 2>/dev/null || true) \
+          MAVEN_PL_UPSTREAM=$(paste -sd, "$MAVEN_PL_UPSTREAM_FILE" 2>/dev/null || true) \
+          MAVEN_PL_CHANGED=$(paste -sd, "$MAVEN_PL_CHANGED_FILE" 2>/dev/null || true) \
```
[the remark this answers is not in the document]
tkobayas: Fixed

tkobayas:
Hi @tiagobento 

> I wonder if those partitions are the most optimal. It seems we have a big opportunity to break down kogito-apps and the default partition as well.

These partition files are just a starting point. Certainly, we can break down kogito-apps and the default partition further, and that flexibility is one of the advantages of the partition approach. However, the current partitions look reasonably balanced (three of them take around one hour). So, could we do further tuning in a separate PR?

> One thing that crossed my mind too: Are we sure that the sum of these partitions result in all the modules being built? In KIE Tools we added a guard clause for it that helped be confident that no module was being forgotten by bugs or misconfiguration.

Thanks, I added the check following the kie-tools approach.

> If we can extract the time each test takes from the logs, we can heuristically determine what's the best partition allocation by number of partitions and members of each partition.

Yes, we can, and I think it would be an interesting exercise. Could we do that in a separate PR?

tkobayas:
@yesamer 
> It's curious that JDK21 is slower than JDK17 :o

Job elapsed times are not very stable, so JDK 21 is sometimes slower than 17. However, from what I've seen, JDK 21 is faster on average. :)

tiagobento:
@tkobayas That's awesome! Thanks a lot for all the work here. Let me spend a few hours digesting the code changes, so that I can do a more in-depth code review.

In the mean time, I think we can establish ci.yaml to be used for commits to main, while the new partitioned CI can be used for PR checks. This way we have a self regulating mechanism to make sure our builds are stable across PR checks and commits to main, while keeping fast execution for PR checks, and guaranteed stability for commits to main.

If either fails when the other succeeds, that's a signal something is wrong between them that we need to address.

tkobayas:
@tiagobento 
> In the mean time, I think we can establish ci.yaml to be used for commits to main, while the new partitioned CI can be used for PR checks. This way we have a self regulating mechanism to make sure our builds are stable across PR checks and commits to main, while keeping fast execution for PR checks, and guaranteed stability for commits to main.

Sure, it sounds good. => Made `ci-parallel.yaml` as default on PR check. `ci.yaml` is used on push to main.

PR #6871, opened 2026-08-04
performance(kie-dmn-feel): Performance regression in Drools v10+ with DMN models  

### LINKED ISSUE #6870: Performance regression in Drools v10+ with DMN models
### Describe the bug

We use Drools with DMN models to evaluate data validity rules on massive data sets (400k+ rows). Our performance bottleneck is the Drools engine itself, the vast majority of the execution time is spent with FEEL evaluating and processing.

When we upgraded from 8.44.2.Final to 10.x.x, we noticed a significant drop in performance in our workload -- for example, runs that previously took around 200s now take over 315s. With another set of inputs, the runs went from around 120s to 200s.

We are fairly certain that the regression is within Drools in version 10.x.x onwards. We tested multiple times and eliminated issues on our end, and found a few examples of changes that were not in release notes that may explain at least some of the regressions.

Do you have automated performance tests, and have you noticed regressions in performance like we have?

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

Linux 6.8.0-124-generic #124~22.04.1-Ubuntu x86_64 x86_64 x86_64 GNU/Linux

### Output of `java -version`

openjdk version "25.0.3" 2026-04-21 LTS

_No response_

_No response_

Apache Maven 3.9.16

### Additional information

We used AI to explore past changes between 8.44.2.Final and 10.2.0. It found the following issue which we confirmed by performance testing again after fixing the issue; it also found other issues, but they are not as straight forward and we didn't thoroughly test them. The other issues were related to e.g. unnecessary iteration over lists.

- Commit 8dac313955a: XQuery implementation instantiates the processor & compiler on every invocation

yesamer:
Closes https://github.com/apache/incubator-kie/issues/6870

### Problem

After upgrading from 8.44.x to 10.x, users observed a **~60% increase in execution time** on
large DMN workloads (e.g. 200 s → 315 s on 400k-row batches). The bottleneck was traced to
`XQueryImplUtil`, which is called on every invocation of the FEEL `matches()` and `replace()`
functions.

Two independent inefficiencies were introduced:

1. **`new Processor(false)` on every call** (commit `8dac313`): Saxon's `Processor` constructor
   initialises the entire Saxon configuration and performs a license check. Creating it per call
   cost ~60 ms per 2,000 rows on Saxon 12.x. Saxon's own documentation recommends one `Processor`
   instance per JVM.

2. **Multi-pass string escaping**: `escapeXmlCharactersReferencesForXPath()` scanned the input
   string up to 11 times (one regex guard scan + one `contains` + one `replace` per special
   character), creating up to 5 intermediate `String` allocations per call. This is invoked 2–3
   times per `matches()`/`replace()` call, so the cost multiplied across the batch.

### Changes

**`XQueryImplUtil.java`**
- Extracted `Processor` and `XQueryCompiler` into `private static final` fields — one instance
  per JVM, shared across all calls. Both are thread-safe per Saxon's API contract.
- Rewrote `escapeXmlCharactersReferencesForXPath()` as a single-pass `StringBuilder` loop with
  lazy allocation: no `StringBuilder` is created at all when the input contains no special
  characters (the common case in DMN), and the original `String` reference is returned unchanged.
  The `Pattern` field and `java.util.regex` import are no longer needed and have been removed.

**`XQueryImplUtilTest.java`**
- Added 15 new test cases covering: explicit empty flags, multi-flag combinations (`si`, `mi`),
  all five XML special characters in input and pattern, special characters in replacement strings,
  and backreference syntax — ensuring the full escape→embed→evaluate pipeline is exercised
  end-to-end against Saxon.

### Measured impact (Saxon 12.10, 2,000 rows/run, median of 5 runs)

| Workload | Original (`main`) | Fixed | Speedup |
|---|---|---|---|
| `matches()` — plain input | 66 ms | 9 ms | **7.7×** |
| `matches()` — XML special chars in input | 67 ms | 10 ms | **6.9×** |
| `matches()` — unique input per row | 67 ms | 9 ms | **7.6×** |
| `replace()` — plain input | 86 ms | 7 ms | **12.8×** |
| `replace()` — XML special chars in input | 68 ms | 10 ms | **6.8×** |
| `replace()` — unique input per row | 65 ms | 9 ms | **7.7×** |

Extrapolated to a 400k-row batch: **~13–17 s → ~1–2 s** per workload type.

_We used a manual warm-up benchmark to measure the regression. The results show a consistent 7–12× improvement across all workload profiles. Results in production could be different._


kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/XQueryImplUtil.java
```diff
@@ -50,9 +58,7 @@ public static String executeReplaceFunction(String input, String pattern, String
 
      static <T> T evaluateXQueryExpression(String expression, Class<T> expectedTypeResult) {
          try {
-             Processor processor = new Processor(false);
-             XQueryCompiler compiler = processor.newXQueryCompiler();
-             XQueryExecutable executable = compiler.compile(expression);
+             XQueryExecutable executable = COMPILER.compile(expression);
```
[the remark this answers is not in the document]
gitgabrio: @yesamer wdyt ? ^^

tkobayas:
Filed https://github.com/apache/incubator-kie/issues/6876 for `CI :: Build` failure.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/XQueryImplUtil.java
```diff
@@ -26,11 +26,27 @@
...
+     * creates a plain Home Edition configuration.
+     */
+    private static final Processor PROCESSOR = new Processor(false);
+
+    /**
+     * Single XQueryCompiler instance. XQueryCompiler is reusable and may in principle
+     * be used concurrently in multiple threads. In practice, concurrent compilations share
+     * the same ErrorReporter, making it difficult to associate error messages with specific
+     * compilations. Since errors here are immediately wrapped and re-thrown as
+     * {@link IllegalArgumentException}, this is not a concern.
```
tkobayas: I think the error message could still be replaced by another thread in a concurrent scenario, even if it's wrapped in an `IllegalArgumentException`. So I think "`this is not a concern`" is too strong a statement.

We may instead acknowledge that we're using a shared instance for performance, with the trade-off that error messages may not always be attributed correctly.

tkobayas:
The fix looks good, but we need to run CI before merging.

Btw, do we have a benchmark for the targeted FEEL operations? It would be greater if we can confirm the improvement (not mandatory).

tkobayas (approved):

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/XQueryImplUtil.java
yesamer: @tkobayas Thank you, I applied your suggestions and a Benchmark.

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/XQueryImplUtil.java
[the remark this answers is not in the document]
yesamer: XQueryCompiler docs says Concurrent use is permitted, but error messages may not be attributed to the correct thread under concurrent error conditions. See Saxon s9api {@link net.sf.saxon.s9api.XQueryCompiler} Javadoc. In this case should be safe.

gitgabrio (commented):
(removed)

tiagobento:
Re-triggered the PR checks after

- #6877 

was merged.

tiagobento:
@yesamer Just FYI, you don't need to merge the base branch with the PR branch anymore. Just re-running after things land on `main` is enough. We have a custom action now that simulates a "Squash and merge" for the PR, and it uses `main` at the moment the CI starts. That's the safest thing to do to prevent semantic conflicts before using Merge Queues.

tkobayas:
Now I see only 1 job is in_progress and 3 jobs (including this PR) are queued for `CI :: Build`:
https://github.com/apache/incubator-kie/actions/workflows/ci.yaml?query=is%3Ain_progress
https://github.com/apache/incubator-kie/actions/workflows/ci.yaml?query=is%3Aqueued
Hmm, is it a temporal resource issue?

tiagobento:
@tkobayas I think GitHub Actions glitched. A new commit in this PR should fix it.

tkobayas:
This PR check looks stuck in a queue : https://github.com/apache/incubator-kie/actions/workflows/ci.yaml?query=is%3Aqueued . Even "Cancel run" doesn't work. I think you need to add an empty commit in this PR to re-trigger the CI. @yesamer

kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/XQueryImplUtil.java
```diff
@@ -66,22 +71,38 @@ static <T> T evaluateXQueryExpression(String expression, Class<T> expectedTypeRe
...
-        if (input != null && XML_CHARACTER_REFERENCES_PATTERN.matcher(input).find()) {
-            input = input.contains("&") ? input.replace("&", "&amp;") : input;
-            input = input.contains("\"") ? input.replace("\"",  "&quot;") : input;
-            input = input.contains("'") ? input.replace("'",  "&apos;") : input;
-            input = input.contains("<") ? input.replace("<",  "&lt;") : input;
-            input = input.contains(">") ? input.replace(">",  "&gt;") : input;
+        if (input == null) {
+            return null;
+        }
+        StringBuilder sb = null;
```
gitgabrio, line 85: HI @yesamer 
IIUC the logic, I seems it would be possible to avoid the usage of `null` semantic; i.e.
1. instantiate directly `sb`
2. here -> 
`return sb != null ? sb.toString() : input;` 
replace with
`return sb.isEmpty() ? input : sb.toString();`

It would a bit clearer (and less code/if): wdyt ?

yesamer: @gitgabrio Your suggestion is more elegant, but in terms of performance that means to instantiate a StringBuilder for any call, even if is not required because the string doesn't contain any special char.

gitgabrio (approved):

PR #6873, opened 2026-08-05
[NO-ISSUE] : CVE-2026-56624: upgrade Apache MINA SSHD (sshd-common/sshd-core) to 2…

athirakm94:
Summary 
CVE-2026-56624: upgrade Apache MINA SSHD (sshd-common/sshd-core) to 2.19.0

quarkus-bom:3.27.4.1 (via wildfly-elytron-parent:2.6.5.Final) hard-pins org.apache.sshd:sshd-common and org.apache.sshd:sshd-core to 2.12.1. No newer Quarkus BOM version is available that resolves this transitively, so explicit dependencyManagement overrides are added to the three build-parent/BOM files that directly import quarkus-bom.

Files modified:
- kie-quarkus-build-parent/pom.xml
- kogito-quarkus/bom/pom.xml
- optaplanner-build/optaplanner-build-parent/pom.xml


kogito-quarkus/bom/pom.xml
```diff
@@ -104,6 +106,17 @@
...
+      <dependency>
+        <groupId>org.apache.sshd</groupId>
+        <artifactId>sshd-common</artifactId>
+        <version>${version.org.apache.sshd}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.sshd</groupId>
+        <artifactId>sshd-core</artifactId>
+        <version>${version.org.apache.sshd}</version>
+      </dependency>
```
yesamer: I guess the changes in quarkus-bom are not necessary //cc @gitgabrio

gitgabrio: Yes. If that dependencyManagemnt is declared with the same `version` variable name in a parent pom, those are not needed - only the variable override -if any.
But in this PR, I see the same variable name/value and dependencyManagement also in a parent pom, so that is strange because there is not even an override

gitgabrio (approved):
Thanks @athirakm94 !

drools-reliability/drools-reliability-tests/pom.xml
```diff
@@ -129,8 +129,18 @@
...
+        <exclusion>
+          <groupId>org.apache.sshd</groupId>
+          <artifactId>sshd-common</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.sshd</groupId>
+      <artifactId>sshd-common</artifactId>
+      <version>2.19.0</version>
```
gitgabrio: HI @athirakm94 
There is an enforcer to avoid version declaration outside the `dependencyManagement` tag.
Moreover, there is another one to limit the creation of `dependencyManagment` tags in general.
The reason for the above is to avoid the uncontrolled multiplication of version declarations, an incredible confusion for which we wasted a lot of time over the years and we invested lot of effort to fix.
So, please, declare this in the _**kie-parent**_ `dependencyManagement` tag. If an override is required, then simply use the property-change approach.
Thanks!

gitgabrio (changes requested):
Hi @athirakm94 : please follow the guidelines about putting all the version declarations in one single place - the kie-server parent, eventually overriding them, with the property approach, only when needed. Thanks!

drools-reliability/drools-reliability-tests/pom.xml
athirakm94: Thanks @gitgabrio . I’ll align this with the repo rules: no standalone version declaration in drools-reliability/drools-reliability-tests/pom.xml, and centralize the version in kie-parent/pom.xml <dependencyManagement>.

gitgabrio (approved):
THanks @athirakm94 !

josedee (approved):

Kusuma04-dev (approved):

PR #6877, opened 2026-08-06
[incubator-kie-6876] CI :: Build failure : NPE in maven-artifact-plug…

### LINKED ISSUE #6876: CI :: Build failure : NPE in maven-artifact-plugin:compare with `-pl` builds
# NPE in maven-artifact-plugin:compare with `-pl` builds

## Symptom

CI build fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-artifact-plugin:3.5.1:compare (compare)
  on project kie-parent: Execution compare of goal
  org.apache.maven.plugins:maven-artifact-plugin:3.5.1:compare failed:
  Cannot invoke "org.apache.maven.project.MavenProject.getBuild()" because "root" is null
```

Reproducible with 3.5.2. **Fixed in 3.6.1** (see Solution below).

https://issues.apache.org/jira/browse/MARTIFACT-79

## Root Cause

The CI workflow (`.github/workflows/ci.yaml`, line 195) runs:

```bash
mvn ... -Dfull -Dreproducible -pl "$pl" install
```

where `$pl` is the list of affected modules (e.g. `kie-parent,kie-api,...`). The root pom (`drools-parent`) is **not** included in `$pl`.

The `-Dreproducible` property activates the `reproducible-build` profile (root `pom.xml`, line 275), which binds `maven-artifact-plugin:compare` to the `install` phase.

In a multi-module reactor, the `compare` goal aggregates results. For the **last** module in the reactor, `CompareMojo.execute()` calls:

```java
checkAgainstReference(artifacts, session.getProjects().size() == 1);
```

Since there are multiple modules, `mono = false`, which leads to:

```java
// CompareMojo.java line 126 (3.5.2)
MavenProject root = mono ? project : getExecutionRoot();
File referenceDir = new File(root.getBuild().getDirectory(), "reference");  // NPE here
```

`getExecutionRoot()` iterates over `session.getProjects()` looking for a project where `isExecutionRoot() == true`. With `-pl`, the root pom (`drools-parent`) is excluded from the reactor, so no project has `isExecutionRoot() == true`, and the method returns `null`.

In 3.6.x the code changed to `session.getTopLevelProject()`, which returns the first project in the reactor (e.g. `kie-parent`) rather than null when the root pom is excluded via `-pl`. This resolves the NPE.

### Code path (3.5.x — broken)

```
AbstractBuildinfoMojo.execute()
  → mono = session.getProjects().size() == 1   // false (multiple -pl modules)
  → last module: calls checkAgainstReference(artifacts, false)
    → CompareMojo.checkAgainstReference()
      → root = getExecutionRoot()               // returns null
      → root.getBuild()                          // NPE
```

```java
// AbstractBuildinfoMojo.java line 342
protected MavenProject getExecutionRoot() {
    for (MavenProject p : session.getProjects()) {
        if (p.isExecutionRoot()) {
            return p;
        }
    }
    return null;  // ← no guard against null
}
```

## Version comparison

| Version | Root resolution method          | NPE with `-pl`? |
|---------|--------------------------------|-----------------|
| 3.5.1   | `getExecutionRoot()`           | Yes             |
| 3.5.2   | `getExecutionRoot()`           | Yes (confirmed) |
| 3.6.1   | `session.getTopLevelProject()` | **No** (confirmed) |

In 3.5.x, `getExecutionRoot()` returns `null` when the root pom is not in the reactor.
In 3.6.x, `session.getTopLevelProject()` returns the first project in the reactor, avoiding the NPE.

## Affected CI step

```yaml
# .github/workflows/ci.yaml line 189-195
- name: "PR CHECK :: BUILD :: Changed and affected modules"
  if: github.event_name == 'pull_request'
  shell: bash
  run: |
    pl=$(paste -sd, "$MAVEN_PL_AFFECTED_FILE")
    [ -z "$pl" ] && echo "No affected modules. Skipping." && exit 0
    mvn --batch-mode --no-transfer-progress -fae -Dsurefire.redirectTestOutputToFile=true -Dfull -Dreproducible -pl "$pl" install
```

Note: the full CI build on `push` (line 200) runs **without** `-pl` and works correctly:

```yaml
- name: "CI :: BUILD :: Full"
  if: github.event_name == 'push'
  run: mvn ... -Dfull -Dreproducible install
```

## Solution

Upgrade `maven-artifact-plugin` to **3.6.1**. No CI workflow changes needed.

In root `pom.xml`:

```xml
<version.maven.artifact.plugin>3.6.1</version.maven.artifact.plugin>
```

### Additional required change

Version 3.6.1's `check-buildplan` goal has an updated plugin database that correctly detects `maven-remote-resources-plugin:3.2.0` as non-reproducible (inherited from the Apache parent POM). Upgrade it to 3.3.0:

```xml
<version.maven-remote-resources-plugin>3.3.0</version.maven-remote-resources-plugin>
```

Without this, the build fails with:

```
[ERROR] plugin with non-reproducible output: org.apache.maven.plugins:maven-remote-resources-plugin:3.2.0, require minimum 3.3.0
```

### Alternative: Add the root pom to `-pl`

If upgrading the plugin is not feasible, adding `.` (the root pom) to `-pl` in the CI workflow also fixes the NPE with 3.5.x:

```bash
mvn ... -Dfull -Dreproducible -pl ".,$pl" install
```

This ensures `drools-parent` is in the reactor so `getExecutionRoot()` finds it. The root pom is `pom`-packaging only, so build overhead is negligible.

### Confirmed

`mvn -Dreproducible -pl kie-parent,kie-api install` succeeds with 3.6.1 **without** adding `.` to `-pl`:

```
[INFO] [Reproducible Builds] rebuild comparison result: 6 files match
[INFO] BUILD SUCCESS
```

tkobayas:
…in:compare with `-pl` builds

closes https://github.com/apache/incubator-kie/issues/6876

### DIFF
--- a/kie-dmn/kie-dmn-ruleset2dmn-parent/kie-dmn-ruleset2dmn-cli/pom.xml
+++ b/kie-dmn/kie-dmn-ruleset2dmn-parent/kie-dmn-ruleset2dmn-cli/pom.xml
@@ -63,7 +63,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
-        <version>3.5.2</version>
         <executions>
           <execution>
             <phase>package</phase>
--- a/kie-parent/pom.xml
+++ b/kie-parent/pom.xml
@@ -314,7 +314,7 @@
     <!-- OpenRewrite plugin -->
     <version.rewrite.maven.plugin>4.42.0</version.rewrite.maven.plugin>
     <version.rewrite.testing.frameworks>1.22.0</version.rewrite.testing.frameworks>
-    <version.shade.plugin>3.3.0</version.shade.plugin>
+    <version.shade.plugin>3.5.3</version.shade.plugin>
     <version.shared.utils>3.4.2</version.shared.utils>
     <!-- simple-jndi is a small library that helps us avoid JNDI error messages during testing -->
     <version.simple-jndi>0.11.4.1</version.simple-jndi>
--- a/kie-parent/pom.xml
+++ b/kie-parent/pom.xml
@@ -314,7 +314,7 @@
     <!-- OpenRewrite plugin -->
     <version.rewrite.maven.plugin>4.42.0</version.rewrite.maven.plugin>
     <version.rewrite.testing.frameworks>1.22.0</version.rewrite.testing.frameworks>
-    <version.shade.plugin>3.5.3</version.shade.plugin>
+    <version.maven-shade-plugin>3.5.3</version.maven-shade-plugin>
     <version.shared.utils>3.4.2</version.shared.utils>
     <!-- simple-jndi is a small library that helps us avoid JNDI error messages during testing -->
     <version.simple-jndi>0.11.4.1</version.simple-jndi>
@@ -2873,6 +2873,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
+          <version>${version.maven-shade-plugin}</version>
         </plugin>
         <!-- This plugin's configuration is used to store Eclipse m2e settings only.
              It has no influence on the Maven build itself. -->
--- a/kogito-build/kogito-build-no-bom-parent/pom.xml
+++ b/kogito-build/kogito-build-no-bom-parent/pom.xml
@@ -109,7 +109,7 @@
     <version.rpkgtests.maven.plugin>1.0.0</version.rpkgtests.maven.plugin>
     <version.resources.plugin>3.1.0</version.resources.plugin>
     <version.site.plugin>3.7.1</version.site.plugin>
-    <version.shade.plugin>3.5.3</version.shade.plugin>
+    <version.maven-shade-plugin>3.5.3</version.maven-shade-plugin>
     <version.sonar.plugin>3.6.1.1688</version.sonar.plugin>
     <version.source.plugin>3.2.1</version.source.plugin>
     <version.surefire.plugin>3.3.1</version.surefire.plugin> <!-- minimum required by JUnit 5 -->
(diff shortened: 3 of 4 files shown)


yesamer (approved):

tkobayas:
Need to bump maven-shade-plugin
```
2026-08-06T02:55:46.1893570Z [INFO] --- artifact:3.6.1:check-buildplan (check-buildplan) @ drools-drlonyaml-schemagen ---
2026-08-06T02:55:46.1894839Z [ERROR] plugin with non-reproducible output: org.apache.maven.plugins:maven-shade-plugin:3.5.2, require minimum 3.5.3
```

klbowers:
Thank you so much for sorting this out so quickly! @tkobayas  Really appreciate it :) Seems like kie-dmn/kie-dmn-ruleset2dmn-parent/kie-dmn-ruleset2dmn-cli/pom.xml has 3.5.2 hardcoded and the kie-parent's pluginManagement declaration is missing <version>${version.shade.plugin}</version> so the 3.5.3 property was never actually applied

This is a problem though.
`2026-08-06T09:13:48.7047975Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-artifact-plugin:3.6.1:compare (compare) on project integration-tests-jobs-service-springboot: Rebuilt artifacts are different from reference -> [Help 1]`
Since we no longer do daily snapshot deploys, compare will always fail with missing reference file for any new module that hasn't been published yet. It might be best removing the compare execution from the reproducible-build profile and just keeping the check-buildplan, as that still enforces all plugins are configured for reproducible output without needing a remote reference. WDYT?

tkobayas:
@kbowers-ibm Thank you. Fixed:
- maven-shade-plugin version
- remove the compare execution 

We will need to use `-Dreproducible.compare` for release CI.

klbowers (approved):

tiagobento:
Thank you @tkobayas! ❤️

tiagobento (approved):

PR #6886, opened 2026-08-07
[NO-ISSUE]: Fix Jackson-databind 3.1.4 → 3.1.5

AnnJoy23:
## Summary

Fixes **CVE-2026-59889** (CVSS 6.5, Medium) — `jackson-databind` `@JsonUnwrapped`/`@JsonView` view bypass.

## Root Cause

`spring-boot-dependencies:4.0.7` imports `tools.jackson:jackson-bom:3.1.4` which resolves `jackson-databind:3.1.4` across the Spring Boot transitive tree.

Dependency path (INTRA-TRANSITIVE):
```
spring-boot-dependencies:4.0.7
  └─ tools.jackson:jackson-bom:3.1.4
       └─ [drools|jbpm|kie-addons]-spring-boot-starter
            → spring-boot-starter-web → spring-boot-starter-jackson
              → spring-boot-jackson → jackson-databind:3.1.4  ← CVE-2026-59889
```

## Fix

Pin tools.jackson.core:jackson-databind to 3.1.5 directly in kogito-springboot/bom/pom.xml. This overrides the vulnerable jackson-databind:3.1.4 carried by spring-boot-dependencies:4.0.7. Only jackson-databind is pinned as it is the only artifact affected by CVE-2026-59889.

**File changed:** `springboot/bom/pom.xml`

## Verification

mvn dependency:tree -Dverbose -pl kogito-apps-springboot/integration-tests-jobs-service-springboot -Dincludes="tools.jackson*" confirms jackson-databind resolves to 3.1.5 ✓

## TODO

Remove force-pin once `spring-boot-dependencies` upgrades its `jackson-bom.version` to `>= 3.1.5`.

## References
- https://nvd.nist.gov/vuln/detail/CVE-2026-59889

### DIFF
--- a/kogito-springboot/bom/pom.xml
+++ b/kogito-springboot/bom/pom.xml
@@ -53,20 +53,16 @@
          https://logging.apache.org/security.html#CVE-2026-49844 -->
     <version.org.apache.logging.log4j>2.26.1</version.org.apache.logging.log4j>
     <!-- CVE-2026-59889: force-pin jackson-databind to 3.1.5.
-         spring-boot-dependencies:4.0.7 imports tools.jackson:jackson-bom:3.1.4 which carries
-         the vulnerable jackson-databind:3.1.4. Importing jackson-bom:3.1.5 before
-         spring-boot-dependencies overrides the transitive version for all spring-boot-* carriers.
-         Remove this pin once spring-boot-dependencies imports jackson-bom >= 3.1.5. -->
-    <version.tools.jackson>3.1.5</version.tools.jackson> <!-- CVE-2026-59889: override jackson-bom before spring-boot-dependencies import -->
+         spring-boot-dependencies:4.0.7 carries the vulnerable jackson-databind:3.1.4.
+         Remove this pin once spring-boot-dependencies imports jackson-databind >= 3.1.5. -->
+    <version.tools.jackson.core>3.1.5</version.tools.jackson.core>
   </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>tools.jackson</groupId>
-        <artifactId>jackson-bom</artifactId>
-        <version>${version.tools.jackson}</version>
-        <type>pom</type>
-        <scope>import</scope>
+        <groupId>tools.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>${version.tools.jackson.core}</version>
       </dependency>
       <dependency>
         <groupId>org.springframework.cloud</groupId>
--- a/kogito-springboot/bom/pom.xml
+++ b/kogito-springboot/bom/pom.xml
@@ -52,9 +52,22 @@
          Override the version property inherited from kie-parent to 2.26.1 (minimum 2.25.5).
          https://logging.apache.org/security.html#CVE-2026-49844 -->
     <version.org.apache.logging.log4j>2.26.1</version.org.apache.logging.log4j>
+    <!-- CVE-2026-59889: force-pin jackson-databind to 3.1.5.
+         spring-boot-dependencies:4.0.7 imports tools.jackson:jackson-bom:3.1.4 which carries
+         the vulnerable jackson-databind:3.1.4. Importing jackson-bom:3.1.5 before
+         spring-boot-dependencies overrides the transitive version for all spring-boot-* carriers.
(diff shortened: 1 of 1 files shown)


yesamer (approved):

gitgabrio:
Hi @AnnJoy23 : could you please check failures ? I do not remember exactly, but we introduced  some customization about the jakson usage in springboot and I'm not sure if this could break them

AnnJoy23:
Hi @gitgabrio ,
Root cause is the image-building modules live under kogito-apps-quarkus which is in a completely different BOM chain (kogito-apps-quarkus → kogito-apps-quarkus-bom), so they are NOT downstream of kogito-spring-boot-bom and are therefore NOT included in the affected pass. The integration tests get built but the Docker images they need are never built.

gitgabrio:
> Hi @gitgabrio , Root cause is the image-building modules live under kogito-apps-quarkus which is in a completely different BOM chain (kogito-apps-quarkus → kogito-apps-quarkus-bom), so they are NOT downstream of kogito-spring-boot-bom and are therefore NOT included in the affected pass. The integration tests get built but the Docker images they need are never built.

Thanks @AnnJoy23 
But, IIUC, on one side your modification should impact only springboot chain, but the breaking modules belongs to the quarkus one... Do I understood correctly ? Why this happen ? 🤔 
Anyway, I think we should not  merge PRs whose build is completely failing

@yesamer @tkobayas wdyt ?

gitgabrio:
Hi @AnnJoy23 
If I may suggest, the way to check actual dependencies on a maven project is:

`mvn dependency:tree -Dverbose {eventual_flags} {-Dincludes}`
1. the `dependency:tree` shows all the inheritance tree
2. `verbose` is used to also show dependencies/versions that have been omitted for some reason
3. the `{eventual_flags} `,  i.e. the flags actually used at build time, are needed because some dependency could be bring in by profiles or some other reason
4. `includes` it is useful to reduce the scope of the tree

In this case, one thing to be aware is the potential mix of different jackson libraries (e.g.)

This one -> 
`mvn help:effective-pom -pl springboot/bom`, IMO, is a bit unreliable

M2C

AnnJoy23:
Hi @gitgabrio ,
Thank you for the suggestion — running mvn dependency:tree -Dverbose -pl kogito-apps-springboot/integration-tests-jobs-service-springboot -Dincludes="tools.jackson*" confirms the concern:

tools.jackson.core:jackson-databind:jar:3.1.5:compile (version managed from 3.1.4)
tools.jackson.core:jackson-core:jar:3.1.4:compile (version managed from 3.1.5)

The pin is scoped to jackson-databind since the CVE (CVE-2026-59889) affects only jackson-databind. jackson-core:3.1.4 has no known vulnerability at this time. This pin is temporary and will be removed once spring-boot-dependencies ships with jackson-databind >= 3.1.5.

gitgabrio (approved):
THanks @AnnJoy23

yesamer (approved):

gitgabrio:
@yesamer This PR provide the standard approach, i.e. use the order of imports in dependencyManagement so that the first one is selected.
I was wondering, for future, to check if it would be possible to achieve the same result including the exclusion inside the import, i.e., instead of ->

```xml
<dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>tools.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
        <version>${version.tools.jackson.core}</version>
      </dependency>
      <dependency>
        <groupId>org.springframework.cloud</groupId>
        <artifactId>spring-cloud-dependencies</artifactId>
        ...
 ```
 (that relies on the order, something a bit frail)
 that -> 
 
 ```xml
 <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.springframework.cloud</groupId>
        <artifactId>spring-cloud-dependencies</artifactId>
        <exclusions>
         <exclusion>
           <groupId>tools.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
        ....
        </dependency>
      <dependency>
       <groupId>tools.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
        <version>${version.tools.jackson.core}</version>
      </dependency>
 ```
 
 Again, just a suggestion for the future, not for this one. And, it should be verified: wdyt ?

yesamer:
@gitgabrio Yes, that makes sense.

PR #6895, opened 2026-08-12
[incubator-kie-issues#2361] Add missing build-ordering dependencies for Data Index integration tests in partial builds

AthiraHari77:
Needed by : https://github.com/apache/incubator-kie/pull/6861

**Problem**

The Data Index integration tests (`integration-tests-data-index-service-quarkus`,
`integration-tests-data-index-service-springboot`, and
`kogito-addons-quarkus-data-index-mongodb-persistence-integration-tests-process`)
fail when run as part of a partial CI build.

These ITs start real Data Index service instances via Testcontainers. The container
images are produced by the `data-index-service-mongodb`, `data-index-service-inmemory`,
and `data-index-service-postgresql` modules. However, those service modules were never
declared as dependencies of the IT modules, so the partial build's dep-graph has no
edge pulling them into the upstream pass — the service binaries are stale or missing
at runtime when the ITs execute.

This does not surface in a full build because reactor ordering always builds the
service modules first. It only appears in partial builds, where build scope is
determined entirely by declared dependency edges.

**Fix**

Add build-ordering sentinel `<dependency>` entries (the same pattern already used in
`kogito-apps-integration-tests/integration-tests-data-index-service/pom.xml`) to the
three affected IT modules. These are `type=pom`, `scope=provided`, with all transitive
dependencies excluded — nothing is added to the classpath, only the build ordering is
corrected.

### DIFF
--- a/kogito-apps-bom/pom.xml
+++ b/kogito-apps-bom/pom.xml
@@ -148,7 +148,25 @@
       </dependency>
       <dependency>
         <groupId>org.kie.kogito</groupId>
+        <artifactId>data-index-service-inmemory</artifactId>
+        <version>${project.version}</version>
+        <type>pom</type>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
         <artifactId>data-index-service-mongodb</artifactId>
+        <version>${project.version}</version>
+        <classifier>sources</classifier>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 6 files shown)


yesamer (approved):
LGTM

josedee (approved):
Thanks @AthiraHari77 for this

PR #6896, opened 2026-08-13
[incubator-kie-6891] DRL10 ANTLR4 parser does not support lambdas and…

### LINKED ISSUE #6891: drools: DRL10 ANTLR4 parser does not support lambdas and methods in functions
### Describe the bug

The ANTLR4 DRL parser (`DRL10ParserWrapper`) cannot parse Java 8 functional syntax inside a `function` declaration's body, while the legacy parser accepts it. Rules using such functions compile today only because `drools.drl.antlr4.parser.enabled` defaults to `false`; they will stop compiling when the ANTLR4 parser becomes the default.

Measured directly against `DRL10ParserWrapper` from `org.drools:drools-drl-parser:10.2.0`, loaded in an isolated classloader over its own ANTLR 4.13.2 runtime, so no other parser could be in play. The same constructs parse cleanly in a **rule consequence**, which is handled as an island, and the legacy ANTLR3 parser accepts them in both positions.

**Scope of the gap** — everything else tried in a function body parses: diamond generics, explicit generics, `var`, enhanced `for`, try-with-resources, `switch`, casts, ternaries and anonymous inner classes. Only lambdas (expression-bodied, block-bodied and typed-parameter forms) and method references (`obj::m`, `Type::m`) fail, plus any stream chain containing one.

**Likely cause** — `DRL10Parser.g4` does carry the alternatives:

```
| lambdaExpression // Java8
| switchExpression // Java17
| drlExpression COLONCOLON typeArguments? drlIdentifier
| typeType COLONCOLON (typeArguments? drlIdentifier | NEW)
| classType COLONCOLON typeArguments? NEW
```

but they hang off `drlExpression`, and a function body is `drlBlock → drlBlockStatement`, whose expression statements do not route through `drlExpression`. So the support appears to be present-but-unreachable from a block statement rather than absent by design — which matches the constructs working in every other position.

### Expected behavior

The ANTLR4 parser supports patterns supported by the ANTLR3 parser, including functions in DRL

### Actual behavior

Error: Line X:Y missing ';' at '<token>'

### How to Reproduce?

any of these as a DRL function body:

```drl
package org.example;

function void f( java.util.List xs, java.util.List out )
{
  xs.forEach(e -> out.add(e));      // lambda
}
```

```drl
function void g( java.util.List xs, java.util.List out )
{
  xs.forEach(out::add);             // method reference
}
```

_No response_

### Output of `java -version`

_No response_

_No response_

10.2.0

_No response_

### Additional information

_No response_

tkobayas:
… methods in functions

Closes https://github.com/apache/incubator-kie/issues/6891

### DIFF
--- a/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
+++ b/drools-drl/drools-drl-parser-tests/src/test/java/org/drools/drl/parser/antlr4/MiscDRLParserTest.java
@@ -4055,6 +4055,104 @@ class MiscDRLParserTest {
     }
 
     @Test
+    void functionWithLambdaExpression() {
+        // incubator-kie-issues#6891 : expression-bodied lambda as a method argument
+        final String text = "function void applyToAll(java.util.List xs, java.util.List out) {\n" +
+                "    xs.forEach(e -> out.add(e));\n" +
+                "}";
+        PackageDescr packageDescr = parseAndGetPackageDescr(text);
+
+        FunctionDescr function = packageDescr.getFunctions().get(0);
+
+        assertThat(function.getName()).isEqualTo("applyToAll");
+        assertThat(function.getParameterTypes()).containsExactly("java.util.List", "java.util.List");
+        assertThat(function.getParameterNames()).containsExactly("xs", "out");
+        assertThat(function.getBody()).isEqualToIgnoringWhitespace("xs.forEach(e -> out.add(e));");
+    }
+
+    @Test
+    void functionWithBlockBodiedLambda() {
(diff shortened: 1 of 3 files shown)


yesamer (approved):

Rikkola (approved):

drccrd:
Thanks all! I also tested locally with the same fix in a local fork of incubator-kie-tools/packages
/drools-lsp/ with a formatter built on top of the ANTLR4 parser and can confirm the issue is resolved.

tkobayas:
@drccrd Thank you for reporting. I'll update the grammar in drools-lsp accordingly.

PR #6633, opened 2026-04-02
[incubator-kie-issues#2029]  Dependency management simplification

gitgabrio:
Fixes: https://github.com/apache/incubator-kie-issues/issues/2029

drools-repo

Needed by:
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4241
https://github.com/apache/incubator-kie-kogito-apps/pull/2317
https://github.com/apache/incubator-kie-tools/pull/3538
https://github.com/apache/incubator-kie-kogito-examples/pull/2197

** Please read ticket description to understand the motivation behind this PR **

Please consider that the motivation of this PR rely on the unitary view of all our codestack (from drools to examples to kie tools) and not on specific considerations of repositories/modules considered by themselves in isolation.

This PR:

1. move all the `dependencyManagement`(s) tags to the `drools-build-parent` pom
2. remove all the `dependencyManagement`(s) tags from all other poms
3. create a `kie-quarkus-build-parent`specific bom that extend `drools-build-parent` one
4. modify the `drools-quarkus-extension` pom to inherit from `kie-quarkus-build-parent`
5. implement a custom enforcer rule `noDependencyManagementRule` to "forbid" declaration of `dependencyManagement` tag in module, not even inside `profile`(s) (it could be overridden, though: see later)

The `kie-quarkus-build-parent` _extends_ the `drools-build-parent` one, instead of importing it, to also inherit the `plugingManagement`. 
It declares the 

`<allowedPomsList>org.kie:kie-quarkus-build-parent</allowedPomsList>`
property to not fail on the `noDependencyManagementRule` check.

The reason for the `noDependencyManagementRule` is to avoid/limit the habitual approach of adding `dependencyManagement` in downstream/leaf modules.

The problem to solve is that, in our code base, we have multiple versions inconsistency due to the fact that:
1. the same dependency is declared, with different versions, in "parent" and "children" modules
2. the same dependency is inherited transitively on some module but explicit declared in other ones
3. at the very bottom of the stack (i.e. final applications/examples) there are multiple convergence issues due to the above
4. the trickiest part is that in most cases the wrong behavior is not at compile time, where it would be easily spotted, but at runtime, maybe in some combination of artifacts/examples that are not even tested in community

Those points makes our code base frail and hard to maintain (e.g. long time required for frameworks migrations, with overall delay in community major versions releases) for two reasons:

1.  there are situations where it is not clear why there is the version clash
[...]

PR #6903, opened 2026-08-18
[incubator-kie-6902] Upgrade Quarkus to 3.33.3.2 (LTS)

### LINKED ISSUE #6902: Upgrade Quarkus to 3.33 LTS (3.33.3.1) — Quarkus 3.27 LTS support ends September 2026
### Description

The repository currently builds on **Quarkus 3.27.5.1**. The Quarkus 3.27 LTS stream reaches
**end of support in September 2026** — after that it receives no further critical bug fixes or
security patches, leaving every KIE/Drools/Kogito/OptaPlanner Quarkus module without a patch
path for future CVEs.

The current recommended production LTS is **Quarkus 3.33**, supported until **March 2027**. Its
latest respin is **3.33.3.1**. Upgrading now keeps us inside a supported window and restores CVE
coverage.

References:
- Quarkus release/support lifecycle: https://quarkus.io/releases/
- LTS support windows: https://endoflife.date/quarkus-framework
- 3.33 LTS release planning: https://github.com/quarkusio/quarkus/wiki/3.33-LTS-Release-Planning

## Proposed target

- `version.io.quarkus` → **3.33.3.1** (latest 3.33 respin) in the three declaring parents
  (`kie-quarkus-build-parent`, `kogito-quarkus/bom`, `optaplanner-build-parent`)
- `version.io.quarkus.camel` → **3.33.2** (newest camel-quarkus in the 3.33 stream; Camel Quarkus
  publishes no respins, so it intentionally differs from the core version)

## Expected scope (from preliminary analysis)

- Re-align the Quarkus-coupled version pins in `kie-parent/pom.xml` with quarkus-bom 3.33.3.1
  (SmallRye family, Mutiny 3.x major, gRPC, Hibernate, MongoDB driver, etc.); keep CVE pins that
  are deliberately ahead of the BOM (jackson, log4j)
- Quarkiverse extension bumps where a 3.33-baseline release exists (openapi-generator,
  reactive-messaging-http)
- Replace `quarkus-smallrye-metrics` (removed from the 3.33 BOM) with Micrometer in data-audit
- Adapt extension deployment modules to 3.33 internal API changes (ArC Gizmo 2 creator API,
  removed constants/classes, runtime-config-in-build-step enforcement, stricter generated-class
  consistency checks affecting generated REST endpoint registration on both REST stacks)
- Deliberate holds requiring no action now: protobuf (BOM moved to 4.x, coupled to
  Infinispan/ProtoStream) and Flyway (BOM moved to 12.x, needs migration-script validation)

### Implementation ideas

_No response_

nrknithin:
**Issue:** Closes https://github.com/apache/incubator-kie/issues/6902

**Related:** 
- https://github.com/apache/incubator-kie-examples/pull/2244 [MERGED]
- https://github.com/apache/incubator-kie-tools/pull/3987 [MERGED]
- https://github.com/apache/incubator-kie-examples/pull/2245

Quarkus 3.27 LTS support ends September 2026. 3.33.3.2 is the latest respin of the next LTS stream
(supported until March 2027). Camel Quarkus 3.33.2 is the matching 3.33-stream release.

## What changed

**Versions**
- `version.io.quarkus` 3.27.5.1 → 3.33.3.2 (`kie-quarkus-build-parent`, `kogito-quarkus/bom`,
  `optaplanner-build-parent`); `version.io.quarkus.camel` → 3.33.2. Quarkiverse: openapi-generator 2.17.0-lts,
  reactivemessaging-http 2.7.1.
- **Quarkus-platform pins live in the Quarkus BOMs, not in kie-parent** (kie-parent keeps the framework-neutral
  defaults; the Spring Boot side is unaffected). `kogito-quarkus/bom` and `kie-quarkus-build-parent` (the drools
  Quarkus extensions do not inherit the kogito BOM) redefine: smallrye-config 3.16.0 (hard requirement — Quarkus 3.33
  references `io.smallrye.config.Config`), jandex 3.5.3 (hard — arc-processor uses `MethodParameterInfo.nameOrDefault`),
  smallrye-open-api 4.2.4 (needs jandex 3.5), commons-io 2.21.0, mutiny 3.1.1, smallrye-reactive-messaging-in-memory
  4.33.0. `kogito-quarkus/bom` additionally: guava 33.5.0-jre, grpc 1.81.0, hibernate 7.2.19.Final, postgresql 42.7.13,
  mongodb-driver-sync 5.6.4 (matches the bson/driver-core quarkus-bom supplies), smallrye-health 4.3.0.
- **kie-parent changes only what is genuinely common:** infinispan 16.0.13 and protostream 6.0.7 — shared jbpm/codegen
  marshalling code (`process-serialization-protobuf`, `kogito-codegen-processes`) had to change for ProtoStream 6, so
  both frameworks run the same version; `kogito-springboot/bom` imports `infinispan-bom` first so the Spring Boot
  Infinispan stack stays coherent. `infinispan-query-dsl` / `infinispan-remote-query-client` dropped (no longer exist in
  Infinispan 16; the query DSL classes now live in `infinispan-commons`).
- Framework-specific pins moved out of kie-parent into their BOM: jcip-annotations, jredisearch, graalvm svm, org.json →
  `kogito-quarkus/bom`; spock-core → `kogito-springboot/bom`; the springdoc duplicate removed (already in the Spring Boot BOM).
  reactive-streams and json-smart stay in kie-parent: core modules resolve them transitively (narayana-jta, json-path) and
  rely on the kie-parent pins.
- Kept on purpose: jackson 2.22.1 and log4j 2.26.1 (CVE pins, ahead of the BOM); protobuf 3.25.5 and
  flyway 11.14.1 (behind the BOM, deliberate); opentelemetry-api 1.62.0 CVE pin not re-added (needs
  Quarkus 4.0's OTel support — DBACLD-238206).
- Redundant micrometer override and its unused `version.io.micrometer` property removed from `optaplanner-build-parent`
  (that tree imports quarkus-bom directly, which already manages micrometer 1.16.6).

**Code — Quarkus 3.33 API/behaviour changes**
- `QuarkusAppPaths`: `JarResultBuildStep.MAIN` was removed → inlined `"main"`.
- `OptaPlannerProcessor`: ConstraintVerifier bean creator moved from Gizmo 1 to Gizmo 2 (ArC creator API
  changed); dead runtime-config parameter removed (3.33 rejects runtime config in build steps).
- `GizmoMemberAccessorEntityEnhancer`: accessor generated once per member (3.33's jar-consistency check
  rejected the per-annotation duplicate).
- `AbstractDroolsAssetsProcessor` / `KogitoAssetsProcessor`: generated REST endpoints are no longer
  registered as beans only when the reactive REST stack is present (it registers them itself → duplicate
  under 3.33); classic RESTEasy still needs the registration.
- Dev-mode hot-reload ITs (rules, decisions): 3.33 dev mode no longer publishes its resolved port for
  tests to read → tests now pick a free port up front (`SocketUtils` from `kogito-test-utils`, same
  approach as the processes hot-reload test) and hand it to the dev-mode app through the test archive's
  `application.properties`. No system-property mutation.
- Knative KSink health-check ITs: dropped the `ConfigProvider` assertions — in a
  `@QuarkusIntegrationTest` the app runs in a separate process, and 3.33's test framework no longer
  copies `@TestProfile` overrides into system properties, so the test JVM cannot see those values
  anymore. The health-endpoint assertions remain and cover the actual behaviour.
- `data-audit-quarkus-service`: `quarkus-smallrye-metrics` was removed from Quarkus → micrometer
  Prometheus registry (`/q/metrics` format changes).
- `quarkus-maven-plugin` pluginManagement gets `<extensions>true</extensions>` (3.31 migration guide).

**Code — aligned dependency majors**
- smallrye-open-api 4.2: `JobServiceModelFilter` null-safe for `components`/`schemas`; the no-schemas
  case (expected for the embedded job service without the REST API) is logged at debug.
- mongodb-driver 5.6.4: `MockMongoClient` implements the new `appendMetadata`.
- ProtoStream 6 / Infinispan 16: marshallers registered through `MarshallerProvider`
  (`SerializableMarshallerProvider` in codegen, `TriggerMarshallerProvider` in jobs-service).
  The jobs-service registration is mandatory — startup fails fast if the ProtoStream
  `SerializationContext` is unavailable, since the Infinispan repository cannot work without it.
- jobs-service infinispan: cache config moved from an inline string in `hotrod-client.properties` (no
  longer honoured) to `cache/job-details-cache.xml`; SASL default SCRAM-SHA-512.
- jobs-service: `runAsync` reworked to `vertx.executeBlocking(callable).toCompletionStage()` — a failing
  supplier now propagates to callers instead of hanging, the Vert.x future resolves properly, and it
  drops an overload that is deprecated in Vert.x 4.5 and removed in Vert.x 5.
- `drools-reliability-tests`: Infinispan 16 removed `InfinispanContainer()` → pass
  `IMAGE_BASENAME + ":" + Version.getMajorMinor()` (same image the old constructor used).
- Import-order-only edits: `ReteCompiler`, `PatternBuilder`, `DroolsJaxbHelperProviderImpl`.


optaplanner-build/optaplanner-build-parent/pom.xml
yesamer: @nrknithin I guess this is now unused, as you removed the declaration.

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/KSinkInjectionHealthCheckEnabledIT.java
```diff
@@ -39,7 +37,6 @@ class KSinkInjectionHealthCheckEnabledIT extends AbstractKSinkInjectionHealthChe
 
     @Test
     void kSinkInjectionHealthEnabled() {
-        assertThat(config.getOptionalValue(CONFIG_ALIAS, Boolean.class)).hasValue(true);
```
yesamer, line 42: @nrknithin Why no longer needed?

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/KSinkInjectionHealthCheckDisabledIT.java
```diff
@@ -40,7 +38,6 @@ class KSinkInjectionHealthCheckDisabledIT extends AbstractKSinkInjectionHealthCh
 
     @Test
     void kSinkInjectionHealthDisabled() {
-        assertThat(config.getOptionalValue(CONFIG_ALIAS, Boolean.class)).hasValue(false);
```
yesamer, line 43: @nrknithin Why no longer needed?

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/AbstractKSinkInjectionHealthCheckIT.java
```diff
@@ -18,22 +18,12 @@
...
 
 import static io.restassured.RestAssured.given;
 
 abstract class AbstractKSinkInjectionHealthCheckIT {
 
-    Config config;
-
-    @BeforeEach
-    void setUp() {
-        config = ConfigProvider.getConfig();
```
yesamer, line 34: @nrknithin Why no longer needed?

kogito-apps-quarkus/jobs-service-quarkus/jobs-service-infinispan/pom.xml
```diff
@@ -66,10 +66,6 @@
             <artifactId>quarkus-infinispan-client</artifactId>
         </dependency>
         <!-- Needs to be declared as Jobs service doesn't leverage persistence-commons-infinispan. -->
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-query-dsl</artifactId>
-        </dependency>
```
yesamer, line 72: @nrknithin Why no longer needed? In such a case, please remove the above comment

optaplanner-build/optaplanner-build-parent/pom.xml
nrknithin: yes — it was only there for those micrometer overrides, so it's unused now. Removed it.

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/KSinkInjectionHealthCheckEnabledIT.java
nrknithin: This is a QuarkusIntegrationTest, so the app runs in its own process and ConfigProvider.getConfig() here reads the test JVM's config, not the app's. Up to Quarkus 3.27 the test framework copied the profile overrides into system properties, so the test could still see the value — 3.33 doesn't do that anymore, so getOptionalValue comes back empty and the assertion fails. It was really only checking a test-harness side effect. The assertHealthChecks call below still verifies the actual behaviour (health check present when enabled), so I dropped the config check.

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/KSinkInjectionHealthCheckDisabledIT.java
nrknithin: Same reason as in the Enabled test — the test JVM doesn't see the profile override anymore on 3.33, so the config read comes back empty. The health endpoint assertion below still verifies the check is actually disabled.

kogito-quarkus/addons/knative/eventing/integration-tests/src/test/java/org/kie/kogito/addons/quarkus/knative/eventing/AbstractKSinkInjectionHealthCheckIT.java
nrknithin: It was only there to feed the config assertions I removed in the subclasses — nothing else reads it, so the field and the setUp became dead code.

kogito-apps-quarkus/jobs-service-quarkus/jobs-service-infinispan/pom.xml
nrknithin: infinispan-query-dsl doesn't exist anymore in Infinispan 16 (which Quarkus 3.33 pulls in) — the query dsl classes moved into infinispan-commons, which we already get transitively from quarkus-infinispan-client. The quarkus bom also stopped managing it, so the build couldn't even resolve it. And you're right about the comment — it belonged to this dependency, removed it.

kogito-apps-quarkus/jobs-service-quarkus/jobs-service-common/src/main/java/org/kie/kogito/jobs/service/repository/impl/BaseReactiveJobRepository.java
```diff
@@ -43,7 +43,14 @@ protected BaseReactiveJobRepository(Vertx vertx, JobEventPublisher jobEventPubli
...
-        vertx.executeBlocking(v -> future.complete(function.get()), r -> {
+        vertx.executeBlocking(v -> {
+            try {
+                future.complete(function.get());
+            } catch (Throwable t) {
+                // without this, any failure would leave the future incomplete and callers waiting forever
+                future.completeExceptionally(t);
+            }
+        }, r -> {
         });
```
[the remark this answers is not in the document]
nrknithin: That pattern was already there before this PR — I only added the try/catch so a failing supplier doesn't leave callers waiting forever. But fair point, I've switched it to vertx.executeBlocking(function::get).toCompletionStage(), which is simpler and completes the vertx future properly.

kogito-apps-quarkus/jobs-service-quarkus/jobs-service-infinispan/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
```diff
@@ -64,6 +70,14 @@ public HealthCheck infinispanHealthCheck(Instance<RemoteCacheManager> cacheManag
     void initializeCaches(@Observes @Priority(Interceptor.Priority.PLATFORM_BEFORE) StartupEvent startupEvent,
             Instance<RemoteCacheManager> remoteCacheManager,
             Event<InfinispanInitialized> initializedEvent) {
+        // ProtoStream 6 no longer indexes interface-typed marshallers by Java class; this provider
+        // keeps Trigger-typed values resolvable (see TriggerMarshallerProvider).
+        try {
+            MarshallerUtil.getSerializationContext(remoteCacheManager.get())
+                    .registerMarshallerProvider(new TriggerMarshallerProvider());
+        } catch (RuntimeException e) {
+            LOGGER.warn("ProtoStream SerializationContext is not available, Trigger marshalling may not work: {}", e.getMessage());
```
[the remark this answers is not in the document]
nrknithin: I went with the fail-fast option instead of the logging fix — if the SerializationContext isn't available, the Infinispan repository can't work at all (JobDetailsMarshaller is itself ProtoStream-based), so logging a warning would just delay the failure to the first job write with a much less obvious error. Removed the try/catch so startup fails immediately with the full stack trace, which also covers the diagnosability concern. Only extra change was stubbing getMarshaller() in InfinispanConfigurationTest since its mock used to go through the catch branch.

kogito-apps-quarkus/jobs-service-quarkus/jobs-service-common/src/main/java/org/kie/kogito/jobs/service/openapi/JobServiceModelFilter.java
```diff
@@ -55,9 +57,18 @@ public class JobServiceModelFilter implements OASFilter {
 
     @Override
     public void filterOpenAPI(OpenAPI openAPI) {
+        // Since smallrye-open-api 4.2 both the components and their schemas map can be null when the
+        // document declares no schemas (e.g. embedded job service without the REST API).
+        Components components = openAPI.getComponents();
+        Map<String, Schema> schemas = components != null ? components.getSchemas() : null;
+        if (schemas == null) {
+            LOGGER.warn("No component schemas are present in the OpenAPI document, skipping job service schema adjustments.");
```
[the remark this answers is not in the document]
nrknithin: Right, no schemas is the normal case for the embedded job service without the REST API — changed it to debug.

tkobayas (approved):

gitgabrio (changes requested):
Hi @nrknithin @tkobayas @yesamer @jstastny-cz @martinweiler  @pefernan 
I think this PR goes, somehow, against the bom-splitting approach and design.
The idea behind it was to remove the binding between quarkus and springboot frameworks/dependencies, but we are falling back to the old habit of mixing everything in one single place, and this will recreate the problem we invest so much effort to fix.

What I see in this PR is that all the dependency changes required for ->quarkus upgrade<- are written in the kie-parent, hence affecting also the springboot framework.
The consequence of that goes outside our own code, because springboot users will find their dependency tree modified - and potentially broken - by this change.
Probably some changes are minimal, but I see some about infinispan and hibernate that could have some impact.
The fact that the build is succeeding, unfortunately, may reflect lack of coverage.

kie-parent/pom.xml
```diff
@@ -148,13 +148,13 @@
...
-    <version.io.smallrye.config.core>3.13.4</version.io.smallrye.config.core>
-    <version.io.smallrye.jandex>3.4.0</version.io.smallrye.jandex>
+    <version.io.smallrye-config>3.16.0</version.io.smallrye-config>
+    <version.io.smallrye-health>4.3.0</version.io.smallrye-health>
+    <version.io.smallrye-open-api>4.2.4</version.io.smallrye-open-api>
+    <version.io.smallrye.config.core>3.16.0</version.io.smallrye.config.core>
+    <version.io.smallrye.jandex>3.5.3</version.io.smallrye.jandex>
     <version.io.smallrye.mutiny>2.9.5</version.io.smallrye.mutiny>
-    <version.io.smallrye.openapi.core>4.0.12</version.io.smallrye.openapi.core>
+    <version.io.smallrye.openapi.core>4.2.4</version.io.smallrye.openapi.core>
```
pefernan: I'm curious to see this versions here in kie-parent. Afaik smallrye is used in quarkus, maybe some of them could be moved to quarkus bom.

nrknithin:
> Hi @nrknithin @tkobayas @yesamer @jstastny-cz @martinweiler @pefernan I think this PR goes, somehow, against the bom-splitting approach and design. The idea behind it was to remove the binding between quarkus and springboot frameworks/dependencies, but we are falling back to the old habit of mixing everything in one single place, and this will recreate the problem we invest so much effort to fix.
> 
> What I see in this PR is that all the dependency changes required for ->quarkus upgrade<- are written in the kie-parent, hence affecting also the springboot framework. The consequence of that goes outside our own code, because springboot users will find their dependency tree modified - and potentially broken - by this change. Probably some changes are minimal, but I see some about infinispan and hibernate that could have some impact. The fact that the build is succeeding, unfortunately, may reflect lack of coverage.

Thanks @gitgabrio  — you're right, and I've changed the PR because of it: kie-parent's inherited management does outrank the BOMs imported below it, so a bump there reaches Spring Boot. Six of the eight packages you'd care about now resolve exactly what they do on main.

- Unchanged for Spring Boot: grpc 1.76.0, in-memory-messaging 4.28.0, guava 33.4.8-jre, mutiny 2.9.5, hibernate 7.1.18.Final, postgresql 42.7.12 — Quarkus gets 1.81.0 / 4.33.0 / 33.5.0-jre / 3.1.1 / 7.2.19.Final / 42.7.13 via a property override in kogito-quarkus-bom (and kie-quarkus-build-parent for the drools-quarkus tree).
- Still changed: infinispan 16.0.13 and protostream 6.0.7. These can't be split — drools-reliability sits under drools-build-parent, neither framework BOM, and declares them with no version; the shared addons are compiled once against ProtoStream 6.
- One real defect found and fixed: the Spring Boot classpath was mixed (protostream-types 5.0.13 under protostream 6.0.7, counter-api 15.2.6 under hotrod 16.0.13). Now coherent via an infinispan-bom import in kogito-springboot/bom.

kie-parent/pom.xml
nrknithin: Good catch @pefernan  — you're right, some of these don't belong in kie-parent.

**Moved out:**

- smallrye-health → Quarkus BOM. Only one pom used it, kogito-quarkus/bom, and kie-parent didn't manage any artifact with it.
- jcip-annotations, json-smart, graalvm svm, jredisearch, org.json → Quarkus BOM. Only Quarkus modules declare them.
- reactive-streams, spock-core → Spring Boot BOM. Only Spring Boot modules declare them.
- springdoc → removed. kogito-spring-boot-bom already had it, so kie-parent's copy was a duplicate.

**Left in kie-parent, with reasons:**

- smallrye-config — kie-parent uses it for smallrye-config-core, and a Spring Boot IT reads the property directly, so it would break if moved.
- smallrye-open-api — kie-parent uses it for smallrye-open-api-ui/-vertx, and kogito-spring-boot-bom uses the same property. Both sides need it.
- hamcrest-all — not all of its consumers inherit from the Quarkus BOM.
- io.fabric8 — kie-parent imports the kubernetes-client BOM with it, and the Spring Boot BOM already overrides it.

gitgabrio: HI @nithin, many thanks for all the care and time you are spending on that. Anyway, I  've a similar question. I understand why some of those smallrye artifacts are required here. What is not clear to me is what the need to ->upgrade<- them here. i.e.: if we override/upgrade them only on the quarkus bom, leaving the "default versions" here as they are, would something be broken ?
The overall design should be:
kie-parent -> "default" (i.e. tested) versions
springboot/quarkus -> version overrides based on framework specific requirement

nrknithin: @gitgabrio 
- What changed: smallrye-config / smallrye-open-api / jandex in kie-parent moved to the quarkus-bom 3.33.3.1 versions (3.16.0 / 4.2.4 / 3.5.3).
- Why in kie-parent: it's an ancestor of kogito-quarkus-bom, and inherited dependencyManagement beats an imported BOM — so kie-parent's version is what every Quarkus module actually gets; the quarkus-bom import can't override it (verified: forcing kie-parent to 3.13.4 makes jobs-service-common resolve 3.13.4).
- Why not leave the old values: Quarkus 3.33 then fails at JVM start with NoClassDefFoundError: io/smallrye/config/Config (0 tests run) — the class isn't in 3.13.4. And these values have always tracked the Quarkus platform (3.20 → 3.11.4, 3.27 → 3.13.4, brought into kie-parent unchanged by #6633); the core modules using them needed no code changes.

nrknithin: @gitgabrio  @pefernan  — you were both right. I’ve updated the branch so the rule is now applied consistently.

Quarkus-specific bumps are now out of kie-parent and overridden in kogito-quarkus/bom and kie-quarkus-build-parent, while the shared Infinispan 16 / ProtoStream 6 constraint stays in kie-parent since core code requires both frameworks to use the same versions.

I also corrected my earlier list: reactive-streams and json-smart need to stay in kie-parent because they’re transitively consumed by core modules. The remaining dependencies were verified to be framework-specific.

gitgabrio:
Thanks @nrknithin !
@yesamer @pefernan : could you please review again ? My left doubt is about infinispan: the version has been changed on kie-parent and then a modification become needed on springboot bom.
I would not expect that for pure-quarkus upgrade...

nrknithin:
> Thanks @nrknithin ! @yesamer @pefernan : could you please review again ? My left doubt is about infinispan: the version has been changed on kie-parent and then a modification become needed on springboot bom. I would not expect that for pure-quarkus upgrade...

@gitgabrio 

Infinispan/ProtoStream are shared. The same code runs on both Spring Boot and Quarkus, and the ProtoStream 6 upgrade also required changes to the shared marshalling code.

The important part is that Infinispan 16 requires ProtoStream 6, so they need to move together. Also, the Infinispan version property drives the server image used by the ITs on both sides, so keeping it framework-specific would require splitting that as well.

The Spring Boot BOM change is only to avoid mixing Infinispan 15 and 16 in the same classpath. 

CC:  @yesamer @pefernan

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ReteCompiler.java
```diff
@@ -22,10 +22,10 @@
 import java.util.Collection;
 
 import org.drools.base.RuleBase;
-import org.drools.compiler.builder.impl.AssetFilter;
-import org.drools.compiler.compiler.PackageRegistry;
 import org.drools.base.definitions.InternalKnowledgePackage;
 import org.drools.base.definitions.rule.impl.RuleImpl;
+import org.drools.compiler.builder.impl.AssetFilter;
```
jstastny-cz, line 25: I don't think these changes should be submitted as part of this PR. Also other files are affected.

drools-quarkus-extension/drools-quarkus-util-deployment/src/main/java/org/drools/quarkus/util/deployment/QuarkusAppPaths.java
```diff
@@ -52,7 +51,8 @@ private enum PathType {
     }
 
     protected QuarkusAppPaths(List<Path> projectPaths, Collection<Path> classesPaths, boolean isJar) {
-        super(projectPaths, classesPaths, isJar, AppPaths.BT, JarResultBuildStep.MAIN, false);
+        // "main" = quarkus-app output subdir (formerly JarResultBuildStep.MAIN, removed in Quarkus 3.33)
+        super(projectPaths, classesPaths, isJar, AppPaths.BT, "main", false);
```
jstastny-cz: Wouldn't AppPaths.MAIN_DIR be better?

drools-reliability/drools-reliability-tests/src/test/java/org/drools/reliability/test/ReliabilityTestBasics.java
```diff
@@ -142,8 +143,8 @@ static Stream<PersistedSessionOption.PersistenceStrategy> strategyProviderFull()
     public void setUp() {
         if (((TestableStorageManager) StorageManagerFactory.get().getStorageManager()).isRemote()) {
             LOG.info("Starting InfinispanContainer");
-            container = new InfinispanContainer()
-                    .withFileSystemBind("infinispan-remote-config", "/user-config")
+            container = new InfinispanContainer(InfinispanContainer.IMAGE_BASENAME + ":" + Version.getMajorMinor());
```
jstastny-cz, line 146: Doesn't the need for this version specification indicate there's a misconfiguration of the image being used?

gitgabrio (approved):
Thanks  @nrknithin : great stuff!

optaplanner-build/optaplanner-build-parent/pom.xml
yesamer, line 150: @nrknithin Can you please double check if this is still required?

optaplanner-build/optaplanner-build-parent/pom.xml
```diff
@@ -262,6 +245,7 @@
           <groupId>io.quarkus</groupId>
           <artifactId>quarkus-maven-plugin</artifactId>
           <version>${version.io.quarkus}</version>
+          <extensions>true</extensions>
```
yesamer, line 248: @nrknithin Why?

kogito-quarkus/bom/pom.xml
```diff
@@ -31,17 +31,71 @@
...
+    <version.io.smallrye-config>3.16.0</version.io.smallrye-config>
+    <version.io.smallrye.config.core>3.16.0</version.io.smallrye.config.core>
+    <version.io.smallrye.jandex>3.5.3</version.io.smallrye.jandex>
+    <!-- quarkus-bom alignment; smallrye-open-api 4.2.4 requires jandex 3.5, so it is pinned in the same scope as jandex. -->
+    <version.io.smallrye-open-api>4.2.4</version.io.smallrye-open-api>
+    <version.io.smallrye.openapi.core>4.2.4</version.io.smallrye.openapi.core>
+    <!-- quarkus-bom alignment; must match the bson/mongodb-driver-core version quarkus-bom supplies. -->
+    <version.mongodb-driver-sync>5.6.4</version.mongodb-driver-sync>
+    <!-- quarkus-bom alignment only (Quarkus 3.33 works on 2.20.0). -->
+    <version.commons-io>2.21.0</version.commons-io>
```
yesamer: @nrknithin The comment is not aligned with the version. BTW, I guess commons can be globally managed in kie-parent, there is a reason to override it for quakus only?

nrknithin: Agreed. There’s no Quarkus-specific reason for the override, and Quarkus 3.33 works fine with 2.20.0. Removed the override from both Quarkus BOMs, so it remains managed globally by kie-parent.

optaplanner-build/optaplanner-build-parent/pom.xml
nrknithin: Required since Quarkus 3.31 for quarkus-maven-plugin to work as a build extension. Added it in both places where the plugin is declared.

optaplanner-build/optaplanner-build-parent/pom.xml
nrknithin: Still needed. Quarkus BOM 3.33.3.2 pins sshd at 2.17.1, which is below the 2.19.0 CVE fix. OptaPlanner doesn't inherit kie-parent, so its version can't override this. Also corrected the comment to reference 2.17.1.

drools-reliability/drools-reliability-tests/src/test/java/org/drools/reliability/test/ReliabilityTestBasics.java
nrknithin: Not a misconfiguration. Infinispan 16 removed the no-arg constructor, and the old constructor used the same IMAGE_BASENAME + ":" + Version.getMajorMinor() value. This just keeps the previous default explicitly. Would u suggest any other way to achieve the same results?

drools-quarkus-extension/drools-quarkus-util-deployment/src/main/java/org/drools/quarkus/util/deployment/QuarkusAppPaths.java
nrknithin: Agreed. AppPaths already defines MAIN_DIR = "main", so switched to AppPaths.MAIN_DIR.

drools-compiler/src/main/java/org/drools/compiler/builder/impl/processors/ReteCompiler.java
nrknithin: These were fixing pre-existing CustomImportOrder issues and aren't related to the upgrade. Reverted all three files.

nrknithin:
@yesamer @jstastny-cz 
thanks for the review have updated the code and added review comments 
Thank you.

drools-reliability/drools-reliability-tests/src/test/java/org/drools/reliability/test/ReliabilityTestBasics.java
jstastny-cz: Nope, thanks for clarifying, this is sufficient explanation, thank you @nrknithin .

jstastny-cz (approved):
Looks good to me.

yesamer (approved):

pefernan (approved):

jomarko (approved):
I do not have findings in this PR. Approving. Thank you.

PR #6907, opened 2026-08-21
[incubator-kie-issues#2394] replace data-isolation inline VALUES constructor

### LINKED ISSUE #2394: replace data-isolation inline VALUES constructor with ANSI-compliant CTE for cross-database compatibility
#### **Problem**
Our query logic builds inline parameter lists using the SQL `VALUES` syntax (e.g., `VALUES (?, ?)`). Some database engines reject this keyword inside subqueries or fail to detect parameter data types. This causes query crashes across both our **runtime persistence** and **data audit storage** components.

#### **Solution**
Replace the `VALUES` syntax with standard, highly compatible ANSI SQL elements:
* Fetch parameters using a standard query anchored to a single indexed row (`MIN(id)`).
* Wrap the first row with standard `CAST(? AS VARCHAR(255))` expressions to guarantee data types are recognized.
* Maintain a flat query structure so that sorting and pagination work correctly down the line.

#### **Impact**
* **Unified Fix:** Resolves query errors for both **runtime persistence** and **data audit storage** simultaneously.
* **Broad Database Support:** Code runs cleanly across multiple databases and local test environments.
* **Zero Performance Loss:** Primary key indexing keeps the parameter selection ultra-fast.

jstastny-cz:
Closes apache/incubator-kie-issues#2394

Making CTE more robust for
* plain JDBC
* JPA native query

Where VALUES constructor is not ANSI standard and not all DBs support its multi-value form.
This is a cheap, performance neutral, change to make CTE SQL query more robust.

### DIFF
--- a/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index-persistence/kogito-addons-quarkus-data-index-persistence-postgresql/integration-tests-process/pom.xml
+++ b/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index-persistence/kogito-addons-quarkus-data-index-persistence-postgresql/integration-tests-process/pom.xml
@@ -92,6 +92,22 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- Ensure data-index-service-postgresql binary is built before running integration tests.
+         Build-ordering sentinel only: type=pom + full exclusions adds nothing to classpath.
+         Without this edge the partial CI build skips the service module and the
+         DataIndexPostgreSqlContainer image is missing at runtime. -->
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>data-index-service-postgresql</artifactId>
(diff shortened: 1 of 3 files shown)


kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index-persistence/kogito-addons-quarkus-data-index-persistence-postgresql/integration-tests-process/pom.xml
```diff
@@ -92,6 +92,22 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- Ensure data-index-service-postgresql binary is built before running integration tests.
```
pefernan, line 95: good catch

pefernan (approved):

martinweiler (approved):
Great work @jstastny-cz

PR #7077, opened 2026-08-28
[NO-ISSUE] bump mvel to 2.5.4.Final

tkobayas:


yesamer (approved):

kie-parent/pom.xml
```diff
@@ -284,7 +284,7 @@
     <version.org.mongo>5.3.1</version.org.mongo>
     <version.org.mozilla.rhino>1.8.1</version.org.mozilla.rhino>
     <!-- Keep synchronized with junit-jupiter (middle and minor should be the same) -->
```
yesamer: @tkobayas Is this message for mvel? it seems misleading...

tkobayas: It was introduced by https://github.com/kiegroup/drools/commit/ab6c0e70a40 for `<version.org.junit.jupiter>` and `<version.org.junit.platform>`. Now they are combined to `<version.org.junit>`, so the comment looks obsolete. Removing...

yesamer: Thanks :)

drools-model/drools-model-codegen/src/test/java/org/drools/model/codegen/execmodel/MvelDialectTest.java
```diff
@@ -763,7 +763,7 @@ public void testCompoundOperatorBigDecimalConstantWithoutLiterals(RUN_TYPE runTy
 
         ksession.insert(john);
         assertThat(ksession.fireAllRules()).isEqualTo(1);
-        assertThat(john.getMoney()).isEqualTo(new BigDecimal( 0 ));
+        assertThat(john.getMoney()).isEqualByComparingTo(new BigDecimal( 0 ));
```
tkobayas, line 766: The result is BigDecimal `0.0`, so failed with `isEqualTo`. The scale difference was introduced by https://github.com/mvel/mvel/issues/321. I think using `isEqualByComparingTo` is rather correct, because it doesn't make sense to be strict about scale when calculating with double.

tkobayas:
Suspending this PR, as I found another bug in mvel. Checking...

tkobayas:
Fixed in mvel 2.5.4.Final, so bumping again.

baldimir (approved):

PR #7099, opened 2026-09-10
[incubator-kie-6902] Align RESTEasy with quarkus-bom 3.33 in kogito-quarkus-bom

nrknithin:
## What

Pin `version.org.jboss.resteasy` to `6.2.18.Final` in `kogito-quarkus/bom/pom.xml`.

## Why

`quarkus-bom:3.33.3.2` manages RESTEasy `6.2.18.Final`, while `kie-parent` pins
`6.2.12.Final` (the version `quarkus-bom:3.27.5.1` shipped). After the Quarkus 3.33
upgrade that pin silently downgrades RESTEasy on the Quarkus side.

`kie-parent` declares `resteasy-client`, `resteasy-jackson2-provider` and
`resteasy-multipart-provider`, which are shared with the Spring Boot tree, so the
override is scoped to the Quarkus BOM instead of bumping `kie-parent`. Same pattern
as the existing `mongodb-driver-sync` and `smallrye-open-api` alignment pins.

## Verification

Effective POMs:

| BOM | `version.org.jboss.resteasy` | managed artifacts |
|---|---|---|
| `kogito-quarkus-bom` | 6.2.18.Final | all 34 at 6.2.18.Final |
| `kogito-springboot-bom` | 6.2.12.Final | unchanged |

Matches `quarkus-bom:3.33.3.2` exactly; Spring Boot resolution is unaffected.

### DIFF
--- a/kogito-quarkus/bom/pom.xml
+++ b/kogito-quarkus/bom/pom.xml
@@ -47,6 +47,7 @@
     <version.io.smallrye.mutiny>3.1.1</version.io.smallrye.mutiny>
     <version.org.hibernate>7.2.19.Final</version.org.hibernate>
     <version.org.postgresql>42.7.13</version.org.postgresql>
+    <version.org.jboss.resteasy>6.2.18.Final</version.org.jboss.resteasy>
     <!-- Quarkus 3.33 hard requirements: smallrye-config 3.16 (quarkus-core/junit-config reference io.smallrye.config.Config)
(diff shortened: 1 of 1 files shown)


yesamer:
@nrknithin Does 6.2.18 contains breaking changes / different behavior than 6.2.12.Final? Usually, when the minor version of a dependency is updated, it means it contains bug fixes and CVE resolution only (not always true... but most of the times :D). Do you think there's a good reason to use different versions? Can we use the same in the overall codebase? thanks

PR #3869, opened 2021-10-01
[DROOLS-6636] reoreder right tuple and blockers in NotNode even when …

mariofusco:
…property reactivity prevents propagation

**JIRA**: https://issues.redhat.com/browse/DROOLS-6636

Issue #7106, opened 2026-09-11
Dev mode: "EntityManagerFactory is closed" after a few live reloads when kogito-addons-quarkus-data-index-jpa is used


thiagoelg:
## Describe the bug

When `kogito-addons-quarkus-data-index-jpa` is on the classpath, a Quarkus dev mode application stops working after a few live reloads. It fails to start with `java.lang.IllegalStateException: EntityManagerFactory is closed` and never recovers.

**What goes wrong:**

`KogitoBeanProducer` produces the CDI `UnitOfWorkManager` bean by returning a JVM-wide static singleton, `StaticUnitOfWorkManger.staticUnitOfWorkManager()`. Quarkus dev mode throws the application away on each reload, but it does not throw that static object away, because the class holding it comes from a dependency JAR that the base runtime ClassLoader keeps loaded.

So the same `UnitOfWorkManager`, and the same `EventManager` inside it, is reused by every reload. Each restart adds its `EventPublisher`s to it, and `addPublisher` only ever adds — there is no matching remove. After a few reloads, publishing an event calls publishers that belong to applications which already shut down. Those old publishers still point at an `EntityManagerFactory` that was closed when their application stopped, which is the error above.

The data index add-on is not the bug. It is just the only publisher that opens a Hibernate session straight away, so it is the first one to fail visibly.

Reproducer, with public dependencies only and full details: https://github.com/thiagoelg/quarkus-dev-mode-emf-closed-repro

## Expected behavior

Each dev mode live reload starts a clean application. Event publishers from earlier reloads are discarded with the application they belong to.

## Actual behavior

The first start and the first reload work. From the second reload the application fails to start, requests return HTTP 500, and it never recovers.

```
Caused by: java.lang.IllegalStateException: EntityManagerFactory is closed
	at org.hibernate.internal.SessionFactoryImpl.validateNotClosed(SessionFactoryImpl.java:587)
	at io.quarkus.hibernate.orm.runtime.session.TransactionScopedSession.find(TransactionScopedSession.java:199)
	at org.kie.kogito.index.jpa.storage.AbstractJPAStorageFetcher.get(AbstractJPAStorageFetcher.java:91)
	at org.kie.kogito.index.service.IndexingService.indexProcessDefinition(IndexingService.java:98)
	at org.kie.kogito.index.addon.event.DataIndexEventPublisher.publish(DataIndexEventPublisher.java:60)
[...]

PR #3870, opened 2021-10-02
DROOLS-6638 DMN OAS generation NPE

tarilabs:
fix NPE bug as described in:

**JIRA**: https://issues.redhat.com/browse/DROOLS-6638

**referenced Pull Requests**: none

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #7107, opened 2026-09-11
[incubator-kie-7106] fix(`kogito-quarkus-workflow-extension-common`): scope the UnitOfWorkManager bean to the application, not to the JVM

thiagoelg:
**Issue**:

* https://github.com/apache/incubator-kie/issues/7106

**Referenced Pull Requests**:

* https://github.com/apache/incubator-kie-kogito-runtimes/pull/3870 (context only — see "Relationship to `#3870`" below)

---

## TL;DR

This fixes a bug I faced while running a Kogito application in dev mode. After changing a BPMN asset a few times, the runtime would fail and throw the following error:
```
IllegalStateException: EntityManagerFactory is closed
```

Removing the data-index JPA add-on made the error stop. The changes to `KogitoBeanProducer` fixed it, and I was able to add the add-on back again.

Read below if you want a deeply technical AI-generated description of the underlying issue.

## What this fixes

`KogitoBeanProducer` builds the CDI `UnitOfWorkManager` bean by returning a static singleton: `StaticUnitOfWorkManger.staticUnitOfWorkManager()`. There is one of those per JVM. The class that holds it comes from a dependency JAR, so Quarkus loads it with the base runtime ClassLoader. A dev-mode live reload does not recreate that ClassLoader, so the manager stays in memory after the application it belongs to has stopped.

That means every reload reuses the same manager, and the same `BaseEventManager` inside it. On each restart `AbstractProcessConfig` adds the new application's `EventPublisher`s and `UnitOfWorkEventListener`s to it. Both `addPublisher` and `register` only add — there is no remove. So the set of publishers grows by one per reload, and `publish()` ends up calling publishers that belong to applications which already shut down. Those old publishers still point at an `EntityManagerFactory` that was closed when their application stopped, and that is the `IllegalStateException: EntityManagerFactory is closed`.

A probe recorded object identities across five dev-mode starts. `Application` is a new object every time, as it should be. The manager and its event manager never are, and the publisher count goes up by one each reload:

| Start | `Application` | `UnitOfWorkManager` | `BaseEventManager` | Publishers |
|---|---|---|---|---|
| restart no:0 | `@2b15af5c` | `DefaultUnitOfWorkManager@135ba` | `@78f53ca9` | 1 |
| restart no:1 | `@41d76578` | `DefaultUnitOfWorkManager@135ba` | `@78f53ca9` | 2 |
| restart no:2 | `@266a08bf` | `DefaultUnitOfWorkManager@135ba` | `@78f53ca9` | 3 |
| restart no:3 | `@6637bbb9` | `DefaultUnitOfWorkManager@135ba` | `@78f53ca9` | 4 |
| restart no:4 | `@72a2ae1f` | `DefaultUnitOfWorkManager@135ba` | `@78f53ca9` | 5 |

The fix gives the producer the `@Singleton` scope and returns a new `DefaultUnitOfWorkManager`. The manager then lives and dies with the application, and nothing from an old reload survives.

Reproducer (public dependencies only, plus the full analysis): https://github.com/thiagoelg/quarkus-dev-mode-emf-closed-repro

## Relationship to [#3870](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3870) — this is not a revert

This changes the line that dcaca447 ([`[Fix #3869] Using the same work and event manager`](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3870)) introduced, so it is worth being explicit about the difference.

Before dcaca447 the producer was a bare `@Produces`, which means `@Dependent`. With that scope **every injection point gets its own manager**, and only the one held by `ProcessConfig` had the publishers on it. That is [#3869](https://github.com/apache/incubator-kie-kogito-runtimes/issues/3869): any code that got its manager from a different injection point sent its events to a manager with no publishers, so those events were lost.

So the real problem in [#3869](https://github.com/apache/incubator-kie-kogito-runtimes/issues/3869) was the **missing scope**. dcaca447 made every injection point share one object by using a static field. This PR makes them share one object through the CDI container instead. What dcaca447 fixed stays fixed; only the lifetime of the object changes.

`@Singleton` is used rather than `@ApplicationScoped` because ArC does not create a client proxy for it. Each injection point gets the object directly, so the unit-of-work path keeps the same shape it had before.

This also explains why dcaca447 only needed to touch Quarkus: Spring `@Bean` methods are singleton-scoped by default, so Spring Boot never had #3869. Its producer still reads `new DefaultUnitOfWorkManager(new CollectingUnitOfWorkFactory())`.

`LightProcessRuntimeServiceProvider` is deliberately left exactly as dcaca447 set it.

## Regression test for [#3869](https://github.com/apache/incubator-kie-kogito-runtimes/issues/3869)

PR [#3870](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3870) shipped no test, so nothing in CI catches this. This PR adds one in `kogito-addons-quarkus-data-index-jpa/integration-tests-process`. The `timerResume` process stops at an intermediate timer, so the instance is saved to the database and then resumed by the job scheduler rather than by the request thread. The test checks that the data index sees the instance complete.

The test was checked in both directions first — a test that cannot fail proves nothing:

| Build | `UnitOfWorkManager` identities | Test |
|---|---|---|
| Pre-dcaca447 (reverse-applied) | config `@3c11533` (1 publisher), fresh lookup `@7fd0231` — a **different object** — static `@4c1a14df` (0 publishers) | **Fails** (`ConditionTimeoutException`, instance holds `ACTIVE`) |
| `main` | all one object, 1 publisher | Passes |
| This PR | config == fresh, 1 publisher | Passes |

## Validation

- Dev-mode reproducer: without the fix it breaks from reload `#2` and never recovers. With the fix, 11 starts out of 11 are clean and the log has no `EntityManagerFactory is closed`.
- The data index really does receive the events. This matters because a change that simply disconnected the publishers would also look clean: no error, HTTP 201. 5 instances were started across 5 generations, and exactly 5 were indexed, all `COMPLETED`.
- `kogito-quarkus-processes-integration-test` (10 tests), `integration-tests-quarkus-usertasks` (22 tests) and `kogito-addons-quarkus-data-index-jpa/integration-tests-process` (4 tests) all pass.

The user-task module is the important one there. `UserTaskConfig` shares the `UnitOfWorkManager` bean with `ProcessConfig`, so a wrong scope shows up in that module.

## Note on the second commit

`[NO-ISSUE] chore: fix the jobs-service IT shared/ ignore rule...` has nothing to do with #7106. The four jobs-service IT modules copy shared BPMN files into `src/main/resources/shared` during the build. There is already a `.gitignore` rule for them, but it still points at `apps-integration-tests/...`, the path used before `kogito-apps` was merged into this repository. It has matched nothing since that merge, so a working copy that has been built shows 24 generated files as untracked. Happy to split it into its own PR if preferred.

---

Disclaimer: Claude Opus 5 was used in the process of research, development, testing, and writing the description of this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,4 +180,4 @@ package-lock.json
 !.rat-excludes
 
 # Jobs Service ITs
-apps-integration-tests/integration-tests-jobs-service/**/src/main/resources/shared/
+kogito-apps-quarkus/integration-tests-jobs-service-quarkus/**/src/main/resources/shared/
--- a/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index/kogito-addons-quarkus-data-index-jpa/integration-tests-process/pom.xml
+++ b/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index/kogito-addons-quarkus-data-index-jpa/integration-tests-process/pom.xml
@@ -80,6 +80,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kogito-addons-quarkus-data-index-postgresql-deployment</artifactId>
       <type>pom</type>
--- a/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index/kogito-addons-quarkus-data-index-jpa/integration-tests-process/src/main/resources/timerresume.bpmn
+++ b/kogito-apps-quarkus/data-index-quarkus/kogito-addons-quarkus-data-index/kogito-addons-quarkus-data-index-jpa/integration-tests-process/src/main/resources/timerresume.bpmn
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+<bpmn2:definitions xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL"
+                   xmlns:drools="http://www.jboss.org/drools"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   id="_timerResumeDefinitions"
+                   targetNamespace="http://www.omg.org/bpmn20">
+  <!-- Parks on a timer so the instance is persisted and resumed by the scheduler. Guards #3869. -->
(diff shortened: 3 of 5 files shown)


pefernan (approved):

martinweiler (approved):
Thank you @thiagoelg !

PR #7116, opened 2026-09-16
fix(drools): exists(A() and not B()) stops firing when sibling package wraps same CE in its own exists

### LINKED ISSUE #7115: Drools: rule with `not` inside `exists` stops firing
### Describe the bug

On Drools 10.x a rule guarded by exists ( A() and not B() ) stops firing even though its condition is satisfied, when another rule in a different package wraps the same leading conditional element in its own exists. The rule is simply never activated: no build error, no warning, nothing in the logs. The same rules fire correctly on 9.44.0.Final.

The whole reproducer is four DRL files. Types (repro.types):

```
package repro.types;

declare Seed end
declare SharedFact end
declare NegatedFact end
declare OtherFact end
declare Result end
```

A rule that derives SharedFact from the single input fact (repro.a):

```
package repro.a;
import repro.types.*;

rule "produce-shared-fact"

    when

        Seed()

    then

        insert(new SharedFact());

end
```

The rule that misbehaves (repro.b):

```
package repro.b;

import repro.types.*;

rule "subject"

    when

        exists ( SharedFact() and not NegatedFact() )

    then

        insert(new Result());

end
```

And a rule that never fires itself, but whose presence is what breaks the one above (repro.c):

```
package repro.c;

import repro.types.*;

rule "sibling"

    when

        exists ( SharedFact() and OtherFact() )

    then

end
```

A single Seed is inserted and fireAllRules() is called. Nothing in the rule set ever produces NegatedFact or OtherFact, so after produce-shared-fact runs, subject's condition is satisfied and it must fire. sibling correctly does not fire, since OtherFact is absent.

On 9.44.0.Final it fires. On 10.0.0, 10.1.0 and 10.2.0 it does not.

This looks like the same family as https://github.com/apache/incubator-kie/issues/7109

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

_No response_

### Output of `java -version`

_No response_

_No response_

_No response_

_No response_

### Additional information

_No response_

Rikkola:
Closes: https://github.com/apache/incubator-kie/issues/7115

## Problem

A rule guarded by `exists(A() and not B())` silently stops firing when a rule in a **different package** also uses `exists(A() and ...)`. No build error, no warning — the rule is simply never activated.

Reproducer (four packages):

```drl
// repro.a
rule "produce-shared-fact" when Seed() then insert(new SharedFact()); end

// repro.b  ← stops firing
rule "subject" when exists(SharedFact() and not NegatedFact()) then insert(new Result()); end

// repro.c  ← its presence breaks the rule above
rule "sibling" when exists(SharedFact() and OtherFact()) then end
```

`NegatedFact` and `OtherFact` are never inserted, so `subject` must fire. It does on 9.44.0.Final. It does not on 10.0–10.2 and current main.

Related: apache/incubator-kie#7109

## Root cause

Two commits combine to produce this bug:

1. **[DROOLS-7238] Rete Buildtime Segmentation** (9b650b1) — introduced `SegmentPrototype.splitEagerProtos` with the guard:
   ```java
   if (proto1WasEager) { // if it wasn't eager before, nothing can be eager after
   ```
2. **Split Beta Nodes** (2aa4b18) — introduced `ExistsRight`/`NotRight` wrapper nodes, enabling cross-subnetwork `JoinNode` sharing.

Packages build sequentially. When `repro.b` is added, its inner segment contains both `JoinNode(SharedFact)` and `NotNode`. The joint presence of both suppresses the eager requirement (`requiresAnEagerSegment() = false`), so the segment is **not marked eager**. When `repro.c` is added later, `processSplit` splits the shared segment:

- proto1 = `[JoinNode(SharedFact)]` — **not eager** (`proto1WasEager = false`)
- proto2 = `[NotNode, TTON_subject]` — **eager** (NotNode without its JoinNode)

`splitEagerProtos(proto1WasEager=false, proto2, TTON_subject)` hits the guard and exits immediately, so **proto2 is never registered** in `TTON_subject.eagerSegmentPrototypes`. At session creation, `initializePathMemory` finds an empty eager list → the `[NotNode, TTON_subject]` segment is never created → the NotNode's "start linked" state is never propagated → `ExistsNode_subject` is permanently unlinked → the outer rule never fires.

## Fix

Add an `else if` branch in `splitEagerProtos` for the case where `proto1` was not eager but `other` (proto2) requires eager:

```java
} else if (other.requiresEager()) {
    // proto1 was not eager, but the split produced an eager proto2.
    SegmentPrototype[] eager = endNode.getEagerSegmentPrototypes();
    SegmentPrototype[] newEager = new SegmentPrototype[eager.length + 1];
    System.arraycopy(eager, 0, newEager, 0, eager.length);
    newEager[newEager.length - 1] = other;
    endNode.setEagerSegmentPrototypes(newEager);
}
```

## Test

`ExistsWithNotSubnetworkCrossPackageTest` — 5 test methods × 2 configs (CLOUD_IDENTITY + CLOUD_EXEC_MODEL) = 10 parameterised instances:

| Test | Purpose |
|------|---------|
| `subjectFiresWhenSiblingInDifferentPackage` | Direct regression: cross-package sibling, insert reproducer |
| `subjectFiresWhenSiblingInDifferentPackageWithRuleChaining` | Rule-chained insert variant |
| `subjectFiresWhenSiblingInSamePackage` | Same-package sibling must still fire (control) |
| `subjectFiresWithoutSibling` | No sibling at all must still fire (control) |
| `subjectFiresWithPlainJavaClasses` | Plain Java classes, not declare-type-specific |

All 10 pass after fix. 310 tests across 9 regression suites all green (including the new `SegmentSplitAccumulateTest` from #7110).


drccrd:
@Rikkola nice thank you - this is another (and the only other) issue I had also noticed after upgrading to drools 10, but couldn't get a reproducer working - I thought I had entered a ticket but can't find it. I will test on the affected rules (which we "fixed" by adding a null check in the second occurrence).

Rikkola:
@drccrd thanks any testing would be great. 

This was reported in Zulip: [#drools > rule with &#96;not&#96; inside &#96;exists&#96; stops firing @ 💬](https://kie.zulipchat.com/#narrow/channel/232677-drools/topic/rule.20with.20.60not.60.20inside.20.60exists.60.20stops.20firing/near/624450038)

I am still keeping this as a draft. I need to investigate that the scope is correct.

Ticket here: https://github.com/apache/incubator-kie/issues/7115

The commit that broke it: Split Beta Nodes https://github.com/apache/incubator-kie/commit/2aa4b184e72

Rikkola:
Improving the test coverage. The reported situation was not the only case where this can be hit. So added different test variants for the same bug.

drccrd:
@Rikkola I had another look at the problems I had but they were actually another pattern: `not( X( $list: things ) and Y() from $list )` shared in the same file with one more adjacent pattern in the second occurrence, and the problem was an intermittent NPE rather than a rule not firing. I will try again on a reproducer for that one, but I don't think it is related to this PR.

Rikkola:
@drccrd I might have a fix for that too. Will be a different PR and not sure if I get to do it this week or next.

tkobayas:
@Rikkola  Thank you for the fix.

One more point, when a rule is removed, we also need to remove the registered eager prototype in `mergeEagerProtos()`.

Test case:
```
@ParameterizedTest(name = "KieBase type={0}")
@MethodSource("parameters")
public void testRemoveSiblingRule_afterTopLevelNegationSegmentSplit(
        KieBaseTestConfiguration kieBaseTestConfiguration) {

    final String drlSubject =
            "package repro.b;\n" +
            "import java.util.ArrayList;\n" +
            "import java.util.LinkedList;\n" +
            "rule \"subject\"\n" +
            "    when\n" +
            "        String()\n" +
            "        ArrayList()\n" +
            "        not LinkedList()\n" +
            "    then\n" +
            "end\n";

    final String drlSibling =
            "package repro.c;\n" +
            "import java.util.ArrayList;\n" +
            "import java.util.HashMap;\n" +
            "rule \"sibling\"\n" +
            "    when\n" +
            "        String()\n" +
            "        ArrayList()\n" +
            "        HashMap()\n" +
            "    then\n" +
            "end\n";

    final InternalKnowledgeBase kbase = (InternalKnowledgeBase) KieBaseUtil.getKieBaseFromKieModuleFromDrl(
            "remove-sibling-top-level-not-test", kieBaseTestConfiguration, drlSubject);
    final KieBase siblingBase = KieBaseUtil.getKieBaseFromKieModuleFromDrl(
            "remove-sibling-top-level-not-sibling", kieBaseTestConfiguration, drlSibling);

    // Add the sibling after the subject's segment exists, forcing a split that makes the NotNode segment eager.
    kbase.addPackages(siblingBase.getKiePackages());
    // Merging the segments again must remove the obsolete eager prototype before a new session is created.
    kbase.removeRule("repro.c", "sibling");

    final KieSession session = kbase.newKieSession();
    try {
        session.insert("start");
        session.insert(new java.util.ArrayList<>());
        assertThat(session.fireAllRules())
                .as("After removing sibling and merging segments, subject must still fire")
                .isEqualTo(1);
    } finally {
        session.dispose();
    }
}
```

Possible fix:
```
diff --git a/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java b/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
index b2953361bff..1d7c727dbb6 100644
--- a/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
+++ b/drools-core/src/main/java/org/drools/core/reteoo/SegmentMemory.java
@@ -922,8 +922,8 @@ public class SegmentMemory extends LinkedList<SegmentMemory>
             }
 
             SegmentPrototype[] eager = endNode.getEagerSegmentPrototypes();
-            if (requiresEager() && proto2.requiresEager()) {
-                // keep proto1 and remove proto2
+            if ((!requiresEager() && proto2WasEager) || (requiresEager() && proto2.requiresEager())) {
+                // remove proto2 from the eager array
                 SegmentPrototype[] newEager = new SegmentPrototype[eager.length - 1];
                 copyWithRemoval(eager, newEager, proto2);
                 endNode.setEagerSegmentPrototypes(newEager);
```

Without the fix, the test case fails with an `ArrayIndexOutOfBoundsException`.

Rikkola:
@tkobayas thanks. I will take a look.

Rikkola:
Thanks @tkobayas appended the fix.

Rikkola:
But did not run build with enough coverage based on the failures. Taking a look.

Rikkola:
Looks like it was not failing because of me, but since I have a temp commit here I converted back to draft.

PR #7126, opened 2026-09-22
[incubator-kie#2406] Fix_#2406 Drools: NPE in PhreakFromNode.doLeftDeletes when a left tuple is not in the from node's left memory

Thebas:
Adds the `leftTuple.getMemory() != null` guard around `ltm.remove(leftTuple)` in
`PhreakFromNode.doLeftDeletes`. A left tuple staged for delete has not necessarily
been added to the node's left tuple memory — it may have been staged and never
actually added, or already removed — and removing it anyway lets
`LinkedList.remove()` miss both the `firstNode` and `lastNode` branches, fall into
the middle-node branch and dereference a null `previous`.

`PhreakFromNode` was the only Phreak node without this guard. `PhreakJoinNode:409`,
`PhreakNotNode:446`, `PhreakExistsNode:432`, `PhreakAccumulateNode:556`,
`PhreakSubnetworkNotExistsNode:193` and `PhreakNodeOperations:219,240` all have it,
with the same `// it may have been staged and never actually added` comment.
`PhreakReactiveFromNode` extends `PhreakFromNode` and does not override
`doLeftDeletes`, so it is covered by the same change.

`PhreakFromNodeDeleteTest` pins the guard from both directions the state can be
reached:

- a left tuple that was never added to the left memory
- a left tuple that was added and already removed — `TupleList.remove` clears
  `memory`, `previous` and `next`

Both throw the NPE from the issue against current `main`. A third case asserts the
removal still happens when the tuple *is* in the memory, so the guard does not turn
a real delete into a no-op.

**NOTE!:** Double-check the target branch for this PR.
The default is `main`.

**Ports** If a forward-port or a backport is needed, paste the forward port PR

* None — targets `main` only.

**Issue**: _(please edit the GitHub Issues link if it exists)_

* [apache/incubator-kie-issues#2406](https://github.com/apache/incubator-kie-is

**Referenced Pull Requests**: _(please edit the URLs of referenced pull request

* None

<details>
<summary>
How to replicate the CI locally
</summary>

The CI does "simple" maven build(s). All commands can be run from the repositor

**Quick build of just your changed modules and their dependents** (mirrors what

```shell
# 1. Build upstream dependencies of your changes (skip tests for speed)
mvn -T 1C --batch-mode -fae -DskipTests -DskipITs \
    -Denforcer.skip=true -Dcheckstyle.skip=true -Dformatter.skip=true -Darchuni
    -pl <upstream-modules> install

# 2. Build your changed modules and their transitive dependents (with tests)
mvn --batch-mode -fae -Dreproducible \
    -pl <affected-modules> install

To compute <upstream-modules> and <affected-modules> automatically (requires JB

jbang script/ci/CiComputeBuildScopes.java \
    <changed-files.txt> \
    <upstream-modules.txt> \
    <affected-modules.txt> \
    <changed-modules.txt>

Full build (equivalent to what runs on every push to main):

mvn --batch-mode -fae -Dfull -Dreproducible install

-Dfull additionally builds the distribution modules and generates Javadoc JARs.

Useful flags:

┌────────────────────────
│          Flag          │                        Effect                         │
├────────────────────────┼
│ -Dquickly              │ Skip tests, Checkstyle, formatter, Enforcer, ArchUnit │
├────────────────────────┼
│ -DskipTests            │ Skip unit tests
├────────────────────────┼
│ -DskipITs              │ Skip integration tests
├────────────────────────┼
│ -Denforcer.skip=true   │ Skip Enforcer plugin              
├────────────────────────┼
│ -Dcheckstyle.skip=true │ Skip Checkstyle                   
├────────────────────────┼
│ -Dformatter.skip=true  │ Skip formatter                      
├────────────────────────┼
│ -Darchunit.skip=true   │ Skip ArchUnit
└────────────────────────┴

</details>

<details>
<summary>
How to retest this PR or trigger a specific build
</summary>

- To re-run CI: push a new commit to the PR (an empty commit is enough: git comger CI").

</details>
```


mariofusco:
The guard that this pull request aims to introduce is not only useless, but also violates one of the core principle of drools code base: fail as fast as possible instead of hiding the dust under the carpet. If a tuple that is not there is going to be removed we want the code the fail with an exception instead of guarding the removal and propagate the error downstream where it will be harder to be investigated and fixed. Also note that the provided test doesn't demonstrate an actual bug of the engine but only create an unreal situation. If you think this could be a real problem of the engine please provide an integration test demonstrating it using drools rules.

PR #7127, opened 2026-09-22
[incubator-kie#2407] Keep facts dynamic across a marshalling round trip

### LINKED ISSUE #2407: Drools: a dynamic fact loses its PropertyChangeListener when the session is marshalled and restored
### Component

Drools

### Versions

Drools: 10.2.0
Java: 21 runtime (locally 21.0.2); Maven compilation target 17
Spring Boot: 3.5.16
Maven: 3.9.16 (locally installed)

### Problem

A fact is dynamic when the entry point is registered on it as a JavaBeans
`PropertyChangeListener`, either through `insert(fact, true)` or because its type is
declared `@propertyChangeSupport`. Its setters then notify the session and the rules
matching it are re-evaluated with no explicit `update` call.

That registration does not survive marshalling. The listener is the entry point
itself, which is not serializable, so `PropertyChangeSupport` silently omits it and
the blob carries an empty listener list. `ProtobufOutputMarshaller` does not record
that the fact was dynamic, and `ProtobufInputMarshaller` re-creates the handle and
asserts it into the network without registering any listener.

The restored fact still holds the right values, which makes this easy to miss: the
object looks correct while the engine was never told it changed. `fireAllRules()`
does not compensate — there is no activation to fire, because nothing re-evaluated
the pattern.

- Fresh session: setter changes the fact → the entry point receives the event →
  affected rules can fire.
- Restored session: the same setter changes the fact → no listener receives the
  event → the fact is never re-evaluated.

### Reproducer

```java
public static class DynamicFact implements Serializable {
    private final PropertyChangeSupport support = new PropertyChangeSupport(this);
    private String name;

    public DynamicFact(String name) { this.name = name; }
    public String getName() { return name; }

    public void setName(String name) {
        String old = this.name;
        this.name = name;
        support.firePropertyChange("name", old, name);
    }
    public void addPropertyChangeListener(PropertyChangeListener l) { support.addPropertyChangeListener(l); }
    public void removePropertyChangeListener(PropertyChangeListener l) { support.removePropertyChangeListener(l); }
}
```

```drl
rule "name changed"
when
    DynamicFact( name == "changed" )
then
end
```

```java
KieSession session = kieBase.newKieSession();
((WorkingMemoryEntryPoint) session.getEntryPoint("DEFAULT"))
        .insert(new DynamicFact("initial"), true);
session.fireAllRules();

ByteArrayOutputStream baos = new ByteArrayOutputStream();
MarshallerFactory.newMarshaller(kieBase).marshall(baos, session);
session.dispose();

KieSession restored = MarshallerFactory.newMarshaller(kieBase)
        .unmarshall(new ByteArrayInputStream(baos.toByteArray()));

DynamicFact fact = (DynamicFact) restored.getObjects().iterator().next();
fact.setName("changed");

restored.fireAllRules();   // returns 0, expected 1
```

The same happens with a plain insert when the type is declared
@propertyChangeSupport.

Expected behaviour

An unmarshalled session behaves like the session that was marshalled: a fact that was
dynamic before the round trip stays dynamic, and its setters go on notifying the
engine. A fact inserted without the dynamic flag stays non-dynamic.

Thebas:
A dynamic fact notifies the session from its setters, so that changing it re-evaluates the rules matching it without an explicit update. That notification is a JavaBeans PropertyChangeListener registration and the listener is the entry point itself, which is not serializable: PropertyChangeSupport drops it on write, and the unmarshalled fact comes back with an empty listener list. The fact still holds the right values, which makes the loss easy to miss, but the engine is never told it changed.

Record the dynamic flag per fact handle in the blob and re-register the entry point as the listener on read, for both ways a fact becomes dynamic: an insert with the dynamic flag, and a type declared @propertyChangeSupport. The second is derivable from the knowledge base and so is also restored for blobs written before this change.

ProtobufMessages.java is regenerated from droolsmessages.proto with protoc 22.5, which reproduces the checked-in file byte for byte, so the diff is confined to FactHandle.

Records the dynamic flag per fact handle in the marshalled blob
(`FactHandle.is_dynamic`, new optional field 13 — 12 is taken by `otn_count`) and
re-registers the entry point as the fact's `PropertyChangeListener` on read.

`ProtobufInputMarshaller.reattachPropertyChangeListener` mirrors what
`NamedEntryPoint.insert` does on a live insertion: it re-registers when the handle
carries the persisted dynamic flag, or when the fact's type is declared
`@propertyChangeSupport`, and passes the flag through so `dynamicFacts` is rebuilt
and `removePropertyChangeListener` on delete still works. It deliberately does not
re-attach every fact that happens to expose `addPropertyChangeListener` — that would
give notification to facts inserted without the flag, which
`plainlyInsertedFact_afterRoundTrip_staysNonDynamic` pins.

`NamedEntryPoint.addPropertyChangeListener` is widened from `protected` to `public`
and `isDynamicFact(handle)` is added, so the marshaller can ask and act. No subclass
overrides either method.

`ProtobufMessages.java` is regenerated from the `.proto` with **protoc 22.5**, which
reproduces the checked-in file byte for byte, so the diff is confined to `FactHandle`.
Note for anyone regenerating: the `-Dproto` profile runs whatever `protoc` is on
`PATH`, and a different major version rewrites several thousand unrelated lines.

`DynamicFactMarshallingTest` covers a live session (control), a round trip after
`insert(fact, true)`, a round trip with a type declared `@propertyChangeSupport`, and
a round trip after a plain insert (control). The two round-trip dynamic cases fail on
`main`; the two controls pass either way.

Blobs written before this change carry no `is_dynamic` field, so it reads as false.
The `@propertyChangeSupport` half is derivable from the knowledge base and therefore
also restores older blobs; the `insert(fact, true)` half only applies to blobs written
after this change.

Fixes https://github.com/apache/incubator-kie-issues/issues/2407


tkobayas:
@Thebas  Thank you very much for reporting this issue and for the PR.

First, I'd like to know whether `insert(fact, true)` is actually used in your real-world use cases. It is an internal method of `WorkingMemoryEntryPoint`, not a public method of `KieSession`.

An old document mentions that this API is deprecated:
https://docs.drools.org/7.4.1.Final/drools-docs/html_single/#__propertychangesupport

(Although the method itself is not annotated with `@Deprecated`.)

```
The boolean parameter that was used in the insert() method in the Drools 4 API is deprecated and does not exist in the drools-api module.
```

If we limit the scope of the fix to `@propertyChangeSupport` only, the fix could be simpler.

Thebas:
Hello @tkobayas 

Indeed.
(we used insert($state, true); before, but we are migrating away our rules to a scalable strategy)

The @PropertyChangeSupport would be sufficient for our use case.

Should I fix the PR? Or you take over?

Thank you very much in advance.

